PR #2, opened 2015-04-03
pytx (pie-tex) library

mgoffin:
This PR adds the pytx library for interacting with the ThreatExchange API in Python.

The library allows you to initialize once with your APP-ID and APP-SECRET. From there all pytx objects as well as those created through the API will have access to those values for submitting requests.

Current classes include:

```python
from pytx import (
    ThreatIndicator
    ThreatExchangeMember
    Malware
)
```

When using the classes to access content you have a few options for how to get results. For example:

```python
from pytx import init
from pytx import ThreatIndicator

init('<app-id>', '<app-secret>')

# Get results which will return a generator which handles
# following the pager for you. Looping over the generator
# will return instantiated ThreatIndicator objects.
results = ThreatIndicator.objects(text="www.facebook.com")

# Get results using __raw__ which allows me to specify my
# GET parameters myself. Will still return a generator like the
# previous example.
results = ThreatIndicator.objects(__raw__={'text': 'www.facebook.com'})

# Get results but instead of using the generator I want the
# response body returned so I can manually handle it
# myself. This does not handle paging for you.
results = ThreatIndicator.objects(text="www.facebook.com", full_response=True)

# Get results and use the generator so we get looping 
# and paging, but instead of returning instantiated 
# ThreatIndicator objects, return each object as a dictionary.
results = ThreatIndicator.objects(text="www.facebook.com", dict_generator=True)
```

This should make things a bit more flexible for people who want more control or a more classful approach to interacting with the API. Also there is a request Broker which can be used directly for people who want full control over the URL and parameters when making requests as well as the full response. The Broker is used internally for the classes, but using it directly allows you to do things like the following:

```python
from pytx import init
from pytx.request import Broker

init('<app-id>', '<app-secret>')

results = Broker.get("https://graph.facebook.com/threat_indicators/",
                     params={'text': 'www.facebook.com'})
```

When querying for details you can also cater to your situation by using an instantiated or uninstantiated class:

```python
i = ThreatIndicator()
i.id = '1234'
i.details()

i = ThreatIndicator.details(id='1234')
```

Depending on the situation you are in it might be more useful to populate your already-instantiated object with the details, or if you don't have one it makes it very easy to get one and get the details in one step.

There is a vocabulary available for developers (and the library itself) to leverage for accessing ThreatExchange properly. It should be kept up-to-date over time as vocabulary changes. This makes it very easy for us to change field names, URLs, limits, etc. without disrupting existing code.

It comes with a `scripts` directory which contains scripts that replicate (as best as I can tell) all of the existing ThreatExchange Python scripts sans the POST script (I didn't want to submit random test data). When performing a `setup.py install` these scripts will also be installed so developers can use them as a reference if they wish.

I've started a `docs` directory with the hopes of getting this on readthedocs, but it seems like readthedocs doesn't like building documentation from subfolders (or I haven't been able to figure out how). It begs the question of whether we should make pytx its own repo/project, or just move all of the documentation to the root of the ThreatExchange repo and use readthedocs from there?

I haven't added this to PyPi as I figured we should get it in its final resting place before setting that up, so `pip install pytx` won't work yet.

*NOTE*: This PR could definitely use a few more eyes and lots of testing to make sure it's working well. There are a coupe TODO's in the vocabulary as some of the API documentation has changed over time and I don't want to remove large chunks of the vocab until it's confirmed it can get nuked. As long as we are happy with the core setup of the library then I'm good with proceeding. I'd hate for it to get accepted and we have a fundamental shift in how we want the library to work after people have started using it :)


mgoffin:
One thing I am debating doing (thought about it today) is abstracting away the request methods into their own "Request Manager" class.

This class would handle the GET, POST, DELETE, response, and returning the appropriate object (whether it be a dict, generator, instantiated object, etc.).

This would allow someone to do something like `ThreatIndicator.objects()` without having to first instantiate a `ThreatIndicator` object.

Thoughts?

jessek:
This is fantastic! Unfortunately I'm out and on mobile only for a while now and can't do a full review. I promise to look this I've over on full when I'm back on the 14th.

pytx/pytx/request.py
```diff
@@ -0,0 +1,277 @@
...
+            int(timestamp)
+            return True
+        except ValueError, e:
+            raise pytxValueError(e)
+
+    @staticmethod
+    def validate_limit(limit):
+        """
+        Verifies the limit provided is valid and within the max limit Facebook
+        will allow you to use (currently 5000).
```
jessek: The limit is now 1,000.

pytx/pytx/common.py
```diff
@@ -0,0 +1,345 @@
...
+        """
+
+        init.Broker.is_timestamp(timestamp)
+        self.set(ti.EXPIRED_ON, timestamp)
+        self.save()
+
+    def false_positive(self, object_id):
+        """
+        Mark an object as a false positive by setting the status to
+        NON_MALICIOUS.
```
jessek: The convention we are trying to establish is that only whitelisted data should be marked as NON_MALICIOUS. When something is a false positive, it should not necessarily be added to the whitelist. As such, when reporting a false positive, it would be better to set the status to UNKNOWN.

pytx/pytx/common.py
```diff
@@ -0,0 +1,345 @@
...
+        NOTE: This method can be used on both instantiated and uninstantiated
+        classes like so:
+
+            foo = ThreatIndicator(id='1234')
+            foo.details()
+
+            foo = ThreatIndicator.details(id='1234')
+
+
+        BE AWARE: Due to the nature of ThreatExchange allowing you to query for
```
jessek: This is a feature, not a bug. :)

pytx/pytx/request.py
```diff
@@ -0,0 +1,277 @@
...
+        :param limit: The limit to validate.
+        :type limit: int, str
+        :param since: The since timestamp to validate.
+        :type since: str
+        :param until: The until timestamp to validate.
+        :type until: str
+        """
+
+        if since:
+            cls.is_timestamp(since)
```
jessek: The since and until parameters don't have to be timestamps/integers. There are some magic values, yesterday; today; and now, which can make searching easier.

pytx/pytx/vocabulary.py
```diff
@@ -0,0 +1,448 @@
...
+    MALICIOUS       = "MALICIOUS"
+
+
+class PrivacyType(object):
+    """
+    Vocabulary for the Threat Indicator Privacy Type.
+    """
+
+    VISIBLE         = "VISIBLE"
+    HASWHITELIST    = "HASWHITELIST"
```
jessek: These values should be HAS_WHITELIST and HAS_BLACKLIST.

jessek:
This code looks great! Thank you for taking the time to work on ThreatExchange. I've found a few nits/typos, above, but in general this library is going to be a great addition. I'm looking forwarding to merging it.

I'll keep testing it over the next few days to ensure the functionality works as intended.

mgoffin:
Thanks! I'll try to get to these soon. Is there a full list of helper vocab for since and until so I can integrate? Or would you prefer to skip validation and let the server respond accordingly?

jessek:
Since and Until are defined in the base Graph API documentation,
https://developers.facebook.com/docs/graph-api/using-graph-api/v2.3#paging.
Look under 'time based pagination'. The short version is 'anything which
strtotime which parse,' which is actually a lot, e.g.

12:00
03/15/2015 12:00
03/15/2015
15 Mar 2015
2 months ago
-1 day
-60 min
-2 years
yesterday
today
now



On Fri, Apr 17, 2015 at 4:40 PM, Mike Goffin <notifications@github.com>
wrote:

> Thanks! I'll try to get to these soon. Is there a full list of helper
> vocab for since and until so I can integrate? Or would you prefer to skip
> validation and let the server respond accordingly?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/facebook/ThreatExchange/pull/2#issuecomment-94098135>.
>



-- 
Jesse Kornblum
jessekornblum@gmail.com

mgoffin:
Looking over the python options, dateutil was the first thing that popped into my head. It also seems like the best options. It would add another dependency outside of `requests` if we choose to add it. If that's not a decent option we could just leave it and let people submit whatever they want and deal with error responses?

pytx/pytx/common.py
mgoffin: Definitely a feature, but worth noting in detail for developers who are confused when they use an ID in the wrong class and wonder why things look messed up. We could get around this if objects returned their type by default without having to query with `&metadata=1` and parsing out the type from there.

mgoffin:
I've left the `is_timestamp` method in the code, but I've removed the exception handling. Now all it does is return `True`. I can remove it completely but my reason for leaving it was to 1) alert developers as to the reasons why `since` and `until` aren't being validated, and 2) what you can actually provide for those values from a programmatic perspective.

I tried python's `dateutil` but it doesn't work in the same way as `strtotime` based on their examples. Without a good way to exactly duplicate `strtotime` I think it's best to remove the check and leave it up to the API user to handle in the event they provide a bad value.

mgoffin:
I have a change I might commit but I wanted to ask before I do so and push.

Given that the API has a version, I wanted to add a version to the vocabulary and then have pytx use it. It would make an install of pytx API-version-specific, but the Broker (since it takes direct input) could be used to leverage other versions directly.

Is this something that would be helpful to include? Or should I leave it out and let the API default to the latest version?

hammem:
@mgoffin adding the version number is a great idea.  We're going to start linking breaking API changes to version numbers, so having support for them in pytx will be really handy soon.

mgoffin:
@hammem sounds good!

jessek:
I'm going to merge this request as is. I'm sure we'll have changes and fixes as time goes on, but let's get this party started.

PR #3, opened 2015-04-06
Cleans up the malware fetcher script

tpott:
Added parsed arguments, allows for single hash queries as well as maintaining support for a file with several hashes. Also adds the ability to include a mix of hash types in the file input format


9b:
Agreed with you tpott. Fair enough.

PR #4, opened 2015-04-20
New PHP upload script and classes for Indicators

hammem:
This adds a few new PHP classes and a CLI script for uploading
indicators to ThreatExchange.

I also cleaned up some errors and things that have changed in the
README and markdown API docs.


jessek:
Looks great, thank you!

hammem:
Fixed up BaseUpload.php, since it was trying to support malware and family uploads with classes that don't yet exist.

PR #5, opened 2015-04-21
Start of an interface for ThreatExchange.

mgoffin:
Initial build of a client-side interface for ThreatExchange.

If you want to hard-code your app-id and app-secret after getting it locally, you can do so in the `js/globals.js` file, otherwise you'll have to enter it every time you refresh.


tiegz:
(this is nice, thanks for adding this!)

PR #6, opened 2015-04-23
Added Ruby Library

maus-:
A basic ruby library to interface with Facebooks ThreatExchange API. 

## Installation
Install required dependencies 

    bundle install

Install this gem locally

    rake install

Add this to your application:

    require 'ThreatExchange'

## Usage

The ThreatExchange library provides a base set of abstractions for the ThreatExchange API. You instanciate a new ThreatExchange::Client and provide it with your access token.

```ruby

access_token = '56356356|367665625651'
TE = ThreatExchange::Client.new(access_token)

```
To query the ThreatExchange API you create a hash with parameters for your query

```ruby

 query = { 
    threat_type: 'COMPROMISED_CREDENTIAL',
	type: 'EMAIL_ADDRESS',
	fields: 'indicator,passwords',
	limit: 30
}
```

Then call the query with the respective method. 
```ruby

result = TE.threat_indicators(query)
```

The result will return either a string, a singular hash or an array of hashes and then from there you can manipulate the data as you like.

Take a look at the examples folder for more usage examples.


ruby/README.md
```diff
@@ -0,0 +1,57 @@
...
+    require 'ThreatExchange'
+
+## Usage
+
+The ThreatExchange library provides a base set of abstractions for the ThreatExchange API. You instanciate a new ThreatExchange::Client and provide it with your access token.
+
+```ruby
+
+access_token = 'abc123'
+TE = ThreatExchange::Client.new(access_token)
```
jessek: Where should a person specify their AppID? Should that be included in the access token? (e.g. 'xxx|yyyy' ?)

ruby/example/ThreatExchangeTest.rb
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env ruby
+# ThreatExchange example script.
+require 'ThreatExchange'
+require_relative 'lib/example'
+access_token = 'abc123|somethingsomething'
+TEClient = ThreatExchange::Client.new(access_token)
+#------------------------------------------------------------------------------
+banner('Example Malware Analysis')
+query = {limit: 1000, text: 'bepush', strict_text: false}
```
jessek: Because the default limit may change per the Platform API, it's probably better not to explicitly set a value here. Maybe offer it as an optional parameter. But we should use the system default by default. ;-)

maus-: The example script is just showing parameters that can be passed into each method, its not a default nor do you have to set it. So yes it is an optional parameter ;)

ruby/README.md
maus-: Exactly. You would just provide it the same way that you generate it from the graph.

ruby/lib/ThreatExchange.rb
```diff
@@ -0,0 +1,148 @@
...
+          paging = result['paging']['cursors']['after']
+        end
+      end 
+      return data
+    end
+
+    def threat_indicators(filter={})
+      filter[:access_token] = @access_token
+      begin
+        RestClient.proxy = "127.0.0.1"
```
tiegz: @maus- why the proxy here?

ruby/lib/ThreatExchange.rb
```diff
@@ -0,0 +1,148 @@
...
+      rescue => e
+        puts e.inspect
+      end
+      data = result['data']
+      paging = result['paging']['cursors']['after']
+      until paging.nil?
+        filter[:after] = paging
+        begin 
+          response = RestClient.get "#{@baseurl}/threat_indicators", 
+          { params: filter, verify_ssl: OpenSSL::SSL::VERIFY_NONE }
```
tiegz: ooooh

ruby/lib/ThreatExchange.rb
maus-: oh total whoops, leftover from debugging. Good catch

ruby/lib/ThreatExchange.rb
maus-: I thought I removed all this. Thanks for noticing :+1:

ruby/README.md
jessek: Can you please change this example to have the format of an actual access token:

    access_token = '56356356|367665625651'

maus-: updated

PR #8, opened 2015-04-28
UI: fix positioning on main content

tiegz:
Fixes the sidebar/results list -- the first item in the results list and the sidebar error were being covered up by the navbar.

### DIFF
--- a/ui/css/dashboard.css
+++ b/ui/css/dashboard.css
@@ -36,7 +36,7 @@ body {
 @media (min-width: 768px) {
   .sidebar {
     position: fixed;
-    top: 51px;
+    top: 100px;
     bottom: 0;
     left: 0;
     z-index: 1000;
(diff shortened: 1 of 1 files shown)


tiegz:
done-so!

mgoffin:
Howdy! Just curious what browser you are using for this? With your changes the search bar is now significantly lower than the nav bar which looks weird since the search bar has a different background color. I was not having the issues you were having prior to your changes. This was built using Chrome so it's possible there's a little conflict causing your browser issues though.

tiegz:
@mgoffin I'd tested this on Chrome and Safari, but I think I see the issue now. My browser was open at <1200px wide, which broke the search bar down to the next line. Still worth fixing imo, but I'll try to think of a better approach to fixing this!

mgoffin:
@tiegz Ahh that would do it! Yea I think we'll need a better solution for small window sizes!

PR #10, opened 2015-05-08
Enforcing PRIVACY_TYPE in pytx/UI

mgoffin:
Enforce the PRIVACY_TYPE requirement for submitting new Indicators in pytx and perform a quick check in the UI (which pre-populates the PRIVACY_TYPE dropdown anyways so unless someone is abusing the JS it should already work).

Added some helper properties to the Malware object:

    .zfh - return a file handle for the b64 decoded zip file
    .zf  - return the raw b64 decoded zip file
    .rfh - return a file handle for the b64 decoded and unzipped file
    .rf  - return the raw b64 decoded and unzipped file

These properties should be flexible enough for people who want to utilize the sample without having to do the decode/unzip dance at whatever stage they are looking to get to. Required adding a `_unique` class attribute to track properties we don't want people to change.

Fixed a couple other bugs with saving objects to ThreatExchange in pytx:

- Don't keep adding the ID to `_DETAILS` making it an invalid URL.
- Use the Broker to submit new content to ThreatExchange.
- `.save()` does not require a url to be passed in.
- Remove the hard version set for pytx/UI to take advantage of the features in newer versions.

### DIFF
--- a/ui/js/vocabulary.js
+++ b/ui/js/vocabulary.js
@@ -1,6 +1,6 @@
 var v_ThreatExchange = {
     URL             : "https://graph.facebook.com/",
-    VERSION         : "", //"v2.4/",
+    VERSION         : "v2.3/",
     ACCESS_TOKEN    : "access_token",
     DEFAULT_LIMIT   : 500,
     MAX_LIMIT       : 5000,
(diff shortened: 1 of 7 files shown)


ui/js/vocabulary.js
```diff
@@ -1,6 +1,6 @@
 var v_ThreatExchange = {
     URL             : "https://graph.facebook.com/",
-    VERSION         : "v2.0/",
+    VERSION         : "", //"v2.4/",
```
jessek: This should be version 2.3, the current version. The next version of the API, 2.4, will change some of the functionality for handling descriptors rather than indicators.

PR #11, opened 2015-05-08
Privacy Type is mandatory for IOC submissions now in ruby library

maus-:
All IOC submissions must contain a privacy type key in the hash submitted to the method now.


jessek:
Looks good, thank you!

PR #13, opened 2015-05-13
bugfixes

aledista:


malware/get_samples.py
```diff
@@ -66,7 +66,7 @@ def run_query(url):
         # the response is a 'data' object, which is a list of objects
         resp = json.loads(urlopen(url).read())['data'][0]
     except Exception as e:
-        print('Failed to load "%s": %s' % (md5, str(e)))
+        print('Failed to load "%s": %s' % ('md5', str(e)))
```
mgoffin: I guess we could just include "md5" as part of the string instead of adding it in the tuple? Nice catch!

mgoffin:
Thanks for cleaning all of this up! Other than the line note, everything else looks great!

malware/get_samples.py
jessek: It's a nit, but it's a good nit. Let's move md5 into the string and I'll merge.

jessek:
Nice catches. Did you find these because you triggered the exceptions?

Thank you for contributing to ThreatExchange!

aledista:
Nope, just looking at the code :)

PR #15, opened 2015-05-20
UI: fix for param parsing in search terms.

tiegz:
Currently a search for `threat_type:X` matches `type:` first and produces this search: `.../threat_?type=X`.


hammem:
Thanks @tiegz !

PR #17, opened 2015-05-21
Add member email to pytx and UI.

mgoffin:
I had a quick second :)

Exposes the email of a ThreatExchangeMember in pytx. Also when you drop
the Members menu down in the UI it will create a mailto link with the
POC email address.

### DIFF
--- a/ui/js/api.js
+++ b/ui/js/api.js
@@ -107,7 +107,8 @@ function get_members() {
                 .css('max-height', '500px')
                 .css('overflow-y', 'auto');
             $.each(json.data, function(idx, member){
-                html.append(member.name + "<br />");
+                var mailto = '<a href="mailto:' + member.email + '">' + member.name + '</a>';
+                html.append(mailto + "<br />");
             });
             $('#member-popover-content').html(html);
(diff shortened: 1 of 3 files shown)


hammem:
Thanks @mgoffin !

ui/js/api.js
```diff
@@ -107,7 +107,8 @@ function get_members() {
                 .css('max-height', '500px')
                 .css('overflow-y', 'auto');
             $.each(json.data, function(idx, member){
-                html.append(member.name + "<br />");
+                var mailto = '<a href="mailto:' + member.email + '">' + member.name + '</a>';
```
jessek: The email field is optional and is not guaranteed to be present. Will this throw an exception if it's not there?

mgoffin:
The link will show "mailto:undefined" but no exception will be thrown. We could check for undefined and not make it a link? I didn't know a good way to say "this org hasn't given a POC yet"

tpott:
Could we have a mouseover tooltip read "No POC available"?

mgoffin:
Seems legit!

PR #20, opened 2015-05-26
add fields selection to object pull

omercnet:
Allow fields selection for object pull, so you can fetch fields that are not pulled by default like passwords
This is useful for large fetches to avoid using ThreatIndicator.details() for each IOC

Also fixed DEFAULT_LIMIT to meet the actual default on the graph api, and then modernized the sample and made some use of DEFAULT_LIMIT

### DIFF
--- a/pytx/scripts/get_compromised_credentials.py
+++ b/pytx/scripts/get_compromised_credentials.py
@@ -11,8 +11,8 @@ from pytx.vocabulary import ThreatType as tt
 from pytx.vocabulary import Types as t
 
 
-app_id = '<your-app-id>'
-app_secret = '<your-app-secret>'
+app_id = '<app-id>'
+app_secret = '<app-secret>'
 
 init(app_id, app_secret)
 
@@ -24,28 +24,23 @@ def get_results(options):
     if options.since is None or options.until is None:
         raise Exception('You must specify both "since" and "until" values')
 
-
-    query = {
-        'threat_type': tt.COMPROMISED_CREDENTIAL,
-        'type' : t.EMAIL_ADDRESS,
-        'fields' : 'indicator,passwords',
-        'since' : options.since,
-        'until' : options.until,
-    }
-
-    results = ThreatIndicator.objects(__raw__=query, dict_generator=True)
+    results = ThreatIndicator.objects(threat_type=tt.COMPROMISED_CREDENTIAL, type_=t.EMAIL_ADDRESS, limit=options.limit,
+                                        fields=['indicator','passwords'], since=options.since, until=options.until)
     return results
 
-def process_results(handle, data):
+def process_result(handle, result):
     '''
     Process the threat indicators received from the server. This version
     writes the indicators to the output file specified by 'handle', if any.
-    Indicators are written one per line, formatted in JSON.
+    Indicators are written one per line.
     '''
-    if handle is None:
-        return
 
-    handle.write('%s:%s\n' % (data['indicator'], data['passwords'][0]))
+    for password in result.passwords:
+        output = '%s:%s\n' % (result.indicator, password)
+        if handle is None:
+            print output,
+        else:
+            handle.write(output)
 
 def run_query(options, handle):
     try:
@@ -58,7 +53,7 @@ def run_query(options, handle):
 
     count = 0
     for result in results:
-        process_results(handle, result)
+        process_result(handle, result)
         count += 1
 
     try:
@@ -80,15 +75,16 @@ def get_parser():
     parser.add_argument('-o', '--output')
     parser.add_argument('-s', '--since')
     parser.add_argument('-u', '--until')
+    parser.add_argument('-l', '--limit')
 
     return parser
 
 if __name__ == '__main__':
     args = get_parser().parse_args()
-    if args.output is not None:
-        handle = open(args.output, 'w')
-    else:
-        handle = None
+    
+    handle = open(args.output, 'w') if args.output else None
+    args.limit = te.DEFAULT_LIMIT if args.limit is None or not args.limit.isdigit() else int(args.limit)
+
     start = int(time.time())
     run_query(args, handle)
     end = int(time.time())
--- a/pytx/scripts/get_compromised_credentials.py
+++ b/pytx/scripts/get_compromised_credentials.py
@@ -82,8 +82,15 @@ def get_parser():
 if __name__ == '__main__':
     args = get_parser().parse_args()
     
-    handle = open(args.output, 'w') if args.output else None
-    args.limit = te.DEFAULT_LIMIT if args.limit is None or not args.limit.isdigit() else int(args.limit)
+    if args.output is not None:
+        handle = open(args.output, 'w')
+    else:
+        handle = None    
+
+    if args.limit is None or not args.limit.isdigit():
+        args.limit = te.DEFAULT_LIMIT
+    else: 
+        int(args.limit)
 
     start = int(time.time())
     run_query(args, handle)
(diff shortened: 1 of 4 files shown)


mgoffin:
If @jessek and/or @hammem can confirm that these changes are appropriate from the Graph perspective, this would simplify things so you can do a single query for data instead of two.

omercnet:
https://developers.facebook.com/docs/graph-api/using-graph-api/v2.3
```
Choosing Fields

By default, not all the fields in a node or edge are returned when you make a query.
You can choose the fields (or edges) you want returned with the "fields" query parameter.
This is really useful for making your API calls more efficient and fast.
```

omercnet:
Hi guys, any update here?

pytx/scripts/get_compromised_credentials.py
```diff
@@ -80,15 +75,16 @@ def get_parser():
...
     return parser
 
 if __name__ == '__main__':
     args = get_parser().parse_args()
-    if args.output is not None:
-        handle = open(args.output, 'w')
-    else:
-        handle = None
+    
+    handle = open(args.output, 'w') if args.output else None
```
jessek: Although this makes the code fit on fewer lines, it makes it, in my opinion, more difficult to read.

jessek:
Adding the 'fields' parameter is great and the same for the 'limit'. We've been talking internally about how we want to handle that. On one hand, we want to do things the Facebook way, which would mean returning 25 results, by default, on each request. That would also ease some of the load on our end for searching, too. But we don't want to make the system unusable by requiring a jillion connections to download all of the indicators which a person wants. Thoughts?

pytx/scripts/get_compromised_credentials.py
omercnet: Given the fact this is meant to serve as a simple example, I agree

omercnet:
I was thinking the same thing.
There should be a total limit for the amount of objects you want to to pull, which works well if it's under 1000 - it results in a single request.
But when you need something between 1000 and everything, you get stuck.

A simple solution to that would be simply removing the [limit check](https://github.com/facebook/ThreatExchange/blob/master/pytx/pytx/request.py#L68), that way if the result is greater than 1000 the Graph will still return only 1000 per page, but the generator will still stop once you reach the limit.

What do you think?

jessek:
I agree that we should remove the restrictions on the limit parameter. Hopefully we can raise the default and max limits on our end as we improve the performance of the system. When that happens, I'd like that benefit to be available via this framework without a code change. 

Let's allow the framework to set the limit parameter and do some basic validation on it (i.e. has to be an integer greater than zero). But we should also make the framework not specify a limit by default; use the default from Facebook unless specifically requested. And, if you want to submit a request with a limit of 999999999, you should be able to do that. Who knows, maybe someday we'll be able to handle that! #aimhigh

omercnet:
so what I did in the [last commit](https://github.com/omercnet/ThreatExchange/commit/89e66c8b56b1eeb234817becebbbbb84dc4b262c) is change it so no limit means 25 (DEFAULT_LIMIT) and -1 (NO_LIMIT) means get all the results 1,000 at a time. Anything specific and greater than zero will still fetch up to 1,000 at at time

jessek:
Right now the code restrict a query to 25 results if the caller does not specify an explicit limit. That number is the same as the Facebook default for query results. I'm hopeful that we can raise the default query limit in the future. When we do that, I think the library should use the new Facebook query limit automatically. The same goes for the maximum query limit. It's currently 1,000, but I'd like to increase it. If/when we do that, this library should use that limit rather than a self-imposed one.

omercnet:
The code doesn't restrict to 25 results, if no there's no limit specified Facebook returns 25.
My last commit added the use of DEFAULT_LIMIT if none is specified.

As for the framework learning the limits automatically, it's either we query the settings from an external source, or run a few queries to check the current limits - either way that adds several queries for every run,to determine something that may change once a year (probably every few years).

Unless there's anything we can solve right now, could you please approve the merge? I need to update my internal tools :)

jessek:
I'm not suggesting that the code queries the server to determine the default and maximum limits. 

I'm suggesting that we should remove DEFAULT_LIMIT AND MAX_LIMIT from the code entirely. Let the server dictate the default and maximum allowed values. Let the person whatever values they want to pass to the server.

mgoffin:
Those values were added due to the language in the TX API docs. If we decide to remove them we should update the docs as well :)

It's also possible to leave those values in pytx vocabulary but not leverage them. People doing development may wish to use them for documentation, responses, error messages, etc. Thoughts?

jessek:
We have a requirement to update our docs when update our code. But I don't think we should have to update pytx at the same time. 

I'm not at all opposed to the library having its own defaults and limits, separate from the server. But we should at least put a comment in about that.

omercnet:
I actually think removing the limits works well, the only issue there is the inability to specifically request all the objects. But we can solve that by requesting 999999999 like you suggested

jessek:
If the client does not specify a limit, they will get all of the objects. There's no need to specify 999999.

omercnet:
Oh that's right, but then we're back to the paging issue
They'll make a ton of requests with 25 results each

jessek:
Exactly, yes. That's tradeoff we're wrestling. Which is better? Lots of requests for 25 results each, or one big request for 1,000 results?

For now, I think that lots of little requests is the way to go.

omercnet:
It makes fetching large amounts of objects painfully slow..

jessek:
But fetching a large amount of objects all at once can cause a timeout and thus no data to be returned. 

Surely there is a happy medium? Something above the Facebook default but doesn't crash the system? 250? 500? I'd be ok with the library having its own default, so long as it's explicitly labeled as to why we're doing it. And stop calling me Shirley. :airplane:

omercnet:
The requests library can handle the timeouts, I don't see a good reason to query small amounts by default..

jessek:
No, the timeout is on the server side. The request takes too long to compute, and so it returns an error. The client doesn't get any data after that for the query.

omercnet:
Bummer.. So why don't you guys fix that? #aimhigh :) 
and we'll settle for hard coded 500 limit ?

jessek:
We *absolutely* want to get that bug fixed! On a completely unrelated note, did you see this? https://www.facebook.com/careers/department?dept=it&req=a0I1200000G4YE5EAN

But yes, I think 500 is a good limit for now. We can always revisit and/or restrict from the server side as necessary.

omercnet:
Thanks man, but I'm pretty happy at Yahoo. Tell Mark and Chad I say hi :)

I'll commit the change tomorrow.

omercnet:
@jessek @mgoffin are we OK to merge this then..?

jessek:
I'd like to remove the MAX_LIMIT parameter and the code which prevents the user from submitting any particular 'limit' parameter. I'd like the server to be the "bad guy" and tell a user they aren't allowed to do something. :)

mgoffin:
I'm in favor of this as well.

Allow people to set whatever page limit they want and let the server determine if:

1) There's no limit provided so use the server default.
2) There's a limit and it's valid so use it.
3) There's an invalid limit and return an error (or do something else sane and document the behavior).

People can always bomb out of the generator if they need/want to :)

omercnet:
You can submit any 'limit' parameter you want,  validate_limit only checks it's an integer

as we previously agreed, MAX_LIMIT is set to 500 and only used when 'limit' is >500, it it only used to paginate the results as MAX_LIMIT per page. Otherwise the server's default of 25 per-page is just awful with large pulls :/

omercnet:
breaking this up to separate pull requests.

PR #21, opened 2015-06-01
Omercnet pip prep

omercnet:
fixed setup.py and pushed to pypi to streamline builds and development of tools using pytx
https://pypi.python.org/pypi/pytx

### DIFF
--- a/pytx/DESCRIPTION.rst
+++ b/pytx/DESCRIPTION.rst
@@ -1,4 +1,1 @@
-pytx
-====
-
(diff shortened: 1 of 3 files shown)


mgoffin:
+1!!

mgoffin:
Also does PyPi point to the official repo? It says 0.1.2 already before this is merged. If not I think it should get changed so it tracks the official repo once this is accepted :)

omercnet:
it does not, if you'll tag the commits here we can link directly to the releases here, which would be awesome :)

PR #30, opened 2015-06-03
Add documentation for readthedocs

mgoffin:
This adds a quickstart guide (could always use more work but it's a good start) and module documentation. This will allow readthedocs to auto-generate documentation for the codebase off of master.


jessek:
This is fantastic, thank you!

For those of us who don't grok readthedocs, can you please add a script/README with instructions on how to regenerate these files? Something which I point people to when I tell them, "please update the documentation for this feature you're adding/changing".

mgoffin:
Sure! readthedocs isn't anything special outside of what we would normally require. Any changes to pytx we would want to make sure the docstrings are updated if necessary and the only focus we will want to keep up on is the `rst` documentation in the `docs` directory.

Does it make sense to put the README in `docs` or would it make more sense to start a `CONTRIBUTING` file that's a bit more standard?

jessek:
It doesn't matter where you put the instructions, but I want something so that an intelligent, but currently inexperienced person (i.e. me) can regenerate the documentation.

I agree that the main README is a bad place for it. But a CONTRIBUTING file sounds like a winner.

mgoffin:
Added a CONTRIBUTING file :)

jessek:
TIL there's a thing called Sphinx.

thank you!

Issue #32, opened 2015-06-04
InsecurePlatformWarning from requests on OSX with default python


ivanlei:
```shell
$ python scripts/get_compromised_credentials.py --since=2015-05-01 --until=2015-06-01
READING https://graph.facebook.com/threat_indicators/
/Users/ivanlei/virtual_envs/ThreatExchange/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
```

This is easily fixed. PR incoming.

ivanlei:
Closed by https://github.com/facebook/ThreatExchange/pull/33

PR #33, opened 2015-06-04
Have pytx use TLS1.0+ to avoid bad SSL warnings and avoid using bad SSL

ivanlei:
**Let's use TLS!**
This is issue #32

The fix here is real simple. Install the option `secure` requirements for the `requests` package.
See more @ http://stackoverflow.com/questions/29099404/ssl-insecureplatform-error-when-using-requests-package

**.gitignore**
Quick update. Add `.DS_Store` which OSX likes to litter the filesystem with.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ pytx/build
 pytx/dist
 pytx/pytx.egg-info
 pytx/docs/_build
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ pytx/dist
 pytx/pytx.egg-info
 pytx/docs/_build
 .DS_Store
+.tox/
+.coverage
--- a/pytx/Makefile
+++ b/pytx/Makefile
@@ -1,0 +1,15 @@
+.DELETE_ON_ERROR:
+
+all:
+	echo >&2 "Must specify target."
+
+test:
+	tox
+
+clean:
+	rm -rf build/ dist/ *.egg-info/ .tox/
+	rm -f .coverage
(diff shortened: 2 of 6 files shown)


mgoffin:
I believe this is for pre-python-2.7.9 users? I wish requests documented that "feature" more as it seems it would be useful for a lot of people.

ivanlei:
Yeah, @mgoffin it's for pre-python-2.7.9 users. I tried to find this documented on http://docs.python-requests.org/en/latest/ or https://pypi.python.org/pypi/requests and I don't think it is. Lemme go fork https://github.com/kennethreitz/requests and document it.

mgoffin:
Is there a situation where requiring `requests[security]` will break for anyone? If not I'm :+1: for this!

ivanlei:
Lemme try to use tox to test that @mgoffin. From https://github.com/kennethreitz/requests/blob/master/setup.py#L65 it seems `requests` is compatible with a wide range of python versions but there's no tests for that.

ivanlei:
Okay. So that last commit adds a `tox.ini` which then installs `pytx` (using both `setup.py` and `requirements.txt`) under:
- py26
- py27
- py33
- py34

As a side benefit, if any further tests are added to `/tests` there's a `tox.ini` ready to run them.

Basically, this verifies there's not a situation where `requests[security]` breaks something.

mgoffin:
Excellent! Looks good :+1:

omercnet:
I concur :) :+1:

hammem:
thanks, @ivanlei !

PR #35, opened 2015-06-04
Add unit tests for pytx.init

ivanlei:
## The Changes
**Unit tests come to `pytx`**
So I want to make some changes to the behavior of `pytx.init` but it seemed like I ought to get some tests in place first so I don't break things too bad. 

It might look like a lot has changed here, but first focus on:
- `pytx/tests/access_token_test.py` 
  - These are them. The unit tests. They do a good job of testing the various ways `init` can be called. And dang, there are a lot of ways!
- `pytx/pytx/access_token.py` 
  - This is a git move of `pytx/pytx/init.py`. Having a file in a package named `init.py` and a public member in the package named `init` basically made it impossible for the unit tests to introspect the private members of the 
  - The changes here are relatively minor. Basically, the code for reading creds from a file has been pulled out into a helper function. This make unit testing a ton easier as mocking `open` is pretty painful.
  - A bit of error handling cleanup. Deal with `IOError` vs `Exception`. [This stack overflow does a good job of explaining why](http://stackoverflow.com/questions/21553327/why-is-except-pass-a-bad-programming-practice)

The file rename of course flows through the rest of the package.

**Let's make it easy to build the docs**
I wanted to refer to a local copy of the HTML docs so I added some build commands.
```shell
$ make docs
```
will now build the pytx docs.

I added a new doc with basics on developing within the pytx module.

**pre-commit - PEP-8 cleanup**
So the rest of the changes are basically whitespace fixes that make the code come into alignment with PEP-8 and basic python linting. This is all accomplished via [pre-commit](http://pre-commit.com).

In brief, after building a virtualenv to do development work out of you wanna the following one time:
```shell
$ pre-commit install
```
After that, whenever you commit, pre-commit will run and either fix up minor issues in your files or let you know about them. Fix the issues, stage the files, and then commit again. Once there are no issues, the commit will succeed. To skip the checks, use:
```shell
$ git commit --no-verify
```
But `--no-verify` is so uncool.

## How I Verified This
- Run the tests: In a clean checkout of the branch, do `$ cd pytx && make test`
- Build the docs: `$ cd pytx && make docs` and then read the docs in Chrome.
- Verify __all__: `$ python -c 'import pytx; print dir(pytx)'` and look for the right exports.
- Run all the scripts in `pytx/scripts`:
  - Add my creds to each script
  - `python scripts/get_compromised_credentials.py --since=2015-05-01 --until=2015-06-01`
  - `python scripts/get_indicators.py --type=IP_ADDRESS --since=2015-05-01 --until=2015-06-01 -o=/dev/stdout`
  - `python scripts/get_members.py`
  - `python scripts/malware_api.py`
  - `echo '681f1b31baa671a81e4b803dbf8a9f10' > md5s.txt && mkdir out && python scripts/malware_grabber.py`
- Clean stuff: `$ make clean` and verify with `$ git status` that no new stuff is around.


malware/get_samples.py
```diff
@@ -15,13 +15,15 @@
...
 from urlparse import urlunparse
 
+
 def get_app_creds():
     """Reads your App ID and App Secret from env variables"""
     return {
-        'AppID' : os.environ['FB_THREATEXCHANGE_APP_ID'],
-        'AppSecret' : os.environ['FB_THREATEXCHANGE_APP_SECRET'],
+        'AppID': os.environ['FB_THREATEXCHANGE_APP_ID'],
+        'AppSecret': os.environ['FB_THREATEXCHANGE_APP_SECRET'],
```
mgoffin: These should probably be changed to use the vocabulary and pull the proper environment variables, similar to how init does it.

pytx/pytx/__init__.py
```diff
@@ -1,5 +1,13 @@
-import vocabulary
-from init import init
+from access_token import init
+from access_token import Broker
```
mgoffin: Why is the Broker being imported from access_token?

ivanlei: Lemme refactor this a bit more. I have a plan for a follow up change to sort this. Lemme just do it now.

mgoffin:
Chatting with @ivanlei about this. We agree that the Broker shouldn't be in `init` but that `access_token.Broker` is very confusing. This is part of the code I wasn't originally happy with (using globals is no good).

We're talking through ways to solve this so both the code and the way to use the code make sense.

ivanlei:
Okay, @mgoffin I think this addresses you concerns and cleans things up nicely?

- pytx.access_token.get_access_token replaces the global __ACCESS_TOKEN__
- Error handling around a missing access token is entirely contained in pytx.access_token.get_access_token
- All use of the access token is restricted to pytx/request.py and pytx/access_token.py

mgoffin:
I think I am good with this!

I ran my personal scripts for testing API requests and they all came back as expected. I ran `make test` and had 9 collected tests with 9 passes. I ran `make docs` and docs were properly generated. `pre-commit` worked as advertised (at least the install and setup in the repo, I have yet to commit anything to verify it works).

The re-arrangement of code to get away from the global access token makes a lot of sense and is much cleaner.

I tested having my token generated automatically from my environment and that worked. I was also able to call `init()` manually and supply my token.

The PEP8 stuff all looks good too.

Unless @ivanlei has anything else he's cleaning up... :+1:!

ivanlei:
I'm not looking to clean up anything else in this. Thanks @mgoffin

jessek:
Y'all make this too easy for me. :+1:

PR #36, opened 2015-06-05
No more access tokens in code

ivanlei:
**Motivation**
- I *almost* checked my access token into the sample scripts twice already. 
- The graph documentation is pretty explicit - don't put your access token in code.
*However*, the sample scripts all require you to add your access token to the code. :-1: 

**The Change**
Speaking with @mgoffin we came up with these rules for how `pytx.access_token.init` should work:
     1. Use the value of the 'TX_ACCESS_TOKEN' environment variable.
     2. Use the concatenation of the 'TX_APP_ID' and 'TX_APP_SECRET' environment variables.
     3. Use the first line of the file '$PWD/.pytx' or ~/.pytx'
     4. Use the concatenation of the app_id and app_secret parameters
     5. Use the first line of the file 'token_file'
So now `pytx.access_token.init` works that way. For the first 3 cases, there is no need to explicitly call `pytx.access_token.init` anymore.

**Script Updates**
Quick cleanup of the scripts such that they all have a clean `main` and have some stuff like help strings for the command line arguments.

**How I Know It Works**
Tests pass.
All scripts work.
A smurf told me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ pytx/docs/_build
 .DS_Store
 .tox/
 .coverage
+.pytx
--- a/pytx/docs/quickstart.rst
+++ b/pytx/docs/quickstart.rst
@@ -128,11 +128,9 @@ leveraging the classes you can do so:
 
 .. code-block :: python
 
-   from pytx import init
    from pytx.request import Broker
    from pytx.vocabulary import ThreatExchange as te
 
-   init()
    b = Broker()
    url = te.URL + te.THREAT_INDICATORS
    params = {te.TEXT: "www.facebook.com"}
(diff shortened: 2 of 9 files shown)


mgoffin:
I reviewed the changes prior to me losing Internet at home. I think this does exactly what we talked about and looks good! :+1:

jessek:
Protecting the :closed_lock_with_key: is very important and I appreciate you working to protect them!

(Yes, I'm playing with icons. :))

tiegz:
:egg: cellent

I almost did the same with [`globals.js`](https://github.com/facebook/ThreatExchange/blob/master/ui/js/globals.js#L6) but can't think of a way to avoid that one, besides remembering the creds in a cookie via JS. :\

tpott:
/malware/get_samples.py went down the route of using concatenated env variables. I can update it to use the same variable names that were added here!

PR #44, opened 2015-06-15
Improve docs for adding connections

hammem:
I added a concrete example of adding a connection between two indicator
nodes.


jessek:
Thank you!

PR #45, opened 2015-06-17
Adds docs for malware_family and more!

hammem:
Adding docs for the malware_family object and cleaned up some errors /
attributes we don’t support.


jessek:
This is great, thank you!

PR #52, opened 2015-07-02
node.js library for FBTE. Follow instructions in README.md

zmallen:
Things to note:

* As per discussion on the Facebook group, for any change in the endpoints/API etc., both the UI and the node app will have to be updated
* It would make sense to have 1 codebase that both the UI & node uses, but that will require dev time
* We could browserify this code and go that route, but I did not investigate this further

Once we have 1 codebase to work from, I can make the node code a thin wrapper over the shared codebase and make a P/R.


zmallen:
Also for npm integration, I made an npm account and once this is merged I can publish the code.

zmallen:
any update or feedback on this?

PR #53, opened 2015-07-07
Adds a limit parameter to get_indicators.py

tpott:
This can be used to increase batch size of requests


hammem:
Thanks, @tpott!

PR #55, opened 2015-07-15
Allow get_samples.py to run on Python 2.6

j-p-c:
Tested that it works in both Python 2.6 and 2.7 now.


hammem:
thanks, @j-p-c !

PR #12, opened 2015-05-12
Allow to query by threat_type

omercnet:
Added threat_type to objects query

PR #57, opened 2015-07-16
Upgraded PHP APIs to use v2.4 of ThreatExchange

hammem:
I updated the indicator search and upload code to use the new
descriptor objects that were included in v2.4

### DIFF
--- a/php/ThreatExchangeConfig.php
+++ b/php/ThreatExchangeConfig.php
@@ -23,7 +23,7 @@ class ThreatExchangeConfig {
   const APP_ID = 0;
   const APP_SECRET = 'secret';
 
-  const FACEBOOK_SERVER = 'https://graph.facebook.com';
+  const FACEBOOK_SERVER = 'https://graph.facebook.com/v2.4';
 
   public static function getAppID() {
     if (!self::APP_ID) {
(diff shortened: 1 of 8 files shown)


php/ThreatExchangeConfig.php
```diff
@@ -23,7 +23,7 @@ class ThreatExchangeConfig {
   const APP_ID = 0;
   const APP_SECRET = 'secret';
 
-  const FACEBOOK_SERVER = 'https://graph.facebook.com';
+  const FACEBOOK_SERVER = 'https://graph.facebook.com/v2.4';
```
jessek: Let's keep this explicit. Just in case there are changes in future versions, I like working code to stay on the current version. Although probably not a factor here, it's a good precedent to set.

hammem: I attempted that originally, but got this error when trying to use descriptors:

{"error":{"message":"(#12) Unable to display the requested object. requires version v2.4 or higher","type":"OAuthException","code":12}}`

jessek:
Oh hey: 

  An unversioned call will default to the oldest available version of the API. 

https://developers.facebook.com/docs/apps/versions

So I guess we need the explicit version.

hammem:
TIL

PR #62, opened 2015-07-23
Add precision and review_status to threat descriptor fields

jessek:
Adds documentation for the precision and review_status fields.


mgoffin:
What are the acceptable values for each?

jessek:
They are both enumerated values described at the bottom of the document.

PR #69, opened 2015-09-05
An attempt to get the error message values out of the json which is returned in the body of the 400.

fuzzball5000:
Currently, if a malware sample isn't found in the database for a given hash, the only message returned to the user is the fact an exception happened. So, grab the json returned with the error, parse out message and associated hash and print.

### DIFF
--- a/malware/get_samples.py
+++ b/malware/get_samples.py
@@ -71,14 +71,15 @@ def run_query(url):
     try:
         # the response is a 'data' object, which is a list of objects
         resp = json.loads(urlopen(url).read())['data'][0]
-    except Exception as e:
-        print('Failed to load md5: %s' % str(e))
-        try:
+    except urllib2.HTTPError as e:
+        if e.code==400:
             error_body = json.loads(e.read())['error']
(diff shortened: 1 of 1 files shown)


malware/get_samples.py
```diff
@@ -71,8 +71,15 @@ def run_query(url):
     try:
         # the response is a 'data' object, which is a list of objects
         resp = json.loads(urlopen(url).read())['data'][0]
-    except Exception as e:
-        print('Failed to load md5: %s' % str(e))
+    except urllib2.HTTPError as e:
+        if e.code==400:
+            error_body = json.loads(e.read())['error']
+            print ('Error message: ',error_body['message'],'for hash: ',error_body['error_data'])
```
jessek: Can you please use printf formatting like the statement below? (i.e. %s for the strings).

jessek:
This will be great to have, thank you!

fuzzball5000:
(i.e. %s for the strings). - Done, submitting another pull request. Also, split the catching to explicitly deal with a 400 and return a generic error for everything else. :)

PR #70, opened 2015-09-08
2.4 API support

mgoffin:
ThreatDescriptor is a new object to reflect the 2.4 API change.
ThreatIndicator has had its vocabulary stripped to reflect the 2.4 API
changes. MalwareFamily support has been added.

Vocabulary was updated with the latest API docs.

The `.save()` method on a cls_or_self has been changed. It now requires
a `params` argument which is a dictionary of all of the things you want
to submit. Given the differences between what you submit and how the
data is modeled in a class, we could no longer support `.save()` being
able to accurately determine how to convert the class into a "save or
update" request. You can now perform `.save()` on an uninstantiated
class for adding new objects to the graph, or `.save()` on an
instantiated class object to submit changes for that specific object.

There is now a `.get_changed()` method which will generate a dict of all
of the changed fields. This is useful for people performing updates and
want to get a "dirty dict" of all of the changes they've made to their
instantiated class.

API version has been locked to 2.4 as all of these changes are not
backwards-compatible with <2.4.

This could use some testing before acceptance. It's a bit of a change
from how pytx was used previously and I could not adequately test
the `.save()` changes without altering real data in the graph.


jessek:
Looks great, thank you!

PR #72, opened 2015-09-14
Fix examples to work with v2.4 and add new class methods.

mgoffin:
Updates the README to use v2.4 examples.

Adds `.new()` and `.send()` class methods to creating a new object or submitting requests to specific object type or details URLs.

`.save()` now only saves changes for an instantiated object. It requires parameters to be passed in, otherwise we will attempt to programmatically figure out what's changed in the object and submit those "as-is" and it is up to the developer to determine if that's what they actually want to happen.

Bumped pytx to version 0.2.0 because the above combined with ThreatDescriptors changes quite a lot of the pytx API and how it is used.

If someone can tag this we can release pytx 0.2.0 :)


jessek:
Sweet!

PR #74, opened 2015-09-15
Add page_limit for a maximum num of page results

RyPeck:
"limit - Defines the maximum size of a page of results. The maximum is 1,000/"
https://developers.facebook.com/docs/threat-exchange/reference/apis/threat-descriptors

The limit parameter when making a call directly to the Graph API limits
the number of results on a single page that is returned.

Limit as used in objects limits the total number of objects returned.

I've added a parameter to the `objects` method that allows you to set a
page result limit - `page_limit` - which will be the parameter passed to
`limit`. The original `limit` will continue to function as it had.

Shout out to everyone that maintains pytx - huge help!

### DIFF
--- a/pytx/pytx/access_token.py
+++ b/pytx/pytx/access_token.py
@@ -15,7 +15,7 @@ def _read_token_file(token_file):
     :param token_file: The full path and filename where to find the access token.
     :type token_file: str
     :returns: str
-    :raises: :class:`errors.pytxIniterror`    
+    :raises: :class:`errors.pytxIniterror`
     """
     try:
         with open(token_file, 'r') as infile:
--- a/pytx/pytx/common.py
+++ b/pytx/pytx/common.py
@@ -137,7 +137,7 @@ class Common(object):
     @classmethod
     def objects(cls, text=None, strict_text=False, type_=None, threat_type=None,
                 fields=None, limit=None, since=None, until=None, __raw__=None,
-                full_response=False, dict_generator=False):
+                page_limit=None, full_response=False, dict_generator=False):
         """
         Get objects from ThreatExchange.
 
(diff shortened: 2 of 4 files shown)


mgoffin:
Thanks for helping out with pytx :) In the Graph API as you noted there is a "limit" field. Am I understanding this right that all you are basically doing is renaming it to "page_limit" in the code and it eventually gets converted back to "limit" in the Broker for submission?

I only ask because I think "page_limit" seems more like "number of pages to return", "limit" sounds like "number of total objects to return", and neither sound like "number of objects per page to return" :) I am undecided as to what makes more sense, but I do think that it would make more sense for the argument to be named similar to the Graph API field; in this case: "limit". It makes going from Graph documentation to pytx implementation.

I'm open to other opinions on this so we can decide if the great work done here is the road we want to go, or if we should stay where we're at.

jessek:
I'm glad you are using and enjoying pytx!

I'm not sure I see the value in adding another name/layer for an existing term. I'd like to keep pytx as an interface to the graph, but not get too far away from the Graph API's terms.

mgoffin:
@jessek Sounds good to me! I say we keep it in line with the Graph API term then!

RyPeck:
Consensus looks to go with the Graph API term. I'll make that switch.

@mgoffin yes you have the correct understanding.

Should there still be functionality built in that is similar to what the `limit` parameter does now in the generator code?

IMHO there shouldn't - if you want to only get a certain number of objects, keep track of that on your own when you are iterating.

I'll close this and open another PR since for the modification.

mgoffin:
The generator works on the set of objects in the results and isn't bound by page limits. It will continue to grab more objects and iterate over the pages. It allows you to continue looping without having to deal with paging. It should stay very simple to allow developers to decide how and when to stop looping.

RyPeck:
@mgoffin yes - but my understanding with the current code is that what happens now is that the limit parameter is passed to the `get_generator` function as `total`. The for loop uses this to eventually stop returning result once the total has been reached.

Code refs -
https://github.com/facebook/ThreatExchange/blob/master/pytx/pytx/request.py#L284
https://github.com/facebook/ThreatExchange/blob/master/pytx/pytx/common.py#L192

mgoffin:
The goal here is that the limit you impose on the total number of results is what ends the generator loop (because you are done with the number of objects the server will give to you). If the limit is "number of results per page" then the generator would be pre-maturely ending after only the first page of results. That would obviously be bad!

PR #79, opened 2015-09-25
Add support for retries

mgoffin:
Retries in `requests` is a bit of a pain, but this should expose the ability for people to retry GET, POST, and DELETE requests as often as they feel they need to in order for the request to succeed. By default we still try and fail after one attempt. This propagates into the generator so for example if we perform a query which will end up with 15 pages and a retry of `3`, we will try to fetch each page as we need to and individually they can be retried up to 3 times before failing.

Failing here is defined by what requests fails on. We create a urllib3 Retry object and specify 500 and 503 responses as ones we will retry.

This PR also adds a logging feature (necessary if people are going to be doing retries and random stuff and want to inspect what pytx is doing/thinking). You can now do something like this:

```python
from pytx import setup_logger

setup_logger("/path/to/my/log/file")
```
After this, any pytx code will log to this file. Any pytx errors thrown will log there, and the generator now has logging to show you the current cursors, number of results in the `data` field, and will let you know if `next` is not in the pager (meaning the generator is over) or `paging` is not in the response (meaning something entirely different?).

I do not know if this will fix the issues people have mentioned about timeouts and putting things in while loops with try/except. If people can reliably generate those issues and can test with this PR to see if it fixes or helps, that would be great! If it does not (in the case where data gets submitted fine and the server returns a valid response with an internal timeout error) then we might have some other work to do but this PR could still be valid.

### DIFF
--- a/pytx/pytx/common.py
+++ b/pytx/pytx/common.py
@@ -309,7 +309,7 @@ class Common(object):
         )
 
     @classmethod
-    def new(cls, params):
+    def new(cls, params, retries=None):
         """
         Submit params to the graph to add an object. We will submit to the
         object URL used for creating new objects in the graph. When submitting
@@ -318,6 +318,8 @@ class Common(object):
 
         :param params: The parameters to submit.
         :type params: dict
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
         :returns: dict (using json.loads())
         """
 
@@ -328,9 +330,9 @@ class Common(object):
             if (params[td.PRIVACY_TYPE] != pt.VISIBLE and
                     len(params[td.PRIVACY_MEMBERS].split(',')) < 1):
                 raise pytxValueError('Must provide %s' % td.PRIVACY_MEMBERS)
-        return Broker.post(cls._URL, params=params)
+        return Broker.post(cls._URL, params=params, retries=retries)
 
-    def save(self, params=None):
+    def save(self, params=None, retries=None):
         """
         Submit changes to the graph to update an object. We will determine the
         Details URL and submit there (used for updating an existing object). If
@@ -339,12 +341,14 @@ class Common(object):
 
         :param params: The parameters to submit.
         :type params: dict
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
         :returns: dict (using json.loads())
         """
 
         if params is None:
             params = self.get_changed()
-        return Broker.post(self._DETAILS, params=params)
+        return Broker.post(self._DETAILS, params=params, retries=retries)
 
     @class_or_instance_method
     def send(cls_or_self, id_=None, params=None, type_=None, retries=None):
@@ -362,7 +366,7 @@ class Common(object):
         :type params: dict
         :param type_: GET or POST
         :type type_: str
-        :param retries: Number of retries to fetch a page before stopping.
+        :param retries: Number of retries to submit before stopping.
         :type retries: bool
 
         :returns: dict (using json.loads())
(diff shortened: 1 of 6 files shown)


pytx/pytx/common.py
```diff
@@ -231,6 +235,8 @@ def details(cls_or_self, id=None, fields=None, connection=None,
         :param dict_generator: Return a dictionary instead of an instantiated
                                object.
         :type dict_generator: bool
+        :param retries: Number of retries to fetch a page before stopping.
+        :type retries: bool
```
jessek: This should be int.

pytx/pytx/common.py
```diff
@@ -307,6 +316,8 @@ def new(cls, params):
 
         :param params: The parameters to submit.
         :type params: dict
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
```
jessek: This should be int.

jessek:
I'm not thrilled about there being so many errors in TX, but am hopeful this will help until we can get that sorted on our end.

There are two nits in the code and then I'll merge this.

pytx/pytx/common.py
```diff
@@ -166,6 +166,9 @@ def objects(cls, text=None, strict_text=False, type_=None, threat_type=None,
         :type full_response: bool
         :param dict_generator: Return a dictionary instead of an instantiated
                                object.
+        :type dict_generator: bool
+        :param retries: Number of retries to fetch a page before stopping.
+        :type retries: bool
```
omercnet: int?

pytx/pytx/common.py
```diff
@@ -351,6 +364,8 @@ def send(cls_or_self, id_=None, params=None, type_=None):
         :type params: dict
         :param type_: GET or POST
         :type type_: str
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
```
omercnet: int?

pytx/pytx/common.py
```diff
@@ -368,64 +383,74 @@ def send(cls_or_self, id_=None, params=None, type_=None):
...
 
-    def expire(self, timestamp):
+    def expire(self, timestamp, retries=None):
         """
         Expire by setting the 'expired_on' timestamp.
 
         :param timestamp: The timestamp to set for an expiration date.
         :type timestamp: str
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
```
omercnet: int? :)

pytx/pytx/common.py
```diff
@@ -368,64 +383,74 @@ def send(cls_or_self, id_=None, params=None, type_=None):
...
-    def false_positive(self, object_id):
+    def false_positive(self, object_id, retries=None):
         """
         Mark an object as a false positive by setting the status to
         UNKNOWN.
 
         :param object_id: The object-id of the object to mark.
         :type object_id: str
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
```
omercnet: s/retries: bool/retries: int/g

omercnet:
seems like a lot of bools instead of ints, or did I fail to understand how requests retries?

pytx/pytx/common.py
```diff
@@ -368,64 +383,74 @@ def send(cls_or_self, id_=None, params=None, type_=None):
...
 
-    def add_connection(self, object_id):
+    def add_connection(self, object_id, retries=None):
         """
         Use HTTP POST and add a connection between two objects.
 
         :param object_id: The other object-id in the connection.
         :type object_id: str
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
```
omercnet: int

pytx/pytx/common.py
```diff
@@ -368,64 +383,74 @@ def send(cls_or_self, id_=None, params=None, type_=None):
...
 
-    def delete_connection(self, object_id):
+    def delete_connection(self, object_id, retries=None):
         """
         Use HTTP DELETE and remove the connection to another object.
 
         :param object_id: The other object-id in the connection.
         :type object_id: str
+        :param retries: Number of retries to submit before stopping.
+        :type retries: bool
```
omercnet: last one

mgoffin:
That should do it. Not sure there's going to be anything else major like these updates. I think after this it would be ok for you to tag v0.2.0 so we can put the pytx release up on PyPi :)

omercnet:
I concur, we're way overdue for a release!

just one confirmation, I've been going over the python logging documentation and looks like FileHandler inherits StreamHandler, this means you do setup_logging() without any filename it'll default to sys.stderr, correct?

mgoffin:
If you do that, log_file will be None, and no logging will be setup.

PR #81, opened 2015-10-29
Change init() to access_token() and add proxies/verify support.

mgoffin:
The name init() didn't make much sense for what it did. It only sets up
your access token if you have a special requirement. setup_logger()
would still need to be run as an "initialization" step if you want to do
logging. So init() has been renamed to access_token() for people who
need to use it.

Folks complained that they need to make connections through proxies or
have special conditions where they need to set verify=False under
certain network conditions. This has been exposed in two ways:

    1) connection(): you can run this and pass a proxies and/or verify
    argument which will be used in all requests.
    2) you can pass a proxies and/or verify argument to class methods
    like .objects(), .details(), etc. which will be used.

    NOTE: passing an argument to class methods will override the global
    settings set by using connection(). This is intentional as it allows
    you to set a general configuration but modify it per-call if
    necessary.

Bumped the version to 0.3.0 due to the API change with init() being
renamed to access_token(). If accepted, someone will need to tag the
commit so the version will bump and we can release 0.3.0.

Also, if there is anyone who can actually test the proxies stuff that
would be great as I am only in a situation to test the verify piece.


brentonchang:
Thanks for this! Works great for my environment.

mgoffin:
The pending check won't clear because the original commit for the PR didn't fix the tests to use the new functions. I believe the tests should pass once accepted.

mgoffin:
Anything I can do to assist in getting this PR merged? After it is in we have some other things to fix (stuff for @9b at the very least).

PR #86, opened 2015-11-02
Added owner and status to the outgoing calls as outlined in the most …

### LINKED ISSUE #83: Unexpected keyword "status" when searching ThreatDescriptors
When searching for ThreatDescriptors using the "objects" interface in pytx, it does not seem to match the documentation displayed on the Facebook developers site - https://developers.facebook.com/docs/threat-exchange/reference/apis/threat-descriptors.

```
from pytx import init
from pytx import ThreatDescriptor
from pytx.vocabulary import Status

init(app_id='<app-id>', app_secret='<app-secret>')
results = ThreatDescriptor.objects(
    text='37.59.224.217',
    strict_text=True,
    status=Status.MALICIOUS
)
```

When running the code above, the following comes back:

```
  File "app/tools/threat_exchange.py", line 35, in <module>
    status=Status.MALICIOUS
TypeError: objects() got an unexpected keyword argument 'status'
```

9b:
Added the "owner" and "status" field as outlined in the latest documentation of ThreatExchange 2.5. Resolves #83 .

### DIFF
--- a/pytx/pytx/common.py
+++ b/pytx/pytx/common.py
@@ -138,7 +138,8 @@ class Common(object):
     @classmethod
     def objects(cls, text=None, strict_text=False, type_=None, threat_type=None,
                 fields=None, limit=None, since=None, until=None, __raw__=None,
-                full_response=False, dict_generator=False, retries=None):
+                owner=None, status=None, full_response=False, 
+                dict_generator=False, retries=None):
         """
         Get objects from ThreatExchange.
 
@@ -187,6 +188,8 @@ class Common(object):
                 limit=limit,
                 since=since,
                 until=until,
+                owner=owner,
+                status=status
             )
         if full_response:
             return Broker.get(cls._URL, params=params, retries=retries)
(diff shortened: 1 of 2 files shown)


mgoffin:
Thanks for noting these missing API parameters! Couple of things:

- docstrings need to be updated to note the new parameters for all modified methods.
- might be more clean to put the parameters before `__raw__` in `.objects()` (everything after and including __raw__ are arguments that don't map directly to a parameter). It would also make the arguments identical to the `build_get_parameters()` order.
- these changes won't actually do anything because the `build_get_parameters()` method doesn't actually check for them and add them to the `params` dictionary.

hammem:
Thanks, @9b !

mgoffin:
This shouldn't have been merged until the above comments were addressed.

hammem:
My fault, @mgoffin , I just reverted it.

mgoffin:
No worries! Thanks! :)

PR #91, opened 2015-11-10
Add owner and status to .objects() and bump to v2.5.

mgoffin:
This adds the ability to search on owner and status using .objects().
Now that v2.5 is out we are also bumping to that.

### DIFF
--- a/pytx/pytx/common.py
+++ b/pytx/pytx/common.py
@@ -137,9 +137,9 @@ class Common(object):
 
     @classmethod
     def objects(cls, text=None, strict_text=False, type_=None, threat_type=None,
-                fields=None, limit=None, since=None, until=None, __raw__=None,
-                full_response=False, dict_generator=False, retries=None,
-                proxies=None, verify=None):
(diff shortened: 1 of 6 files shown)


pytx/pytx/common.py
```diff
@@ -137,9 +137,9 @@ def to_dict(self):
 
     @classmethod
     def objects(cls, text=None, strict_text=False, type_=None, threat_type=None,
```
jessek: This is getting cluttered. Can you please put one argument per line?

pytx/pytx/request.py
```diff
@@ -150,7 +150,7 @@ def validate_get(cls, limit, since, until):
     @classmethod
     def build_get_parameters(cls, text=None, strict_text=None, type_=None,
```
jessek: This is getting cluttered. Can you please put one argument per line?

mgoffin:
Will do! I'll fix that up.

mgoffin:
Yay for passing tests and easier to read function arguments :) Thanks for pointing that out @jessek. It's a heck of a lot more readable now!

Issue #93, opened 2015-11-23
Failed malware analysis example on documentation


9b:
Attempting to run the example in the documentation (https://developers.facebook.com/docs/threat-exchange/reference/apis/threat-indicator/v2.5) fails inside the graph explorer.

https://graph.facebook.com/v2.4/768629009848617/malware_analyses/?access_token=555|aSdF123GhK

jessek:
Are you able to view 768629009848617 by itself?

9b:
I get the following when requesting it directly:

```
{
  "error": {
    "message": "Unsupported get request. Please read the Graph API documentation at https://developers.facebook.com/docs/graph-api",
    "type": "GraphMethodException",
    "code": 100,
    "fbtrace_id": "Bil+JpDLoV4"
  }
}
```

hammem:
Thanks, @9b.  It should be fixed now.

hammem:
Dah, I fixed a different issue in the docs :)  I'll fix this specific one as well.

hammem:
Ok, for real this time.  It's fixed.

PR #94, opened 2015-11-23
Capture error responses better

mgoffin:
I was annoyed duplicating #93 and having to manually parse the
response string which wasn't already JSON the way pytx was
returning.

When an error occurs, we capture the contents of it. We now build up a
dictionary containing all of the error contents from the server response
and we add in the `status_code` and the `url` that the request was made
to. This should make it much easier to parse out and automatically use
within scripts since it's already a dictionary, not a large string that
isn't in JSON.

Example:

```python
from pytx import Malware
from pytx.errors import pytxFetchError

m = Malware()
m.id = "19374-19841-4813-408"
response = None
try:
    m.details()
except pytxFetchError, e:
    response = e.message
```

The `response` dictionary looks like this:

```python
{
    'code': 803,
    'fbtrace_id': 'BTS22ImzDvd',
    'message': '(#803) Some of the aliases you requested do not exist: 19374-19841-4813-408',
    'status_code': 404,
    'type': 'OAuthException',
    'url': 'https://graph.facebook.com/v2.5/19374-19841-4813-408/?access_token=555|aSdF123GhK'
}
```

### DIFF
--- a/pytx/docs/quickstart.rst
+++ b/pytx/docs/quickstart.rst
@@ -204,3 +204,28 @@ The Broker will also allow you to POST and DELETE if you need to.
 One thing you might notice is the constant use of vocabulary. pytx comes with a
 vocabulary which will allow you to write your code using class attributes so if
 ThreatExchange ever changes a string your code will still function properly.
+
+Error responses can be acquired and leveraged as a dictionary. Here is an
+example:
+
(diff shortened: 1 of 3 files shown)


jessek:
MOVE FAST! (Wow, thank you!)

PR #100, opened 2015-12-27
Added Rails UI for ThreatExchange

skills4ever:
Added a Ruby on Rails UI for Facebook ThreatExchange. Allows user to easily query and share share threat indicators without any programming experience. Code was originally released at https://github.com/skills4ever/ThreatExchangeUI.


maus-:
Hey @skills4ever 
In its current state, we're unable to accept this pull request into our official repo. We noticed the ThreatExchange API tokens are stored in the user model. This raises security concerns if you're collecting a company's AppID/Secret or providing it to a third party hosting this app. We suggest moving the API token from the user model to the application config so someone deploying this WebUI can configure it to their organization. Ultimately, AD/LDAP integration would be even better. 

We also noticed you took code/patterns from the Ruby gem provided, but did not integrate it back into the library. Contributions are most helpful when they can be pushed back upstream to the gem, creating greater impact for the community and reducing duplicate work on the backend. If there is a feature or function you'd like to see, please add to the Ruby library and push it back to the upstream repo.

hammem:
@skills4ever , please let us know if you would like to address the feedback, otherwise I'll close this one out.

skills4ever:
@maus- @hammem , I stored the AppID and secret in the user model because I originally envisaged it as a multi-company platform. I will make it single company, and move the App ID and secret from the user model to the app config as an environment variable. Will it be acceptable to FB if I make these changes?

maus-:
@skills4ever 
We would also like you to utilize the Ruby ThreatExchange library in the app. This allows us to easily refactor the clients when we make back end changes. If there is an absent feature from the gem that you require I encourage you to add it and contribute it back upstream.

skills4ever:
K. Thanks @maus- My original intent was to make an application that someone with limited or no coding experience could easily download, setup and use to query or submit threat intel. It is really just a front end that ties into the Ruby library.

PR #101, opened 2015-12-29
pytx: lots of vocabulary updates.

mgoffin:
Keeping up with vocabulary changes. These were all found in the latest developer docs.

### DIFF
--- a/pytx/pytx/malware_family.py
+++ b/pytx/pytx/malware_family.py
@@ -20,7 +20,7 @@ class MalwareFamily(Common):
         mf.PRIVACY_TYPE,
         mf.SAMPLE_COUNT,
         mf.SHARE_LEVEL,
-        mf.SUBMITTER_COUNT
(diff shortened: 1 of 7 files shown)


pytx/pytx/malware_family.py
```diff
@@ -17,8 +17,10 @@ class MalwareFamily(Common):
         mf.FAMILY_TYPE,
         mf.MALICIOUS,
         mf.NAME,
+        mf.PRIVACY_TYPE,
         mf.SAMPLE_COUNT,
-        mf.SHARE_LEVEL
+        mf.SHARE_LEVEL,
+        mf.SUBMITTER_COUNT
```
jessek: Please add a trailing comma so that we can preserve blame. We haven't really modified the line above, but it's included in this commit because we're adding the trailing comma now.

mgoffin:
Fixed it in a couple places :) Thanks for catching!

PR #102, opened 2015-12-30
Adding sample_size and sample_size_compressed.

mgoffin:
These are new vocab terms for Malware which provide the size of the
sample and the size of the compressed sample.


jessek:
MOVE FAST!

PR #109, opened 2016-01-28
Adding bootstrapping scripts for all endpoints

motiwarifacebook:
Set your `TX_ACCESS_TOKEN` environment variable and run any of the python scripts in the get_all_data folder as follows:

`python <script_name> --help`

to get info on the options.

The script will then begin downloading the necessary data from ThreatExchange. For example:

`python get_threat_indicators.py --days_back 10`

will get all threat_indicators uploaded over the last 10 days and save them in 10 separate CSVs, one for each day


get_all_data/get_threat_descriptors.py
```diff
@@ -0,0 +1,131 @@
...
+                        lambda x: x if type(x) == int else
+                            (x.encode('utf-8') if x else ""),
+                        [
+                            r.get(TD.ID),
+                            r.get(TD.ADDED_ON),
+                            r.get(TD.CONFIDENCE),
+                            r.get(TD.DESCRIPTION),
+                            r.get(TD.EXPIRED_ON),
+                            r.get(TD.INDICATOR)['indicator'],
+                            r.get(TD.INDICATOR)['type'],
```
hammem: Use the vocab defined fields here.

motiwarifacebook:
Done, thanks for input

get_all_data/common_field_transforms.py
```diff
@@ -0,0 +1,40 @@
+import dateutil.parser
+import datetime
+
+def convert_to_header(field):
```
hammem: Should we move these functions into Pytx?  I'd imagine most anyone will want this functionality.

get_all_data/get_malware_analyses.py
```diff
@@ -0,0 +1,86 @@
...
+
+from pytx import Malware
+from pytx.vocabulary import Malware as MA
+
+import common_field_transforms as CFT
+from datetime import datetime
+
+def get_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-T", "--sample_type", help="The type of malware sample \
```
hammem: Add a -O  --owner_id search option.

get_all_data/get_malware_analyses.py
```diff
@@ -0,0 +1,86 @@
...
+    for day_counter in range(s.days_back):
+        until_param, until_param_string, since_param, since_param_string = \
+            CFT.get_time_params(s.end_date, day_counter, format_)
+
+        output_file = 'malware_analyses_' + since_param_string + '_to_' + \
+            until_param_string + '.csv'
+        with open(output_file,'wb') as fout:
+            writer = csv.writer(fout)
+
+            # TODO: Remove this once querying the fields related to a sample
```
hammem: make fetching the full sample a command line arg option as well, we can return an error if it's enabled for now.

get_all_data/get_malware_analyses.py
```diff
@@ -0,0 +1,86 @@
...
+                sample_type=s.sample_type,
+                share_level=s.share_level,
+                text=s.text,
+                status=s.status,
+                strict_text=s.strict_text,
+                since=since_param_string,
+                until=until_param_string,
+            )
+
+            fields_list = [
```
hammem: Is it worth making this subset a class method, getDefaultFields(), on pytx.vocabulary.Malware?

get_all_data/get_malware_families.py
```diff
@@ -0,0 +1,64 @@
...
+        until_param, until_param_string, since_param, since_param_string = \
+            CFT.get_time_params(s.end_date, day_counter, format_)
+
+        output_file = 'malware_families_' + since_param_string + '_to_' + \
+            until_param_string + '.csv'
+        with open(output_file,'wb') as fout:
+            writer = csv.writer(fout)
+            results = MalwareFamily.objects(
+                fields=MalwareFamily._fields,
+                limit=1000,
```
hammem: make this an option as well?

get_all_data/get_threat_descriptors.py
```diff
@@ -0,0 +1,103 @@
...
+                [TD.OWNER, TE.NAME],
+                [TD.OWNER, TE.EMAIL],
+                TD.PRECISION,
+                TD.RAW_INDICATOR,
+                TD.REVIEW_STATUS,
+                TD.SEVERITY,
+                TD.SHARE_LEVEL,
+                TD.STATUS,
+                TD.THREAT_TYPE,
+                TD.TYPE,
```
hammem: this will always be the same as indicator.type, no need to include it.

get_all_data/get_malware_analyses.py
mgoffin: You can already do `Malware._fields` to acquire a list of the available fields for a Malware object:

https://github.com/facebook/ThreatExchange/blob/master/pytx/pytx/malware.py#L16

This goes for any object you can query for :)

hammem: Yep, but what we have here is a subset (it basically is everything but getting the actual sample, since that can quickly become prohibitive).  And, I'd imagine it's a commonly desired subset.

hammem:
@motiwarifacebook, Thanks for working on this!

get_all_data/common_field_transforms.py
mgoffin: I think it would be a great idea to provide some common formatting and transformation functions that people can leverage.

get_all_data/get_malware_analyses.py
mgoffin: Ahh. I think that should be doable!

get_all_data/get_malware_families.py
motiwarifacebook: Are you suggesting we make the "limit" parameter an option? I think that is a bad idea. Most people will be using these scripts to get all the data for particular day(s) in TX. Allowing this to be anything less than 1000 would increase the number of roundtrip queries to TX and thus make the script take longer.

motiwarifacebook:
Note to self: Also have to add information about these scripts to the readme

hammem:
Thanks, @motiwarifacebook !

Issue #111, opened 2016-02-01
pytx support for batch queries


mgoffin:
The Graph API allows for this and right now the only way to do so is using the __raw__ argument to .objects() (I'm not sure if that works without also having to set full_response or dict_generator to True).

pytx should have a simple interface for making these types of requests.

motiwarifacebook:
Solved by https://github.com/facebook/ThreatExchange/pull/112

PR #112, opened 2016-02-01
Add batch request capability.

mgoffin:
:warning: DO NOT ACCEPT YET!

This adds the ability to submit batch requests to the Graph.

There is a new `Batch` class which has a `.submit()` classmethod. This
takes an arbitrary number of dictionaries as arguments. Each dictionary
should contain the method, URL (full or relative), and potentially a
body. For kwargs it has detailed in the docstrings that it takes similar
named arguments as methods like `.objects()` insofar as modifying how
the requests library submits your requests (modifying headers, proxy,
verify, retries). It also takes an `include_headers` named argument
which allows you to enable headers in the response.

Added a feature to common class methods like `.objects()` where you can
now pass a `request_dict` argument (boolean). If you do, instead of
actually submitting a request to the Graph, it will instead return a
dictionary which is formatted specifically for passing as an argument to
the `Batch.submit()` method. Example:

    from pytx import Batch, ThreatDescriptor

    print Batch.submit(ThreatDescriptor.objects(text='foo',
                                                request_dict=True),
                       ThreatDescriptor.objects(text='bar',
                                                request_dict=True))

Fixed a bug where "include_expired" was not converting to a string so
when it got encoded by requests it was stripping some of the value.

Bumped copyright year.

Bumped version to 0.4.0 as this is a larger addition to the library.

Related to #111


mgoffin:
@motiwarifacebook: I'll need your help testing this one. I want to make sure that this catches edge cases properly and it adheres to the request formats for submitting batch requests to the Graph. Would you be able to give it a test drive and see if it works properly?

pytx/pytx/batch.py
```diff
@@ -0,0 +1,85 @@
...
+            verify = kwargs.get('verify', None)
+
+            return Broker.post(t.URL,
+                               params=params,
+                               retries=retries,
+                               headers=headers,
+                               proxies=proxies,
+                               verify=verify)
+        except:
+            raise pytxFetchError('Error with batch request.')
```
motiwarifacebook: It's quite possible that a single request in the batch errors (ie returns something other than code 200). As written, the code would exit fine and provide a valid JSON with the error code embedded. I think we should error out if a single request errors, no?

mgoffin:
I thought that due to the responses being in a list and loop able by the client that they could acquire the successful information and handle the error responses individually. It would allow them to resubmit if they wish, or at least report back to whatever interface so the user knows what succeeded and what failed.

Or did I misinterpret how the API works? :)

motiwarifacebook:
Good point - your design decision makes more sense. Test driving now.

pytx/pytx/common.py
```diff
@@ -592,6 +617,10 @@ def send(cls_or_self,
         if params is None:
             params = {}
         if type_ == 'GET':
+            if request_dict:
+                return Broker.request_dict('GET',
+                                           url,
```
motiwarifacebook: cls._URL,?

pytx/pytx/common.py
```diff
@@ -599,6 +628,10 @@ def send(cls_or_self,
                               proxies=proxies,
                               verify=verify)
         else:
+            if request_dict:
+                return Broker.request_dict('POST',
+                                           url,
```
motiwarifacebook: cls._URL,?

pytx/pytx/batch.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Class for making Batch requests to the API.
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Initialized the object.
+        """
+
+    @classmethod
```
motiwarifacebook: I notice in a few places the use of @classmethod instead of @staticmethod. It doesn't appear that we use cls in this method's logic - should we make this and other similar instances @staticmethod?

motiwarifacebook:
Other than the comments I just made, everything looks good. Test driving now.

motiwarifacebook:
Oh and I see your point that its only for the initial POST, not for the individual batch requests

pytx/pytx/common.py
mgoffin: The `send()` method allows you to use it whether the object is instantiated or not. For example, both of these could work:

```python
ThreatDescriptor.send()
ThreatDescriptor().send()
```

In each case the URL is determined in a different way. In the former, the `cls._URL` is used for the URL, in the latter the `self._DETAILS` is used for the URL.

pytx/pytx/common.py
mgoffin: See above comment :)

pytx/pytx/batch.py
mgoffin: You are correct. In deciding between `@staticmethod` and `@classmethod`, I went with the latter because if we decide to add any default attributes to the `Batch()` class I wanted to be able to start using them where needed without having to go back and adjust all of the methods. Granted right now we don't have a ton of code, but I was thinking long-term :)

pytx/pytx/common.py
motiwarifacebook: Ah nice, I see the code now. Awesome.

pytx/pytx/batch.py
motiwarifacebook: That sounds good (y)

motiwarifacebook:
The following code gives an error (at the bottom): http://pastebin.com/pa9pMTBD. Do we support POSTs in Batch.submit? What about mixed GETs and POSTs?

1. I am confused as to the difference between send() and save() in common.py. Am I correct in that save() only functions on ThreatExchange OBJECTS, whereas send() applies to URLs we might want to send parameters to? But send() can also "save" if we POST certain params via send(), right?

2. How do I access a single node (like https://graph.facebook.com/v2.5/955977237771521) and create a ThreatDescriptor object out of it using pytx?

mgoffin:
When I read the Batch request API documentation (https://developers.facebook.com/docs/graph-api/making-multiple-requests) it looked like each batch request was its own dictionary. That dictionary had to contain a method (GET, POST, etc.), a URL (relative to https://graph.facebook.com), and an optional body (for POST parameters). So one batch request would look something like:

```python
{
    'type': 'GET',
    'url': 'threat_descriptors/?access_token=foo|bar&text=foo'
}
```

So there are two (bad example since we're not doing multiple requests) ways of doing this in pytx:

```python
from pytx import Batch, ThreatDescriptor

# using the classes (note the request_dict argument being set to True to create the dict)
print Batch.submit(ThreatDescriptor.objects(text='foo', request_dict=True))

# building your own custom dictionary
print Batch.submit({'type': 'GET', 'url': 'threat_descriptors/?access_token=foo|bar&text=foo'})
```

The former seemed a lot easier on the developer to leverage, but the latter allows some more general flexibility if you need to do something specific.


`send()` allows you to send parameters to the appropriate URL. By using the appropriate object (ThreatDescriptor, etc.) we just ensure the parameters provided are going to the appropriate Graph URL with your access_token appended. So if you use `ThreatDescriptor.send(type_='GET')` we will ensure it GETs from https://graph.facebook.com/threat_descriptors/, whereas if you use `ThreatDescriptor(id=123).send(type_='POST')` we will ensure it POSTs to https://graph.facebook.com/123/.

`save()` is used to update an existing ThreatExchange Object. Sure you can do the same thing with `send()` but instead of having to specify the type_, it will know it's a POST. It also requires an instantiated class object with a defined ID (to generate the `self._DETAILS` url). `.save()` will also try to unintelligently determine what's changed on the Object since it was instantiated and attempt to submit those changes as `params` if you don't provide `params` as an argument.

Accessing a single node can be done like this:

```python
from pytx import ThreatDescriptor

# fields argument can be omitted if you just want the default fields)
t = ThreatDescriptor(id='955977237771521').details(fields=ThreatDescriptor._fields)
```

Hope that answers everything! :)

mgoffin:
I made some code changes to allow for named arguments to `.submit()`. Example:

```python
Batch.submit(foo=ThreatDescriptor.objects(text='foo', request_dict=True)
```

When doing this, we will add `'name': 'foo'` to the request. This is noted in the FB documentation as being useful for chaining batch requests where one request relies on the results of another in the same Batch.

This was already able to be done by manually adding the `name` key to a custom dictionary, but this makes it a bit easier for the above example.

I also added support for `omit_response_on_success`.

motiwarifacebook:
This all looks great. After extensive testing, I am happy to approve this! Thanks for your patience @mgoffin, glad we could resolve all the (non-)issues ;)

PR #114, opened 2016-02-08
Moving utility functions to pytx

motiwarifacebook:
@mgoffin what do you think?

### DIFF
--- a/pytx/docs/pytx.utils.rst
+++ b/pytx/docs/pytx.utils.rst
@@ -1,0 +1,10 @@
+pytx.utils package
+=========================
+
+Module contents
+---------------
+
+.. automodule:: pytx.utils
+    :members:
+    :undoc-members:
+    :show-inheritance:
--- a/pytx/pytx/__init__.py
+++ b/pytx/pytx/__init__.py
@@ -1,4 +1,5 @@
 import access_token
+import utils
 from connection import connection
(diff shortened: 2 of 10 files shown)


mgoffin:
I think the code looks good.  However, we should probably add a `pytx.utils.rst` into the `docs` directory (copy any one of the ones that's there and there might be one or two changes in the file to make). Also, we might need to add this to `__init__.py` (similar to importing `access_token` in that file) so you can do this:

```python
from pytx.utils import get_time_param
```

If we have that then I think this would be good to commit :)

motiwarifacebook:
@mgoffin  I added your suggested changes and also added the _default_fields we discussed in https://github.com/facebook/ThreatExchange/pull/109

hammem:
@mgoffin / @motiwarifacebook - Let's bump the pytx version number, since we're adding a new dependency.  Unless you want to just wait until #112 gets merged in.

mgoffin:
Yea I think when #112 gets merged we can do a full 0.4.0 release which would include all of this stuff and a few PRs we accepted recently. Just waiting to ensure that #112 is doing things right and there's not any bells/whistles we want to include in it :)

But I think this looks good to go! Thanks @motiwarifacebook!

hammem:
The ayes have it, the motion passes.

PR #118, opened 2016-02-26
Fix docs to include latest stuff.

mgoffin:
Missing several files from the index causing them to be excluded from generated documentation.


motiwarifacebook:
utils was my mistake. Thanks for catching this!

PR #120, opened 2016-03-04
Add Attack Type field on ThreatDescriptor

hammem:
We recently added support for setting/getting an AttackType on a
descriptor.  It’s intent is to describe the class of attack an
indicator was associated with (e.g. Malware, Phishing, etc)

AttackType docs:
https://developers.facebook.com/docs/threat-exchange/reference/apis/attack-type/

### DIFF
--- a/pytx/pytx/threat_descriptor.py
+++ b/pytx/pytx/threat_descriptor.py
@@ -11,6 +11,7 @@ class ThreatDescriptor(Common):
 
     _fields = [
         td.ADDED_ON,
+        td.ATTACK_TYPE,
         td.CONFIDENCE,
         td.DESCRIPTION,
         td.EXPIRED_ON,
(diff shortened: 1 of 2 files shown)


mgoffin:
Excellent! Is the vocabulary located [here](https://github.com/facebook/ThreatExchange/blob/master/pytx/pytx/vocabulary.py#L303) accurately represent the available options?

hammem:
Yeah, I'm really glad we got this shipped out in the API.  I'm hoping it's the start of killing off THREAT_TYPE, which is a mashup of TYPE and ATTACK_TYPE right now...

mgoffin:
Oh interesting. We've found Threat Type in tandem with Attack Type to be very useful in uniquely identifying the use-case of an Indicator. I'll have to keep up on where that goes!

PR #122, opened 2016-03-16
[RFC] Support for fetching summarized descriptor data

hammem:
We recently added support for fetching ‘summary’ data from the
ThreatIndicator -> ThreatDescriptor connection in the ThreatExchange
RESTful APIs.

The summary object will return rolled up data for the descriptors
attached to a ThreatIndicator.  For now, there’s two summary fields:
- count (the number of descriptors tied to the indicator)
- preferred (when there are multiple descriptors, the one that is
  ‘preferred’)

Why do this?  When you encounter an indicator with multiple
descriptors, you have to do your own munging to handle it gracefully.
Rather than everyone have do the work on their own, the APIs offer a
way to do it for you (if you want it to).

But how does the underlying ThreatExchange API determine the
‘preferred’ one?

Through a set of lightweight, pre-defined heuristics, which can be
selected from in the API request.  Initially, the following heuristics
are supported:
- MINE_OR_NEWEST: the preferred descriptor should be ‘mine’, if I’ve
  submitted one.  failing that, return the newest published descriptor;
- NEWEST:  return the newest published descriptor;
- OLDEST:  return the oldest published descriptor;
- FROM_LIST_OR_NEWEST: the preferred descriptor should be one
  published by the member IDs from a provided list (order matters), if
  they’ve submitted one.  failing that, return the newest published
  descriptor.

We can support more heuristics in time, but started simple from the
methods we’d heard from those rolling their own logic.

There will also be support for summary data on other connections (e.g.
malware to indicators, malware family to variants, etc), so the support
is embedded in the Common connection() code in Pytx.


pytx/pytx/common.py
```diff
@@ -411,6 +413,8 @@ def connections(cls_or_self,
         :type verify: bool, str
         :param metadata: Get extra metadata in the response.
         :type metadata: bool
+        :param summary: Get a summarized view of the connection.
+        :type summary: bool
```
mgoffin: We will need to add `summary_param` below this so people will know what to pass in.

pytx/pytx/common.py
```diff
@@ -429,6 +433,15 @@ def connections(cls_or_self,
             params[t.FIELDS] = ','.join(f.strip() for f in fields)
         if metadata:
             params[t.METADATA] = 1
+        if summary and fields is not None:
+            params[t.FIELDS] = 'summary'
+        if summary:
+            params[t.FIELDS] = params[t.FIELDS] + ',summary'
```
mgoffin: This conditional will always overwrite the one above it. Should the one above it be `if summary and fields is None` and turn this one into an `elif summary`?

mgoffin:
Other than the few comments, without knowing what should be passed in as `summary_params` it looks ok. One question: would it be easier to just assume if the developer has set `summary_params` then they want a summary, otherwise they don't? It would save on the extra `summary` argument to `connections()`. Or use `summary` as the only argument and it contains the parameters being defined?

pytx/pytx/common.py
hammem: will do

pytx/pytx/common.py
hammem: yep, good catch.

hammem:
This is, likely, the first summarized edge we'll support.  This one has params that you can provide, but others may not, which is why I kept the two.  It's clumsy for this example, is there a better way ppl usually handle this?

mgoffin:
Interesting. I didn't think about that. I suspect there might be other ways but a "yes/no" with optional params seems legit?

hammem:
Given that we don't have one yet, I'll yank it and wait until we actually need it.

hammem:
Updated with @mgoffin's changes.

pytx/pytx/common.py
```diff
@@ -411,6 +412,12 @@ def connections(cls_or_self,
         :type verify: bool, str
         :param metadata: Get extra metadata in the response.
         :type metadata: bool
+
+        :param summary_params: Parameters to indicate a summary of the
+                               connections is desired and how to compute the
+                               summary.  Use pytx.vocabulary.SummaryConnection
+                               for the available params.
+        :type summary_params: map
```
mgoffin: `map` should be `dict` here I believe?

RyPeck:
A few heuristics suggestions -

- Highest confidence
- Highest review rating status type (manual review being the highest)
- Lowest or highest StatusType (based on a scale of NON_MALICIOUS, SUSPICIOUS and MALICIOUS).

hammem:
Thanks, @RyPeck .  For the last two, does having ties get broken by timestamp suffice?  That is, if there are two MALICIOUS descriptors, we just pick the newer of the two?

RyPeck:
@hammem that suffices. Sorry for the delayed response.

Dcallies:
Hey all, looks like this has staled out. I'm going to close this out unless you want to try and revive it!

Dcallies:
```
               ...
             ;::::;
           ;::::; :;
         ;:::::'   :;
        ;:::::;     ;.
       ,:::::'       ;           OOO\
       ::::::;       ;          OOOOO\
       ;:::::;       ;         OOOOOOOO
      ,;::::::;     ;'         / OOOOOOO
    ;:::::::::`. ,,,;.        /  / DOOOOOO
  .';:::::::::::::::::;,     /  /     DOOOO
 ,::::::;::::::;;;;::::;,   /  /        DOOO
;`::::::`'::::::;;;::::: ,#/  /          DOOO
:`:::::::`;::::::;;::: ;::#  /            DOOO
::`:::::::`;:::::::: ;::::# /              DOO
`:`:::::::`;:::::: ;::::::#/               DOO
 :::`:::::::`;; ;:::::::::##                OO
 ::::`:::::::`;::::::::;:::#                OO
 `:::::`::::::::::::;'`:;::#                O
  `:::::`::::::::;' /  / `:#
   ::::::`:::::;'  /  /   `#
```

PR #124, opened 2016-03-18
Fix ruby client pagination, and refactor.

tiegz:
Pagination in the ruby client was broken because it checked for a Symbol instead of String in the parsed JSON. A few other minor changes in the library.


hammem:
Thanks, @tiegz !

tiegz:
thanks for the quick merge! :sunglasses:

PR #126, opened 2016-03-22
Removing "submitter_count" from viable fields and vocabulary

mgoffin:
This was recently removed so I also bumped the bugfix version so we can put out a new release to compensate.


jessek:
Proper indenting makes me so happy. Thank you!

PR #131, opened 2016-04-14
Making pytx work with python >= 2.7.

mgoffin:
:warning: Testing and review needed!

These changes allow me to run my test script using python 2.7 and python 3.5.

This is my first run at making something work in both 2.7 and 3.x, so we should probably do some more testing/poking/researchy stuffs to make sure I am not enraging the python3 gods with what I've done (For example: the importing feels really strange going with relative instead of absolute).

Tox should now test with python 2.7 and 3.5 and Flake8 was happy with these changes.

Bumping version to 0.5.0. Developers "shouldn't" see any changes that would impact their code, but this opens up a can of worms in the python 3.x realm we now need to be able to deal with.


mgoffin:
As a head's up, checks will continue to fail because Travis doesn't support python3.5 natively (https://github.com/travis-ci/travis-ci/issues/4794). There's a bunch of hacks to get python 3.5 installed during the test process but I'd rather not rely on them if we don't have to.

wxsBSD:
FYI, I plan on looking into getting this working with Travis in early May. As you said, the importing feels strange but might actually be the right way to do it. I'll do more research when I'm back in a few weeks. :)

wxsBSD:
I'm inclined to just accept this but it will break out travis setup and I see no obvious way to fix that in travis-land. Is it possible to see if we are using python 3 and import differently if we are? If that's possible I'll accept this.

mgoffin:
I could drop the version down to 3.4 if they support that. If that works are we ok with saying that's "good enough"?

mgoffin:
Ugh, sorry about all of the silly little commits. Once I got travis to work it then revealed some issues with the test code. It was foreign to me how to fix and had to incrementally get there and test travis each time :(

wxsBSD:
Ten points for @mgoffin! Thanks again for all you do on this! :)

PR #132, opened 2016-05-13
Adding sort order and updating vocab.

mgoffin:
sort_order has been added to the API so that is now supported. Bumped
vocab to include API version 2.6 as well as additions for sort_order and
the change to Attack Types to keep things singular.

One thing to note is for the singular vocab changes. Historically we have said
if the vocab changes in this way we will keep the class attribute the same and
change the string so code doesn't break. However, we are bumping soon to a
pytx v0.5 release. If we wanted to adjust the class attribute to stay true to the
value, now would be the time to do it. I have left the attribute as-is in this PR 
but if we feel differently and want to adjust the attribute as well I can make a 
quick change and include it here too.

### DIFF
--- a/pytx/pytx/common.py
+++ b/pytx/pytx/common.py
@@ -245,7 +245,7 @@ class Common(object):
                 status=status,
                 review_status=review_status,
                 share_level=share_level,
-                sort_order=sort_order
+                sort_order=sort_order,
             )
         if request_dict:
             return Broker.request_dict('GET',
--- a/pytx/pytx/common.py
+++ b/pytx/pytx/common.py
@@ -153,6 +153,7 @@ class Common(object):
                 status=None,
                 review_status=None,
                 share_level=None,
+                sort_order=None,
                 __raw__=None,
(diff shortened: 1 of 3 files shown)


pytx/pytx/common.py
```diff
@@ -241,7 +244,8 @@ def objects(cls,
                 owner=owner,
                 status=status,
                 review_status=review_status,
-                share_level=share_level
+                share_level=share_level,
+                sort_order=sort_order
```
jessek: Can you please add a trailing comma? That will let us keep the blame correct. This PR doesn't really modify the share_level line above, for example.

pytx/pytx/vocabulary.py
```diff
@@ -5,7 +5,7 @@ class ThreatExchange(object):
     """
 
     URL = 'https://graph.facebook.com/'
-    VERSION = 'v2.5/'
+    VERSION = 'v2.6/'
```
motiwarifacebook: +1

PR #136, opened 2016-06-30
ruby: remove duplication in request methods, and make params uniform

tiegz:
Some ruby [client] refactoring:

* break out `get`, `post`, and `delete` http methods to reuse
* moves the exception-catching boilerplate into ^ methods
* swap `filter` and `data` naming with `params`, to make it clear we're talking about HTTP request params

### DIFF
--- a/ruby/lib/ThreatExchange.rb
+++ b/ruby/lib/ThreatExchange.rb
@@ -18,8 +18,6 @@ module ThreatExchange
     end
 
     def malware_analyses(params={})
-      params = params.merge(access_token: @access_token)
-
       result = get "#{@baseurl}/malware_analyses", params
       data = result['data']
       cursor = result.fetch('paging', {}).fetch('cursor', {}).fetch('after', nil)
@@ -38,8 +36,6 @@ module ThreatExchange
     end
 
     def threat_indicators(params={})
-      params = params.merge(access_token: @access_token)
-
       result = get "#{@baseurl}/threat_indicators", params
 
       data = result['data']
@@ -58,34 +54,26 @@ module ThreatExchange
     end
 
     def indicator_pq(params={})
-      params = params.merge(access_token: @access_token)
-
       get "#{@baseurl}/#{params[:id]}/", params
     end
 
     def members(params={})
-      params = params.merge(access_token: @access_token)
-
       get "#{@baseurl}/threat_exchange_members/", params
     end
 
     def new_relation(params={})
-      params = params.merge(access_token: @access_token)
       id = params.delete(:id)
 
       post "#{@baseurl}/#{id}/related", params
     end
 
     def remove_relation(params={})
-      params = params.merge(access_token: @access_token)
       id = params.delete(:id)
 
       delete "#{@baseurl}/#{id}/related/", params
     end
 
     def new_ioc(params={})
-      params = params.merge(access_token: @access_token)
-
       if params.has_key?(:privacy_type)
         post "#{@baseurl}/threat_indicators", params
       else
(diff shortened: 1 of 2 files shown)


ruby/lib/ThreatExchange.rb
```diff
@@ -17,27 +17,16 @@ def initialize(access_token=nil)
...
-      filter[:access_token] = @access_token
-      begin
-        response = RestClient.get "#{@baseurl}/malware_analyses",
-        { params: filter  }
-        result = JSON.parse(response)
-      rescue => e
-        puts e.inspect
-      end
+    def malware_analyses(params={})
+      params = params.merge(access_token: @access_token)
```
tiegz: also toyed w/the idea of moving this line from each method into the HTTP methods -- thoughts?

tiegz:
@mgoffin got more where this came from, once this is merged 😎

mgoffin:
Sounds good :) Although I don't maintain the Ruby stuff ;)

tiegz:
aha! That's the key, then I conjure @maus-

ruby/lib/ThreatExchange.rb
```diff
@@ -17,27 +17,14 @@ def initialize(access_token=nil)
...
       until cursor.nil?
-        filter[:after] = cursor
-        begin
-          response = RestClient.get "#{@baseurl}/malware_analyses",
-          { params: filter }
-          result = JSON.parse(response)
-        rescue => e
-          puts e.inspect
-        end
+        params[:after] = cursor
```
tomdz: I understand this was in here before, but it is generally considered bad style to change the input parameters. Would be better if this copies stuff, e.g.
```
query_params = (params || {}).merge(:after : cursor)
```

ruby/lib/ThreatExchange.rb
```diff
@@ -17,27 +17,14 @@ def initialize(access_token=nil)
...
+        result = get "#{@baseurl}/malware_analyses", params
```
tomdz: This style was there before, but for consistency would be better to use `()` in this method invocation similar to all the other method invocations in this code.

ruby/lib/ThreatExchange.rb
```diff
@@ -17,27 +17,14 @@ def initialize(access_token=nil)
...
-    def malware_analyses(filter={})
-      filter[:access_token] = @access_token
-      begin
-        response = RestClient.get "#{@baseurl}/malware_analyses",
-        { params: filter  }
-        result = JSON.parse(response)
-      rescue => e
-        puts e.inspect
-      end
+    def malware_analyses(params={})
```
tomdz: Not sure about the styleguide used here, but most Ruby styleguides prescribe spaces around `=` for the method arguments.

ruby/lib/ThreatExchange.rb
```diff
@@ -78,72 +53,62 @@ def threat_indicators(filter={})
...
-    def new_relation(data={})
-      data[:access_token] = @access_token
-      id = data.delete(:id)
-      begin
-        response = RestClient.post "#{@baseurl}/#{id}/related", data
-      rescue => e
-        e.inspect
-      end
+    def new_relation(params={})
+      id = params.delete(:id)
```
tomdz: Same as above, changing the passed in hash is bad form, especially for a library.

ruby/lib/ThreatExchange.rb
```diff
@@ -78,72 +53,62 @@ def threat_indicators(filter={})
...
-          response = RestClient.post "#{@baseurl}/threat_indicators", data
-          return response
-        rescue => e
-          e.inspect
-        end
+    def new_ioc(params={})
+      if params.has_key?(:privacy_type)
+        post "#{@baseurl}/threat_indicators", params
       else
         puts "You must set a privacy_type in your query"
```
tomdz: I understand this was here before, but would be great if this could be changed to raise an exception instead (e.g. `ArgumentError`).

ruby/lib/ThreatExchange.rb
```diff
@@ -78,72 +53,62 @@ def threat_indicators(filter={})
...
     end
 
+    ##
+    ## Request Methods
+    ##
+    private def get(url, params={})
+      response = RestClient.get url, params: params.merge(access_token: @access_token)
+      JSON.parse(response)
+    rescue => e
+      puts e.inspect
```
tomdz: This should use a logger, ideally one that can be passed in to the constructor.

ruby/lib/ThreatExchange.rb
```diff
@@ -78,72 +53,62 @@ def threat_indicators(filter={})
...
+       id = params.delete(:id)
+
+      result = post "#{@baseurl}/#{id}", params.merge(access_token: @access_token)
     end
 
+    ##
+    ## Request Methods
+    ##
+    private def get(url, params={})
+      response = RestClient.get url, params: params.merge(access_token: @access_token)
```
tomdz: `params` is not guaranteed to be non-`nil`. For instance, the user can explicitly pass in `nil`. Better than the default argument would be this:
```
query_params = (params || {}).merge(access_token: @access_token)
```

tiegz:
@tomdz thx for looking this over! I've implemented your suggestions, and I also have another changeset after this PR that cleans up the code more.

maus-:
Awesome - the refactoring looks good. Glad to see this is getting some polish. I don't think this is a result of your changes but we still have some neglected functionality. IOC Submissions / Updates seem to be throwing 400 errors. I'll take a closer look shortly here.

tiegz:
@maus- thx, I have a PR incoming for that next! (with updates to the example script, etc)

PR #140, opened 2016-07-12
ruby: reorganize lib, assert valid params, update api calls.

tiegz:
Followup to https://github.com/facebook/ThreatExchange/pull/136, with more updates to the ruby lib:

* version 0.0.9 ~> 0.0.10
* change client's constructor method signature
* updated ThreatExchangeTest.rb
* replace `new_ioc` with `new_descriptor` to reflect this API change
* rename `update_ioc` with `update_indicator` to be more consistent with API naming
* adds `member_threat_privacy_groups` method to get privacy groups of which you're a member
* adds `owned_threat_privacy_groups` method to get privacy groups which you own
* adds `new_threat_privacy_group` method to add a new privacy group
* adds `get_indicator` method to fetch an Indicator
* adds `malware_families` method to get groupings of malware

cc @bryongloden @maus- @tomdz


ruby/lib/ThreatExchange/version.rb
```diff
@@ -1,3 +1,3 @@
 module ThreatExchange
-  VERSION = "0.0.9"
+  VERSION = "0.0.10"
```
tiegz: wondering if we should update minor or even major instead, since the constructor signature changed

ruby/lib/ThreatExchange/http_methods.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module ThreatExchange
+  class Client
+    module HttpMethods
+      # Common GET Options:
+      #   limit - Defines the maximum size of a page of results. The maximum is 1,000.
+      #   since - Returns malware collected after a timestamp
+      #   until - Returns malware collected before a timestamp
+      #   fields - A list of fields to return in the response
+      #   sort_order - A given SortOrderType
```
tiegz: To avoid duplication in each of the methods I just pulled out these common params?

ruby/lib/ThreatExchange/version.rb
maus-: +1

maus-:
this is rad and works on my end. Thanks for your contributions @tiegz.

ruby/Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ThreatExchange (0.0.9)
+    ThreatExchange (0.0.10)
```
jessek: Does this need to match the 1.0.0 below?

tiegz: woops! Thanks @jessek

PR #141, opened 2016-07-20
Ruby client fix for assert_params()

tiegz:
My last PR added `ThreatExchange::Client#assert_params`, but I mixed up the calculation. This fixes it and gives the args a better name.


wxsBSD:
Thanks for the fix! :)

PR #145, opened 2016-07-29
pytx/docs Add reference to Facebook Docs

RyPeck:
Adding reference to the Facebook Documentation for making multiple
requests and using JSONPath to extract data from those requests.

Took me a bit to find these docs searching - thought a link would be
helpful for anyone else that comes along.


mgoffin:
:+1:

wxsBSD:
It got a 👍  from @mgoffin? That's always good enough for me! <3

RyPeck:
@mgoffin is updating readthedocs.io a manual step?

mgoffin:
At this time yes. They tried making a github hook to kick off the build but it never worked and I don't think we've looked at it any further. I'll kick one off now.

PR #149, opened 2016-10-05
Add full print usage message

divingboy:


php/upload.php
```diff
@@ -16,8 +16,8 @@
 ThreatExchangeConfig::init();
 
 // Get the command line options
-$options = getopt('d:f:g:hm:p:t:');
-if (!isset($options['t'])) {
+$options = getopt('d:f:g:h:m:p:t:');
```
jessek: I don't understand why you've made the -h flag require an argument here. Typo?

divingboy: Oh, I see, -h flag doesn't require an argument. I will fix it. Thanks!

PR #151, opened 2016-10-11
Add SSL/TLS support for the RTUListener.

mgoffin:
This adds support for SSL/TLS through Flask using the RTUListener. Developers can create their own custom `SSLContext` with whatever options they deem necessary for their use case and pass it into the RTUListener's `ssl_context` argument.

Also added documentation referencing the Whitelisting of Facebook IPs in the Webhook docs as it pertains to securing your RTU listener.

### DIFF
--- a/pytx/docs/quickstart.rst
+++ b/pytx/docs/quickstart.rst
@@ -297,3 +297,20 @@ build up the listener, specifying the host, port, URL suffix we plan on using,
 the custom GET response we configured for our Webhook, and the callback
 function. The custom GET response is necessary so ThreatExchange can validate
 the Webhook with your server. After that we start the listener. That's it!
+
+You can also create your own SSLContext to pass into the RTUListener's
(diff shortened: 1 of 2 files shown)


wxsBSD:
Does it make more sense to default to SSL with reasonable options and have a "NO_SSL" flag here instead? Just trying to point out that we should encourage SSL by default unless explicitly asked.

mgoffin:
We could, but we would have to provide our own cert chain and I suspect most people would prefer to generate/provide that on their own given the issues that come along with using someone else's pre-generated pem and key file.

wxsBSD:
Good point, we certainly don't want to be doing that. I was just thinking it may be worth it to take the hit of having an extra step (generate a keypair) just to make sure things are secure by default. However, since this isn't meant to be what people use for production, but instead just for testing that it's working, I think this may be fine to accept as is.

PR #152, opened 2016-10-21
Support for 2.8 changes for tags and query sorting.

mgoffin:
Update vocabulary to match 2.8 changes. Allow for the 'tagged_objects'
connection for ThreatTag. Add the 'sort_by' capability to `.objects()`
for sorting by RELEVANCE or CREATE_TIME.

This should get us to where we need to be to push a 0.5.3 release.


pytx/pytx/vocabulary.py
```diff
@@ -5,7 +5,7 @@ class ThreatExchange(object):
     """
 
     URL = 'https://graph.facebook.com/'
-    VERSION = 'v2.6/'
+    VERSION = 'v2.8/'
```
gsingh93: This is technically a breaking change, because v2.8 doesn't allow users to set the attack type and threat type. If we're going to do this, maybe we should wait until the 60 day time limit we gave people is up? And should we also bump the major version number since it's a breaking change? Not sure what versioning scheme we've decided to follow.

mgoffin: This was something I was considering while developing this PR. I had intentions of removing `attack_type` and `threat_type` from the code base as well as bumping to `0.6.0` and updating the documentation.

However, after looking at the `2.8` developer documentation I noticed that `attack_type` and `threat_type` are still referenced as valid fields. Also, when submitting queries, you can still request and see the values for those fields in the response data. As a side note didn't see any `tags` for the returned objects which I was expecting to at least have the `attack_type` and `threat_type` there as part of the migration effort to deprecate the `attack_type` and `threat_type`.

I'm open to changing the code to better reflect how we expect the API to be used, just let me know!

mgoffin: Correct. When they go to POST they'll get whatever the response is from TX (success or failure). Since Descriptors are returning those fields still in `2.8` it's necessary to keep the class attributes so there's no failures building the object.

I also figured leaving the vocabulary would be useful to keep. Even if it's not tied to a specific object, they might be good reference material for tags.

motiwarifacebook:
Thanks @mgoffin !

PR #155, opened 2016-11-22
Add a new script for dumping out a malware family from ThreatExchange

hammem:
This adds a new script for quickly dumping out the members of a malware
family in ThreatExchange: as CSV of the variant hashes or the variants
themselves.

### DIFF
--- a/pytx/scripts/malware_family_grabber.py
+++ b/pytx/scripts/malware_family_grabber.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python
+
+import argparse
+import base64
+import cStringIO
+import csv
+import os
+import zipfile
+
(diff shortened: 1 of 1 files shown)


mgoffin:
Nice! I noticed in your save_file you do a bunch of stuff to get to the binary. The Malware class has methods for accessing the binary in several states. It might help out with making that section of code simpler.

hammem:
Thanks, @mgoffin , for reminding me.  I'll do that.

I'd like to specify the file name, for consistency.  Mind if I also update `Malware.zfh()` to support passing in a file name and maintain the existing logic as the default?

mgoffin:
Seems reasonable to me!

PR #156, opened 2016-11-22
New Malware.rfh() hotness

hammem:
Allow you to specify the filename as a param to `Malware.rfh()`


mgoffin:
:+1:

hammem:
It's a holiday, no one is around.  As @mgoffin as my witness, I'm totally self-committing.

mgoffin:
Roger that!

PR #157, opened 2017-01-04
Remove threat type and attack type from pytx

jessek:
As noted in https://www.facebook.com/groups/threatex/permalink/573588456165790/, we have removed threat type and attack type from ThreatExchange. These options are no longer supported and attempting to use them results in an error. As such, this PR removes them from pytx.

### DIFF
--- a/pytx/README.rst
+++ b/pytx/README.rst
@@ -44,7 +44,7 @@ Quick Example
    access_token('<app-id>', '<app-secret>')
    results = ThreatDescriptor.objects(text='www.facebook.com')
    for result in results:
-       print result.get(td.THREAT_TYPE)
+       print result.get(td.CONFIDENCE)
 
    # type is type_ because type is a reserved word.
    results = ThreatDescriptor.objects(type_='IP_ADDRESS',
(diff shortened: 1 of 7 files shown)


mgoffin:
This looks good. One thing I might throw out there and see what you think: should we remove the vocabulary? It might be useful for folks looking to describe an attack or threat via tags to have some references they can use. It would slightly help with the consistency issue if the recommended tag list have some representation in the vocabulary.

I'm not suggesting they stay as they are,  just that having something might be helpful.

RyPeck:
@mgoffin I like that suggestion. I would add that perhaps the most popular tags (some of which are carried over from attack types) as documented could be included.

Bottom of - https://developers.facebook.com/docs/threat-exchange/reference/apis/threattags/v2.8.

jessek:
If we're going to add new code, I'd prefer to do it in another PR. Is it ok to land this one as it is, or is it a must have to add the popular tags?

mgoffin:
That's fine. I was only suggesting that if we thought keeping a vocabulary for tags was a good idea, that we not remove the vocabulary classes yet and just change them over in another PR.

PR #159, opened 2017-02-06
Adjusted default fields for main objects.

### LINKED ISSUE #158: pytx should set sane defaults for fields returned and use them by default
Each main class has an `_default_fields` attribute which is a list of fields that are supposed to be "default fields returned by ThreatExchange when you perform a query."

We should change this so that `_default_fields` is a list of "useful" fields determined by those that use the data. Once those lists are fixed up, we should change `.objects()` to use what is in `cls._default_fields` if the `fields` argument is still `None`. This will allow people to override and specify the fields they want to use, otherwise they will get what pytx considers the default ones.

mgoffin:
Default fields will now be more sane with providing useful information.
If you do not provide the `fields` argument, we will use the
`_default_fields`...by default...

Updated some documentation bits and the Copyright year.

Added a PopularTags class to vocabulary for people who want to leverage
some of the more popularly-used tags in ThreatExchange.

Fixed a bug introduced in a recent commit which attempted to add a
provided filename while reading from a zip handler. Since the filename
you provide might not be one of the filenames in the zip handler, this
will break as it's meant for reading an existing filename, not writing
to a new filename.

Bumping to 0.5.4.

Closes #158.


wxsBSD:
<3

wxsBSD:
Looks like the 0.5.4 release still reports to be 0.5.3 because it wasn't bumped in https://github.com/facebook/ThreatExchange/blob/master/pytx/pytx/__init__.py. Any chance we can get that bumped to either 0.5.4 and re-rolled on pypi or just go with 0.5.5 everywhere?

PR #160, opened 2017-02-15
Add new "similar_malware" edge.

mgoffin:
There is a new edge for malware called "similar_malware". This adds it to the vocabulary and as a viable connection.


wxsBSD:
Thanks for bumping to 0.5.5 too! :)

mgoffin:
I knew there was a reason for me to do so ;) Plus PyPi's policy is to not allow you to re-upload the same version of your project.

PR #164, opened 2017-03-29
Fix Python 3 compatability

seanthegeek:
The ``basestring`` builtin type was removed in Python 3.


mgoffin:
I think this should be fine. If the Travis check comes back clean :)

PR #167, opened 2017-04-24
Consolidated python scripts into pytx/scripts/

mfvonh:
Consolidated functionality from various scripts

Updated API endpoint to v2.9


pytx/scripts/get_data.py
```diff
@@ -0,0 +1,264 @@
...
+                parameters = dict(
+                    fields=MalwareFamily._fields,
+                    limit=result_limit,
+                    text=args.text,
+                    strict_text=args.strict_text,
+                    since=since_str,
+                    until=until_str
+                )
+
+            elif args.object == 'threat_description':
```
mgoffin: Should we change this to "threat_descriptor" so it aligns with how the other object types are used?

mgoffin:
This is great! I will note that we should probably bump the pytx version (in all of the necessary places) so that this can be published to PyPi and people can start using both the scripts and the new API endpoint.

pytx/scripts/get_data.py
mfvonh: Agreed, fixed

PR #172, opened 2017-09-22
Support ThreatTag Filters in Search Parameters

smrojas:
This enables the tags vocabulary to be used in malware/descriptor filtering. By adding the such a filter (i.e. tags="malicious_ip,malware"), the search results will be filtered to the the data which contains any of the tags in the list.


hammem:
thanks, @smrojas !

mgoffin:
This doesn't include any of the bumps to version so I can't push this to PyPi. Now that it's merged we can always create another PR for bumping it.

PR #186, opened 2019-08-08
add some initial tests for ruby client

gregors:
Whaaaat? No tests for the ruby client?

* updated rest-client to recent version
* add minitest
* add webmock
* add basic tests for `#members` and `#malware_analyses`
* add to README

I really just wanted a starting point to get some basic test coverage. There are some things I'd like to clean up in client, but not without coverage first.

The updated Gemfile shows 1.0.0 => 1.0.1 It was and is 1.0.1 The last update didn't check in the updated Gemfile.

### DIFF
--- a/ruby/Gemfile.lock
+++ b/ruby/Gemfile.lock
@@ -1,37 +1,53 @@
 PATH
   remote: .
   specs:
-    ThreatExchange (1.0.0)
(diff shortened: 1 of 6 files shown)


johnkerl:
Thank you!!!

PR #190, opened 2019-08-09
Switch to pagewise-streaming model for bulk download

johnkerl:
Previously a query would get the list of all queried-for descriptor IDs, then go back through and pull details for each. This was a simple coding model but it had the expense of high latency from start of query to arrival of first detail. Switching to a callback model makes the code a bit less obvious but means that we start retrieving details almost immediately.

Test scenario involves running the commands in `hashing/te-tag-query-java/README.md`.

### DIFF
--- a/hashing/te-tag-query-java/README.md
+++ b/hashing/te-tag-query-java/README.md
@@ -37,7 +37,7 @@ java com.facebook.threatexchange.TETagQuery tag-to-details \
   --page-size 10 \
   --hash-dir ./tmk-hash-dir \
   --hash-type tmk \
-  media_priority_test
+  media_type_long_hash_video
 ```
 
 # Contact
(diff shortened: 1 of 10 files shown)


johnkerl:
Test plan for 4th commit:

```
$ echo 2290659920988866 | java com/facebook/threatexchange/TETagQuery -a TX_ACCESS_TOKEN_MHST -s -v ids-to-details 2290659920988866
TETagQuery ids-to-details: extraneous arguments.
IDs must be provided on standard input, one per line.
Usage: TETagQuery ids-to-details [options]
Options:
--no-print-hash -- Don't print the hash to the terminal
--hash-dir {d} -- Write hashes as {ID}.{extension} files in directory named {d}
Please supply IDs one line at a time on standard input.

$ echo 2290659920988866 | java com/facebook/threatexchange/TETagQuery -a TX_ACCESS_TOKEN_MHST -s -v ids-to-details
(... processes without error)
```

PR #199, opened 2019-08-30
tmk Makefile support for query with faiss.

bbhopesh:
Added minimal support for faiss in Makefil which seems to be working for me. Please take a look.

### DIFF
--- a/hashing/tmk/cpp/Makefile
+++ b/hashing/tmk/cpp/Makefile
@@ -52,6 +52,7 @@ EXE= \
  tmk-show-pair-offsets \
  tmk-clusterize \
  tmk-query \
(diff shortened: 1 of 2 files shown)


johnkerl:
Thank you!! There may be differences for different platforms, e.g. this working on Linux but not Mac for example, but we can deal with those as a follow-on.

bbhopesh:
Thanks. I have no way to check on Mac OS. I don't own one. Hopefully, some else can contribute that.
 Another thing, my changes need faiss already installed. I forgot to add this to README file. I will add this and submit another PR.

johnkerl:
Definitely. A coworker and I did some MacOS work  We can follow up no problem. Thanks!!

PR #202, opened 2019-08-31
Update readme files to include tmk-query-with-faiss.

bbhopesh:
Updating README files. A followup of #199.


johnkerl:
Thank you!! :)

PR #226, opened 2020-04-19
Fix for not being able to use filehasher in a loop.

bbhopesh:
There are known [problems](https://unix.stackexchange.com/questions/36310/strange-errors-when-using-ffmpeg-in-a-loop) with using `ffmpeg` in a bash loop. As, `tmk-hash-video` binary uses `ffmpeg`, I couldn't use it in a bash script loop. This PR is a fix for that.


bbhopesh:
@johnkerl Can you please take a look at this?

johnkerl:
Hi @bbhopesh -- thanks! This looks fine as long as it doesn't break reading from `stdin` on `popen` (which isn't really `stdin` I suppose). Let me check it out ...

johnkerl:
Looks good to me -- `make` in `tmk/hashing/cpp/` passes regression.

PR #229, opened 2020-04-30
Include more fields in the output: review_status, description, last_updated

johnkerl:
Testing: re-ran all the commands at https://github.com/facebook/ThreatExchange/blob/master/hashing/te-tag-query-java/README.md


johnkerl:
Thanks @BarrettOlson ! :)

PR #230, opened 2020-04-30
rename descriptor-filter to indicator-type-filter

johnkerl:
since it operates on data before all fields have been read.


johnkerl:
Thanks @mounikabodapati !

PR #231, opened 2020-04-30
--since/--until -> --tagged-since/--tagged-until for query by tag

johnkerl:


johnkerl:
Thanks @mounikabodapati !

PR #233, opened 2020-05-05
Added some bare-curl examples

johnkerl:
Typing out bare URLs is error-prone -- which is one of the main reasons I created `TETaqQuery.java` in the first place.

That said, to get some bare-curl documentation in place, we can leverage `TETagQuery.java` to produce the bare URLs and show their JSON responses.

This should be helpful to people wanting to port this code to other languages/contexts.


johnkerl:
You can see what the rendered markdown looks like at

https://github.com/johnkerl/ThreatExchange/blob/master/hashing/te-tag-query-java/README.md

johnkerl:
Thanks @BarrettOlson !

PR #237, opened 2020-05-12
te-tag-query-ruby

johnkerl:
Verificaiton: run all the commands in `te-tag-query-ruby/README.rb`


johnkerl:
Thanks @BarrettOlson ! Ruby-wise I think I'm it on the team ;). I'll let the test plan serve for remaining verification. :)

PR #244, opened 2020-06-04
Retry transient 500s on the Python side

johnkerl:
I tried running all of `te-tag-query-ruby tag-to-details media_type_video` and `te-tag-query-python tag-to-details media_type_video` and realized the Python netlib is apparently more brittle with respect to transient 500s than Ruby's or Java's. Mean time to failure is around every 30K/50K descriptors retrieved so this issue only arises on get-all-the-data pulls of bigger datasets.


johnkerl:
TYVM! :)

PR #245, opened 2020-06-09
Python/Ruby created-at filters

johnkerl:
Details/rationale in comments.

Tested via

```
te-tag-query-python tag-to-details \
  --created-since 2020-06-01T00:00:00Z \
  --created-until 2020-06-03T00:00:00Z \
  tag-name-goes-here \
| jq .added_on

te-tag-query-python tag-to-details \
  --created-since -2weeks \
  --created-until -1week \
  tag-name-goes-here \
| jq .added_on
```

and checking timestamps are in the desired range.

### DIFF
--- a/hashing/te-tag-query-python/TE.py
+++ b/hashing/te-tag-query-python/TE.py
@@ -10,6 +10,8 @@ import urllib.request
 import urllib.error
 import json
 import copy
+import datetime
+import re
 
 # ================================================================
 # HTTP-wrapper methods
@@ -159,9 +161,9 @@ class Net:
       "&limit=" + str(pageSize)
 
     if taggedSince != None:
-      startURL += "&tagged_since=" + taggedSince
+      startURL += "&tagged_since=" + urllib.parse.quote(taggedSince)
     if taggedUntil != None:
-      startURL += "&tagged_until=" + taggedUntil
+      startURL += "&tagged_until=" + urllib.parse.quote(taggedUntil)
 
     nextURL = startURL
     pageIndex = 0;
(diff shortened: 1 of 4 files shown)


hashing/te-tag-query-python/TETagQuery.py
```diff
@@ -460,6 +480,36 @@ def handle(self, args, options):
...
+          (self.progName, self.verbName))
+        sys.exit(1)
+      options['taggedSince'] = options['createdSince']
+      options['createdSinceEpochSeconds'] = TE.Net.parseTimeStringToEpochSeconds(options['createdSince'])
+    if options.get('createdUntil') != None:
+      if options.get('taggedUntil') != None:
+        eprint("%s %s: Please specify at most one of --tagged-until and --created-until." %
+          (self.progName, self.verbName))
+        sys.exit(1)
+      options['taggedUntil'] = None
```
BarrettOlson, line 510: doesn't this already have to be None based on the conditional above it?

hashing/te-tag-query-python/TE.py
```diff
@@ -159,9 +161,9 @@ def processDescriptorIDsByTagID(self, tagID, idProcessorCallback, **kwargs):
       "&limit=" + str(pageSize)
 
     if taggedSince != None:
-      startURL += "&tagged_since=" + taggedSince
+      startURL += "&tagged_since=" + urllib.parse.quote(taggedSince)
     if taggedUntil != None:
-      startURL += "&tagged_until=" + taggedUntil
+      startURL += "&tagged_until=" + urllib.parse.quote(taggedUntil)
```
BarrettOlson, line 166: nit: it might be worth a comment noting that if 'taggedUntil' is None it is treated as 'now' (if my understanding is correct that is)

johnkerl: Thanks!

PR #250, opened 2020-06-17
Improve indicator-type filtering for the Java impl

johnkerl:
This design has always been three-pass:

* Tag text to tag ID
* Tag ID to descriptor ID/indicator-text (only)
* Descriptor ID to full descriptor details

The `--indicator-type` filtering has all along been on pass 2, trying to draw inferences on the indicator text without access to the type. That worked OK for a small list of types; for larger lists of types, we simply need to jettison the clever inferencing from stage 2 and replace it with direct type-field checking in stage 3 (where there is a type field available).

Testing: run the commands in the `README.md` file.


hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
```diff
@@ -334,20 +328,6 @@ public void handle(
           pageSize = Integer.valueOf(args[0]);
           args = Arrays.copyOfRange(args, 1, args.length);
 
-        } else if (option.equals("--indicator-type")) {
```
jeberl, line 337: Should we make "--indicator-type" a constant?

johnkerl: we could? all the flags are literals & we could constantize them? personally i prefer to constantize things which appear more often in a method; "rule of three" -- ? :)

johnkerl:
Thanks @jeberl !!

hashing/te-tag-query-java/com/facebook/threatexchange/IndicatorTypeFilterer.java
```diff
@@ -0,0 +1,144 @@
...
+
+  // The uppercase names are the ones coming back in the 'type' field in TE API
+  // requests. The lowercase names are standard shorthands.
+  //
+  // Note: this data structure Could be a hashmap-literal but the syntax gets
+  // neater in Java 9 & let's keep this code backward-compatible with various
+  // Java versions.
+  private static final String[][] NAMES = {
+    {"ADJUST_TOKEN", "adjustToken"},
+    {"API_KEY", "api_key"},
```
Dcallies, line 24: In the comment, it says that lower case names are "standard shorthands"

How do you decide whether to camelCase or lower_with_underscore names?

johnkerl: @Dcallies thanks! These were taking from a TE-internal hashmap. For purposes of this program, though, they're solely conveniences and there is no need to preserve compatibility with the source enum. We can change them, for sure ...

PR #252, opened 2020-06-30
[te-tag-query][py] Add __name__ == '__main__'

Dcallies:
This PR adds a python convention for files which contains the main, so that they can be loaded without executing the main if loaded by another library. More details can be found here: https://stackoverflow.com/questions/419163/what-does-if-name-main-do

Test plan:
` ./te-tag-query-python --help
Usage: ./TETagQuery.py [options] {verb} {verb arguments}
Downloads descriptors in bulk from ThreatExchange, given
...
`

### DIFF
--- a/hashing/te-tag-query-python/TETagQuery.py
+++ b/hashing/te-tag-query-python/TETagQuery.py
@@ -1162,4 +1162,5 @@ class CopyHandler(AbstractPostSubcommandHandler):
 
 # Rememmber that sys.argv includes the program name in Python,
 # like C/C++/Go and unlike Java or Ruby.
-MainHandler(sys.argv[0]).handle(sys.argv[1:])
+if __name__ == '__main__':
+  MainHandler(sys.argv[0]).handle(sys.argv[1:])


johnkerl:
@Dcallies this is awesome!! Can you please do the same for the Ruby impl? (Sibling directory, very similar.)

Dcallies:
I don't have much Ruby experience, I googled for an equivalent and found: https://stackoverflow.com/questions/2249310/if-name-main-equivalent-in-ruby

Which mentions that the pattern isn't common, but it exists:

`if __FILE__ == $0`

Is this what you want done?

johnkerl:
@Dcallies looks great! I realized I should have pointed you to https://github.com/facebook/ThreatExchange/blob/master/hashing/te-tag-query-ruby/TENet.rb#L686

johnkerl:
You can go ahead and do a second commit on top of the current one and `git push`, and that'll show up here in this PR automatically. :)

johnkerl:
Will do Ruby version on a separate PR

PR #253, opened 2020-06-30
Make Ruby impl 2.0-compatible

johnkerl:
Initially developed on 2.6. Recently run on a 2.0 system.

A few things:

* `require "cgi"` not `require "CGI"`
* `[...].filter` is a new alias for the older `[...].select`
* `|` in access-token fields needs explicit `CGI::escape` in order to build a `URI` object
* `:,` kwarg syntax didn't work in 2.0
* `:` as dict delimiters results in symbol-valued keys in 2.6, syntax error in 2.0. We should always use `=>`.

Testing: run all the commands in the `README.md`

Note: My restack-fu isn't nearly as strong in Git as in Mercurial and as a result there is some in-progress Java work mixed in here too.

### DIFF
--- a/hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
+++ b/hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
@@ -870,6 +870,9 @@ public class TETagQuery {
       if (args.length != 1) {
         usage(1);
       }
+      if (args[0].equals("-h") || args[0].equals("--help")) {
+        usage(0);
+      }
       String startURL = args[0];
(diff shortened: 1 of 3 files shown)


johnkerl:
If it's easier I can cancel this PR and submit a clean Ruby-only one on a separate feature branch. Didn't mean to mix up the Java refactor stuff in with this. :^/

johnkerl:
I'll make a new one on a separate branch as noted above.

johnkerl:
https://github.com/facebook/ThreatExchange/pull/254 instead

PR #256, opened 2020-07-01
AbstractPostSubcommandHandler refactor

johnkerl:
Step 1 in a sequence of PRs to add some recent Ruby-port/Java-port features to the Java port.

This commit moves some submit-handler/update-handler logic to a shared parent class so that it will be easier to reuse code for a new copy-handler on a subsequent PR.

Testing: run all the commands in the `README.md`.

### DIFF
--- a/hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
+++ b/hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
@@ -1154,9 +1154,6 @@ public class TETagQuery {
           dryRun = true;
 
         } else if (option.equals("-I")) {
-          if (args.size() < 1) {
-            usage(1);
-          }
           indicatorTextFromStdin = true;
         } else if (option.equals("-i") || option.equals("--indicator")) {
           if (args.size() < 1) {
--- a/hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
+++ b/hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
@@ -904,66 +904,235 @@ public class TETagQuery {
   }
 
   // ----------------------------------------------------------------
-  // NOTE: SubmitHandler and UpdateHandler have a lot of the same code but also
-  // several differences. I found it simpler (albeit more verbose) to duplicate
-  // rather than do an abstract-and-override refactor.
+  // Some code-reuse for all subcommand handlers that do POSTs.
+  public static abstract class AbstractPostSubcommandHandler extends CommandHandler {
+    public static final String POSTER_NAME_SUBMIT = "submit";
+    public static final String POSTER_NAME_UPDATE = "update";
 
-  public static class SubmitHandler extends CommandHandler {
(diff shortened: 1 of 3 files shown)


johnkerl:
Will do Ruby equiv on a separate PR

hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
```diff
@@ -984,138 +1154,46 @@ public void handle(
           dryRun = true;
 
         } else if (option.equals("-I")) {
-          if (args.length < 1) {
+          if (args.size() < 1) {
```
bodnarbm, line 1157: Nit: Can be addressed here or in separate PR.

This looks to be an existing issue, but I think this check is not right. `-I` does not itself require any additional arguments. I think we should remove this if check here.

hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
```diff
@@ -904,78 +904,248 @@ public void handle(
...
       int numIDsPerQuery,
       boolean verbose,
       boolean showURLs,
       DescriptorFormatter descriptorFormatter
     ) {
       boolean dryRun = false;
       boolean indicatorTextFromStdin = false;
-      DescriptorPostParameters  params = new DescriptorPostParameters();
+      DescriptorPostParameters postParams = new DescriptorPostParameters();
+      ArrayList<String> args = new ArrayList<String>(Arrays.asList(argsAsArray));
```
bodnarbm, line 1144: Non-blocking

Would using an Iterator in place of a list here be helpful? I'm thinking it could help clean up some of the code that consumes additional arguments while parsing through the command line options be using `hasNext()` and `next()` instead of modifying the array list.

hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
```diff
@@ -904,78 +904,248 @@ public void handle(
...
+    public AbstractPostSubcommandHandler(String verb) {
+      super(verb);
+    }
+
+    protected void commonPosterUsage(int exitCode, String posterName) {
+      PrintStream o = Utils.getPrintStream(exitCode);
+
+      o.printf("Usage: %s %s [options]\n", PROGNAME, this._verb);
+
+      if (posterName.equals(POSTER_NAME_SUBMIT)) {
```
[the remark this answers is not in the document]
johnkerl: nice

hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
johnkerl: good catch, thanks!

hashing/te-tag-query-java/com/facebook/threatexchange/TETagQuery.java
johnkerl: hmm, food for thought. seems kind of six of one, half a dozen of the other ...

PR #263, opened 2020-08-04
[OpenSource] Create dummy package for all-in-one py code

Dcallies:
This PR creates an empty package that will hold the all-in-one scripts.

I went back and forth on various names. I think the best thing to do is to use the existing naming convention from the pytx package in this same repo. 

I copied a lot of the conventions from pytx, and hopefully got all the references.

Test Plan:
1. Set up a python virtual environment
```
$ source ~/.venv/threatexchange_install/bin/activate
```
2. Install from testpypi
```
$ python3 -m pip install --index-url https://test.pypi.org/simple/ pytx3
```
3. Verify it's there
```
$ python3
>>> import pytx3
```


bodnarbm:
@Dcallies This looks fine to me. Tested install locally and it worked to install the skeleton. Only thing I wanted to add to the basic setup is ensuring we are ignoring the build artifacts in git. Added that as #264.

PR #273, opened 2020-08-10
[pytx] Make more config fields optional

Dcallies:
Summary:
This diff removes the content_types field (didn't get used),
and also makes all fields except labels optional.

This makes the smallest functional config something like:
```
{"labels": ["media_priority_samples"]}
```

Test Plan: Used the config from the summary, successfully fetched.

### DIFF
--- a/pytx3/pytx3/collab_config.py
+++ b/pytx3/pytx3/collab_config.py
@@ -18,20 +18,17 @@ class CollaborationConfig:
     KEY_PRIVACY_GROUPS = "privacy_groups"
     KEY_LABELS = "labels"
     KEY_SAMPLE_TAG = "sample_tag"
-    KEY_CONTENT_TYPES = "content_types"
 
     def __init__(
         self,
         name: str,
         labels: t.Dict[str, t.Any],
         privacy_groups: t.List[int],
-        content_types: t.Dict[str, t.Dict[str, t.Any]],
         sample_tag: t.Optional[str] = None,
     ):
         self.name = name
         self.privacy_groups = privacy_groups
         self.labels = labels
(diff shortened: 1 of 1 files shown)


pytx3/pytx3/collab_config.py
```diff
@@ -42,10 +39,9 @@ def default_state_dir_name(self) -> str:
     def load(cls, file: t.IO):
         content = json.load(file)
         return cls(
-            content[cls.KEY_NAME],
+            content.get(cls.KEY_NAME, "Sample Config"),
             content[cls.KEY_LABELS],
```
bodnarbm, line 43: Since labels is required in the config and not the other fields, do we want to raise an exception here with a descriptive message that the config must include the labels information? This is what the stack trace looks like right now.

```
$ pytx3 fetch 
  File "<clip>/bin/pytx3", line 11, in <module>
    load_entry_point('pytx3', 'console_scripts', 'pytx3')()
  File "<clip>/pytx3/pytx3/cli/main.py", line 137, in main
    execute_command(namespace)
  File "<clip>/pytx3/pytx3/cli/main.py", line 69, in execute_command
    cfg = init_config_file(namespace.config)
  File "<clip>/pytx3/pytx3/cli/main.py", line 124, in init_config_file
    return CollaborationConfig.load(f)
  File "<clip>/pytx3/pytx3/collab_config.py", line 43, in load
    content[cls.KEY_LABELS],
KeyError: 'labels'
```

Dcallies: Will punt for a future change, I haven't figured out what the flow for creating/editing configs should be.

PR #277, opened 2020-08-24
[pytx3] Implement simple normalization + distance on text matching

Dcallies:
Summary:
Text matching has been a problem for partners, and we had been
telling them "just use levenshtein distance with 95% match", based on
what would map closest to Facebook's systems. Since then, we realized
there was actually a significant amount of normalization happening first.

We sent partners some rough ideas, and there are plenty of articles on
the internet (and plenty of python libraries) that do some kinds of
normalization and distance.

My goals are the following:
1. The suggested normalization should be simple enough that someone
could easily port it to another language from reading this code.
2. Take on the minimum amount of dependency complexity
3. The highest possible incremental value over raw string comparison.
4. Spend less than 3 hours implementing it

This is attempt at fulfilling those goals.

Test Plan:
I potentially need help coming up with a benchmarking set here
(preferably one I can put in the open source repo or talk about
publicly).

I added some debugging printouts with distance, which are not
in the final version.
```
$ cat ~/samples/raw_text.te
The quick brown fox jumps over the lazy dog, ...
"We the People of the United States, in Order to form a more perfect
 Union, establish Justice, ensure domestic Tranquility, provide for the
 common defence, promote the general Welfare, and secure the Blessings of
 Liberty to ourselves and our Posterity, do ordain and establish this
 Constitution for the United States of
 America.", ...
bball now?, ...

$ pytx3 match text "cdallies5"
DCallies5 - 0 cdallies5
  * >26 thequickbr
  * >259 wethepeopl
  * 6 bballnow

$  pytx3 match text "a bb áll no w"
a bb áll n - 0 abballnow
  * >26 thequickbr
  * >259 wethepeopl
  * 1 bballnow

$ pytx3 match text -T  "We the people of the United States, in order to
form a more perfect union, establish justice, insure domestic
tranquility, provide for the common defense, promote the general
welfare, and secure the blessings of liberty to ourselves and our
posterity, do ordain and establish this Constitution for the United
States of America."
We the peo - 13 wethepeopl
  * >233 thequickbr
  * 2 wethepeopl
  * >260 bballnow
3428971913788104 raw_text media_priority_samples

# Against a test dataset with 50Mb of 30k samples
time pytx3 -c ~/5mb_30k_samples.te.cfg match text <a text within average size of text>

2829833393779198 raw_text media_priority_test_dataset true_positive

real    0m33.456s
user    0m32.942s
sys     0m0.507s
```

### DIFF
--- a/pytx3/pytx3/signal_type/raw_text.py
+++ b/pytx3/pytx3/signal_type/raw_text.py
@@ -5,8 +5,13 @@
 Wrapper around the raw text signal type.
 """
 
-import typing as t
+import math
 import pathlib
+import re
+import typing as t
+import unicodedata
+
+import Levenshtein
 
 from ..descriptor import SimpleDescriptorRollup, ThreatDescriptor
 from . import signal_base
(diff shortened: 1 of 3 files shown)


pytx3/pytx3/signal_type/raw_text.py
```diff
@@ -24,5 +29,69 @@ class RawTextSignal(signal_base.SimpleSignalType, signal_base.StrMatcher):
...
     TYPE_TAG = "media_type_text"
 
+    def __init__(self) -> None:
+        super().__init__()
+        self.normal_to_raw = {}
+
     def match(self, content: str) -> t.List[signal_base.SignalMatch]:
         return self.match_hash(content)
+
+    def match_hash(self, signal_str: str) -> t.List[signal_base.SignalMatch]:
```
jeberl, line 39: why separate these functions? If the purpose is it is easy to remove the normalization why not add an optional `normalize : bool` argument?

pytx3/pytx3/signal_type/raw_text.py
```diff
@@ -24,5 +29,69 @@ class RawTextSignal(signal_base.SimpleSignalType, signal_base.StrMatcher):
...
+        normalized_str = self._normalize_string(signal_str)
+        # Match considered if 95% match
```
jeberl, line 41: Should we make 95% a constant?

jeberl:
Some small comments but also wanted to make sure you had considered all the the normalization techniques mark proposed. If so I'm good with this merge!

pytx3/pytx3/signal_type/raw_text.py
Dcallies: Eh, I prefer not to constant-tize if only used in one place in the code.

In the long term, I think we'll likely made this configurable in some way, but I'm not sure how yet (I originally had an entry in the collab config for it).

pytx3/pytx3/signal_type/raw_text.py
Dcallies: Check out the SignalBase, StrMatcher, and the `match` command to see how match and match_hash are used differently. I'm not sure if the way it's currently used is the best final form of it, but these two methods are separate for a reason.

Dcallies:
We discussed the normalization techniques offline, and why which ones were included or not. Some will likely be added as followups.

PR #283, opened 2020-08-26
remove legacy prototype/proof of concept for UI

jennylin:
full details of current TE UI are already instead at https://developers.facebook.com/docs/threat-exchange/ui


jeberl:
LGTM! 

Merge it

PR #284, opened 2020-08-26
[api-examples] Move node to api-examples

Dcallies:
Summary: Node appears to be javascript, so make a new directory for
js and move it there with `git mv`

Test Plan: /shrug


jeberl:
This looks good! I think the summary/test plan comment is a useful habbit 

Merge away!

PR #286, opened 2020-08-26
move te-tag-query examples into api-reference-examples

jennylin:
* created new folders for each language that didn't have it in examples
* moved examples from hashing to their appropriate examples folder


Dcallies:
lgtm!

PR #287, opened 2020-08-26
Move /hashing/tmk => /tmk AND /hashing/pdq => /pdq

BarrettOlson:
Installing the things I need to run Make for both right now. I will update here once I am able to run the tests


BarrettOlson:
### Test Plan
Run make for each:
```
cd pdq/cpp/
make

...

ALL REGRESSION TESTS PASSED
Tests completed: 41
```
```
cd tmk/cpp
make

...

REGRESSION PASSED
```

bodnarbm:
I do not know why the PR looks like this with the extra commits and files being affected. But when I pull this down locally the diff looks fine to me. So this should still be okay to merge.

PR #288, opened 2020-08-26
Fix go packaging to be under a module

bodnarbm:
Go files need to be packaged under a directory that is used for the import module name when used.


bodnarbm:
I had to test locally using the existing directory structure just to confirm that `go get` worked with

```
go get github.com/Facebook/ThreatExchange/api-reference-examples/go
```

Once this merges, we will be able to do the 

```
go get github.com/Facebook/ThreatExchange/api-reference-examples/go/threatexchange
```

bodnarbm:
Confirmed this after merge with

```
go get -v github.com/Facebook/ThreatExchange/api-reference-examples/go/threatexchange
```

Output:
```
github.com/Facebook/ThreatExchange (download)
```

PR #289, opened 2020-08-26
[pytx3] Fix merge conflict bug in fetch

Dcallies:
Summary: In a previous stack, I had a nasty conflict that I appear to
have fixed incorrectly. This re-adds the code that drains the rest of
the queue at the end of tagging.

Test Plan: pytx3 fetch, gets data instead of erroring.


Dcallies:
Well, if anything else is broken, it will be hard to spot until we figure out unittests.

PR #291, opened 2020-08-26
[pytx3] Refresh readme

Dcallies:
There's more to be said here, but this an incremental improvement.


BarrettOlson:
LGTM - incremental goodness

PR #292, opened 2020-08-26
Adding github action for Go API Example

bodnarbm:
This adds a github action to run go tests on the go example in the api-references-example folder. This actually caught an issue with our existing code that is fixed in this PR also.

Test Plan:

Action should run on this PR and we should see a green checkmark when it is completed in the below status checks.

### DIFF
--- a/.github/workflows/golang-ci.yaml
+++ b/.github/workflows/golang-ci.yaml
@@ -1,0 +1,31 @@
+name: API Reference Examples CI - Go
+on:
+  push:
+    branches:
+    - master
+    paths:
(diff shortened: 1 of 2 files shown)


Dcallies:
image

BarrettOlson:
I saw, `apiVersion = "v2.5"` in the go code, do we want to update that?

bodnarbm:
> I saw, `apiVersion = "v2.5"` in the go code, do we want to update that?

@BarrettOlson, most likely. I will take a further look at the go code and will try to update it in a later PR.

PR #294, opened 2020-08-27
[go] fix json unmarshalling for expired_on for ThreatDescriptor

### LINKED ISSUE #196: Error unmarshaling json, ThreatDescriptor.expired_on
https://github.com/facebook/ThreatExchange/blob/06806faa683a14417e31d93657ed3bc33b10a460/golang/threatexchange/threatexchange.go#L97

ExpiredOn should be a string instead of an int, otherwise i'm getting unmarshaling errors.

`cannot unmarshal string into Go struct field ThreatDescriptor.expired_on of type int`

the expired_on fields are now string dates `"expired_on":"2019-09-20T00:00:00+0000"`

bodnarbm:
Facebook graph returns date by default in a string representation
using the ISO-8601 format. For now, we can fix unmarshalling the json
by using a string instead of an int.

Fixes #196

Test Plan:

Tested this locally by setting the environment variables TX_APP_ID and TX_APP_SECRET for a test application and then executed the following in the go implementation folder

```
$ go test -v -run TestQueryThreatDescriptorsOfIpsProxy
=== RUN   TestQueryThreatDescriptorsOfIpsProxy
--- PASS: TestQueryThreatDescriptorsOfIpsProxy (0.77s)
=== RUN   TestQueryThreatDescriptorsOfIpsProxyByFBAdminOwner
--- PASS: TestQueryThreatDescriptorsOfIpsProxyByFBAdminOwner (1.19s)
PASS
ok      _/Users/bodnarbm/github/ThreatExchange/api-reference-examples/go/threatexchange 2.116s
```

### DIFF
--- a/api-reference-examples/go/threatexchange/threatexchange_test.go
+++ b/api-reference-examples/go/threatexchange/threatexchange_test.go
@@ -3,12 +3,13 @@ package threatexchange
 
 import (
 	"github.com/stretchr/testify/assert"
+	"os"
 	"testing"
 )
 
-const (
-	appId     = "" // Fill this to make tests run
-	appSecret = "" // Fill this to make tests run
+var (
+	appId     = os.Getenv("TX_APP_ID")     // Set this for integration tests
+	appSecret = os.Getenv("TX_APP_SECRET") // Set this for integration tests
 )
 
 func TestQueryThreatDescriptorsOfIpsProxy(t *testing.T) {
--- a/api-reference-examples/go/threatexchange/threatexchange.go
+++ b/api-reference-examples/go/threatexchange/threatexchange.go
@@ -95,7 +95,7 @@ type ThreatDescriptor struct {
 	ShareLevel   string          `json:"share_level,omitempty"`
 	AddedOn      string          `json:"added_on,omitempty"`
 	Confidence   int             `json:"confidence,omitempty"`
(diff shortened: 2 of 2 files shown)


api-reference-examples/go/threatexchange/threatexchange_test.go
```diff
@@ -3,12 +3,13 @@ package threatexchange
...
 	"github.com/stretchr/testify/assert"
+	"os"
 	"testing"
 )
 
-const (
-	appId     = "" // Fill this to make tests run
-	appSecret = "" // Fill this to make tests run
+var (
+	appId     = os.Getenv("TX_APP_ID")     // Set this for integration tests
```
bodnarbm, line 11: It looks like all these tests are actually integration tests requiring hitting the live api which has test data that can change. We should look at moving these to actual unit tests that don't rely on the live system.

BarrettOlson: Agreed, right now how does this test working non-locally without these being set?

bodnarbm: Basically, all the tests below skip if these are not set (that was the behavior from before with the empty string), and the overall test run passes solely do to no failures.

So the current test runs in CI are basically just checks on if it compiles.

PR #300, opened 2020-10-21
[pytx3] add expiremental_fetch command

InderH:
Create a new fetch2 command that downloads threat data via the new threat_updates endpoint.

### DIFF
--- a/pytx3/pytx3/TE.py
+++ b/pytx3/pytx3/TE.py
@@ -181,27 +181,31 @@ class Net:
     # Gets threat updates for the given privacy group.
     @classmethod
     def getThreatUpdates(self, privacy_group, **kwargs):
-        if 'next' in kwargs:
-            url = kwargs['next']
+        if "next" in kwargs:
+            url = kwargs["next"]
         else:
             url = (
-                self.TE_BASE_URL + '/'
+                self.TE_BASE_URL
+                + "/"
                 + str(privacy_group)
                 + "/threat_updates/"
                 + "?access_token="
                 + self.APP_TOKEN
-                + '&fields=id,indicator,type,creation_time,last_updated,is_expired,expire_time,tags,status,applications_with_opinions'
+                + "&fields=id,indicator,type,creation_time,last_updated,is_expired,expire_time,tags,status,applications_with_opinions"
             )
             for arg in kwargs:
                 if kwargs.get(arg) is not None:
-                    if arg == 'additional_tags':
-                        url = url + '&additional_tags=' + ",".join([str(app) for app in kwargs[arg]])
-                    elif arg == 'threat_type':
-                        url = url + '&threat_types=' + ",".join(kwargs[arg])
+                    if arg == "additional_tags":
+                        url = (
+                            url
+                            + "&additional_tags="
+                            + ",".join([str(app) for app in kwargs[arg]])
+                        )
+                    elif arg == "threat_type":
+                        url = url + "&threat_types=" + ",".join(kwargs[arg])
                     else:
-                        url = url + '&' + arg + '=' + str(kwargs[arg])
+                        url = url + "&" + arg + "=" + str(kwargs[arg])
         return self.getJSONFromURL(url)
-
 
     # ----------------------------------------------------------------
     # Returns error message or None.
--- a/pytx3/pytx3/TE.py
+++ b/pytx3/pytx3/TE.py
@@ -178,6 +178,32 @@ class Net:
         return descriptors
 
     # ----------------------------------------------------------------
+    # Gets threat updates for the given privacy group.
+    @classmethod
+    def getThreatUpdates(self, privacy_group, **kwargs):
+        if 'next' in kwargs:
+            url = kwargs['next']
+        else:
+            url = (
+                self.TE_BASE_URL + '/'
+                + str(privacy_group)
+                + "/threat_updates/"
+                + "?access_token="
+                + self.APP_TOKEN
+                + '&fields=id,indicator,type,creation_time,last_updated,is_expired,expire_time,tags,status,applications_with_opinions'
+            )
+            for arg in kwargs:
+                if kwargs.get(arg) is not None:
+                    if arg == 'additional_tags':
+                        url = url + '&additional_tags=' + ",".join([str(app) for app in kwargs[arg]])
+                    elif arg == 'threat_type':
+                        url = url + '&threat_types=' + ",".join(kwargs[arg])
+                    else:
+                        url = url + '&' + arg + '=' + str(kwargs[arg])
+        return self.getJSONFromURL(url)
+
+
+    # ----------------------------------------------------------------
     # Returns error message or None.
     # This simply checks to see (client-side) if required fields aren't provided.
     @classmethod
--- a/pytx3/pytx3/TE.py
+++ b/pytx3/pytx3/TE.py
@@ -193,18 +193,16 @@ class Net:
                 + self.APP_TOKEN
                 + "&fields=id,indicator,type,creation_time,last_updated,is_expired,expire_time,tags,status,applications_with_opinions"
             )
-            for arg in kwargs:
-                if kwargs.get(arg) is not None:
+            for arg, value in kwargs.items():
+                if value is not None:
                     if arg == "additional_tags":
-                        url = (
-                            url
-                            + "&additional_tags="
-                            + ",".join([str(app) for app in kwargs[arg]])
+                        url += "&additional_tags=" + ",".join(
+                            [str(app) for app in value]
                         )
                     elif arg == "threat_type":
-                        url = url + "&threat_types=" + ",".join(kwargs[arg])
+                        url += "&threat_types=" + ",".join(value)
                     else:
-                        url = url + "&" + arg + "=" + str(kwargs[arg])
+                        url += "&" + arg + "=" + str(value)
         return self.getJSONFromURL(url)
 
     # ----------------------------------------------------------------
--- a/pytx3/pytx3/cli/experimental_fetch.py
+++ b/pytx3/pytx3/cli/experimental_fetch.py
@@ -11,8 +11,10 @@ from ..indicator import ThreatIndicator
 from ..content_type import meta
 from . import command_base
 
-class Fetch2Command(command_base.Command):
+class ExperimentalFetchCommand(command_base.Command):
     """
+    WARNING: This is experimental, you probably want to use "Fetch" instead.
+
     Download content from ThreatExchange to disk.
 
     Using the CollaborationConfig, identify ThreatPrivacyGroup that
--- a/pytx3/pytx3/cli/experimental_fetch.py
+++ b/pytx3/pytx3/cli/experimental_fetch.py
@@ -11,6 +11,7 @@ from ..indicator import ThreatIndicator
 from ..content_type import meta
 from . import command_base
 
+
 class ExperimentalFetchCommand(command_base.Command):
     """
     WARNING: This is experimental, you probably want to use "Fetch" instead.
@@ -20,6 +21,7 @@ class ExperimentalFetchCommand(command_base.Command):
     Using the CollaborationConfig, identify ThreatPrivacyGroup that
     corresponds to a single collaboration and fetch related threat updates.
     """
+
     PROGRESS_PRINT_INTERVAL_SEC = 30
 
     @classmethod
@@ -64,15 +66,18 @@ class ExperimentalFetchCommand(command_base.Command):
         self.threat_types = threat_types
         self.additional_tags = additional_tags
         self.signal_types_by_name = {
-            name: signal()
-            for name, signal in meta.get_signal_types_by_name().items()
+            name: signal() for name, signal in meta.get_signal_types_by_name().items()
         }
         self.last_update_printed = 0
         self.counts = collections.Counter()
 
     def execute(self, dataset: Dataset) -> None:
-        #TODO: [Potential] Force full fetch if it has been 90 days since the last fetch.
-        self.start_time = dataset.get_indicator_checkpoint() if self.start_time is None else self.start_time
+        # TODO: [Potential] Force full fetch if it has been 90 days since the last fetch.
+        self.start_time = (
+            dataset.get_indicator_checkpoint()
+            if self.start_time is None
+            else self.start_time
+        )
         self.stop_time = int(time.time()) if self.stop_time is None else self.stop_time
         dataset.load_indicator_cache(self.signal_types_by_name.values())
         more_to_fetch = True
@@ -88,10 +93,10 @@ class ExperimentalFetchCommand(command_base.Command):
                 additional_tags=self.additional_tags,
                 next_page=next_page,
             )
-            if 'data' in result:
-                self._process_indicators(result['data'])
-            more_to_fetch = 'paging' in result and 'next' in result['paging']
-            next_page = result['paging']['next'] if more_to_fetch else None
+            if "data" in result:
+                self._process_indicators(result["data"])
+            more_to_fetch = "paging" in result and "next" in result["paging"]
+            next_page = result["paging"]["next"] if more_to_fetch else None
 
         for signal_name, signal_type in self.signal_types_by_name.items():
             if signal_name not in self.counts:
@@ -108,16 +113,16 @@ class ExperimentalFetchCommand(command_base.Command):
         """Process indicators"""
         for ti_json in indicators:
             ti = ThreatIndicator(
-                int(ti_json.get('id')),
-                ti_json.get('indicator'),
-                ti_json.get('type'),
-                int(ti_json.get('creation_time')),
-                int(ti_json.get('last_updated')),
-                ti_json.get('status'),
-                ti_json.get('is_expired'),
-                ti_json.get('tags'),
-                [int(app) for app in ti_json.get('applications_with_opinions')],
-                int(ti_json.get('expire_time')) if 'expire_time' in ti_json else None
+                int(ti_json.get("id")),
+                ti_json.get("indicator"),
+                ti_json.get("type"),
+                int(ti_json.get("creation_time")),
+                int(ti_json.get("last_updated")),
+                ti_json.get("status"),
+                ti_json.get("is_expired"),
+                ti_json.get("tags"),
+                [int(app) for app in ti_json.get("applications_with_opinions")],
+                int(ti_json.get("expire_time")) if "expire_time" in ti_json else None,
             )
 
             match = False
--- a/pytx3/pytx3/cli/fetch2.py
+++ b/pytx3/pytx3/cli/fetch2.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import requests
+import typing as t
+from .. import TE
+from ..dataset import Dataset
+from . import command_base
+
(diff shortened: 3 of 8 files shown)


pytx3/pytx3/cli/fetch2.py
```diff
@@ -0,0 +1,58 @@
...
+            "--stop-time",
+            type=int,
+            help="Fetch updates that occured before this timestamp",
+        )
+        ap.add_argument(
+            "--owner",
+            type=int,
+            help="Only fetch updates for indicators that the given app has a descriptor for",
+        )
+        ap.add_argument(
```
InderH, line 44: The argument parsing for "--threat-types" and "--additional-tags" is fixed in a later commit (d3b7431)

pytx3/pytx3/indicator.py
```diff
@@ -0,0 +1,93 @@
...
+    threat_type: str
+    creation_time: int
+    last_updated: int
+    status: str
+    is_expired: bool
+    tags: t.List[str]
+    applications_with_opinions: t.List[int]
+    expire_time: int
+
+    # def __init__(
```
InderH: Forgot to remove this comment, it is removed in a later commit (fb823db)

pytx3/pytx3/cli/fetch2.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import requests
+import typing as t
+from .. import TE
```
jeberl, line 10: For my own edification, what's the difference between `from . import XXX` and `from .. import XXX`

pytx3/pytx3/cli/fetch2.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        ap.add_argument(
+            "--stop-time",
+            type=int,
+            help="Fetch updates that occured before this timestamp",
+        )
+        ap.add_argument(
+            "--owner",
+            type=int,
+            help="Only fetch updates for indicators that the given app has a descriptor for",
```
jeberl, line 42: Can we change the comments here and across pytx to use less specific TE terminology and more general terminology for how we think about the data model. Some examples:
"Indicators" -> "Threats"
"descriptors" -> "opinions"
"privacy group" -> "collaboration"

Here i'd like to see something more like: "Only fetch updates for threats which the given app has added opinions about.

pytx3/pytx3/cli/fetch2.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        ap.add_argument(
+            "--threat-types",
+            type=t.List[str],
+            help="Only fetch updates for indicators of the given type",
```
jeberl, line 47: "Only fetch updates for threats of the given type"

pytx3/pytx3/cli/fetch2.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        ap.add_argument(
+            "--additional-tags",
+            type=t.List[str],
+            help="Only fetch updates for indicators that have a descriptor with each of these tags",
```
jeberl, line 52: "Only fetch updates for threats which some application has given these tags to"

jeberl (changes requested):
Some changes

pytx3/pytx3/cli/fetch2.py
```diff
@@ -56,3 +59,45 @@ def __init__(
...
+            self.start_time,
+            self.stop_time,
+            self.owner,
+            self.threat_types,
+            self.additional_tags
+        )
+
+    def _fetch_threat_updates(
+        self,
+        privacy_group: int,
```
jeberl, line 137: Eventually, we're going to want to change this to a collaboration object right?

pytx3/pytx3/cli/fetch2.py
jeberl: fixed how? What does `nargs='+'` do?

pytx3/pytx3/indicator.py
```diff
@@ -0,0 +1,93 @@
...
+      "tags": [
+        "tag1",
+        "tag2",
+        "tag3"
+      ],
+      "applications_with_opinions": [
+        "123456789055502"
+      ],
+      "expire_time": 9601513796
+    }
```
jeberl, line 35: Good job giving an example

pytx3/pytx3/indicator.py
```diff
@@ -0,0 +1,93 @@
...
+        return (
+            self.id,
+            self.indicator,
+            self.threat_type,
+            self.creation_time,
+            self.last_updated,
+            self.status,
+            self.is_expired,
+            " ".join(self.tags),
+            " ".join([str(app) for app in self.applications_with_opinions]),
```
jeberl, line 60: Why do we join these and then later split them. Why not keep nested structure?

pytx3/pytx3/signal_type/signal_base.py
```diff
@@ -52,6 +52,14 @@ def process_descriptor(self, descriptor: t.Dict[str, t.Any]) -> bool:
...
         return False
 
+    def process_indicator(self, indicator: t.Dict[int, t.Any]) -> bool:
+        """
+        Add ThreatIndicator to the state of this type, if it is for this type.
+
+        Return true if the indicator was used.
+        """
+        return False
+
```
jeberl, line 62: What's this used for?

pytx3/pytx3/signal_type/signal_base.py
```diff
@@ -103,6 +111,7 @@ class SimpleSignalType(SignalType, HashMatcher):
 
     def __init__(self) -> None:
         self.state: t.Dict[str, SimpleDescriptorRollup] = {}
+        self.indicator_state: t.Dict[int, ThreatIndicator] = {}
```
jeberl, line 114: Why does indicator state have int key but descriptor state has string key?

pytx3/pytx3/cli/fetch2.py
```diff
@@ -59,16 +65,72 @@ def __init__(
         self.owner = owner
         self.threat_types = threat_types
         self.additional_tags = additional_tags
+        self.signal_types_by_name = {
+            name: signal()
+            for name, signal in meta.get_signal_types_by_name().items()
+        }
+        self.last_update_printed = 0
```
jeberl, line 72: why not start at `time.time()` instead of 0?

pytx3/pytx3/cli/fetch2.py
```diff
@@ -78,7 +140,11 @@ def _fetch_threat_updates(
         owner: int,
         threat_types: t.List[str],
         additional_tags: t.List[str],
-    ) -> None:
+        next_page: str
+    ):
```
jeberl, line 144: Can we set the return type?

jeberl (commented):
I'm having a hard time understanding what this file is used for. When is self.INDICATOR_TYPE set? What does self.TYPE_TAG refer to? Perhaps best to go through together

jeberl:
In general this looks quite good. One major question though is how are you testing this? There's a number of test cases I'd like to see to show just that this pulls any data in a paginated way and stores/loads it correctly. In future PRs and in backend code 

I'd also like @bodnarbm to review as the python/git guru. I believe he was also the reviewer for davids original work.

pytx3/pytx3/indicator.py
InderH: Following the format used in descriptor.py.

Alternatively I can store the list directly and it would be saved into the file like "[123,456,789]".

pytx3/pytx3/cli/fetch2.py
bodnarbm: Basic version is:

`from . import XXX` means from the directory that houses this file, import the module defined in the file XXX. So here it would import the file XXX.py from `pytx3/pytx3/cli`

`from .. import xxx` means from the parent directory of the directory that houses this file, import the module defined in the file XXX. So here it would import the file XXX.py from `pytx3/pytx3`

`from ..dataset import XXX` means from the module dataset that inside the parent directory of the directory of this file, import the class or symbol XXX. So here it would import the class or symbol XXX that is defined in the file `pytx3/pytx3/dataset.py`

There's other details I'm probably missing, but that's a basic gist of its use here.

pytx3/pytx3/cli/fetch2.py
```diff
@@ -0,0 +1,169 @@
...
+import pathlib
+import time
+import typing as t
+from .. import TE
+from ..dataset import Dataset
+from ..indicator import ThreatIndicator
+from ..content_type import meta
+from . import command_base
+
+class Fetch2Command(command_base.Command):
```
bodnarbm, line 16: I'm unsure what the final command name should be here, but for now, we could go with something like `ExperimentalFetchCommand` (and change the related cli invocation words to match) to make it clear this is not intended for production use yet.

pytx3/pytx3/cli/fetch2.py
bodnarbm: I like the direction of changes to terminology, but I think these should happen first on the ThreatExchange api side. To prevent cognitive mismatch when comparing the pytx3 documentation to the ThreatExchange documentation, I think it may be good to keep the existing terminology and then overall them together (or around the same time).

Just my thought.

pytx3/pytx3/cli/fetch2.py
bodnarbm: Alteratively, we could do something similar to `fetch.py` that uses

```python
# Cause first print to be after ~5 seconds
self.last_update_printed = now - self.PROGRESS_PRINT_INTERVAL_SEC + 5
```

pytx3/setup.py
```diff
@@ -35,6 +35,7 @@
     packages=find_packages(exclude=["tests"]),
     install_requires=[
         "python-Levenshtein",
+        "requests",
```
bodnarbm, line 38: Per comment above, once this switches over to urllib, go ahead and remove the requests requirement.

pytx3/pytx3/indicator.py
bodnarbm: Since this is for translation to the CSV format and is the same approach used by SimpleDescriptorRollup.as_row in descriptor.py, I'm good with keeping the join method here. I'm not a fan of having to escape things in CSV, so if we can avoid having to escape the commas, I'm all for it.

bodnarbm (changes requested):
Overall this direction looks good to me, especially given that this will continue to be in progress work along with other changes to the TE api.

My main request for changes center around removing the `requests` dependency and moving the API code into the `TE.py` module. I also suggest naming the command something along the lines of ExperimentalFetch to put users on notice that this is still in development. But any runtime way of warning users is fine with me.

Before merging this, can you verify that the cli tool parses pre-existing state files created by older versions of the tool. The changes you made to dataset storage should be fine and backwards compatible, but we should test that a dataset cache created by an older version of the tool will still process in this newer version using the exist fetch command. Just don't want to break users of the existing fetch.

pytx3/pytx3/cli/fetch2.py
InderH: nargs means take in arguments as a list. The "+" means expect at least 1 value if the argument flag is given (i.e. don't allow --threat-types to be an empty list)

pytx3/pytx3/cli/fetch2.py
InderH: @jeberl are you ok with punting on the terminology update as @bodnarbm mentioned?

pytx3/pytx3/cli/fetch2.py
InderH: yup

pytx3/pytx3/signal_type/signal_base.py
InderH: Place holder for the super class - if a child class doesn't implement process_indicator it defaults to false.

pytx3/pytx3/signal_type/signal_base.py
InderH: indicator_state is keyed by indicator id, state (for descriptors) is keyed by the indicator (only 1 descriptor is stored per indicator)

pytx3/pytx3/cli/fetch2.py
InderH: Used 0 as an initial value so we would print right after the first fetch so the user gets some immediate feedback (so they know the command is doing stuff) vs having to wait 30s before the feedback.

Could do something similar to fetch.py if you prefer but I don't think it makes much of a difference

bodnarbm (commented):
LGTM.

I'll leave the remainder to @jeberl to review if his items are fully addressed, but the changes you made @InderH resolved my prior comments. 

**My only additional ask is to please run the `black` formatter on the files in `pytx3` to make sure the code format is consistent with the existing files.**

```shell
$ black pytx3/**/*.py
```

assuming you have black installed with `pip install black`

Re testing, I tested locally that an existing fetch checkpoint and dataset from the version in master was able to be processed with the `fetch` command. My test plan was as follows.

```
❯ git checkout master
Already on 'master'
Your branch is up to date with 'origin/master'.
❯ pip uninstall -y pytx3 && pip install -e pytx3
Found existing installation: pytx3 0.0.8
Uninstalling pytx3-0.0.8:
  Successfully uninstalled pytx3-0.0.8
Obtaining file:///home/bodnarbm/repos/facebook/ThreatExchange/pytx3
Requirement already satisfied: python-Levenshtein in /home/bodnarbm/.venv/threatexchange/lib/python3.8/site-packages (from pytx3==0.0.8) (0.12.0)
Requirement already satisfied: setuptools in /home/bodnarbm/.venv/threatexchange/lib/python3.8/site-packages (from python-Levenshtein->pytx3==0.0.8) (44.0.0)
Installing collected packages: pytx3
  Running setup.py develop for pytx3
Successfully installed pytx3
❯ pytx3 fetch --clear
❯ pytx3 fetch --until-timestamp $TIMESTAMP_1
photo_md5: 1
video_md5: 4
❯ pytx3 fetch --until-timestamp $TIMESTAMP_2
Doing an incremental update from ****************
This will miss some updates and deletes of previously fetched data, use --full to force a refetch.
photo_md5: 4
pdq: 5
video_md5: 877
❯ pytx3 fetch --clear
❯ pytx3 fetch --until-timestamp $TIMESTAMP_1
photo_md5: 1
video_md5: 4
❯ git checkout InderH/fetch2
Switched to branch 'InderH/fetch2'
❯ pip uninstall -y pytx3 && pip install -e pytx3
Found existing installation: pytx3 0.0.8
Uninstalling pytx3-0.0.8:
  Successfully uninstalled pytx3-0.0.8
Obtaining file:///home/bodnarbm/repos/facebook/ThreatExchange/pytx3
Requirement already satisfied: python-Levenshtein in /home/bodnarbm/.venv/threatexchange/lib/python3.8/site-packages (from pytx3==0.0.8) (0.12.0)
Requirement already satisfied: setuptools in /home/bodnarbm/.venv/threatexchange/lib/python3.8/site-packages (from python-Levenshtein->pytx3==0.0.8) (44.0.0)
Installing collected packages: pytx3
  Running setup.py develop for pytx3
Successfully installed pytx3
❯ pytx3 fetch --until-timestamp $TIMESTAMP_2
Doing an incremental update from ************
This will miss some updates and deletes of previously fetched data, use --full to force a refetch.
photo_md5: 4
video_md5: 877
pdq: 5
```

I don't see any regression, so this looks safe to merge from a production code path viewpoint.

Note: I did not test the experimental fetch itself, and I leave that to you and @jeberl on what testing is needed for the experimental path given planned changes on the API end of the work.

pytx3/pytx3/cli/fetch2.py
bodnarbm: I should note that my comment was also specific to across pytx3. For this command, given it's intended use, I'm less concerned about the changes. And I'm fine either way, but would prefer to defer the terminology change across pytx3 until we are ready to make that change in the API layer (to prevent potential churn in exact wording).

pytx3/pytx3/cli/fetch2.py
bodnarbm: Either way is fine with me. We can iterate on the UX for the command too once it is out of experimental.

bodnarbm (approved):

pytx3/pytx3/cli/fetch2.py
jeberl: genius. thank you

pytx3/pytx3/cli/fetch2.py
jeberl: @InderH, I'll leave to you

pytx3/pytx3/cli/fetch2.py
jeberl: Ok. I'm taking a documentation refresh on as a project this half so can come back and make these edits

pytx3/pytx3/indicator.py
jeberl: sgtm

pytx3/pytx3/cli/experimental_fetch.py
```diff
@@ -11,6 +11,7 @@
 from ..content_type import meta
 from . import command_base
 
+
```
jeberl, line 14: Delete extra line

jeberl (commented):
Left a bunch few comments through out and Resolved the comments that I think you fix. Once those are set should be good.

Still getting the hang of how to review via git

pytx3/pytx3/cli/fetch2.py
InderH: Function no longer exists; if you look at the newest changes I moved it the fetching into TE.py

pytx3/pytx3/cli/fetch2.py
```diff
@@ -73,6 +73,9 @@ def __init__(
         self.counts = collections.Counter()
 
     def execute(self, dataset: Dataset) -> None:
+        #TODO: [Potential] Force full fetch if it has been 90 days since the last fetch.
+        self.start_time = dataset.get_indicator_checkpoint() if self.start_time is None else self.start_time
+        self.stop_time = int(time.time()) if self.stop_time is None else self.stop_time
```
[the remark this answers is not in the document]
InderH: Yup, threat_updates defaults it to request.time

pytx3/pytx3/cli/fetch2.py
```diff
@@ -73,6 +73,9 @@ def __init__(
         self.counts = collections.Counter()
 
     def execute(self, dataset: Dataset) -> None:
+        #TODO: [Potential] Force full fetch if it has been 90 days since the last fetch.
```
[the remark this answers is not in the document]
InderH: Added into task 2.10 of my milestone 2 plan, I can send you a link if you'd like.

pytx3/pytx3/TE.py
```diff
@@ -177,6 +177,32 @@ def getInfoForIDs(self, ids, **kwargs):
 
         return descriptors
 
+    # ----------------------------------------------------------------
```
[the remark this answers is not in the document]
InderH: The actual `experimentalfetch` command is ~37% of the loc of `fetch`, granted this may change a little bit as I update experimentalfetch once threat_updates is updated.

The code is also much simpler as threat_updates is a much more direct way to download data compared to what `fetch` does.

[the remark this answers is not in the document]
jeberl: Incredible!!!

jeberl (approved):

jeberl:
Great stuff! Your first GitHub PR at fb!

Merge it...

PR #304, opened 2020-10-23
Add initial PDQ matcher class using faiss flat binary indexes

bodnarbm:
## Summary
The Faiss library provides some binary indexes that can be very
useful for matching PDQ hashes against a target dataset. This is
an initial implementation using the Flat Binary Index which will
perform an exhaustive search for a set of query PDQ hashes against
a static set of target hashes.

Later work will be done to create another class for Faiss binary
indexes using the MIH indexing technique to provide optimistically
sublinear searches. Additionally, other work will be needed to allow
serialization of these data structures to allow reuse of the index
across processes.

I am also deferring documenting these changes in the pytx3 Readmes until after the PDQ matcher work is completed.

## Test plan

Created new unit tests for the PDQFlatHashIndex methods using python unit tests.

After installing `numpy` and `faiss-cpu` to allow use of faiss, along with pytest as the test runner. I ran the following from the repo root.

```python
> py.test pytx3
============================================================================================== test session starts ==============================================================================================
platform darwin -- Python 3.8.6, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
rootdir: /Users/bodnarbm/github/ThreatExchange/pytx3
collected 7 items                                                                                                                                                                                               

pytx3/tests/hashing/test_pdq_faiss_matcher.py .......                                                                                                                                                     [100%]

=============================================================================================== 7 passed in 0.17s ===============================================================================================
```

**Note:** neither numpy nor faiss-cpu will be required for other users of pytx3 not wishing to use the faiss methods, but it will be required for unit tests.

### DIFF
--- a/.github/workflows/pytx3-ci.yaml
+++ b/.github/workflows/pytx3-ci.yaml
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.x"
+          python-version: "3.8"
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/pytx3-ci.yaml
+++ b/.github/workflows/pytx3-ci.yaml
@@ -32,3 +32,17 @@ jobs:
       - name: Check code format
         run: |
           black --check .
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[test]"
+      - name: Test with pytest
+        run: |
+          py.test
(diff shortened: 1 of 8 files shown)


.github/workflows/pytx3-ci.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.x"
+          python-version: "3.8"
```
BarrettOlson, line 41: This change is just for CI right? 
Could this at all affect pytx3 usability (or at least signal of compatibility) on <python3.8?

BarrettOlson (approved):
LGTM! Exciting! Apologies if my order of review created merge conflicts.

.github/workflows/pytx3-ci.yaml
bodnarbm: Correct, the change is just for CI builds. This is to prevent 3.9 from being used in CI since numpy does not distribute a wheel for 3.9 yet (3.9 having just been released this month). 3.9 can be used with numpy, but the install takes a while on CI.

I'll do a separate PR to add a matrix for testing 3.6 and 3.7 along with 3.8. (3.6 being the lowest version of python that we state we support).

PR #306, opened 2020-10-23
Initial CODEOWNERS file

bodnarbm:
For now, just adding in my self so that I am auto tagged on PRs for
the repo. Others can add themselves if they wish to be auto requested
on PRs to the repo.


BarrettOlson (approved):
Cool! Might be worth talking about as a team expectations around what names we put here. As a rule our oncall should be reviewing PR. I don't want it to end up falling more on you because you're currently the only username we have here.

PR #307, opened 2020-10-26
Support Python 3.6 and 3.7 in CI tests

bodnarbm:
Currently our CI tests run only against python 3.8. This adds
support for running CI tests with Python 3.6 and 3.7 also. (Python
3.6 being the lowest we state we support for pytx3).


BarrettOlson (approved):
Yep yep yep

PR #308, opened 2020-10-28
Create PDQ Hash Index using Multi-Index Hashing

bodnarbm:
Summary
========

Creates a PDQMultiHashIndex that uses the faiss IndexBinaryMultiHash
class. This will use Multi-Index Hashing for searches on hashes, which
should provide for optimistically sub-linear hash lookups.

Test Plan
=========

Create additional unit tests to cover this new class, and run locally with

```shell
$ py.test
```


BarrettOlson (approved):
LGTM!

PR #309, opened 2020-10-28
Use faiss index to reconstruct hash in search results

bodnarbm:
Summary
========
Presently, we need to keep an companion index in memory to convert
the faiss index search results (which are indices in the index) back
to the PDQ hash string.

This change uses the reconstruct method on the two types of faiss indexes
we are using to recompute those hash values on search results. This
should reduce the memory requirements for these classes (and the
eventual size of the classes on disk once we support serializing them)
by around n*32 bytes, where n is the number of hashes in the index.

Test Plan
=======
Added a common test to the indexes for looking up the hash at a given place, and ensured
that the prior tests all still pass.


BarrettOlson (approved):
LGTM. Let me know if you want me to pull these down and test them locally.

PR #313, opened 2020-10-30
Add pickle support to pdq hash indexes

bodnarbm:
Summary
=======

The underlaying faiss indexes for the PDQHashIndex classes do not
natively support pickling (serialization) of the data to bytes.
Adding support for pickling will allow us to easily save and load
these indexes from files or over a network connection (for example
during a map-reduce or batch job).

This work is dependent upon #309. I will rebase off of #309 once that merges.

Test Plan
======

Added additional testing that runs the common PDQHashIndex tests against a reserialized form of the test subject. This is to ensure that functionality remains the same with the index once it was be serialized and then deserialized.


BarrettOlson (approved):
Love a good pickle

PR #316, opened 2020-11-06
Mark and close stale issues

bodnarbm:
Summary
---------

Add GitHub action to mark and comment on stale issues, and later close them if there is no activity.

Set the current timeframes to 90 days with no activity before marking an issue as stale, and 2 weeks graceperiod before closing.

Test Plan
---------

Unfortuately, this is difficult to test as the action needs to run in a cron job. Mostly just compared the settings with the examples at https://github.com/actions/stale and ensured that my IDE yaml linter did not identify any errors.

### DIFF
--- a/.github/workflows/repo-stale-issues.yml
+++ b/.github/workflows/repo-stale-issues.yml
@@ -1,0 +1,24 @@
+name: Close stale issues
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


BarrettOlson (approved):
clean up, clean up!

PR #317, opened 2020-11-10
Refactor PDQHashIndex Common Tests

bodnarbm:
Summary
---------

To prepare for changes related to supporting optional custom ids for the PDQHashIndexes, this PR refactors
the existing tests to

1) Make checks on whether the index is picklable a commont test; and
2) Make tests for extracting hashes by id via `hash_at` index specific.

Making these changes in advance should make it easier to add the custom ids support without added noise in
the test code.

Test Plan
---------

`py.test`


jeberl:
This looks good to me but I can't say I'm an expert at pickle in case you'd like to wait for another re

BarrettOlson (approved):
LGTM as well

PR #322, opened 2020-11-12
Add basic ci for testing PDQ cpp code.

bodnarbm:
Summary
---------

Adds a very simple CI workflow to test the pdq cpp build using the current `make` default target. This will test both building the code and running the regression tests that the cpp PDQ already has.

Test Plan
---------

This PR should pass its status checks, including the new CI job.


BarrettOlson (approved):
Yay tests

PR #324, opened 2020-11-13
Add option to return only the id when searching PDQ hashes

bodnarbm:
Summary
---------

Useful especially in the cases where a custom id has been set for the hashes
in the index, allowing the search results to be the id instead of the hash
will help callers that are using that id to lookup the information in another
datastore.

Test Plan
---------

Added tests and ran them with `py.test`


BarrettOlson (approved):
LGTM!

PR #325, opened 2020-11-14
Add do-not-reap support for stale issue action

bodnarbm:
Summary
---------

Adding support for the existing do-not-reap tag to have it not auto close issues that we mark with that tag. This is useful for issues for example where we are requesting community help with addressing but there is no current timeline for.

Additionally, minor fix of the yaml formatting for the messages to make them not insert new lines on the GitHub messages themselves.

Test Plan
---------

yaml linting only. Needs to run on GitHub itself after merging.


BarrettOlson (approved):
Cool

PR #330, opened 2020-11-19
Specify dists folder for pypi packages

bodnarbm:
Summary
---------

The pypa/gh-action-pypi-publish action does not respect the working-directory setting, so we need to let it know where the dist folder will be from the repo root.

Test Plan
---------

[Executed action on my local fork](https://github.com/bodnarbm/ThreatExchange/runs/1426584148?check_suite_focus=true) (without the github secret) and verified that it make it through the finding the dist folder phase. Note: action still failed before there is no upload token in my fork.


BarrettOlson (approved):
cool

PR #334, opened 2020-12-04
Start releasing to pypi

bodnarbm:
Summary
---------

Adds pypi publishing to the release job and bumps version for threatexchange client library to trigger this job on merge.

Test Plan
---------

- Validate workflow settings against [example usage](https://github.com/pypa/gh-action-pypi-publish/blob/master/README.md#customizing-target-package-dists-directory) (note that user is not required as the default is already `__token__`).
- Checked the OSS repo settings to ensure the pypi api secret is already set.
- Check that existing workflow has been publishing to [Test PyPI](https://test.pypi.org/project/threatexchange/) already.


jeberl (approved):
Yeah, Good, Ok.

bodnarbm:
@jeberl Will demo and will also add releasing documentation to the Read Me.

PR #343, opened 2020-12-16
Update experimentalfetch to work with the new version of threat_updates

InderH:
Summary
---------

Update experimentalfetch to work with the new version of threat_updates. Make experimentalfetch more resilient by adding error handling and more thorough checkpointing. Update the indicator "storage system" to no longer rely on signal types and instead store based on indicator type.

Test Plan
---------

1. verify fetch still works:
Ran: "threatexchange -c ~/large_dataset_collab.te.cfg fetch --until-timestamp `date -j -vMarm -v1d -v2018y '+%s'`" to competition.
Also used fetch to query the entire large dataset set.

2. Verify experimentalfetch gets everything:
Ran: "threatexchange -c ~/large_dataset_collab.te.cfg experimental-fetch" on the large dataset collaboration and it queried to completion getting ~1.93m records

3. Verify experimentalfetch  --start-time and --end-time
Ran with various combinations of --start-time and --end-time and things worked as expected (only data updated within the time range was returned)

4. Verify experimentalfetch --threat-types
Ran "threatexchange -c ~/other_collab.te.cfg experimental-fetch --threat-types HASH_PDQ URL" on another collaboration and things worked as expected (only data with indicator types of HASH_PDQ or URL were returned)

5. Verify experimentalfetch --continuation
Ran "threatexchange -c ~/other_collab.te.cfg experimental-fetch --continuation" and this works as described:
i.e.
* When no url/cursor is check-pointed threat_updates is queried using the check-pointed last_stop_time as the new start-time and the check-pointed indicator types (if given) as filters
* When a url/cursor is check-pointed we continue fetching from that url/cursor

6. Verify experimentalfetch forces a full update if you haven't fetched in 85 days
* Modified the checkpoint to be > 85 days old
* Tried to run with --continuation
* Got the following message "It's been a long time since a full fetch, forcing one now." and started to fetch the entire collaboration

7. Tested error handling by both explicitly throwing errors and mucking with the url being queried

### DIFF
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -180,9 +180,9 @@ class Net:
     # ----------------------------------------------------------------
     # Gets threat updates for the given privacy group.
     @classmethod
-    def getThreatUpdates(self, privacy_group, **kwargs):
-        if kwargs["next_page"] is not None:
-            url = kwargs["next_page"]
+    def getThreatUpdates(self, privacy_group, next_page=None, **kwargs):
+        if next_page is not None:
+            url = next_page
         else:
             url = (
                 self.TE_BASE_URL
@@ -195,7 +195,7 @@ class Net:
             )
             for arg, value in kwargs.items():
                 if value is not None:
-                    if arg == "threat_type":
+                    if arg == "threat_types":
                         url += "&threat_types=" + ",".join(value)
                     else:
                         url += "&" + arg + "=" + str(value)
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -191,15 +191,11 @@ class Net:
                 + "/threat_updates/"
                 + "?access_token="
                 + self.APP_TOKEN
-                + "&fields=id,indicator,type,creation_time,last_updated,is_expired,expire_time,tags,status,applications_with_opinions"
+                + "&fields=id,indicator,type,creation_time,last_updated,should_delete,tags,status,applications_with_opinions"
             )
             for arg, value in kwargs.items():
                 if value is not None:
-                    if arg == "additional_tags":
-                        url += "&additional_tags=" + ",".join(
-                            [str(app) for app in value]
-                        )
-                    elif arg == "threat_type":
+                    if arg == "threat_type":
                         url += "&threat_types=" + ",".join(value)
                     else:
                         url += "&" + arg + "=" + str(value)
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -181,8 +181,8 @@ class Net:
     # Gets threat updates for the given privacy group.
     @classmethod
     def getThreatUpdates(self, privacy_group, **kwargs):
-        if "next" in kwargs:
-            url = kwargs["next"]
+        if "next_page" in kwargs and kwargs["next_page"] is not None:
+            url = kwargs["next_page"]
         else:
             url = (
                 self.TE_BASE_URL
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -195,8 +195,8 @@ class Net:
             )
             for arg, value in kwargs.items():
                 if value is not None:
-                    if arg == "threat_types":
-                        url += "&threat_types=" + ",".join(value)
+                    if arg == "types":
+                        url += "&types=" + ",".join(value)
                     else:
                         url += "&" + arg + "=" + str(value)
         return self.getJSONFromURL(url)
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -181,7 +181,7 @@ class Net:
     # Gets threat updates for the given privacy group.
     @classmethod
     def getThreatUpdates(self, privacy_group, **kwargs):
-        if "next_page" in kwargs and kwargs["next_page"] is not None:
+        if kwargs["next_page"] is not None:
             url = kwargs["next_page"]
         else:
             url = (
--- a/python-threatexchange/threatexchange/cli/experimental_fetch.py
+++ b/python-threatexchange/threatexchange/cli/experimental_fetch.py
@@ -52,8 +52,8 @@ class ExperimentalFetchCommand(command_base.Command):
         ap.add_argument(
             "--page-size",
             type=int,
-            help="The number of updates to fetch per request, defaults to 100",
-            default=100,
+            help="The number of updates to fetch per request, defaults to 500",
+            default=500,
         )
 
     def __init__(
@@ -93,7 +93,6 @@ class ExperimentalFetchCommand(command_base.Command):
 
         more_to_fetch = True
         remaining_attempts = 5
-
         while more_to_fetch:
             try:
                 result = TE.Net.getThreatUpdates(
@@ -126,6 +125,9 @@ class ExperimentalFetchCommand(command_base.Command):
         dataset.record_indicator_checkpoint(
             privacy_group, self.stop_time, request_time, self.threat_types, next_page
         )
+
+        if remaining_attempts == 5:
+            print("Just like that we are done!")
         print("\nHere is a summary from this run:")
         for threat_type in self.counts:
             print(f"{threat_type}: {self.counts[threat_type]}")
--- a/python-threatexchange/threatexchange/cli/experimental_fetch.py
+++ b/python-threatexchange/threatexchange/cli/experimental_fetch.py
@@ -101,12 +101,11 @@ class ExperimentalFetchCommand(command_base.Command):
                 ti_json.get("indicator"),
                 ti_json.get("type"),
                 int(ti_json.get("creation_time")),
-                int(ti_json.get("last_updated")),
+                int(ti_json.get("last_updated")) if "last_updated" in ti_json else None,
                 ti_json.get("status"),
-                ti_json.get("is_expired"),
+                ti_json.get("should_delete"),
                 ti_json.get("tags"),
                 [int(app) for app in ti_json.get("applications_with_opinions")],
-                int(ti_json.get("expire_time")) if "expire_time" in ti_json else None,
             )
 
             match = False
--- a/python-threatexchange/threatexchange/cli/experimental_fetch.py
+++ b/python-threatexchange/threatexchange/cli/experimental_fetch.py
@@ -61,7 +61,7 @@ class ExperimentalFetchCommand(command_base.Command):
     def execute(self, dataset: Dataset) -> None:
         privacy_group = dataset.config.privacy_groups[0]
         self.indicator_signals = signal_base.IndicatorSignals(privacy_group)
-        self.indicator_signals.load_indicators()
+        dataset.load_indicator_cache(self.indicator_signals)
 
         # TODO: [Potential] Force full fetch if it has been 90 days since the last fetch.
         # self.start_time = (
@@ -89,7 +89,7 @@ class ExperimentalFetchCommand(command_base.Command):
             more_to_fetch = "paging" in result and "next" in result["paging"]
             next_page = result["paging"]["next"] if more_to_fetch else None
 
-        self.indicator_signals.store_indicators()
+        dataset.store_indicator_cache(self.indicator_signals)
         for threat_type in self.counts:
             print(f"{threat_type}: {self.counts[threat_type]}")
         return
@@ -109,7 +109,9 @@ class ExperimentalFetchCommand(command_base.Command):
                 ti_json.get("status"),
                 ti_json.get("should_delete"),
                 ti_json.get("tags") if "tags" in ti_json else [],
-                [int(app) for app in ti_json.get("applications_with_opinions")] if "applications_with_opinions" in ti_json else [],
+                [int(app) for app in ti_json.get("applications_with_opinions")]
+                if "applications_with_opinions" in ti_json
+                else [],
             )
 
             self.indicator_signals.process_indicator(ti)
--- a/python-threatexchange/threatexchange/cli/experimental_fetch.py
+++ b/python-threatexchange/threatexchange/cli/experimental_fetch.py
@@ -25,8 +25,10 @@ class ExperimentalFetchCommand(command_base.Command):
     """
 
     PROGRESS_PRINT_INTERVAL_SEC = 30
-    # We use a DEFAULT_REFETCH_SEC of 85 days because threat_updates tombstones last for 90 days
-    # If you don't fetch for updates within the 90 days you may miss some deletion notifications
+    # If a client does not resume tailing the threat_updates endpoint fast enough, 
+    # deletion records will be removed, making it impossible to determine which
+    # records should be retained without refetching the entire dataset from scratch.
+    # The current implementation will retain for 90 days: TODO: Link to documentation
     DEFAULT_REFETCH_SEC = 3600 * 24 * 85  # 85 days
     MAX_CONSECUTIVE_RETRIES = 5
 
--- a/python-threatexchange/threatexchange/cli/experimental_fetch.py
+++ b/python-threatexchange/threatexchange/cli/experimental_fetch.py
@@ -71,15 +71,14 @@ class ExperimentalFetchCommand(command_base.Command):
         privacy_group = dataset.config.privacy_groups[0]
         self.indicator_signals = signal_base.IndicatorSignals(privacy_group)
         dataset.load_indicator_cache(self.indicator_signals)
-
-        # TODO: Consider threat_type in checkpoints
         self.stop_time = request_time if self.stop_time is None else self.stop_time
         next_page = None
         if self.continuation:
             checkpoint = dataset.get_indicator_checkpoint(privacy_group)
             self.start_time = checkpoint["last_stop_time"]
-            next_page = checkpoint["url"]
             self.stop_time = request_time
+            self.threat_types = checkpoint["threat_types"]
+            next_page = checkpoint["url"]
             if request_time - checkpoint["last_run_time"] > self.DEFAULT_REFETCH_SEC:
                 print("It's been a long time since a full fetch, forcing one now.")
                 self.start_time = 0
@@ -116,7 +115,7 @@ class ExperimentalFetchCommand(command_base.Command):
             remaining_attempts = 5
 
         dataset.record_indicator_checkpoint(
-            privacy_group, self.stop_time, request_time, next_page
+            privacy_group, self.stop_time, request_time, self.threat_types, next_page
         )
         dataset.store_indicator_cache(self.indicator_signals)
         for threat_type in self.counts:
--- a/python-threatexchange/threatexchange/cli/experimental_fetch.py
+++ b/python-threatexchange/threatexchange/cli/experimental_fetch.py
@@ -7,6 +7,7 @@ import time
 import typing as t
 from .. import TE
 from ..dataset import Dataset
+from ..signal_type import signal_base
 from ..indicator import ThreatIndicator
 from ..content_type import meta
 from . import command_base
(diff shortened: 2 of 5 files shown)


python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,102 @@ def init_argparse(cls, ap) -> None:
...
+        privacy_group = dataset.config.privacy_groups[0]
+        self.indicator_store = dataset.get_indicator_store(privacy_group)
+        dataset.load_indicator_cache(self.indicator_store)
+        self.stop_time = request_time if self.stop_time is None else self.stop_time
         next_page = None
+        if self.continuation:
+            checkpoint = dataset.get_indicator_checkpoint(privacy_group)
+            self.start_time = checkpoint["last_stop_time"]
+            self.stop_time = request_time
+            self.threat_types = checkpoint["threat_types"]
```
jeberl, line 88: Can you print the params being used for the continuation?

python-threatexchange/threatexchange/dataset.py
```diff
@@ -76,19 +80,47 @@ def get_fetch_checkpoint(self) -> FetchCheckpoint:
...
+            values["threat_types"] = (
+                values["threat_types"].split(" ")
+                if values["threat_types"] != ""
+                else None
+            )
+            values["url"] = f.readline().strip()
+            values["url"] = values["url"] if values["url"] != "" else None
+        return values
+
+    def record_indicator_checkpoint(
```
jeberl, line 106: Might recommend calling this `set_indicator_checkpoint` so we have a get and a set

jeberl (approved):
This all looks good to me. Thanks for a very thorough test plan! Would like either @bodnarbm or @Dcallies to review also but if it gets too late just send it

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,102 @@ def init_argparse(cls, ap) -> None:
...
+                if remaining_attempts > 0:
+                    print(f"\nTrying again {remaining_attempts} more times.")
+                    time.sleep(5)
+                    continue
+                else:
+                    print(
+                        "\n5 consecutive errors occured, saving state and shutting down!"
+                    )
+                    break
+            remaining_attempts = 5
```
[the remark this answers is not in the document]
InderH: Yup, 5 was just an arbitrary number I came up with - this is just to make sure that if there is some TAO transient error we don't stop querying and instead try again which most of the time fixes things, but if after 5 attempts you're still running into errors it will checkpoint whats left to do and a continuation can be manually started using the continuation option

Something to note is I haven't actually hit any non intentional errors (aside form when my internet went down), so overall I think this is pretty stable

python-threatexchange/threatexchange/TE.py
```diff
@@ -181,8 +181,8 @@ def getInfoForIDs(self, ids, **kwargs):
     # Gets threat updates for the given privacy group.
     @classmethod
     def getThreatUpdates(self, privacy_group, **kwargs):
-        if "next" in kwargs:
-            url = kwargs["next"]
+        if kwargs["next_page"] is not None:
```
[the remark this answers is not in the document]
InderH: The way experimentalfetch is written next_page is always passed in with None meaning start the query using the other given params (i.e. start time, stop time, threat types) and a non-none value meaning fetch the next page, so it was slightly cleaner to not check next_page in kwargs.

Can update if that style is preferred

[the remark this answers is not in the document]
bodnarbm: Change Request:

Since it is an expected parameter and not one that we are allowing for dynamic support in the URL building below (like the other keyword arguments), can you please switch `next_page` to be a optional parameter instead of a kwarg.

Something like 

```python
def getThreatUpdates(self, privacy_group, next_page = None, **kwargs):
```

python-threatexchange/threatexchange/TE.py
```diff
@@ -191,15 +191,11 @@ def getThreatUpdates(self, privacy_group, **kwargs):
...
+                + "&fields=id,indicator,type,creation_time,last_updated,should_delete,tags,status,applications_with_opinions"
             )
             for arg, value in kwargs.items():
                 if value is not None:
-                    if arg == "additional_tags":
-                        url += "&additional_tags=" + ",".join(
-                            [str(app) for app in value]
-                        )
-                    elif arg == "threat_type":
+                    if arg == "threat_type":
```
bodnarbm, line 198: Nit: missed this before, but should the expected `kwarg` be `"threat_types"` instead of `"threat_type"`?

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,102 @@ def init_argparse(cls, ap) -> None:
...
-        self.start_time = (
-            dataset.get_indicator_checkpoint()
-            if self.start_time is None
-            else self.start_time
-        )
-        self.stop_time = int(time.time()) if self.stop_time is None else self.stop_time
-        dataset.load_indicator_cache(self.signal_types_by_name.values())
-        more_to_fetch = True
+        request_time = int(time.time())
+        privacy_group = dataset.config.privacy_groups[0]
```
bodnarbm, line 79: Question: Since the configs allow for multiple privacy groups, should we be looping over these groups?

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -117,23 +152,22 @@ def _process_indicators(
...
                 ti_json.get("status"),
-                ti_json.get("is_expired"),
-                ti_json.get("tags"),
-                [int(app) for app in ti_json.get("applications_with_opinions")],
-                int(ti_json.get("expire_time")) if "expire_time" in ti_json else None,
+                ti_json.get("should_delete"),
+                ti_json.get("tags") if "tags" in ti_json else [],
+                [int(app) for app in ti_json.get("applications_with_opinions")]
+                if "applications_with_opinions" in ti_json
+                else [],
```
bodnarbm, line 161: Non-blocking suggestion: 

you can do `ti_json.get("applications_with_opinions", [])` to provide a default value of the empty array if the the field is not there.

bodnarbm (commented):
Just a couple of comments in addition to the items that @jeberl pointed out.

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,102 @@ def init_argparse(cls, ap) -> None:
...
 
-        for signal_name, signal_type in self.signal_types_by_name.items():
-            if signal_name not in self.counts:
-                continue
-            dataset.store_indicator_cache(signal_type)
-            print(f"{signal_name}: {self.counts[signal_name]}")
+        dataset.store_indicator_cache(self.indicator_store)
+        dataset.record_indicator_checkpoint(
+            privacy_group, self.stop_time, request_time, self.threat_types, next_page
+        )
```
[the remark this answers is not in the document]
InderH: They are privacy group specific

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,102 @@ def init_argparse(cls, ap) -> None:
...
 
-        dataset.record_indicator_checkpoint(self.stop_time)
+        if remaining_attempts == 5:
+            print("Just like that we are done!")
+        print("\nHere is a summary from this run:")
```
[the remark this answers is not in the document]
InderH: This is a summary of all updates that were fetched during this pull. Getting an overall summary is possible by counting when loading the indicator cache but sadly will need to punt on this for now and will come back to it depending on how timings play out tomorrow

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -23,9 +25,15 @@ class ExperimentalFetchCommand(command_base.Command):
...
 
     PROGRESS_PRINT_INTERVAL_SEC = 30
+    DEFAULT_REFETCH_SEC = 3600 * 24 * 85  # 85 days
 
     @classmethod
     def init_argparse(cls, ap) -> None:
+        ap.add_argument(
+            "--continuation",
+            action="store_true",
+            help="Fetch updates that you are missing, this overrides all other given options",
```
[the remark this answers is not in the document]
InderH: So it wont necessarily use the same endtime - this is only true if there was an error and you are using the cursor/url that it failed at. If you are continuing from the last success it will query from that point until the end (i.e. so people can get periodic updates easily)

will go with "Continue fetching updates from where you last stopped with the same threat types. This will ignore all other options given"

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: I can but this might get a little weird if the continuation is based on a url/cursor:
So the url would contain the original start time, stop time, threat types, limit, but the user would need to know how cursors work in order to understand that we are not actually fetching all data from start time to stop time.

The easier thing here would probably be if we are starting from a url/cursor then print out the corresponding url, if we are starting a new time-range based query print out the start time, stop time, threat types, and limit, thoughts?

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: Thanks!

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: Added

python-threatexchange/threatexchange/TE.py
InderH: Nice catch! updated

python-threatexchange/threatexchange/dataset.py
InderH: Updated :)

python-threatexchange/threatexchange/cli/experimental_fetch.py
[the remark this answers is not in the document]
InderH: Added

python-threatexchange/threatexchange/TE.py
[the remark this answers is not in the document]
InderH: Updated

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -23,9 +25,15 @@ class ExperimentalFetchCommand(command_base.Command):
     """
 
     PROGRESS_PRINT_INTERVAL_SEC = 30
+    DEFAULT_REFETCH_SEC = 3600 * 24 * 85  # 85 days
```
Dcallies, line 28: (blocking): I think this is fundamentally incorrect. Your tombstones only last for 30 days, right? At 31 days you'll no longer be able to see the delete update that would put the dataset in the right state.

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -23,9 +25,15 @@ class ExperimentalFetchCommand(command_base.Command):
...
 
     @classmethod
     def init_argparse(cls, ap) -> None:
+        ap.add_argument(
+            "--continuation",
```
Dcallies, line 33: (not blocking): I think this should be the default behavior. The indicator types you care about should be in the collaboration config, and it should just always try fetching them all if called with no arguments.

Will accept as a #todo if you agree with that path

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,118 @@ def init_argparse(cls, ap) -> None:
...
-            dataset.get_indicator_checkpoint()
-            if self.start_time is None
-            else self.start_time
-        )
-        self.stop_time = int(time.time()) if self.stop_time is None else self.stop_time
-        dataset.load_indicator_cache(self.signal_types_by_name.values())
-        more_to_fetch = True
-        next_page = None
+        request_time = int(time.time())
+        privacy_group = dataset.config.privacy_groups[0]
```
Dcallies, line 76: (blocking): Remove this line now that you loop over the privacy groups.

python-threatexchange/threatexchange/cli/experimental_fetch.py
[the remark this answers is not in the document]
Dcallies: (not blocking): I think that this type of retry will probably be fragile, and the root problem is we need to switch to using the `requests` module in the netcode rather than handrolling with python less-powerful default libs

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,118 @@ def init_argparse(cls, ap) -> None:
...
+                    print(f"The following error occured:\n{e}")
+                    if type(e) is urllib.error.HTTPError:
+                        print(e.read())
+                    if remaining_attempts > 0:
+                        print(f"\nTrying again {remaining_attempts} more times.")
+                        time.sleep(5)
+                        continue
+                    else:
+                        print(
+                            "\n5 consecutive errors occured, saving state and shutting down!"
```
Dcallies, line 126: (blocking): Move this into a constant so if we change the number we don't have to hunt through the code to find all the print statements to update

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,118 @@ def init_argparse(cls, ap) -> None:
...
+            dataset.set_indicator_checkpoint(
+                privacy_group,
+                self.stop_time,
+                request_time,
+                self.threat_types,
+                next_page,
+            )
 
-        dataset.record_indicator_checkpoint(self.stop_time)
+            (print(f"\nFor privacy group {privacy_group}:"))
```
Dcallies, line 140: What's with the parenthesis here?

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,118 @@ def init_argparse(cls, ap) -> None:
...
+            )
+            print(f"\nThis brings your total dataset to:")
+            total = 0
+            for threat_type in self.indicator_store.state:
+                size = len(self.indicator_store.state[threat_type])
+                total += size
+                print(f"{threat_type}: {size}")
+            print(f"Total: {total}")
+            print(
+                "\nYou can run 'threatexchange -c {config} experimental-fetch --continuation' to continue from the last successful point."
```
Dcallies, line 157: (not blocking): In general, this tool is much too chatty now. A useful property of CLI tools is their ability to feed into one another from their input and output. In order to get this property, you have to be cognizant of what messages you output on STDOUT

python-threatexchange/threatexchange/dataset.py
```diff
@@ -76,19 +80,47 @@ def get_fetch_checkpoint(self) -> FetchCheckpoint:
...
+            values["last_run_time"] = int(f.readline().strip())
+            values["threat_types"] = f.readline().strip()
+            values["threat_types"] = (
+                values["threat_types"].split(" ")
+                if values["threat_types"] != ""
+                else None
+            )
+            values["url"] = f.readline().strip()
+            values["url"] = values["url"] if values["url"] != "" else None
+        return values
```
Dcallies, line 104: (maybe blocking): This style of read/writing files is super fragile, and it's easy to make mistakes or make backwards incompatible changes.

Please instead switch to reading and writing json dicts, which are easier to create forwards/backwards compabile behavior.

python-threatexchange/threatexchange/dataset.py
```diff
@@ -76,19 +80,47 @@ def get_fetch_checkpoint(self) -> FetchCheckpoint:
...
+
+    def set_indicator_checkpoint(
+        self,
+        privacy_group: int,
+        stop_time: int,
+        request_time: int,
+        threat_types: t.List = [],
+        url: str = "",
```
Dcallies, line 112: (blocking): If you are trying to store the "next" URLs, these are not guaranteed to be viable forever. See https://developers.facebook.com/docs/graph-api/using-graph-api/ under "Don't store cursors")

Dcallies (changes requested):
Still doing a readthrough, but figured I'd put the initial items blocking items in.

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -36,75 +44,118 @@ def init_argparse(cls, ap) -> None:
...
+            self.deleted_count = 0
+            self.indicator_store = dataset.get_indicator_store(privacy_group)
+            dataset.load_indicator_cache(self.indicator_store)
+            self.stop_time = request_time if self.stop_time is None else self.stop_time
+            next_page = None
+            if self.continuation:
+                checkpoint = dataset.get_indicator_checkpoint(privacy_group)
+                self.start_time = checkpoint["last_stop_time"]
+                self.stop_time = request_time
+                self.threat_types = checkpoint["threat_types"]
```
Dcallies, line 89: I don't think threat_type is a concept in the API. This appears to be indicator type? https://developers.facebook.com/docs/threat-exchange/reference/apis/threat-indicator/v9.0

python-threatexchange/threatexchange/TE.py
```diff
@@ -180,9 +180,9 @@ def getInfoForIDs(self, ids, **kwargs):
     # ----------------------------------------------------------------
     # Gets threat updates for the given privacy group.
     @classmethod
-    def getThreatUpdates(self, privacy_group, **kwargs):
-        if "next" in kwargs:
-            url = kwargs["next"]
+    def getThreatUpdates(self, privacy_group, next_page=None, **kwargs):
```
Dcallies, line 183: (not blocking): I generally recommend against passing **kwargs to functions, since it erases type information, and makes it hard to discovered when you mistype an argument name

python-threatexchange/threatexchange/cli/experimental_fetch.py
Dcallies: Ah I see, unfortunately this is a problem with the threat_updates API and not this python code

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: Currently tombstones last for 90 days

python-threatexchange/threatexchange/dataset.py
Dcallies: More guidance: Just storing the time should be enough, as long as it's inclusive, since it's safe to refetch and re-apply old updates.

python-threatexchange/threatexchange/cli/experimental_fetch.py
Dcallies: Chatted with offline, the current impl is indeed 90 days. Suggest putting a comment here about why refetch is needed in general (and it's relation to the tombstone duration.

Dcallies (approved):

python-threatexchange/threatexchange/cli/experimental_fetch.py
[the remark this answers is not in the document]
InderH: I initially wrote the first version of experimentalfetch using `requests` but IIRC it was asked that I switch back to using the existing way of doing things, unfortunately I don't have the time to update this atm but would you like me to leave a TODO?

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: Added todo

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: Added comment

python-threatexchange/threatexchange/cli/experimental_fetch.py
InderH: Updated both to be "types" instead, may take some time to propagate

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -23,9 +25,19 @@ class ExperimentalFetchCommand(command_base.Command):
     """
 
     PROGRESS_PRINT_INTERVAL_SEC = 30
+    # We use a DEFAULT_REFETCH_SEC of 85 days because threat_updates tombstones last for 90 days
+    # If you don't fetch for updates within the 90 days you may miss some deletion notifications
```
Dcallies, line 29: (This is basically a repeat of the comment below): Don't reference the literal value of things in your code comments. If we ever need to change them, it becomes a pain to try and trace them down.

```suggestion
    # If a client does not resume tailing the threat_updates endpoint fast enough, 
    # deletion records will be removed, making it impossible to determine which
    # records should be retained without refetching the entire dataset from scratch.
    # The current implementation will retain for 90 days: TODO: Link to documentation
```

Dcallies:
Talked with Inder offline, we'll land and iterate. Lint failure, but we'll fix as a followup.

PR #347, opened 2020-12-21
Use requests instead of urllib for TE.py

schatten:
Summary
---------
Because requests is much more high-level, it allows easily setting up
retries / timeouts and can be configured going ahead to support proxies
and .netrc and all the goodness!

Test Plan
---------
Two changes were made. The GET and POST commands.

For GET, I removed 'threatexchange_samples' and ran a random command. The directory was setup as expected using the requests code.

**Output**
```shell
$> threatexchange match text "ball now?"                                                                          ~
Looks like you are running this for the first time. Fetching some sample data.
raw_text: 3
url: 1
pdq: 138
video_tmk_pdqf: 1
photo_md5: 1
video_md5: 2
trend_query: 1  
```

For the POST request, I'm still working on it. 
```shell
$> threatexchange label false_positive descriptor 3261912580534814 
```

This did not throw any errors, but I'll have to check if it is getting correctly received on the server.

### DIFF
--- a/python-threatexchange/setup.py
+++ b/python-threatexchange/setup.py
@@ -53,6 +53,7 @@ setup(
     packages=find_packages(exclude=["tests*"]),
     install_requires=[
         "python-Levenshtein",
+        "requests",
     ],
     extras_require=extras_require,
     entry_points={"console_scripts": ["threatexchange = threatexchange.cli.main:main"]},
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -68,6 +68,7 @@ def get_fb_graph_api():
 
     TODO: Identify if requests Session object can be used across threads in a
     `concurrent.futures.ThreadPoolExecutor`
+    Filed:  https://github.com/facebook/ThreatExchange/issues/348
     - because a session is created per call to get_fb_graph_api(), the underlying conn
       pool can't be used.
     - this might become a performance concern if the pre-flight latencies become
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -3,7 +3,7 @@
 """
 This is an entire copy of a file from ThreatExchange/hashing
 
-TODO: Slim down to only what we need, and switch to using requests
+TODO: Slim down to only what we need
 """
 
 import copy
@@ -45,12 +45,12 @@ class TimeoutHTTPAdapter(HTTPAdapter):
 
 
 DEFAULT_TE_BASE_URL = "https://graph.facebook.com/v6.0"
-retry_strategy = Retry(
+_retry_strategy = Retry(
     total=4,
     status_forcelist=[429, 500, 502, 503, 504],
     method_whitelist=["HEAD", "GET", "OPTIONS"],
 )
-adapter = TimeoutHTTPAdapter(timeout=60, max_retries=retry_strategy)
+_adapter = TimeoutHTTPAdapter(timeout=60, max_retries=_retry_strategy)
 
 
 def get_fb_graph_api():
@@ -78,7 +78,7 @@ def get_fb_graph_api():
       requests.get() equivalents
     """
     session = requests.Session()
-    session.mount(DEFAULT_TE_BASE_URL, adapter=adapter)
+    session.mount(DEFAULT_TE_BASE_URL, adapter=_adapter)
     return session
 
 
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -73,7 +73,7 @@ class Net:
 
     @classmethod
     def getJSONFromURL(self, url):
-        """ Perform an HTTP GET request, and return the JSON response payload.
+        """Perform an HTTP GET request, and return the JSON response payload.
         Same timeouts and retry strategy as `fb_graph_api` above.
         """
         return fb_graph_api.get(url).json()
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -14,11 +14,33 @@ import json
 import os
 import re
 import requests
+
+from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
 import urllib.parse
 
-from .common import TimeoutHTTPAdapter
+
+DEFAULT_TIMEOUT = 5  # seconds
+
+class TimeoutHTTPAdapter(HTTPAdapter):
+    """
+    Plug into requests to get a well-behaved session that does not wait for eternity.
+    H/T: https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/#setting-default-timeouts
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.timeout = DEFAULT_TIMEOUT
+        if "timeout" in kwargs:
+            self.timeout = kwargs["timeout"]
+            del kwargs["timeout"]
+        super().__init__(*args, **kwargs)
+
+    def send(self, request, **kwargs):
+        timeout = kwargs.get("timeout")
+        if timeout is None:
+            kwargs["timeout"] = self.timeout
+        return super().send(request, **kwargs)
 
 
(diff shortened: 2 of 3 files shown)


Dcallies:
Your test plan looks fine to me. If we wanted to be crazy we could trace all the existing locations to try and reproduce all the existing behavior. I know from your messages you attempted a label as well and found different bugs.

python-threatexchange/setup.py
```diff
@@ -53,6 +53,7 @@
     packages=find_packages(exclude=["tests*"]),
     install_requires=[
         "python-Levenshtein",
+        "requests",
```
Dcallies, line 56: <3

python-threatexchange/threatexchange/TE.py
```diff
@@ -13,15 +13,32 @@
...
+    method_whitelist=["HEAD", "GET", "OPTIONS"],
+)
+adapter = TimeoutHTTPAdapter(timeout=60, max_retries=retry_strategy)
+
+# `fb_graph_api`: Custom requests session that provides
+# - retries: 4 retries on GET requests for 429 or 5XX error codes
+# - timeout: 60 seconds. Can be adjusted at the fb_graph_api.get(.., timeout=FOO) level
+
+fb_graph_api = requests.Session()
+fb_graph_api.mount(DEFAULT_TE_BASE_URL, adapter=adapter)
```
Dcallies, line 37: Hmm, in general I would say avoid doing work in the initialization of a module. However, I haven't read enough on requests to see if this is doing any work or if it's just populating fields on an unopened connection. Do you know whether this is doing anything significant?


The alternative would be to have a get_client() method which is memoized.

python-threatexchange/threatexchange/TE.py
```diff
@@ -352,17 +349,7 @@ def _postThreatDescriptor(self, url, postParams, showURLs, dryRun):
...
-
-            # Decode the outputs from the POST
-            # This is a Python 'bytes'
-            response = response.read()
-            # Now make it a string
-            response = response.decode("utf-8")
-            responseBody = json.loads(response)
-            responseCode = None
-
-            return [None, None, responseBody]
```
Dcallies, line 365: lol

python-threatexchange/threatexchange/TE.py
```diff
@@ -54,32 +71,12 @@ class Net:
...
-
-    @classmethod
-    def tryGET(self, url):
-        try:
-            # The timeout is a heuristic
-            response = urllib.request.urlopen(url, None, 60)
-            return [response, None]
-        except urllib.error.HTTPError as e:
-            return [None, e]
+        """Perform an HTTP GET request, and return the JSON response payload.
```
Dcallies, line 76: Nit: I think most of the code uses

```
"""
This style of docbloc
"""

"""Instead of this style
"""

File: python-threatexchange/threatexchange/common.py [hunk 5]
```diff
@@ -26,3 +27,25 @@ def camel_case_to_underscore(name: str) -> str:
...
+    """Plug into requests to get a well-behaved session that does not wait for eternity.
+    H/T: https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/#setting-default-timeouts
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.timeout = DEFAULT_TIMEOUT
+        if "timeout" in kwargs:
+            self.timeout = kwargs["timeout"]
+            del kwargs["timeout"]
+        super().__init__(*args, **kwargs)
```
Dcallies, line 45: Nit: I think this should live in TE.py, since all the network stuff is currently isolated there, and I doubt you'll reuse it anywhere else.

python-threatexchange/threatexchange/common.py
```diff
@@ -26,3 +27,25 @@ def camel_case_to_underscore(name: str) -> str:
     """
     s1 = re.sub(r"(.)([A-Z][a-z]+)", r"\1_\2", name)
     return re.sub(r"([a-z0-9])([A-Z])", r"\1_\2", s1).lower()
+
+
+DEFAULT_TIMEOUT = 5  # seconds
+
+
+class TimeoutHTTPAdapter(HTTPAdapter):
+    """Plug into requests to get a well-behaved session that does not wait for eternity.
```
Dcallies, line 36: .

(. = repeat of previous comment)

python-threatexchange/threatexchange/TE.py
```diff
@@ -13,15 +13,32 @@
 # General Python dependencies
```
Dcallies, line 13: (Can't actually comment on L6)

Change the TODO to no longer mention requests

Dcallies (approved):
Overall LGTM,

Can you additionally hunt down the retry code in the `experimentalfetch` function that is hand rolling its retries and add a TODO there?

python-threatexchange/threatexchange/TE.py
schatten: The 'Session' Object would use connection pooling when connecting to the same endpoint.

It is populating the timeout value, but also implementing a retry strategy.

About doing work in module init, the only issue I see is with reusing the same Session object when doing concurrent fetches as we are doing in [FetchCommand](https://github.com/facebook/ThreatExchange/blob/master/python-threatexchange/threatexchange/cli/fetch.py#L204). I'll check if reusing the same Session object is possible/advisable in a ThreadPoolExecutor environment.

FWIW: `requests.get` [creates a new session each time](https://github.com/psf/requests/blob/master/requests/api.py#L60). I'll update the PR with my findings.

python-threatexchange/threatexchange/TE.py
```diff
@@ -13,15 +13,77 @@
...
+        return super().send(request, **kwargs)
+
+
+DEFAULT_TE_BASE_URL = "https://graph.facebook.com/v6.0"
+retry_strategy = Retry(
+    total=4,
+    status_forcelist=[429, 500, 502, 503, 504],
+    method_whitelist=["HEAD", "GET", "OPTIONS"],
+)
+adapter = TimeoutHTTPAdapter(timeout=60, max_retries=retry_strategy)
```
Dcallies, line 53: Nit, not blocking: would prefer this also be scoped to the get_fb_graph_api function to prevent from adding things to the global namespace, or at the very least prefixed with underscore (_) to indicate that it's part of the implementation detail.

Dcallies (approved):

python-threatexchange/threatexchange/TE.py
```diff
@@ -13,15 +13,77 @@
...
+    Ideally, should be used within a context manager:
+    ```
+    with get_fb_graph_api() as fb_graph_api:
+        fb_graph_api.get()...
+    ```
+
+    If using without a context manager, ensure you end up calling close() on the returned value.
+
+    TODO: Identify if requests Session object can be used across threads in a
+    `concurrent.futures.ThreadPoolExecutor`
```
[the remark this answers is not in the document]
schatten: I think I know why you've not faced issues.

tl;dr: the concern is when **multiple** hosts are connected to via the same PoolManager. We only connect to a single host.
---
PoolManager [1] transparently manages creating and destroying ConnectionPools. One ConnectionPool per host [2]. 

The issue reported on [github](https://github.com/urllib3/urllib3/issues/1252) talks about using multiple hosts and how the ConnectionPool retrieved from the PoolManager might get closed if too many hosts are connected to, frequently. Not an issue for us because we will not be connecting to multiple hosts.

Now the one on [stackoverflow](https://stackoverflow.com/questions/18188044/is-the-session-object-from-pythons-requests-library-thread-safe/32503750) is different. That is about the individual threads having access to Session internals and overwriting something like cookies, which it does without taking locks.

So, while the usage of requests is not thread-safe, the way we use it, it is. However, I'm still hesitant to re-use sessions across threads because of how our usage might evolve. I'd want to see if there are legit performance issues before we attempt re-using the same session across threads.

[1]: https://urllib3.readthedocs.io/en/latest/reference/urllib3.poolmanager.html
[2]: https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html

python-threatexchange/threatexchange/TE.py
schatten: This was done.

PR #350, opened 2020-12-23
Prototype HMA - First Pass

bodnarbm:
Summary
---------

This is an initial pass at a prototype for the hasher-matcher-actioner (HMA) system. In it's rough state, it requires some manual setup and data uploads, but it has successfully indexed, hashed, and matched PDQ hashes in a test environment. This will be a continuing prototype.

Test Plan
---------

1. Built and pushed the lambda function code as a docker image using the provided script.
2. Deployed the environment using `terraform apply` with the prefix of `bodnarbm`
3. Uploaded the threatexchange samples data set from the `threatexchange cli` default config to the `threat_exchange_data/` folder of the S3 bucket created for my environment.

Screen Shot 2020-12-23 at 4 44 08 PM

4. Confirmed the pdq indexer lambda created the pdq hash index

Screen Shot 2020-12-23 at 4 44 45 PM

5. Subscribed my email account to the `matches` sns topic to test email sending there.

6. Uploaded some photos to the `images/` folder of the S3 bucket.

7. Waited a few minutes and received the matches emails from the SNS topic.

8. Explored the dynamodb table in the AWS console and confirmed that it has the content hashes and match information.


bodnarbm:
Just wanted to add a quick note that a lot of this, especially the python code, can use some additional work (for example optimizing when we pull down the index in the matcher versus reusing what we already have in the lambda). So this is a fairly rough first pass.

Also, to clean up the git history, I suggest when we merge this, we do so with a squash commit.

hasher-matcher-actioner/hashing-data/main.tf
```diff
@@ -0,0 +1,222 @@
...
+
+resource "aws_sns_topic_policy" "image_notification_topic_policy" {
+  arn = aws_sns_topic.image_notification_topic.arn
+
+  policy = data.aws_iam_policy_document.image_notification_topic_policy.json
+}
+
+# DynamoDB Datastore
+
+resource "aws_dynamodb_table" "datastore" {
```
BarrettOlson, line 166: Is it worth documenting here or elsewhere what these keys are intended to be associated with? I think one would be able to figure it out by reviewing the lambda code but it's a little confusing.

BarrettOlson (commented):
Looks good! added a few preliminary comments/questions

hasher-matcher-actioner/lambdas/requirements.txt
```diff
@@ -0,0 +1,2 @@
+boto
```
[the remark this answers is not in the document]
bodnarbm: Good catch. boto3 is preinstalled in the docker container, hence this was not an issue. But yes, we should specify the correct package here.

hasher-matcher-actioner/hashing-data/main.tf
bodnarbm: Probably not here in the terraform code, but we should probably document this in the overall design documentation, which we should publish here in the repo when we are closer to a release.

BarrettOlson (approved):
LGTM!

hasher-matcher-actioner/CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Config Files
+
+Before using terraform, you will need to provide some additional configuration, examples of which are provided in this folder using the same name, but with example suffixed to the end.
+
+1. `terraform.tfvars`: In this file, you will want to define the values for the variables defined in [variables.tf](variables.tf). Only variables without defaults are required, though if you are setting up an environment for developing on the HMA prototype, you likely want to override the default `prefix` to allow having an isolated environment. See the terraform docs on [input variables](https://www.terraform.io/docs/configuration/variables.html) for more information on providing variable values and overrides.
```
Dcallies, line 23: One thing that threw me off while I was getting onboard was that you can't finish backend.tf until you've done the building of the image, but the order is reversed here.

I assume that this problem goes away if we can check in an image?

hasher-matcher-actioner/lambdas/pdq_matcher.py
```diff
@@ -0,0 +1,89 @@
...
+import datetime
+
+from threatexchange.hashing.pdq_faiss_matcher import PDQFlatHashIndex, PDQMultiHashIndex
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+s3_client = boto3.client("s3")
+sns_client = boto3.client("sns")
+dynamodb = boto3.resource("dynamodb")
```
Dcallies, line 17: For the future, recommend putting the logging setup into a function and then calling the function under __main__ (or whatever the lambda equivalent is) - you'll never know when you'll want to import the file as a library (usually as part of testing).

Calling getLogger() without context will get you the root logger, and then this code mutate state on that logger - it becomes hard to finely control logging setup when importing this file.

hasher-matcher-actioner/CONTRIBUTING.md
bodnarbm: Good point. Going to swap these two sections (config files and building and pushing the docker image).

And yes, once we have a production image that we are using, we can provide that as the default for the `hma_lambda_docker_uri` variable and then the order will not matter as much.

hasher-matcher-actioner/lambdas/pdq_matcher.py
bodnarbm: This is a good call out. Unfortunately, the lambda bootstrapping process as far as I am aware does not support `main` or any other one-time run setups outside of the global context. We could invoke a function at the start of each request in the handler and have it do setup based on if a global variable is false and then set that variable to true. That is one common pattern I have seen in the past.

For the logger specifically though, we could also look at using the logger AWS provides in AWS Lambda Powertools for Python (https://awslabs.github.io/aws-lambda-powertools-python/core/logger/). That has been support for children and root loggers and can help address that concern too.

Dcallies: Sounds good to me!

PR #351, opened 2020-12-28
Add support for PDQ_OCR signal type

BarrettOlson:
Summary
---------

This PR adds fetching and matching functionality for the PDQ_OCR signal type. 
(Experimental: If https://github.com/tesseract-ocr/tesseract is installed it can also attempt to get the text from a given file and use it for matching.)

Test Plan
---------

```
$ threatexchange fetch --full 
video_md5: 2
raw_text: 3
pdq_ocr: 2
video_tmk_pdqf: 1
photo_md5: 1
url: 1
trend_query: 1
pdq: 138
```

```
$ threatexchange match photo -H '<PDQ-HASH>,This is a sample text string' --show-false-positives 
<ThreatDescriptor ID>  pdq_ocr disputed
```

### DIFF
--- a/python-threatexchange/setup.py
+++ b/python-threatexchange/setup.py
@@ -25,10 +25,7 @@ extras_require = {
         "pdqhash>=0.2.2",
         "Pillow",
     ],
-    "pdq_ocr": [
-        "numpy",
-        "pdqhash>=0.2.2",
-        "Pillow",
+    "ocr": [
         "pytesseract",
     ],
 }
--- a/python-threatexchange/setup.py
+++ b/python-threatexchange/setup.py
@@ -25,6 +25,12 @@ extras_require = {
         "pdqhash>=0.2.2",
         "Pillow",
     ],
+    "pdq_ocr": [
+        "numpy",
+        "pdqhash>=0.2.2",
+        "Pillow",
+        "pytesseract",
+    ],
 }
 
 all_extras = set(sum(extras_require.values(), []))
--- a/python-threatexchange/threatexchange/hashing/ocr_utils.py
+++ b/python-threatexchange/threatexchange/hashing/ocr_utils.py
@@ -3,7 +3,7 @@
 
 """
 Util file for Optical character recognition (OCR) related functions.
-Use of pytesseract will required additional libaries, see https://github.com/madmaze/pytesseract#installation
+Use of pytesseract requires additional libaries already be installed, see https://github.com/madmaze/pytesseract#installation
 """
 
 import pdqhash
--- a/python-threatexchange/threatexchange/hashing/ocr_utils.py
+++ b/python-threatexchange/threatexchange/hashing/ocr_utils.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Util file for Optical character recognition (OCR) related functions.
+Use of pytesseract will required additional libaries, see https://github.com/madmaze/pytesseract#installation
+"""
+
+import pdqhash
+import pathlib
+import pytesseract
+from PIL import Image, ImageOps
+
+
+def text_from_image_file(path: pathlib.Path):
+    """
+    Given a path to a file return predicted OCR text
+    Current Supported file types: jpg
+    """
+    img_pil = Image.open(path)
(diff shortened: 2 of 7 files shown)


python-threatexchange/threatexchange/signal_type/pdq.py
```diff
@@ -50,6 +50,8 @@ def match_file(self, file: pathlib.Path) -> t.List[signal_base.SignalMatch]:
         return self.match_hash(pdq_hash)
 
     def match_hash(self, signal_str: str) -> t.List[signal_base.SignalMatch]:
+        if len(signal_str) != BITS_IN_PDQ / 4:
```
bodnarbm, line 53: As I was thinking through why someone would provide a hash with the number of hex characters for the signal_str, I was thinking this might be a copy and paste error. Is this a case where we would want to print an error message to the user to let them know we are ignoring the hash because it's not the right length?

python-threatexchange/threatexchange/hashing/ocr_utils.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Util file for Optical character recognition (OCR) related functions.
+Use of pytesseract will required additional libaries, see https://github.com/madmaze/pytesseract#installation
```
bodnarbm, line 6:
```suggestion
Use of pytesseract requires additional libaries already be installed, see https://github.com/madmaze/pytesseract#installation
```

python-threatexchange/setup.py
```diff
@@ -25,6 +25,12 @@
         "pdqhash>=0.2.2",
         "Pillow",
     ],
+    "pdq_ocr": [
```
bodnarbm, line 28: (Ignorable) This is fine with me. An alternative though would be that if they want to do PDQ_OCR signal types, they should install both the `pdq_hasher` extra and an `ocr` extra. Then this would reduce to just installing `pytesseract`. Might just help cut down a little on the dependency duplication in case you need to update it in the future.

python-threatexchange/threatexchange/signal_type/pdq_ocr.py
```diff
@@ -0,0 +1,98 @@
...
+    additional 'distance' using levenshtein string comparison.
+
+    Accuracy of matches will depend on quality of OCR implementations and selected thresholds.
+    (ocr_utils uses the open source pytesseract wrapper of tesseract)
+    """
+
+    INDICATOR_TYPE = "HASH_PDQ_OCR"
+    TYPE_TAG = "media_type_photo"
+
+    # This may need to be updated (TODO make more configurable)
```
bodnarbm, line 36: Depending on how we make this more configurable in the future, this could help inform what is needed in the HMA system too. For example, if we find that these settings need to be specific to a collaboration config, then that helps use know that our PDQ matchers in the HMA system also need allow for collab specific settings here. But if we find that it is only user specific, then for the HMA system this could be set as a terraform variable.

Just wanted to point out how the outcome for the cli tool can help us understand the needs for the HMA system too.

bodnarbm (approved):
LGTM. We might want to start adding instructions in the README file about the different installation choices and any system level dependencies required for them to work (such as having tesseract installed).

BTW, if you ever wanted to add some tests requiring tesseract to the python-threatexchange test runner, you can install tesseract on the workflow runner [using the same method as pytesseract](https://github.com/madmaze/pytesseract/blob/v0.3.7/.github/workflows/ci.yaml#L52-L53) (just maybe with the `eng` language files instead of the `fra` language files depending on your test cases).

Thanks @BarrettOlson

python-threatexchange/threatexchange/signal_type/pdq.py
BarrettOlson: This was added because without it, if you give it an photo md5 or pdq_ocr string using the -H flag, for 'match photo' it will assert because it tries to match against all types photo hashes. I can add a comment.

python-threatexchange/threatexchange/hashing/ocr_utils.py
BarrettOlson: Thanks, for now I am going to leave it at this and update the readme in another PR that will provide context on this and the other libraries we added.

PR #354, opened 2020-12-30
[hma] Add initial Makefile

Dcallies:
Summary
---------

I don't know a lot about Make, but I think it's a useful way to record setup steps. The main advantage here is that you can bootstrap the dev instance as a Facebook dev without reading too much.

Test Plan
---------

Run all make commands locally.

### DIFF
--- a/hasher-matcher-actioner/Makefile
+++ b/hasher-matcher-actioner/Makefile
@@ -1,0 +1,47 @@
+.PHONY: all docker upload_docker clean
+
+shell-or-die = $\
+	$(eval sod_out := $(shell $(1); echo $$?))$\
+	$(if $(filter 0,$(lastword $(sod_out))),$\
+		$(wordlist 1, $(shell echo $$(($(words $(sod_out)) - 1))), $(sod_out)),$\
+		$(error $(1) exited with status $(lastword $(sod_out))))
+
+USER ?= $(call shell-or-die,whoami)
+REPOSITORY_NAME = hma-lambda-dev
+DOCKER_TAG ?= ${USER} # todo - change to ${USER}-latest?
+DOCKER_URI = $(call shell-or-die,docker images --filter=reference='*/${REPOSITORY_NAME}:${DOCKER_TAG}' --format='{{.Repository}}')
+
+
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/Makefile
```diff
@@ -0,0 +1,47 @@
+.PHONY: all docker upload_docker clean
```
Dcallies, line 1: Need to add the rest of the phonies

hasher-matcher-actioner/Makefile
```diff
@@ -0,0 +1,47 @@
+.PHONY: all docker upload_docker dev_create_configs dev_create_instance dev_destroy_instance dev_clear_configs clean
+
+shell-or-die = $\
```
BarrettOlson, line 3: Could you add a comment explaining this command?  (even if it is ~common make/shell magic)

hasher-matcher-actioner/Makefile
```diff
@@ -0,0 +1,47 @@
...
+	$(eval sod_out := $(shell $(1); echo $$?))$\
+	$(if $(filter 0,$(lastword $(sod_out))),$\
+		$(wordlist 1, $(shell echo $$(($(words $(sod_out)) - 1))), $(sod_out)),$\
+		$(error $(1) exited with status $(lastword $(sod_out))))
+
+USER ?= $(call shell-or-die,whoami)
+REPOSITORY_NAME = hma-lambda-dev
+DOCKER_TAG ?= ${USER} # todo - change to ${USER}-latest?
```
BarrettOlson, line 11: Makefiles aren't known for their help text but either somewhere here or in contributing.md could you explain how to configure this? (i.e. example command usage)

BarrettOlson (approved):
LGTM (also not overly familiar with Make though) I can tell those command shortcuts are going to be useful! Made some nit comments: could you add a command example? Thanks!

Dcallies:
I made slightly too many forks of the repo and am having trouble figuring out the right git magic to get the state I've put here. I'm going to do the comments as a followup.

PR #355, opened 2020-12-31
[hma] sync lockfile

Dcallies:
I'm not entirely sure how this happened, but I think I need to commit this.


BarrettOlson (approved):
Looks not unexpected based on: https://www.terraform.io/docs/configuration/dependency-lock.html#new-provider-package-checksums

PR #358, opened 2021-01-29
[py-tx] Fix formatting of top level help

Dcallies:
### Summary 
Was just bothering me. I also wrote the code that broke it,
so /shrug

### Test Plan
Before:
```
$ threatexchange --help
verbs:
  {fetch,experimental-fetch,match,label}
                        which action to do
    fetch               download content from ThreatExchange to disk
    experimental-fetch  wARNING: This is experimental, you probably want
to
                        use "Fetch" instead
...
```

After:
```
$ threatexchange --help
verbs:
  {fetch,experimental-fetch,match,label}
                        which action to do
    fetch               download content from ThreatExchange to disk
    experimental-fetch  warning: This is experimental, you probably want
to
                        use "Fetch" instead
...
```

Summary
---------

Test Plan
---------

### DIFF
--- a/python-threatexchange/threatexchange/cli/command_base.py
+++ b/python-threatexchange/threatexchange/cli/command_base.py
@@ -71,8 +71,11 @@ class Command:
     def get_help(cls) -> str:
         """The short help of the command"""
         line = cls.get_description().strip().partition("\n")[0]
+        # Good luck debugging this! Slightly reformat short description
+        # (toplevel --help)
         if line[0].isupper():
-            line = line[0].lower() + line[1:]
+            first_word, sp, rem = line.partition(" ")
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/cli/command_base.py
```diff
@@ -71,8 +71,11 @@ def get_description(self) -> str:
     def get_help(cls) -> str:
         """The short help of the command"""
         line = cls.get_description().strip().partition("\n")[0]
+        # Good luck debugging this! Slightly reformat short description
```
BarrettOlson, line 74: :(

BarrettOlson (approved):

Issue #357, opened 2021-01-28
[py-tx] Malformed trend_queries will crash `threatexchange fetch`


Dcallies:
If you have a malformed trend query (invalid regex), it crashes the entire fetch.

Instead, it should just say that it's not a trend query/extract no value.

Now might be a good time to put an "fail to parse" / error condition/counter/log.

PR #360, opened 2021-01-29
[py-tx] Paper over signal type deserialization errors

Dcallies:
Summary
---------

Facebook is exporting malformed trend_queries in some cases. While they shouldn't do that, the library probably also shouldn't crash.

Fixes: https://github.com/facebook/ThreatExchange/issues/357

Test Plan
---------

Run fetch on collaboration with malformed trend_queries, doesn't asplode

### DIFF
--- a/python-threatexchange/threatexchange/cli/fetch.py
+++ b/python-threatexchange/threatexchange/cli/fetch.py
@@ -206,6 +206,7 @@ class FetchCommand(command_base.Command):
             seen_td_ids = set()
 
             counts = collections.Counter()
+            errors = collections.Counter()
 
             tags_to_fetch = dataset.config.labels
             only_first_fetch = False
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/cli/fetch.py
```diff
@@ -295,6 +301,10 @@ def consume_descriptors(dq: collections.deque) -> int:
                 continue
             dataset.store_cache(signal_type)
             print(f"{signal_name}: {counts[signal_name]}")
+        if errors:
```
BarrettOlson, line 304: Is it worth (optionally) logging the first N-IDs that created the errors?

BarrettOlson (approved):
👍  Not asploding is good.

python-threatexchange/threatexchange/cli/fetch.py
Dcallies: Potentially, but the place I would want that is in a logging file or a verbose option, both of which seemed out of scope for now.

Issue #359, opened 2021-01-29
[py-tx] Connection Pool is 'sploding


Dcallies:
Looks like @schatten's comment about pooling was right! https://github.com/urllib3/urllib3/issues/1252

Have to figure out how I patched this over in a previous project:

Traceback (most recent call last):
  File "/home/dcallies/.venv/threatexchange/bin/threatexchange", line 33, in <module>
    sys.exit(load_entry_point('threatexchange', 'console_scripts', 'threatexchange')())
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 149, in main
    execute_command(namespace)
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 87, in execute_command
    command.execute(dataset)
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/cli/fetch.py", line 279, in execute
    consume_descriptors(pending_futures)
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/cli/fetch.py", line 227, in consume_descriptors
    descriptors = item.result()
  File "/usr/lib64/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/lib64/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/lib64/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/cli/fetch.py", line 319, in _fetch_descriptors
    for td_json in TE.Net.getInfoForIDs(td_ids):
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/TE.py", line 198, in getInfoForIDs
    response = self.getJSONFromURL(url)
  File "/data/users/dcallies/fork_te/ThreatExchange/python-threatexchange/threatexchange/TE.py", line 126, in getJSONFromURL
    return fb_graph_api.get(url).json()
  File "/home/dcallies/.venv/threatexchange/lib64/python3.6/site-packages/requests/sessions.py", line 543, in get
    return self.request('GET', url, **kwargs)
[...]

PR #361, opened 2021-01-29
[py-tx] Don't reuse adapters

Dcallies:
Summary
---------

The connect pool is controlled by the adapter, which is closed when the session is closed. If you use new sessions, but reuse the adapter, you'll keep closing the connection pool, which then leads to exceptions.

Fixes: https://github.com/facebook/ThreatExchange/issues/359

Test Plan
---------

Do a fetch on a collaboration with at least 10k items, asplodes before, doesn't asplode afterwards.

### DIFF
--- a/python-threatexchange/threatexchange/TE.py
+++ b/python-threatexchange/threatexchange/TE.py
@@ -13,15 +13,12 @@ import json
 # General Python dependencies
 import os
 import re
-import requests
-
-from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util.retry import Retry
-
+import threading
 import urllib.parse
 
-
-DEFAULT_TIMEOUT = 5  # seconds
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 
 class TimeoutHTTPAdapter(HTTPAdapter):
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/TE.py
```diff
@@ -13,35 +13,28 @@
 # General Python dependencies
 import os
 import re
-import requests
+import threading
```
BarrettOlson, line 16: I see this import was added but I am missing where it was used.

python-threatexchange/threatexchange/TE.py
```diff
@@ -13,35 +13,28 @@
...
     Plug into requests to get a well-behaved session that does not wait for eternity.
     H/T: https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/#setting-default-timeouts
     """
 
-    def __init__(self, *args, **kwargs):
-        self.timeout = DEFAULT_TIMEOUT
-        if "timeout" in kwargs:
-            self.timeout = kwargs["timeout"]
-            del kwargs["timeout"]
+    def __init__(self, *args, timeout=5, **kwargs):
```
BarrettOlson, line 30: nit/ignorable: this 5 and 60 below are kind of magic numbers but imo fine until it's not a single point of maintenance

python-threatexchange/threatexchange/TE.py
```diff
@@ -13,35 +13,28 @@
...
         super().__init__(*args, **kwargs)
 
-    def send(self, request, **kwargs):
-        timeout = kwargs.get("timeout")
+    def send(self, request, *, timeout=None, **kwargs):
         if timeout is None:
-            kwargs["timeout"] = self.timeout
-        return super().send(request, **kwargs)
+            timeout = self.timeout
+        return super().send(request, timeout=timeout, **kwargs)
```
BarrettOlson, line 37: 👍

BarrettOlson (approved):
LGTM but added @schatten as a reviewer as well since based on the issue, he has more context :)

schatten (approved):
LGTM!

python-threatexchange/threatexchange/TE.py
Dcallies: Whoops, leftover from a different attempt at fixing this (threadlocal).

python-threatexchange/threatexchange/TE.py
Dcallies: Yea, I prefer to inline unless it needs to be referenced. Otherwise you have to hunt to the place in the code where it used to tell what it means anyways.

PR #363, opened 2021-01-30
[py-tx] fix directory creation with experimental-fetch

Dcallies:
Summary
---------

If you try and use experimental-fetch right off the bat, it fails to find the existing statefile directories. 

Test Plan
---------

```
rm ~/state_dir
threatexchange experimentalfetch
```


BarrettOlson (approved):
LGTM

PR #369, opened 2021-02-10
[py-tx] Refactor TE.py to api.py

Dcallies:
Summary
---------

TE.net is a dinosaur from another era, and it does a lot of interesting things that need various cleanups (class/module state, java-style class wrapper, naming conventions, directly using GET/POSTS in some commands). We're still a couple of rounds away from getting it into a good shape.

Test Plan
---------

I didn't test this super hard, so it's possible it breaks.
```
$ threatexchange fetch --full # sample data
$ threatexchange experimental-fetch --full # sample data
$ threatexchange label false_positive descriptor 3805925856125337 
```

### DIFF
--- a/python-threatexchange/threatexchange/descriptor.py
+++ b/python-threatexchange/threatexchange/descriptor.py
@@ -28,6 +28,7 @@ class ThreatDescriptor(t.NamedTuple):
         ...
     }
     """
+
     # TODO - do something smarter than this - static
     #        class variable problematic, currently set in main.py
     MY_APP_ID = -1
--- a/python-threatexchange/threatexchange/descriptor.py
+++ b/python-threatexchange/threatexchange/descriptor.py
@@ -8,8 +8,6 @@ Wrappers for the json returned by the ThreatExchange API to typed objects.
 import collections
 import typing as t
 
-from . import TE
-
 
 class ThreatDescriptor(t.NamedTuple):
     """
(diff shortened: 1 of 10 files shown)


python-threatexchange/threatexchange/descriptor.py
```diff
@@ -68,8 +68,7 @@ def is_false_positive(self) -> bool:
     @property
     def is_mine(self) -> bool:
         """This Descriptor is my App's Opinion"""
-        # TODO - come up with a way to do this that doesn't use class state
```
BarrettOlson, line 71: comment is still relevant or no?

python-threatexchange/threatexchange/descriptor.py
```diff
@@ -30,6 +28,8 @@ class ThreatDescriptor(t.NamedTuple):
         ...
     }
     """
+    # TODO - do something smarter than this
```
BarrettOlson, line 31: More context might be good for when we come back and fix this... e.g. 'Class attribute set by CLI, currently pulled from app_token in main.py (pending refactor).'

BarrettOlson (approved):
LGTM. I don't like the mess around `descriptor.ThreatDescriptor.MY_APP_ID` but can't think of a better alternative atm. 
Added a few comment nits. 

Run the formater before merge? Thanks!

python-threatexchange/threatexchange/descriptor.py
Dcallies: Still using class state :/

Dcallies: Moved the comment, but still functionally similar

PR #371, opened 2021-02-12
First stab at pull-request, master-push gh actions

schatten:
Summary
---------
There are two CI hooks as part of this PR.
1. PR Validator (runs `terraform fmt` and `terraform validate` without using a remote backend.
2. Master push validate and test (runs `terraform apply` and creates an integration test friendly environment with a custom prefix "tftest".

Note: Concurrent execution of workflows (multiple terraform apply) on the same prefix will have undefined results. This can be solved, if we see a lot of activity on master, we'll have another stab at this.

Test Plan
---------
Locally, `terraform fmt` was run, but to actually evaluate, the workflows must be run. Since we tie the execution of the workflow to the 'pull_request_target' [1] event, we must first merge this PR into master and raise a subsequent pull request to validate the PR. 'pull_request_target' workflows are run in the context of the merge base which in our case is the master branch.

1: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/hma-master-push.yaml
+++ b/.github/workflows/hma-master-push.yaml
@@ -1,0 +1,41 @@
+name: HMA Master push
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "hasher-matcher-actioner/**"
+      - ".github/workflows/hma-ci.yaml"
+defaults:
+  run:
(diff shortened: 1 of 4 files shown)


BarrettOlson (approved):
LGTM

PR #372, opened 2021-02-13
[py-tx] Big ol' Data Model and experimental-fetch Rewrite

Dcallies:
Summary
---------
I put up this PR after a few days of rewriting this over and over again. To try and finish it off, I'm forcing myself to commit what I consider a workable version before I continue on this path.

This diff does a few interesting things:
1. Recycles most of the previous implementation of experimental-fetch
2. Attempts a partial rewrite of the command to make it more similar to fetch
3. Defines the storage layer for threat_updates, which is basically to store a minified version of the raw json from the endpoint

My default plan is then to do:
```
Sync -> Raw json output
 => SimpleDescriptorRollup (since it works on tag fetch endpoint too) - or a variation thereof

HMA:
  => SimpleDescriptorRollup -> DB
    => SignalType marks which rollups belong to a signal type
       => Output file that is simple Hash -> DB_ids (many options here)
          => SignalType.buildIndex(ofile) -> Index
             => Index goes to matcher
               => SignalType.hash(content) -> Hash -> Index.match(Hash) -> {db_id: dist}
                  => Matcher fetched from DB to determine what to do

CLI:
fetch -> stores minified raw json
  => SimpleDescriptorRollup
    => SignalType marks which rollups belong to a signal type
       => Store indexes for signal type
match
  => determine SignalType from content type
    => load indexes
      => most indexes will be linear scan SignalType.hash(content) -> SignalType.match(content, stored_hash)
        => output match results from SimpleDescriptorRollup
```
However, much of the code is in placeholder locations, and not everything is connected.

Test Plan
---------
tl;dr:
1. Run fetch
2. Run experimental-fetch
See that the files are the same size (though not quite same content)

```
$  time threatexchange -c ~/test.te.cfg fetch

$  time threatexchange -c ~/test.te.cfg experimental-fetch
Currently at ages long past
Downloaded 501 updates. Currently at 67 days 17h16m52s ago
^C
Exception occurred! Attempting to save...

real    0m7.159s

# Resumes
$  time threatexchange -c ~/test.te.cfg experimental-fetch
Currently at 67 days 17h14m25s ago
Downloaded 1001 updates. Currently at 67 days 17h12m03s ago
...

$ ls ~/output_dir
total 900K
-rw-r--r-- 1 dcallies users   51 Feb 17 13:50 7.....threat_updates.te
-rw-r--r-- 1 dcallies users 330K Feb 16 18:32 DEBUG_STRING.te
-rw-r--r-- 1 dcallies users 132K Feb 16 18:32 HASH_PDQ_OCR.te
-rw-r--r-- 1 dcallies users  77K Feb 16 18:32 HASH_PDQ.te
-rw-r--r-- 1 dcallies users 344K Feb 16 18:32 HASH_TMK.te
-rw-r--r-- 1 dcallies users  750 Feb 16 18:32 RAW_URI.te

$ threatexchange -c ~/test.te.cfg fetch
# Doesn't asplode

$ ls # Edited with vim
  69 531732984141635.threat_updates.checkpoint
  37 fetch_checkpoint.te
132M raw_text.te
132M simple.TEXT_STRING.te
133M simple.DEBUG_STRING.te
 54K video_md5.te
 60K simple.HASH_MD5.te
 260 simple.HASH_VIDEO_MD5.te
 54K pdq_ocr.te
 56K simple.HASH_PDQ_OCR.te
 85K pdq.te
 89K simple.HASH_PDQ.te
 66M video_tmk_pdqf.te
 66M simple.HASH_TMK.te
 80K trend_query.te
5.1K simple.TREND_QUERY.te  # Backfill hasn't completed
3.5M simple.URI.te  # Regular URL has a bug!
```


Dcallies:
I've partially put this up so I can see my own file, will add a bunch of comments on it as well (feel free to join in).

There's probable a commit or two left after this.

python-threatexchange/threatexchange/api.py
```diff
@@ -68,16 +100,24 @@ class ThreatExchangeAPI:
...
+    def app_id(self):
+        return int(self.api_token.partition("|")[0])
 
-    def getJSONFromURL(self, url):
+    def getJSONFromURL(self, url, params=None, *, json_obj_hook: t.Callable = None):
         """
         Perform an HTTP GET request, and return the JSON response payload.
         Same timeouts and retry strategy as `_get_session` above.
         """
         with self._get_session() as session:
```
Dcallies, line 118: This needs to throw exceptions

python-threatexchange/threatexchange/api.py
```diff
@@ -68,16 +100,24 @@ class ThreatExchangeAPI:
...
-            return session.get(url).json()
+            x = requests.get(url, params=params or {})
+            return x.json(object_hook=json_obj_hook)
```
Dcallies, line 120: Leftover debug crap

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -1,191 +1,347 @@
...
-from ..signal_type import signal_base
+from ..descriptor import ThreatDescriptor, SimpleDescriptorRollup
 from ..indicator import ThreatIndicator
-from ..content_type import meta
+from ..signal_type import signal_base
 from . import command_base
 
 
-class ExperimentalFetchCommand(command_base.Command):
+class ThreatUpdate(t.NamedTuple):
```
Dcallies, line 22: Adds TODOs about where these live long term

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -1,191 +1,347 @@
...
+
+    @staticmethod
+    def te_threat_updates_fields():
+        return SimpleDescriptorRollup.te_threat_updates_fields()
+
+
+class ThreatUpdatesDelta:
+    def __init__(
+        self, privacy_group: int, start: int = 0, end: t.Optional[int] = None
+    ) -> None:
```
Dcallies, line 52: Explain the whole delta + store relationship in comments

Dcallies (commented):
Add a simple unittest that shims the API

Dcallies (commented):
Still need a test plan, noob.

python-threatexchange/threatexchange/threat_updates.py
```diff
@@ -0,0 +1,375 @@
...
+        api: ThreatExchangeAPI,
+        *,
+        limit: int = 0,
+        progress_fn=lambda x: None,
+    ) -> None:
+        """
+        Fetch from threat_updates to get a more up-to-date copy of the data.
+        """
+
+        # TODO actually implement fancy threading logic
```
BarrettOlson, line 204: ignorable: flagging that more error handingly is probably needed if you thread this because we try to save on error (might we worth a reminder in the comment).

python-threatexchange/threatexchange/threat_updates.py
```diff
@@ -0,0 +1,375 @@
...
+        # alternative - instead make the API give hints about where to start
+        # fetches, which will mean that fancy threading logic will be simpler
+        while not self.done:
+            for update in self.one_fetch(api):
+                progress_fn(update)
+                if limit:
```
BarrettOlson, line 210: ignorable: not a huge fan of this check i.e.
limit = 0 at start ==> no limit
number reduced to 0 ==> return 
has a bit of a smell but that could just be me. Does None not work because of typing?

python-threatexchange/threatexchange/threat_updates.py
```diff
@@ -0,0 +1,375 @@
...
+        self.privacy_group = privacy_group
+        self.checkpoint = None
+
+    @property
+    def fetch_checkpoint(self):
+        return self.checkpoint.fetch_checkpoint
+
+    def reset(self) -> None:
+        """Toss old state and begin anew"""
+        self.checkpoint = ThreatUpdateCheckpoint(time.time(), 0)
```
BarrettOlson, line 273: nit: why do we set `last_fetch_time` to `time.time()` here? 
shouldn't we wait for at least one successful fetch before then?

BarrettOlson (approved):
Added some inline questions/nit but nothing that seemed to be blocking. Please thoroughly update the Test Plan with runs of both fetch and experimental fetch. 

We definitely need unit tests but as we discussed since this only changes the logic of experimental I'm okay with your request to punt until the next PR/breaking changes.

python-threatexchange/threatexchange/threat_updates.py
```diff
@@ -0,0 +1,375 @@
...
+    A class for tracking a raw stream of /threat_updates
+
+    Any integration with ThreatExchange involves the creation of a local copy
+    of the data. /threat_updates sends changes to that data in either the form
+    of an insert/update or a delete.
+
+    A delta is a stream of updates and deletes, which when applied to an
+    existing database will give you the current set of live records.
+
+    As a parallelization trick, if you need to fetch between t1 and t3,
```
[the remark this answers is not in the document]
Dcallies: It is not yet

python-threatexchange/threatexchange/threat_updates.py
Dcallies: Will punt on this until the threading is written, but agree

python-threatexchange/threatexchange/threat_updates.py
Dcallies: Was probably trying to be too cheeky to avoid a nullable (since 0 is not a valid value).

python-threatexchange/threatexchange/threat_updates.py
Dcallies: Thought it made most sense here, but can move it somewhere else.

PR #374, opened 2021-02-23
[py-tx] index base

Dcallies:
Summary
---------

This is part of the diff series to move state out of SignalType directly and into its own class, though overall functionality will be basically the same.

A followup diff will add a simple linear search + pickle solution as a default implementation.

Test Plan
---------

black, mypy

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/index.py
+++ b/python-threatexchange/threatexchange/signal_type/index.py
@@ -27,6 +27,16 @@ T = t.TypeVar("T")
 
 
 class IndexMatch(t.Generic[T]):
+    """
+    Wrapper around a match to the index, which may or may not
+    be an actual match based on the distance settings for a source.
+
+    The index is free to optimize not returning values that are
+    greater than would be considered matches by a downstream (i.e.
+    31 is probably the max distance that would be used in PDQ)
+
+    NamedTuple can't be made generic, so here's an equivalent
+    """
 
     __slots__ = ["distance", "metadata"]
     distance: int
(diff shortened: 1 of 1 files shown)


schatten (changes requested):
Open questions:
1: Why is IndexMatch not conclusive? Why does it not have a boolean?
2: Do we expect the matcher to only get hashes? Could it get complex objects? eg. TextMMS / PDQ OCR.

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -0,0 +1,96 @@
...
+    Abstraction for an efficient matching technique.
+
+    This class can be thought of as just a Dict[hash, List[T]] interface
+    that can return the flattened result of multiple entries on a get().
+
+    The T can be whatever metadata might be useful, although be warned
+    that not all T may serialize correctly.
+
+    In cases where your underlying index has limited type support for
+    values, the easiest workaround is to just generate your own
```
Dcallies, line 50: Specifically mention FAISS

schatten (approved):
Based on the offline discussion with @Dcallies, I'm changing my decision:

1. The decision will actually be taken at the action stage, so index need not return a boolean.
2. For now, the types we are doing are string. Doing more than string has historically proven to be a scale blocker.

So, LGTM now. Go ahead.

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
```
jeberl, line 1: Looks good. Thanks for the explanation

jeberl (approved):

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -0,0 +1,96 @@
...
+            entry = internal_mapping[idx]
+            ret.append(entry)
+        return ret
+
+    def add(hash: str, meta: T):
+        idx = fancy_index_only_to_int.put(hash)
+        internal_mapping[idx] = meta
+    """
+
+    def query(self, hash: str) -> t.List[IndexMatch[T]]:
```
Dcallies, line 69: Add comment explaining why hash type is str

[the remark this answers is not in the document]
Dcallies: Chatted offline - in the long term, may need specialized hashes, but for now, can work around by doing work in the indexer itself. The reason for this is the specialized hashing is multi-pass. PDQ+OCR is pdq, then post-process if match. Text might be nilsimsa followed by levenshtein. If we start doing that becomes harder to generalize, and so would want to punt until this started becoming a problem.

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -0,0 +1,96 @@
...
+        Note that there may be duplicates of the hash type, i.e.
+
+        H1: M1
+        H1: M2
+
+        A later call to query(H1) should return both M1 and M2
+        """
+        raise NotImplementedError
+
+    def serialize(self, fout: t.BinaryIO) -> None:
```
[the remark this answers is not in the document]
Dcallies: The reason to do it this way is if data is going to be really huge where storing a copy of it in memory might be prohibitive. By using a buffer you back the buffer by a file and only have small chunks in memory.

PR #375, opened 2021-02-23
Move lambdas to hmalib: a top-level python package

schatten:
# Summary
Moved lambdas into a python package that is 'brought-in' to make a docker image.

## Why a python package?
a) Easier to create common pieces like a configured logger
b) More natural to write test cases (debatable). Because you are nudged  to write more unit-testable pieces.
c) Natural home for hma-cli should that become a reality
d) Will feel unnatural to hardcode credentials or resources

## Why move Dockerfile? What happened to `lambdas/`?
The main build artifacts of the hasher-matcher-actioner directory is the *code* that we expect partners to run. Potentially deliver this as a single Docker image with multiple entrypoints.

In that line of thought, the Dockerfile and requirements are *key*. So they deserve to be at the root and not within the lambdas directory. The lambdas directory is now nuked.

## What happens to terraform files?
I plan to have another diff out which will move the terraform files into their own directory 'iac' or 'terraform'? and clean up the root of the project.

# Test plan
Deployed the new lambda Docker image to a prefixed instance.

```shell
$> make docker
$> make upload_docker
$> terraform apply -var "prefix=$(whoami)"
..<types>yes
```

* Subscribed my work email to the topic `arn:aws:sns:us-east-1:<aws-account>:dipanjanm-matches20210223205422088500000002`
* Uploaded data, verified that index gets created.
* Uploaded images, verified that I get an email for matches.


Dcallies:
Super dumb question: What is a python package in this case? I'm conflating pip packaging and a few other things I think.

hasher-matcher-actioner/Makefile
```diff
@@ -16,7 +16,7 @@ all:
 	@echo >&2 "Must specify target. ${FAIL}"
 
 docker:
-	docker build -f lambdas/Dockerfile -t ${REPOSITORY_NAME} lambdas
+	docker build -f Dockerfile -t ${REPOSITORY_NAME} .
```
Dcallies, line 19: Thanks for updating this!

Dcallies (commented):
I think this is fine, but I'm not fully sure what I am looking at (super thanks for writing the descriptive summary).

schatten:
> Super dumb question: What is a python package in this case? I'm conflating pip packaging and a few other things I think.

[python docs](https://docs.python.org/3/tutorial/modules.html#packages) Define them as nested modules so you can refer to them as foo.bar.baz.

`pip packaging` makes packages distributable. I agree. This can get confusing. :P

hasher-matcher-actioner/Dockerfile
```diff
@@ -13,4 +13,4 @@ FROM public.ecr.aws/lambda/python:3.8 as prod
 ARG DEPS_PATH
 COPY --from=build $DEPS_PATH $DEPS_PATH
 ENV PYTHONPATH $DEPS_PATH
-COPY . .
+COPY hmalib ./hmalib
```
BarrettOlson, line 16: What does the `.` do here?

BarrettOlson (approved):
LGTM! 
Our first pass was focused on moving quickly, glad you are adding some structure.

hasher-matcher-actioner/Dockerfile
schatten: `.` refers to the present working directory both in docker context and in the docker image. I'm kind of certain omitting it would make no difference, though.

BarrettOlson: got it, if it is superfluous (as we both think it is) and not part of some weird best practices, we're probably better off removing it

PR #378, opened 2021-02-25
Create dedicated terraform directory

schatten:
Summary
---
Terraform files were littered across the codebase and unlike docker, they do not depend on context. Having a dedicated dir gives a cleaner root and helps you immediately identify terraform files.

You need to move your `.terraform` directory, backend.tf and terraform.tfvars files as well. You can use the following ~one~-two-liner to achieve said goal.

```
$> cd hasher-matcher-actioner
$> mv .terraform/ backend.tf terraform.tfvars terraform/
```

Test Plan
---
After doing the above changes, I ran `$> terraform apply -var "prefix=$(whoami)"` from within the new terraform directory.

It correctly identified the current state and suggested 0 changes.

So, it was a smooth transition from the root of hma to hma/terraform.

Updates to test plan
---
After making changes to the Makefile, I ran the changed targets (except dev_create_instance because that does a docker build 😡). All works!


BarrettOlson (commented):
Update the Makefile?

schatten:
> Update the Makefile?

Duh! I'll get right to it. It "felt" like I'm missing something.

BarrettOlson (approved):
LGTM

PR #379, opened 2021-02-26
Working flask app with dockerfile

schatten:
Summary
---
The flask app doesn't do anything but print "Hello, from the threatexchange API".

Adds a Dockerfile, adds flask to requirements and orchestrates everything with uwsgi, one of the canonical webservers writting in python.

Test Plan
---
```
$> docker build -t threatexchange-api:latest -f ThreatExchangeAPI.Dockerfile .
$> docker run -d -p 8080:80 threatexchange-api
```

And then localhost:8080 on your browser.

If you want to develop locally, use flask. First,
```
$> pip install -r requirements.txt # ensures you have flask installed. Only run once.
$> FLASK_APP=hmalib.service.app:app FLASK_ENV=development flask run
```

And go to the URL you are presented with in the console logs.


Dcallies (changes requested):
Discussed offline - would want to try and keep everything in one folder, rather than adding it in root folder. I think we'll copy stuff out after today, but makes cleanup harder. Don't think we'll need anything from hmalib in this pass (but could be wrong)

PR #381, opened 2021-02-26
Add mock APIs for matching single and multiple hashes

schatten:
Summary
---
Add a flask blueprint that provides one endpoint accessible via GET or POST as detailed in API.md

Test plan
---
First run the application
```
$> FLASK_APP=hmalite.app:app FLASK_ENV=development flask run
```

Then run the tests. Because the matches are random, you might need to run the tests a few times to get the outputs you want.

GET Request
```
$> curl -v "http://localhost:5000/v1/hashes/query?hash=tyui"
```

Note the status header.

POST Requests
```
$> curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"hashes":["xyz", "abc"]}' \
  http://localhost:5000/v1/hashes/query
```

### DIFF
--- a/hasher-matcher-actioner/hma-lite/API.md
+++ b/hasher-matcher-actioner/hma-lite/API.md
@@ -1,0 +1,18 @@
+Here's a super-rough first cut for an API
+
+# Match a new hash against the index
+
+```
+GET /v1/hashes/?hash=<PDQ-hash-in-hex>
+{
+  "hash": "<PDQ Hash>"
+}
+```
+
+`200 OK` with following shape if found, `404 Not Found` if not.
+
+```
+{
+  "hash": "<PDQ Hash as string>"
+}
+```
--- a/hasher-matcher-actioner/hma-lite/API.md
+++ b/hasher-matcher-actioner/hma-lite/API.md
@@ -1,15 +1,29 @@
 Here's a super-rough first cut for an API
 
-# Match a new hash against the index
+# Match a single hash against the index
 
 ```
-GET /v1/hashes/?hash=<PDQ-hash-in-hex>
+GET /v1/hashes/query?hash=hash_in_hex_string
 ```
 
-`200 OK` with following shape if found, `404 Not Found` if not.
+`200 OK` if found, `404 Not Found` if not.
+
+# Match multiple hashes against the index
+
+```
+POST /v1/hashes/query
+{
+    "hashes": [
+        ...hashes in hex string
+    ]
+}
+```
+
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hma-lite/API.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Match multiple hashes against the index
+
+```
+POST /v1/hashes/query
+{
+    "hashes": [
+        ...hashes in hex string
+    ]
+}
```
Dcallies, line 19: Think that this is okay to drop on first pass, but I think there's low cost to including

hasher-matcher-actioner/hma-lite/API.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+`200 OK` with following shape.
+
+```
+{
+  "matched_hashes": ["hash1 in hex string", "hash2 in hex string"],
+  "unmatched_hashes": ["hash3 in hex string", "hash4 in hex string"],
```
Dcallies, line 27: We'll probably rewrite later for HMA, but seems fine

hasher-matcher-actioner/hma-lite/API.md
```diff
@@ -0,0 +1,29 @@
+Here's a super-rough first cut for an API
+
+# Match a single hash against the index
+
+```
+GET /v1/hashes/query?hash=hash_in_hex_string
+```
+
+`200 OK` if found, `404 Not Found` if not.
```
Dcallies, line 9: LOL @ 404.

However, worried this is more confusing than a 200 with a json blob sent back, but don't think it's too different, and we can get feedback from externals first.

Dcallies (approved):
Its_happening.jpeg!

hasher-matcher-actioner/hma-lite/hmalite/matcher.py
```diff
@@ -0,0 +1,24 @@
...
+        hashes = request.json['hashes']
+        matches = [ hash for hash in hashes if _mock_is_index_match(hash)]
+        return jsonify({
+            "matched_hashes": matches,
+            "unmatched_hashes": [hash for hash in hashes if hash not in matches]
+        })
+    else:
+        hash = request.args.get('hash')
+        if _mock_is_index_match(hash):
+            return ('', 200)
```
eriksendc, line 17: Eventually there should be something returned in the body, I think, not just empty string (maybe more metadata about what the hash matches?) Just thinking out loud.

PR #382, opened 2021-02-26
Adds pdq index wrap for faiss index lib to signal_type

BarrettOlson:
Summary
---------

See title. 
~Draft as I am worried this might break imports. 

Test Plan
---------
Created Test
`pytest python-threatexchange/threatexchange/signal_type/tests/test_pdq_index.py`

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq_index.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq_index.py
@@ -1,17 +1,15 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 """
-Implementation of SignalTypeIndex abstraction for PDQ by wrapping hashing.pdq_faiss_matcher.
+Implementation of SignalTypeIndex abstraction for PDQ by wrapping 
+hashing.pdq_faiss_matcher.
 """
 
 import typing as t
 import pickle
 
-from threatexchange.signal_type.index import SignalTypeIndex, IndexMatch
+from threatexchange.signal_type.index import SignalTypeIndex, IndexMatch, T as IndexT
 from threatexchange.hashing.pdq_faiss_matcher import PDQMultiHashIndex
-
-T = t.TypeVar("T")
-PDQ_CONFIDENT_MATCH_THRESHOLD = 31
 
 
 class PDQIndex(SignalTypeIndex):
@@ -19,13 +17,16 @@ class PDQIndex(SignalTypeIndex):
     Wrapper around the pdq faiss index lib
     """
 
+    PDQ_CONFIDENT_MATCH_THRESHOLD = 31
+    T = IndexT
+
     def __init__(self, entries: t.Iterable[t.Tuple[str, T]]) -> None:
         super().__init__()
-        self.local_id_to_entry = dict()
-        hashes = list()
-        for i in range(len(entries)):
-            self.local_id_to_entry[i] = entries[i]
-            hashes.append(entries[i][0])
+        self.local_id_to_entry = {}
+        hashes = []
+        for i, entry in enumerate(entries):
+            self.local_id_to_entry[i] = entry
+            hashes.append(entry[0])
         self.index = PDQMultiHashIndex.create(
             hashes, custom_ids=self.local_id_to_entry.keys()
         )
@@ -34,10 +35,12 @@ class PDQIndex(SignalTypeIndex):
         """
         Look up entries against the index, up to the max supported distance.
         """
+
+        # query takes a signal hash but index supports banch quries hence [hash]
         results = self.index.search(
-            [hash], PDQ_CONFIDENT_MATCH_THRESHOLD, return_as_ids=True
+            [hash], self.PDQ_CONFIDENT_MATCH_THRESHOLD, return_as_ids=True
         )
-        matches = list()
+        matches = []
         for result_ids in results:
             for id in result_ids:
                 # distance = -1 (index does not currently support distance)
--- a/python-threatexchange/threatexchange/signal_type/pdq_index.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq_index.py
@@ -1,0 +1,65 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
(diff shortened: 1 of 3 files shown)


Dcallies:
I'm pretty sure it won't break imports (the only tests we have on the CLI would probably catch this(?), since you don't import this new module anywhere, and unimported python code is basically nonexistent code.

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Implementation of SignalTypeIndex abstraction for PDQ by wrapping hashing.pdq_faiss_matcher.
```
Dcallies, line 4: nit: linewrap at 79

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+import typing as t
+import pickle
+
+from threatexchange.signal_type.index import SignalTypeIndex, IndexMatch
+from threatexchange.hashing.pdq_faiss_matcher import PDQMultiHashIndex
+
+T = t.TypeVar("T")
```
Dcallies, line 13: If you want to be silly, this also works:

from threatexchange.signal_type import index
T = index.T

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+PDQ_CONFIDENT_MATCH_THRESHOLD = 31
```
Dcallies, line 14: Put this in the class, or even better in the SignalType (it might already be in the SignalType)

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+
+class PDQIndex(SignalTypeIndex):
+    """
+    Wrapper around the pdq faiss index lib
+    """
+
+    def __init__(self, entries: t.Iterable[t.Tuple[str, T]]) -> None:
+        super().__init__()
+        self.local_id_to_entry = dict()
+        hashes = list()
```
Dcallies, line 25: Nit: it's considered best practice to use the sugar for this if you aren't taking advantage of constructor arguments:
{} = dict()
[] = list()

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+        for i in range(len(entries)):
```
Dcallies, line 26: blocking: This is actually not safe, since Iterable is a generator, it won't have a length.

Instead do

```
for i, entry in enumerate(entries):
```

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+    def query(self, hash: str) -> t.List[IndexMatch[T]]:
+        """
+        Look up entries against the index, up to the max supported distance.
+        """
+        results = self.index.search(
+            [hash], PDQ_CONFIDENT_MATCH_THRESHOLD, return_as_ids=True
+        )
+        matches = list()
```
Dcallies, line 40: .

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+        for result_ids in results:
+            for id in result_ids:
+                # distance = -1 (index does not currently support distance)
```
Dcallies, line 43: :O

BarrettOlson:
> I'm pretty sure it won't break imports (the only tests we have on the CLI would probably catch this(?), since you don't import this new module anywhere, and unimported python code is basically nonexistent code.

Good point, I just meant to call out this depends on faiss but since we don't use it anywhere that doesn't matter atm

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+        Convert the PDQ index into a bytestream (probably a file).
+        """
+        return pickle.dumps(self)
+
+    @classmethod
+    def deserialize(cls, fin: t.BinaryIO) -> "SignalTypeIndex[T]":
+        """
+        Instanciate an index from a previous call to serialize
+        """
+        return pickle.loads(fin)
```
Dcallies, line 65: Bootiful

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -0,0 +1,65 @@
...
+    def build(cls, entries: t.Iterable[t.Tuple[str, T]]) -> "SignalTypeIndex[T]":
+        """
+        Build an PDQ index from a set of entries.
+        """
+        return PDQIndex(entries)
+
+    def serialize(self, fout: t.BinaryIO) -> None:
+        """
+        Convert the PDQ index into a bytestream (probably a file).
+        """
```
Dcallies, line 57: (ignorable): Usually fine to leave off docstring when it's just a copy of the parent

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
Dcallies, line 1: yaytests

Dcallies (approved):
See inlines, but also feel free to land if it's going to take too long.

eriksendc (approved):

python-threatexchange/threatexchange/signal_type/pdq_index.py
BarrettOlson: leaving incase we want to add pylint later...

python-threatexchange/threatexchange/signal_type/pdq_index.py
BarrettOlson: thoughts on `... import, T as IndexT` ?

PR #385, opened 2021-02-27
fix pdq_index.serialize to use bytestream param

BarrettOlson:
Summary
---------

Update serialize for pdq_index to actually follow its type interface 

Test Plan
---------

pytest was green and hmalite ui (to build index still work)


Dcallies (approved):
shipit

PR #386, opened 2021-02-27
[hmalite] flask config

Dcallies:
Summary
---------

Adds a simple flask config. I'm not a huge fan of it being untyped, but not sure if it's easy to fix either.

Test Plan
---------

Spent several hours learning various ways to fail to run. The winning strategy ended up being:
1. Code in devserver
2. Test by running flask directly and then curling it 
2. push to fork repo
3. pull from fork repo onto laptop (in git-bash)
4. in powershell, build a docker image
5. run the docker image with `docker run -d -p 5000:80 lite`
6. visit localhost:5000 in browser

I know there are many possible optimizations to the above, but this is the only route that works so far. I am not a good engineer.

### DIFF
--- a/hasher-matcher-actioner/hma-lite/hmalite/app.py
+++ b/hasher-matcher-actioner/hma-lite/hmalite/app.py
@@ -5,6 +5,7 @@ from hmalite.matcher import matcher_api
 from threatexchange.signal_type import index, pdq_index
 from threatexchange.hashing.pdq_hasher import pdq_from_file
 import os
+import os.path
 import sys
 import csv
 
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hma-lite/hmalite/app.py
```diff
@@ -5,20 +5,28 @@
 from threatexchange.signal_type import index, pdq_index
 from threatexchange.hashing.pdq_hasher import pdq_from_file
 import os
+import os.path
+import sys
```
Dcallies, line 9:
```suggestion
```

hasher-matcher-actioner/hma-lite/hmalite/app.py
```diff
@@ -5,20 +5,28 @@
...
+    app.config.from_object("hmalite.config.HmaLiteDevConfig")
+app.config.from_envvar('HMALITE_CONFIG_FILE', silent=True)
 
 
 @app.route("/")
 def index():
-    files = os.listdir("hmalite/data/")
+    files = []
+    if os.path.exists(app.config["INDEX_FILE"]):
+        files = [app.config["INDEX_FILE"]]
```
Dcallies, line 29: @BarrettOlson - Did this because you said you were planning on optimizing to only use a single index file

BarrettOlson (approved):
Configuration Sensation!

PR #388, opened 2021-02-28
[ise][hma-lite] config

Dcallies:
Summary
---------

This is the output of some number of hours of disorganized hacking, and so is not the cleanest diff.

Somewhere in this pile, there is:
1. Handling configs (supported patterns = written as file, overridden with environ)
2. A bug fix for the PDQ index library
3. Moving things around

Test Plan
---------

Finally figured out how to connect from my browser to the devserver:

ssh -L 5000:localhost:5000

will do the trick. In my case, it took me a few hours to learn that something was reserving 5000, and so needed to switch to 5001.
May contain random bugs.

### DIFF
--- a/hasher-matcher-actioner/hma-lite/hmalite/app.py
+++ b/hasher-matcher-actioner/hma-lite/hmalite/app.py
@@ -1,31 +1,47 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
-from flask import Flask, render_template, request, jsonify, send_from_directory
-from hmalite.matcher import matcher_api
-from threatexchange.signal_type import index, pdq_index
-from threatexchange.hashing.pdq_hasher import pdq_from_file
+import csv
 import os
 import os.path
-import csv
+import shutil
 
-CONFIG_ENV = "HMALITE_CONFIG_FILE"
+from flask import Flask, render_template, request, jsonify, send_from_directory
+from hmalite.config import HmaLiteDevConfig, HmaLiteProdConfig
+from hmalite.matcher import matcher_api
+from threatexchange.hashing.pdq_hasher import pdq_from_file
+from threatexchange.signal_type import pdq_index
+
+# This doesn't really behave the way you might think -
+# instances of flask are destroyed and created all the time
+# Someday we'll have to figure out the correct way to persist it
+_INDEX = pdq_index.PDQIndex(())  # The saddest empty index
+
 
 app = Flask(__name__)
 app.register_blueprint(matcher_api, url_prefix="/v1/hashes")
 
-
+# Set up those configs
 if app.config["ENV"] == "production":
-    app.config.from_object("hmalite.config.HmaLiteProdConfig")
+    config_cls = HmaLiteProdConfig
+    app.config.from_object(HmaLiteProdConfig())
 else:
-    app.config.from_object("hmalite.config.HmaLiteDevConfig")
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hma-lite/hmalite/app.py
```diff
@@ -1,31 +1,64 @@
...
+    for name in config_cls._fields:
+        app.logger.info("%s = %s", name, app.config[name])
+
+
+####################### INDEX HACKS #######################
+# This doesn't really behave the way you might think -
+# instances of flask are destroyed and created all the time/thread
+# local-d, so you'll get a bunch of copies of this(?)
+# Someday we'll have to figure out the correct way to stage it
+_INDEX = None
```
BarrettOlson, line 38: Yeah not ideal, I wonder if something like a global 'session' or some kind of cache could work, probably not easily between different API requests...

schatten: AFAIK, flask application contexts are created and discarded at every request. [1]. Their recommendation for storing data across requests:

> Use the session or a database to store data across requests.

And before you say "session", even that is implemented using memcached/redis. This has always puzzled me about python/php etc. Because of the request-first lifecycle management, there is no way to share state across requests without serialization / deserialization overhead.

Short of implementing a sidecar pattern [2], I do not see a way out of this.

In the mean time, can we `timeit` [3] the initialization and keep logging?

1: https://flask.palletsprojects.com/en/1.1.x/appcontext/#storing-data
2: https://www.oreilly.com/library/view/designing-distributed-systems/9781491983638/ch02.html
3: https://docs.python.org/3/library/timeit.html

schatten: If this is crucial, I'd second that flask is not the appropriate tech for it. Event Driven systems could be more appropriate. eg. https://www.twistedmatrix.com/trac/

Twisted has long lived resources and they display it in its full-glory on their home page. We might want to give that a shot.

schatten (approved):
Approving to unblock so that we can build something off of this, but like you @Dcallies, I'm skeptical of the help flask can provide us here.

hasher-matcher-actioner/hma-lite/hmalite/app.py
```diff
@@ -1,31 +1,64 @@
...
+# Set up those configs
 if app.config["ENV"] == "production":
-    app.config.from_object("hmalite.config.HmaLiteProdConfig")
+    config_cls = HmaLiteProdConfig
+    app.config.from_object(HmaLiteProdConfig())
 else:
-    app.config.from_object("hmalite.config.HmaLiteDevConfig")
-app.config.from_envvar("HMALITE_CONFIG_FILE", silent=True)
+    config_cls = HmaLiteDevConfig
+better_config = config_cls.init_with_environ()
```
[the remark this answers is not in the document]
Dcallies: Will fix in a followup diff

hasher-matcher-actioner/hma-lite/hmalite/app.py
Dcallies: 1. Application contexts are discarded every requests, but the memory is not cleared. This trick works because it's in the outer scope of memory
2. Re: timit - A few seconds at the larger size, but it will only load once each thread.
3. Re: Twisted - I don't think we should invest in hmalite much more than has been done. Having dug into it a bit, I think lambda is probably good enough, and would want to see measurements before we started changing stuff.

PR #391, opened 2021-03-02
[hma] Upload sample data

Dcallies:
Summary
---------

Upload some sample data in a simple format, compatible with the CLI and hma

Test Plan
---------

Got a warning that line endings were being changed, but that it wouldn't be reproduced locally - not sure if these will fail to be read with dialect=excel later.


jeberl (approved):
Lgtm

schatten (approved):

Dcallies:
Summary
---------

Upload some sample data in a simple format, compatible with the CLI and hma

Test Plan
---------

Got a warning that line endings were being changed, but that it wouldn't be reproduced locally - not sure if these will fail to be read with dialect=excel later.

PR #393, opened 2021-03-02
Adds tests script for deployed hma

BarrettOlson:
Summary
---------

This is an attempt to add a smoke test to our hma development flow by adding a Make target `dev_test_instance` This target will pass the new root level `output.tf` values added in this PR to a new python script `tests/scripts/check_deployed_instance.py`

For the script to work you will need to have successfully deployed (via `terraform apply` or `make dev_create_instance` AND have non-expired AWS credentials locally.   

(This PR also copies b.jpg from the pdq directory to be used in the test) 

I spent too much time trying to be smart with Make and realized that is probably beyond me. Open to suggestions on the flow. Various parts of this will need to change as we improve/change pieces of HMA but the methods should be mostly reusable.  

Additional:
I changed the 'SK' for hash sent to the db back to 'type#pdq' from 'type:pdq'

Test Plan
---------

```
$ make dev_test_instance
TF_DYNAMODB_TABLE="barrett-HMADataStore" \
        TF_BUCKET_NAME="barrett-hashing-data20210302200326365300000003" \
        TF_PREFIX="barrett" \
        ./tests/utils/check_deployed.py

Checking status of deployed instance...
Running against instance using prefix: barrett
Attempting to upload threat_exchange_data/pdq.te...
Looking for index/pdq_hashes.index in barrett-hashing-data20210302200326365300000003 (check every 5s x20)...
Attempting to upload images/test_photo.jpg...
Looking for {'PK': 'c#images/test_photo.jpg', 'SK': 'type#pdq'} in dynamodb.Table(name='barrett-HMADataStore') (check every 5s x20)...
Looking for {'PK': 'c#images/test_photo.jpg', 'SK': 'te#5555555555555555'} in dynamodb.Table(name='barrett-HMADataStore') (check every 5s x20)...
Cleaning up...   
Deleted s3 objects and datastore items

Success! Deployed instance seems to behave as expected!
Remember to run "terraform destroy" or "make dev_destroy_instance" if you are done with your deployed instance.
```


hasher-matcher-actioner/terraform/outputs.tf
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+# These root outputs can be used as inputs for tests of delpoyed instances.
```
BarrettOlson, line 3: ...seplling

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
```diff
@@ -0,0 +1,184 @@
...
+            WaiterConfig={"Delay": wait_time, "MaxAttempts": retries},
+        )
+    except ClientError as e:
+        raise TestError(e.msg)
+
+
+def wait_for_db_item(
+    table, key: t.Dict, attributes: t.List, wait_time=5, retries=20
+) -> t.Dict:
+    """
```
jeberl, line 80: Why do we do this in a loop. Why might it not be there first time?

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
```diff
@@ -0,0 +1,184 @@
...
+            if "Item" in result:
+                return result["Item"]
+            time.sleep(wait_time)
+            retries -= 1
+        return {}
+    except ClientError as e:
+        raise TestError(e.msg)
+
+
+def run(bucket: str, table) -> bool:
```
jeberl, line 99: New to python testing. Is this standard to have 1 function that runs all the tests

jeberl (commented):
This seems right to me. The empty file is supposed to be there right?

Going to wait for one other person to review before accepting

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
BarrettOlson: The lambda needs time to do its job and for things to propagate as both uploads trigger an event the call the lambdas

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
BarrettOlson: Not for a pytest tests but this script is basically just running a demo of the system so I think it makes sense here... could be convinced otherwise.

hasher-matcher-actioner/hmalib/dto.py
```diff
@@ -24,7 +24,7 @@ def get_dynamodb_pk(key: str):
     def to_dynamodb_item(self) -> dict:
         return {
             "PK": PDQHashRecord.get_dynamodb_pk(self.content_key),
-            "SK": "type:pdq",
+            "SK": "type#pdq",
```
schatten, line 27: 🙏🏽

hasher-matcher-actioner/terraform/outputs.tf
```diff
@@ -0,0 +1,14 @@
...
+
+output "bucket_name" {
```
schatten, line 5: Can you describe the bucket? Would love it if the variable name gave an idea of what I'm supposed to do with the bucket.

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    ~End-to-end test of HMA that:
+    uploads TE data -> looks for the index
+    uploads a photo -> looks for the hash and a match
+
+    throws TestError if it encoutners unexpected results
+    """
+    # Upload TE Data
+    upload_s3_object(bucket, THREAT_EXCHANGE_PDQ_DATA_KEY, TEST_TE_DATA_PDQ)
```
schatten, line 108: If I am testing an instance the second time, this object will already exist. In those cases, can we check that the index exists and proceed? 

Or perhaps you think that is not an end-to-end test. What do you think?

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # Index created?
+    wait_for_s3_object(bucket, PDQ_INDEX_KEY)
+
+    # Upload Photo
+    with open(os.path.dirname(__file__) + "/../data/b.jpg", "rb") as data:
+        upload_s3_object(bucket, TEST_PHOTO_KEY, data.read())
```
schatten, line 115: This also assumes the images/ S3 prefix-directory-thingy is empty.

For this test to be repeatable, I'd add some randomness to the file name.

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
```diff
@@ -0,0 +1,184 @@
...
+    )
+    if "ContentHash" not in result:
+        raise TestError("Failed to find hash")
+    if result["ContentHash"] != TEST_PHOTO_EXPECTED_HASH:
+        raise TestError("Found Incorrect Hash")
+
+    # match found?
+    result = wait_for_db_item(
+        table,
+        key={"PK": f"c#{TEST_PHOTO_KEY}", "SK": f"te#{TEST_PHOTO_EXPECTED_ID}"},
```
schatten, line 131: Non-blocking. If you can find the time and the answers, can you document the dynamodb table design somewhere? Having it handy would help.

schatten (approved):
This is already in pretty good shape. I have one nit in the inline comments. But you can address that. Approving to unblock.

Feedback
---
I don't find the design of the check_deployed_instance.py file "super" open to re-use. There are methods, but they are generic names, so hard to understand what they would do. 

eg. What does "run()" do? If I were to re-use this method, what would happen....

Also, I could create a BaseTest module which maintains a set of `s3_resource_keys` and `dynamodb_item_pks` and provides methods to clean up after you are done. The smoke test and other integration tests would extend this class and benefit from being focused on what they do rather than do the boilerplaty things.

Similarly, each integration test can be thought of as being an list of sub-flows. Each sub flow is (action, wait-for-something, assert). But it is too early to tell.

However, and this is important
---
I feel when we are writing the second test (integration or smoke), we should be able to refactor as required. Right now, I completely understand that it is too early to tell what the "right" abstractions are.

hasher-matcher-actioner/terraform/outputs.tf
BarrettOlson: Followed up offline, keeping names generic for now as we use the same bucket for various different purposes. (someday it might make sense to have 'data_upload_bucket' and 'index_store_bucket' for example).

BarrettOlson (commented):
Thanks @schatten and @jeberl! add some inline comments and agree we should break it up more in the future.

PR #377, opened 2021-02-25
Refactor hasher lambda, define hash store interface, mypy

schatten:
Summary
---------
The PR turned out to be larger than I had imagined. It does a lot.
1. Adds mypy as a dependency and begins to add type hints to the lambdas. Adds some typing for boto using `boto3-stubs`
1. Converts arbitrary wire-messages (SQS, Dynamo) for hasher lambda into formal data structures
1. Defines a simple wrapper (HashStore) over dynamodb for maintaining hash records, expect more methods from the CLI/Admin surface to go into the wrapper
1. docstring for hasher lambda

Test Plan
---------
```
$> mypy
$> make upload_docker
$> edit hma_lambda_docker_uri variable to point to new sha256 in ECR
$> terraform apply -var "prefix=$(whoami)"
```

Followed by uploading an image to S3 and receiving a match on email.

Note: This was also the first time I edited an existing terraform instance. And it worked out well. The lambdas were switched out and the architecture picked things up automatically.

PR #398, opened 2021-03-03
used signal_type index and generalize hashstore

BarrettOlson:
Summary
---------

Updates the dynamodb 'hashstore' interface to the more general 'dynamostore' and updates pdq_matcher to use it.
This also changes the index interface to the one in signal_type (same one used in hma-lite)

I can split the changes into two PRs if that helps, my git-foo is limited though so I might need help

(Included small fix to Makefile as well)

Test Plan
---------

The new
`make dev_test_instance`
came back green.

Also tested manually: behavior appears the same with slight changes to log messages.


hasher-matcher-actioner/hmalib/dto.py
```diff
@@ -1,39 +1,98 @@
...
+        return {
+            "PK": PDQMatchRecord.get_dynamodb_content_key(self.content_key),
+            "SK": PDQMatchRecord.get_dynamodb_te_key(self.te_id),
+            "ContentHash": self.content_hash,
+            "Timestamp": self.timestamp.isoformat(),
+            "TEHash": self.te_hash,
+            "GSI1-PK": PDQMatchRecord.get_dynamodb_te_key(self.te_id),
+            "GSI1-SK": PDQMatchRecord.get_dynamodb_content_key(self.content_key),
+            "HashType": self.SIGNAL_TYPE,
+            "GSI2-PK": PDQMatchRecord.get_dynamodb_type_key(self.SIGNAL_TYPE),
```
Dcallies, line 93: Nit: use `self` here instead of PDQMatchRecord

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -67,12 +67,21 @@ def lambda_handler(event, context):
         codecs.getreader("utf-8")(pdq_data_file["Body"]),
         fieldnames=PDQ_DATA_FILE_COLUMNS,
     )
-    pdq_data = [(row["hash"], int(row["id"])) for row in pdq_data_reader]
+    pdq_data = [
+        (
+            row["hash"],
+            # Also add hash to metadata for easy look up on match
```
Dcallies, line 73: Explain this to me? Why isn't the ID enough for easy lookup on match?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -65,25 +52,29 @@ def lambda_handler(event, context):
...
-                key, hash_str, result
-            )
-            logger.info(message_str)
-            for te_id in result:
-                te_hash = hash_index.hash_at(te_id)
-                save_match_to_datastore(
-                    table, key, te_id, hash_str, current_datetime, te_hash
+        results = hash_index.query(hash_str)
+
+        if len(results) > 0:
```
Dcallies, line 58: nit:

```
if results:
```

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -65,25 +52,29 @@ def lambda_handler(event, context):
...
+            match_ids = []
+            for match in results:
+                metadata = match.metadata
+                logger.info(
+                    f"Match found for key: {key}, hash: {hash_str} -> {metadata}"
```
Dcallies, line 63: ignoreable: logger by default supports string formatting, but it's less nice than f-strings, and easier to put the wrong number of arguments, with the possible tradeoff that if the log isn't printed, objects aren't converted to strings

```
logger.info("Match found for key: %s, hash %s -> %s", key, hash_str, metadata)
```

hasher-matcher-actioner/hmalib/storage/dynamostore.py
```diff
@@ -1,19 +1,22 @@
...
-    Stores all kinds of hashes into a dynamodb.
+    Stores all kinds of item into a dynamodb.
     """
 
     def __init__(self, table: Table):
         self._table = table
 
-    def add_hash(self, record: _IDynamoDBItem):
+    def add_item(self, record: _IDynamoDBItem):
         self._table.put_item(Item=record.to_dynamodb_item())
```
Dcallies, line 22: I'm not sure using this class directly should be common, since you generally want the thing you put into the store to be typed to what the table is doing. One advantage of using a class dedicated to the table is that you know what you are interacting with, and can tailor the interfaces to it (for example, instead of _IDynamoDBItem, it can take the appropriate item).

hasher-matcher-actioner/hmalib/storage/dynamostore.py
```diff
@@ -1,19 +1,22 @@
 from mypy_boto3_dynamodb.service_resource import Table
 
-class _IDynamoDBItem():
+
+class _IDynamoDBItem:
```
Dcallies, line 4: Nit: It's generally understood in python that prefixing module or class-level variable with an underscore means that it should be treated as "private", internal-details, or may be refactored in the future without warning.

hasher-matcher-actioner/hmalib/storage/dynamostore.py
```diff
@@ -1,19 +1,22 @@
...
     """
     Clowny type as an interfaces. True interfaces in python are nightmares.
```
Dcallies, line 6: This is not clowny (in python), but the normal way to do it.

Dcallies (commented):
Toplevel:

- I don't think this necessarily needs to go through another round (in diff), though maybe @schatten  knows something I don't
- I'm not a fan of using a "Database" class that takes arbitrary records. I'd rather have an abstract database base class (if it was even needed) and have dedicated objects for each table. It's the same reason why you don't use raw sql very often, even though mysql().select() might work for any datastore.
- I'm still have lingering naming confusion over the "HashStore", since it doesn't store hases, but instead is a logging table for content that is put through the system. Is that correct, or are there some usecases I'm missing?

hasher-matcher-actioner/hmalib/storage/dynamostore.py
Dcallies: Discussed offline, the model is:

Record objects contains all the details for how to store and iteract with it, this class is used for dependency management.

David was expecting an interface to a specific table, and data objects that didn't know their storage details for some reason.

BarrettOlson:
Some of this discussion might make more sense in the context of #377 

@schatten can you speak to the value of having hashstore/dynamostore.py and why it should take a table? @Dcallies asked me and I didn't have a good answer. 

I like your addition of `dto.py` to move the representation of the db key out of the lambda methods but I am not sure the value of having `hashstore.py` compared to just calling `table.put_item(record.to_dynamodb_item()` in the lambda. Is it something do to with our ability to have test shims?

Thanks!

schatten (approved):
I think this is in reasonably good shape. Well done!

The design change in metadata schema for the index is an open question from @Dcallies . @BarrettOlson and I discussed this offline and I'm confident that we can proceed ahead with the PR as it is right now. 

Highlights of the discussion:
1. Adding the hash to the metadata is helpful because:
a. The [IndexMatch](https://github.com/facebook/ThreatExchange/blob/master/python-threatexchange/threatexchange/signal_type/index.py#L29) object does not, at present, include the matched hash. 
b. The [SignalTypeIndex](https://github.com/facebook/ThreatExchange/blob/dabbcfafa2b60a567bf6d6a4b6dff43825c2419c/python-threatexchange/threatexchange/signal_type/index.py#L50) interface does not require impl of the `hash_at` method.
2. Adding the hash to the metadata only affects the size of the pickle files, not of the index.
3. We need to figure out how to change either the `SignalTypeIndex` interface or the `IndexMatch` class. The pickle datastructure can addressed after we make that decision.

hasher-matcher-actioner/hmalib/dto.py
```diff
@@ -1,39 +1,98 @@
...
         }
+
+
+@dataclass
+class PDQMatchRecord(PDQRecord):
+    """
+    Successful execution at the matcher produces this record.
+    """
+
+    te_id: int
```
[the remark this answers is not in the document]
BarrettOlson: yes, but I am going to wait and make it general later by adding a different class. (the db item key is 'te#<id-val>' so to a certain extent that ship has sailed until a larger refactor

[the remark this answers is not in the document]
schatten: Disagree that the ship has sailed because partners are not yet using HMA.

If there are interventions we can do now that make for an easier upgrade path for partners later, let's do it. We can clear our own dynamodb and recreate, but can't do the same for partner systems once it has data. Would urge you to reconsider this.

hasher-matcher-actioner/hmalib/dto.py
```diff
@@ -1,39 +1,98 @@
...
+
+    def to_dynamodb_item(self) -> dict:
+        raise NotImplementedError
+
+    def to_sqs_message(self) -> dict:
+        raise NotImplementedError
+
+
+@dataclass
+class PDQHashRecord(PDQRecord):
```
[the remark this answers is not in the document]
BarrettOlson: Changed name to `PipelinePDQHashRecord`

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
BarrettOlson: See @schatten's note below.

hasher-matcher-actioner/hmalib/storage/dynamostore.py
BarrettOlson: Removed 'store' class altogether!

hasher-matcher-actioner/hmalib/dto.py
```diff
@@ -1,39 +1,98 @@
...
+Classes in this module should implement methods `to_dynamodb_item(self)` and
 `to_sqs_message(self)`
 """
 
-class PDQHashRecord(t.NamedTuple):
+# TODO / review note: should I create a general base version and replace '_IDynamoDBItem' ?
+
+
+@dataclass
+class PDQRecord:
```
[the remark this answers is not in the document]
BarrettOlson: added Base to name, good enough?

schatten (approved):
LGTM. Shipit!

PR #401, opened 2021-03-08
First running version of the HMA UI.

eriksendc:
Summary
---------

Created a trivial React app with Create React App, then added some trivial "hello world" functionality leveraging a simple API Gateway-based HTTP API.

Test Plan
---------

Nothing to test yet. Need to get this in so that it's in the codebase to be built and terra-formed onto an S3 static website.


hasher-matcher-actioner/webapp/.watchmanconfig
```diff
@@ -0,0 +1,2 @@
+{
```
eriksendc, line 1: Couldn't run tests without this file. Error that was showing led to the creation of this empty .watchmanconfig file.

hasher-matcher-actioner/webapp/public/index.html
```diff
@@ -0,0 +1,15 @@
...
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" integrity="sha384-9gVQ4dYFwwWSjIDZnLEWnxCjeSWFphJiwGPXr1jddIhOegiu1FwO5qRGvFXOdJZ4" crossorigin="anonymous">
+    <title>HMA</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.0/umd/popper.min.js" integrity="sha384-cs/chFZiN24E4KMATLdqdvsezGxaGsi4hLGOzlXwp5UZB1LY//20VyM2taTB4QvJ" crossorigin="anonymous"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/js/bootstrap.min.js" integrity="sha384-uefMccjFJAIv6A+rW+L4AHf99KvxDjWSu1z9VI8SKNVmz4sk7buKt/6v9KI65qnm" crossorigin="anonymous"></script>
```
schatten, line 13: Were these there by default? What are we using these libraries for? 

jQuery is not declarative and it will conflict with react. 

Event if we decide to add more libraries, IMO, it should be via npm packaging.

schatten: If you want to use bootstrap goodness, may I recommend https://react-bootstrap.github.io/

hasher-matcher-actioner/webapp/src/App.js
```diff
@@ -0,0 +1,55 @@
...
+        </div>
+      </main>
+    </>
+  );
+}
+
+function MetricsTable() {
+  const [metrics, setMetrics] = useState(null);
+  useEffect(() => {
+    fetch("https://m1mx5hd9h0.execute-api.us-east-1.amazonaws.com/hma-api-function1")
```
schatten, line 35: Would request this be moved out into a separate environment variable or config global. You can use an env var to pass the URL to the 'npm start' script and can then figure out how to pass that on to the javascripty bits. 

[1] https://create-react-app.dev/docs/adding-custom-environment-variables/#referencing-environment-variables-in-the-html

schatten (dismissed):
Looks great! Can't wait to get started on hacking UI!

Request we keep things strictly declarative. No DOM manipulation from jQuery/popper/what-have-you.

Also, would love to see the URL extracted into a "safe" envvar.

hasher-matcher-actioner/webapp/public/index.html
eriksendc: I don't disagree. Trying to take baby steps and have small PRs. There's definitely a "right way" to bring in bootstrap that includes using npm to manage things. Will do that in a subsequent PR.

eriksendc: To memorialize a conversation, we do need to circle back on whether or not we'll use Bootstrap, jQuery, and any other framework. For now these are included for quick progress on a mockup / demo.

hasher-matcher-actioner/webapp/src/App.js
eriksendc: Good recommendation on an eventual solution for this kind of thing.

PR #403, opened 2021-03-08
[hma] base terraform configs for fetcher

Dcallies:
Summary
---------

Pair programmed with @jeberl - base state for the fetcher, with maybe the right permissions, but missing the cloud event for the timer.

Test Plan
---------

```
$ terrform init
$ terraform plan
$ terraform apply
```
Test the lambda by triggering manually


jeberl (commented):
Some wonderful lines of code

PR #406, opened 2021-03-09
[hma] terraform for fetch configs

Dcallies:
Summary
---------

I did some terraform things to connect a DynamoDB table to the fetcher. 

Test Plan
---------

terraform apply-ied to my dev instance. 
Will test that it works in a followup diff.


jeberl (approved):
Looks right to me

eriksendc (approved):

PR #408, opened 2021-03-09
Add Fetch Trigger event to terraform

jeberl:
Summary
---------

Adds to the fetcher terraform, the fetcher rule that triggers the fectcher function to be called on some regular cadence this will 

Test Plan
---------

`terraform validate`
`terraform apply`
AWS looks correct correctly

Set variable to be 1 minute and see that the lambda `jeberl_fetcher` in aws is called every minute by viewing monitoring here: https://console.aws.amazon.com/lambda/home?region=us-east-1#/functions/jeberl_fetcher?tab=monitoring

### DIFF
--- a/hasher-matcher-actioner/terraform/fetcher/main.tf
+++ b/hasher-matcher-actioner/terraform/fetcher/main.tf
@@ -11,6 +11,8 @@ provider "aws" {
   profile = var.profile
 }
 
+### Lambda for fetcher ###
+
 data "aws_iam_policy_document" "lambda_assume_role" {
   statement {
     effect  = "Allow"
@@ -21,8 +23,6 @@ data "aws_iam_policy_document" "lambda_assume_role" {
     }
   }
 }
-
-# Lambda for fetcher
 
 resource "aws_lambda_function" "fetcher" {
   function_name = "${var.prefix}_fetcher"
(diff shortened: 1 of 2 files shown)


Dcallies (approved):
As best as I can tell, this looks like the thing.

Thanks for commenting out these blocks!

I realized that a lot of the current permissions don't have prefixes, and so are likely causing all sorts of weird conflicts as we are all working. Need to do a diff to add prefixes if someone hasn't already :/

hasher-matcher-actioner/terraform/fetcher/variables.tf
```diff
@@ -46,3 +46,9 @@ variable "log_retention_in_days" {
   description = "How long to retain cloudwatch logs for lambda functions in days"
   type        = number
 }
+
+variable "fetch_frequency_min" {
+  description = "How many minutes to wait between queries to ThreatExchange for updates"
+  type        = number
+  default     = 15
```
eriksendc, line 53: Thinking about this number makes me wonder about how a long-running fetcher run interacts with a subsequent run. Is there a CloudWatch feature that understands the concept of not wanting to run two instances of the fetcher lambda concurrently? At some point we need to know what the current behavior is and if necessary we need to prevent unexpected behavior caused by concurrency issues.

eriksendc (approved):
I'm good with not addressing the concurrency issue I brought up, but we do need a plan for dealing with that at some point.

hasher-matcher-actioner/terraform/fetcher/variables.tf
jeberl: good call out

hasher-matcher-actioner/terraform/fetcher/main.tf
```diff
@@ -187,4 +187,23 @@ resource "aws_iam_policy" "fetcher_trigger" {
 resource "aws_iam_role_policy_attachment" "fetcher_trigger" {
   role       = aws_iam_role.fetcher_trigger.name
   policy_arn = aws_iam_policy.fetcher_trigger.arn
+
```
Dcallies, line 190: Hmm, this is the first automatic merge failure I've seen - ate a closing brace.

PR #413, opened 2021-03-10
Webapp terraform cloudfront

eriksendc:
Summary
---------

This PR introduces the conditional terraforming of the webapp's CloudFront distribution.

Test Plan
---------

1. Inspect your local terraform.tfvars file. If it has an entry for `include_cloudfront_distribution`, remove it.
2. Run `terraform plan` and validate that you don't see this:
````# module.webapp.aws_cloudfront_distribution.webapp[0] will be created
3. In your local terraform.tfvars file, add this line:
````include_cloudfront_distribution = false
2. Run `terraform plan` and validate again that you don't see the line regarding the cloudfront distribution being created.
3. In your local terraform.tfvars file, switch the added line to this:
````include_cloudfront_distribution = true
4. Run `terraform apply`
5. When terraforming is finished, sign in to the AWS console.
6. Navigate to the CloudFront service: https://console.aws.amazon.com/cloudfront/home?region=us-east-1
7. select the ID for the distribution that was just created.
8. In the General tab, look for the property labeled Domain Name. It should be something like this: d1oghakifba3zm.cloudfront.net (the first bit will be different).
9. Open a browser to http://d1oghakifba3zm.cloudfront.net (using your version identified in step 8) and you should see this:
![image](https://user-images.githubusercontent.com/8974354/110665696-99c78780-8196-11eb-9c1b-d47b5405f619.png)

### DIFF
--- a/hasher-matcher-actioner/terraform/webapp/variables.tf
+++ b/hasher-matcher-actioner/terraform/webapp/variables.tf
@@ -1,17 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
-variable "region" {
-  description = "AWS region to deploy to"
-  type        = string
-  default     = "us-east-1"
-}
-
-variable "profile" {
-  description = "AWS profile to use for authentication"
-  type        = string
-  default     = null
-}
-
 variable "prefix" {
   description = "Prefix to use for resource names"
   type        = string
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: hasher-matcher-actioner/terraform/webapp/variables.tf [hunk 1]
```diff
@@ -1,17 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
-variable "region" {
```
schatten, line 3: Love that this could be removed! Files had started to look awfully similar.

schatten (approved):

PR #415, opened 2021-03-11
add simple routes

BarrettOlson:
Summary
---------

Add routes to `status_api.py` 

Test Plan
---------
(Ask me for a demo)

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/status_api.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/status_api.py
@@ -2,30 +2,76 @@
 
 import logging
 import os
-
+import bottle
 import boto3
 import json
+from boto3.dynamodb.conditions import Attr
+from apig_wsgi import make_lambda_handler
 
+app = bottle.default_app()
+apig_wsgi_handler = make_lambda_handler(app)
 
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
-# s3_client = boto3.client("s3")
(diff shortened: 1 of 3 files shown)


Dcallies (approved):
shipit

eriksendc (approved):

hasher-matcher-actioner/hmalib/lambdas/api/status_api.py
```diff
@@ -2,30 +2,76 @@
...
-    return {
-        'statusCode': '200',
-        'body': 'Hello World, HMA',
-        'headers': {
-            'Content-Type': 'application/json',
-        },
-    }
\ No newline at end of file
+
+def check_db_for_matches():
```
eriksendc, line 62: I'm assuming that eventually these will go into some other .py.

hasher-matcher-actioner/hmalib/lambdas/api/status_api.py
```diff
@@ -2,30 +2,76 @@
...
+def hashes(key=None):
+    results = check_db_for_hash(key)
+    logger.info(results)
+    return json.dumps(results)
 
 
 def lambda_handler(event, context):
     """
-    Status of delpoyed HMA 
+    Status of delpoyed HMA
```
eriksendc, line 54: When this was a "hello world" the term of art "status" was being used. I'm wondering if at this point this file should be called api.py? Eventually when we break things up into .py's that map to different categories of endpoints we'll have different names that tie to the category.

PR #417, opened 2021-03-12
Create python file to load secrets

jeberl:
Summary
---------
A new python file and class to load AWS secrets. 

As we add more secrets we should add functions to the AWSSecrets class to load them

secrets can either be stored as strings or raw binary and so have

Test Plan
---------
Ran locally by adding line to bottom `print(AWSSecrets().te_api_data().api_token)`
and confirming that the correct access token secret is printed out

Need to hard code `AWS_REGION` to "us-east-1"

### DIFF
--- a/hasher-matcher-actioner/hmalib/aws_secrets.py
+++ b/hasher-matcher-actioner/hmalib/aws_secrets.py
@@ -12,8 +12,6 @@ secrets_client = session.client(
     service_name='secretsmanager',
 )
 
-from threatexchange.api import ThreatExchangeAPI
-
 class AWSSecrets:
     """
     A class for reading secrets stored in aws
--- a/hasher-matcher-actioner/hmalib/aws_secrets.py
+++ b/hasher-matcher-actioner/hmalib/aws_secrets.py
@@ -7,8 +7,6 @@ import boto3
 import base64
 import typing as t
 
-import json
-
 session = boto3.session.Session()
 secrets_client = session.client(
     service_name='secretsmanager',
--- a/hasher-matcher-actioner/hmalib/aws_secrets.py
+++ b/hasher-matcher-actioner/hmalib/aws_secrets.py
@@ -1,0 +1,65 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+"""
+Wrapper functions for reading secrets stored in AWS
+"""
+
+import boto3
+import base64
+import typing as t
+import ast
+
+import json
+
+session = boto3.session.Session()
+secrets_client = session.client(
+    service_name='secretsmanager',
+)
+
+from threatexchange.api import ThreatExchangeAPI
+
+class AWSSecrets:
+    """
+    A class for reading secrets stored in aws
+    """
+
+    @classmethod
+    def te_api_key(self) -> str :
+        """
+        get the ThreatExchange API Key
+        """
+        secret_name = "threatexchange/api_keys"
+        key_name = "default"
+        api_key = get_key_value_secret(secret_name, key_name)
+        return api_key
+
+def get_bin_secret(secret_name: str) -> bytes:
+    """
+    For secerts stored in AWS Secrets Manager as binary
+    """
+    response = get_secret_value_response(secret_name)
+    decoded_binary_secret = base64.b64decode(get_secret_value_response['SecretBinary'])
+    return decoded_binary_secret
+
+def get_str_secret(secret_name: str) -> str:
+    """
+    For secerts stored in AWS Secrets Manager as strings
+    """
+    response = get_secret_value_response(secret_name)
+    str_response = response['SecretString']
+    return str_response
+
+def get_key_value_secret(secret_name: str, key: str) -> str:
+    """
+    For secerts stored in AWS Secrets Manager as dicts
+    """
+    str_response = get_str_secret(secret_name)
+    dict_response = json.loads(str_response)
+    return dict_response[key]
+
+def get_secret_value_response(secret_name: str):
+    get_secret_value_response = secrets_client.get_secret_value(
+            SecretId=secret_name
+        )
+    return get_secret_value_response
+
+print(AWSSecrets.te_api_key())
--- a/hasher-matcher-actioner/hmalib/aws_secrets.py
+++ b/hasher-matcher-actioner/hmalib/aws_secrets.py
@@ -6,7 +6,6 @@ Wrapper functions for reading secrets stored in AWS
 import boto3
 import base64
 import typing as t
-import ast
 
 import json
 
@@ -28,8 +27,7 @@ class AWSSecrets:
         get the ThreatExchange API Key
         """
         secret_name = "threatexchange/api_keys"
-        key_name = "default"
-        api_key = get_key_value_secret(secret_name, key_name)
+        api_key = get_str_secret(secret_name)
         return api_key
 
 def get_bin_secret(secret_name: str) -> bytes:
@@ -48,18 +46,8 @@ def get_str_secret(secret_name: str) -> str:
     str_response = response['SecretString']
     return str_response
 
-def get_key_value_secret(secret_name: str, key: str) -> str:
-    """
-    For secerts stored in AWS Secrets Manager as dicts
-    """
-    str_response = get_str_secret(secret_name)
-    dict_response = json.loads(str_response)
-    return dict_response[key]
-
 def get_secret_value_response(secret_name: str):
     get_secret_value_response = secrets_client.get_secret_value(
             SecretId=secret_name
         )
     return get_secret_value_response
-
-print(AWSSecrets.te_api_key())


hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+    assert response and 'SecretBinary' in response
+    decoded_binary_secret = base64.b64decode(get_secret_value_response['SecretBinary'])
+    return decoded_binary_secret
+
+def get_str_secret(secret_name: str) -> str:
+    response = get_secret_value_response(secret_name)
+    assert response and 'SecretString' in response
+    str_response = response['SecretString']
+    dict_response = ast.literal_eval(str_response)
+    if type(dict_response) is dict:
```
Dcallies, line 34: In general you want this to be an exception, and these kind of checks are usually fragile. In what cases does this happen?

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_bin_secret(secret_name: str) -> t.BinaryIO:
+    response = get_secret_value_response(secret_name)
+    assert response and 'SecretBinary' in response
+    decoded_binary_secret = base64.b64decode(get_secret_value_response['SecretBinary'])
+    return decoded_binary_secret
+
+def get_str_secret(secret_name: str) -> str:
+    response = get_secret_value_response(secret_name)
+    assert response and 'SecretString' in response
```
Dcallies, line 31: (opinion - for this and all the other asserts): In general, this can erase some of the context around the error that python might be giving you for free. For example, simple doing line 32 will lead to a "KeyError: SecretString" pointing to the line, which is actually more descriptive than the error that will be displayed for the assert failure.

This is one reason you don't see too many asserts in python code.

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+import boto3
+import base64
+import typing as t
+import os
+import ast
+from botocore.exceptions import ClientError
+
+from threatexchange.api import ThreatExchangeAPI
+
+AWS_REGION = os.environ["AWS_REGION"]
```
Dcallies, line 12: (not sure): I suspect that boto actually has a method to get this loaded by its own internal mechanism, which might be more flexible than the environment.

my_session = boto3.session.Session()
my_region = my_session.region_name

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+        return dict_response["default"]
+    return str_response
+
+def get_secret_value_response(secret_name: str):
+
+    # Create a Secrets Manager client
+    session = boto3.session.Session()
+    secrets_client = session.client(
+        service_name='secretsmanager',
+        region_name=AWS_REGION
```
Dcallies, line 44: I don't think this is required (see above)

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import boto3
+import base64
+import typing as t
+import os
+import ast
+from botocore.exceptions import ClientError
+
```
Dcallies, line 9: I wasn't able to find a definitive answer on the tradeoffs, but many pieces of code I've seen simply create a resource at the top level, which under the hood has a client.

```
secrets = boto.resource("secretsmanager")
...
def foo():
  secrets.get_secret_value(val)
```

This can potentially save you some boilerplate

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+class AWSSecrets():
+    """
+    A class for reading secrets stored in aws for interacting with ThreatExchange
+    """
+    def te_api_data(self) -> "ThreatExchangeAPI" :
+        secret_name = "threatexchange/api_keys"
+        secret_str = get_str_secret(secret_name)
+        return ThreatExchangeAPI(secret_str)
+
+def get_bin_secret(secret_name: str) -> t.BinaryIO:
```
Dcallies, line 23: I think this returns a literal `bytes` object rather than a BinaryIO, which is a different interface

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+import typing as t
+import os
+import ast
+from botocore.exceptions import ClientError
+
+from threatexchange.api import ThreatExchangeAPI
+
+AWS_REGION = os.environ["AWS_REGION"]
+
+class AWSSecrets():
```
Dcallies, line 14: Nit: When not inheriting from anything, parenthesis should be ignored

```
class AWSSecrets:
```

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    A class for reading secrets stored in aws for interacting with ThreatExchange
+    """
+    def te_api_data(self) -> "ThreatExchangeAPI" :
+        secret_name = "threatexchange/api_keys"
+        secret_str = get_str_secret(secret_name)
+        return ThreatExchangeAPI(secret_str)
```
Dcallies, line 21: Alt, ignorable: a common pattern is a that alternative constructors are stored as classmethods

```
ThreatExchangeAPI:
  def __init__(self, x):
    self.x = x

  @classmethod
  def fromSecret(cls):
    return cls(get_str_secret("threatexchange/api_keys"))

api = ThreatExchangeAPI::fromSavedSecret
```

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
```
Dcallies, line 2: (ignorable): Short docstring

Dcallies (approved):
I think you can simplify some parts of the diff (inline comments), but seems fine.

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+    response = get_secret_value_response(secret_name)
+    assert response and 'SecretBinary' in response
+    decoded_binary_secret = base64.b64decode(get_secret_value_response['SecretBinary'])
+    return decoded_binary_secret
+
+def get_str_secret(secret_name: str) -> str:
+    response = get_secret_value_response(secret_name)
+    assert response and 'SecretString' in response
+    str_response = response['SecretString']
+    dict_response = ast.literal_eval(str_response)
```
schatten, line 33: This is overkill and potentially risky. AST is for meta-programming where the program is reading/writing python code. From a look at the actual secret, it appears to be JSON, easy to confuse with the python dict literal. 

Consider using `json.loads`.

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
schatten: Would second this. At present, the separation of concerns between secrets and the API is unclear. I would go further and have the inference being done at callsite.

eg.
```python
te_secret = hmalib.secrets.get_secret('threatexchange/secret')
api = ThreatExchangeAPI(te_secret)
```

That way, ThreatexchangeAPI does not need to know we are fetching secrets from AWS and hmalib.aws_secrets need not know we are using secrets to create ThreatExchangeAPI objects.

schatten (changes requested):
As David and I have pointed out, there are some digressions from common python idioms. Would love to see those fixed.

Additionally, IMO, the hmalib.lambdas module should contain lambdas and nothing else. This module is better placed elsewhere. Perhaps `hmalib.secrets`?

Also, it appears the partner is expected to store the secrets on their own. Might I suggest we make this Class/API do reads _and_ writes. Then, we can provide a wrapper script to help them get started.

That way the serialization / deserialization / keys of the secrets stays an internal detail of the class and does not leak out. At present, the deserialization is located here, while the serialization is outside.

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
jeberl: This would be if trying to read a string secret that is actually stored as binary. I think removing the assert could be fine

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
jeberl: Good to know!

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
jeberl: Got it. I was following the python example here: https://console.aws.amazon.com/secretsmanager/home?region=us-east-1#!/secret?name=threatexchange%2Fapi_keys

But moving to top level makes sense

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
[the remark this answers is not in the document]
jeberl: No plans at the moment but saw this as I was researching AWS secrets and thought to include the method now incase we want to use later. Happy to remove if confusing

hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
...
+from threatexchange.api import ThreatExchangeAPI
+
+AWS_REGION = os.environ["AWS_REGION"]
+
+class AWSSecrets():
+    """
+    A class for reading secrets stored in aws for interacting with ThreatExchange
+    """
+    def te_api_data(self) -> "ThreatExchangeAPI" :
+        secret_name = "threatexchange/api_keys"
```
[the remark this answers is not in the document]
jeberl: I think we should set up the secret using terraform to ensure partner is always using the same convention

jeberl:
Left comments above but a few other responses
> Additionally, IMO, the hmalib.lambdas module should contain lambdas and nothing else. This module is better placed elsewhere. Perhaps `hmalib.secrets`?
agreed. will move

> Also, it appears the partner is expected to store the secrets on their own. Might I suggest we make this Class/API do reads _and_ writes. Then, we can provide a wrapper script to help them get started.
> 
> That way the serialization / deserialization / keys of the secrets stays an internal detail of the class and does not leak out. At present, the deserialization is located here, while the serialization is outside.
I think we will want to set this up using terraform or AWS UI not python. Terraform makes the most sense to me because then we can insure they are store in a standardized way. But the python class should only be for reading secrets

hasher-matcher-actioner/hmalib/aws_secrets.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+"""
+Wrapper functions for reading secrets stored in AWS
+"""
+
+import boto3
+import base64
+import typing as t
+import ast
```
jeberl, line 9: will remove remove

schatten (approved):
LGTM. #shipit

PR #418, opened 2021-03-12
Add details on how to build docker based on local python-threatexchange

jeberl:
Summary
---------
In order to get docker to load the local version of `threatexchange` we first must copy the local version into the same directory as docker. We copy to a new directory called `local_threatexchange`.

In our dockerfile we now first check if this directory is a python module (does it have a `setup.py` file) and if so we copy the directory into our docker image. Lastly, we install the local package using pip. If we do this before we pip install the `requirements.txt` file the local version of `threatexhcange` is used rather than the published one

I added functions to the make file to do the above copy steps and also to delete the copied file after docker is made. Additionally edited the `make docker` command to create an empty directory because docker will fail if it tries to copy a directory that doesn't exist and there is no way to conditionally copy

Test Plan
---------
1. Edit the `fetcher.py` class to depend on some local changes to the `threatexchange` 
2. Add the following line to bottom of docker file to ensure that can be correctly executed 
```
RUN python3 ./hmalib/lambdas/fetcher.py
```
2. `$ make docker` fails why trying to run the `fetcher.py` file because it is pointed to the wrong version of `threatexchange` module
3. `$ make docker_local_te` and should succeed because it is pointed at the correct version of TE


hasher-matcher-actioner/hmalib/lambdas/aws_secrets.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
jeberl, line 1: Ignore this file. It is already in [another pull request](https://github.com/facebook/ThreatExchange/pull/417) that will be merged first

hasher-matcher-actioner/Dockerfile
```diff
@@ -14,3 +27,5 @@ ARG DEPS_PATH
 COPY --from=build $DEPS_PATH $DEPS_PATH
 ENV PYTHONPATH $DEPS_PATH
 COPY hmalib ./hmalib
+
+RUN python3 ./hmalib/lambdas/fetcher.py
```
schatten, line 31: This seems off. `fetcher.py` does not have an `if __name == "__main__":` block.

Did you accidentally check this in?

hasher-matcher-actioner/Makefile
```diff
@@ -11,17 +11,28 @@ REPOSITORY_NAME = hma-lambda-dev
...
 
 all:
 	@echo >&2 "Must specify target. ${FAIL}"
 
 docker:
+	if [ ! -d ${LOCAL_TE_DIR} ] ; then mkdir ${LOCAL_TE_DIR} ; fi
 	docker build -f Dockerfile -t ${REPOSITORY_NAME} .
 
+docker_with_local_te:
+	cp -r ../python-threatexchange ${LOCAL_TE_DIR}
```
schatten, line 24: Would a symlink work better, perhaps? That way, we don't need to keep syncing things between the two copies of threatexchange.

This might make things weird for @Dcallies though. @Dcallies does your setup do symlinks?

schatten (approved):
Approve to unblock:
1) Would prefer symlinks
2) Last line in the dockerfile is probably something to look at.

hasher-matcher-actioner/Makefile
```diff
@@ -11,17 +11,28 @@ REPOSITORY_NAME = hma-lambda-dev
 DOCKER_TAG ?= ${USER} # todo - change to ${USER}-latest?
 DOCKER_URI = $(call shell-or-die,docker images --filter=reference='*/${REPOSITORY_NAME}:${DOCKER_TAG}' --format='{{.Repository}}')
 
+LOCAL_TE_DIR = local_threatexchange # Should match LOCAL_THREAT_EXCHANGE ARG in Dockerfile
 
 all:
 	@echo >&2 "Must specify target. ${FAIL}"
 
 docker:
+	if [ ! -d ${LOCAL_TE_DIR} ] ; then mkdir ${LOCAL_TE_DIR} ; fi
```
Dcallies, line 20: Hmm, not a fan of this. Would rather have two dockerfiles. Does Dockerfile.dev solve this issue? https://stackoverflow.com/questions/39093806/what-is-a-dockerfile-dev-and-how-is-it-different-from-dockerfile

hasher-matcher-actioner/Dockerfile
jeberl: Yes. This was how I was testing it could run correctly. Removing

jeberl:
Given that we no longer want to use the same docker image, I believe that the work required to make this functional will be too high to do entirely programmatically instead, if you want to build your docker image using local version of threat exchange do the following

1. from hasher-matcher-actioner directory run
```
cp -r ../python-threatexchange local_threatexchange
```
2. edit the MakeFile to include the following lines **before** the pip install requirements:
```
ARG LOCAL_THREAT_EXCHANGE=./local_threatexchange
COPY $LOCAL_THREAT_EXCHANGE $LOCAL_THREAT_EXCHANGE
RUN python3 -m pip install ./local_threatexchange --target "${DEPS_PATH}"
```

3. Run `make docker`
4. run `rm -r local_threatexchange`

PR #419, opened 2021-03-12
Adds new simple auth and upload endpoint

BarrettOlson:
Summary
---------

Adds new simple auth and upload endpoint + rename status api (apologies for not breaking this up more, I had code in the works when we made the call about auth and used the changes to test.)

Don't have my heart set of the naming but I think api_<something> matches the pdq_<something> nicely. Also wasn't sure what we wanted for authorized response context (not leaving `context": {"AuthInfo": "Customer1"}}` but wanted something there to test)

Test Plan
---------

1) update local `terraform.tfvars` to contain:
`api_access_token="<token-value>"`

2) apply
```
hasher-matcher-actioner $ make upload_docker
hasher-matcher-actioner $ terraform -chdir=terraform apply
...   
yes
...
```
3) look up the "Invoke URL" from the default stage of the API gateway in the AWS Console (might be worth adding to `outputs.tf`)

4) Visit in browser or Postman:
`<Invoke URL>/?access_token=<token-value>`

5) Modify db state in AWS console and test endpoints again... repeat


eriksendc (approved):

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import logging
+import os
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
```
Dcallies, line 6: Please don't keep repeating this pattern, I believe it to be an antipattern.

logging.getLogger() will get the "root" logger
logger.setLevel() will change the  level of the "root" logger.

Most applications want to control the overall logging behavior of the application by changing settings on the root logger. This will overwrite that.

If you do want use logger, do the following

```
logger = logging.getLogger(__name__)
# don't set level

# or
def get_logger():
   """This pattern prevents creates implicitly creating a root logger by creating the sub-logger named __name__"""
  return logging.getLogger(__name__)
```

BarrettOlson: All of our lambda that do logging have this pattern... if I don't set the level somewhere it won't generate the specific logs. for the `def get_logger()` options, can this code be shared between lambdas?

PR #420, opened 2021-03-12
Support image file bytes as input to PDQ hashing

schatten:
Summary
---
In HMA, we fetch file bytes from s3 and then hash them. Writing them to an interim temp file only to read them again is inefficient and has proven to be a source of errors when the file is not flushed after writes.

This allows directly passing along bytes from the python program to the hasher.

also: Tests and typing were added to existing functions.

Test Plan
---
```
$ py.test
```
Screen Shot 2021-03-12 at 4 32 02 PM


BarrettOlson (approved):
LGTM! Fewer steps and adds typing, what's not to love? :)

PR #424, opened 2021-03-15
Store TE api access token in AWS as a secret

jeberl:
Summary
---------
Added secret that user inputs as a terraform variable

Test Plan
---------
1. Edit `terraform.tfvars` to include the variables `te_api_token`
2. 
```
terraform init
terraform validate
terraform apply
```
3. validate that the new secret shows up correctly in aws

### DIFF
--- a/hasher-matcher-actioner/terraform/variables.tf
+++ b/hasher-matcher-actioner/terraform/variables.tf
@@ -46,3 +46,14 @@ variable "include_cloudfront_distribution" {
   type        = bool
   default     = false
 }
+
+variable "te_app_id" {
+  description = "The App ID used to access to ThreatExchange"
+  type = string
+}
+
+variable "te_app_secret" {
+  description = "The secret token used to authenticate your access to the ThreatExchange app. You can find this by navigating to https://developers.facebook.com/apps/<YOUR-APP-ID>/settings/basic/ "
+  type = string
+  sensitive = true
+}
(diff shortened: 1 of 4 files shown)


Dcallies:
I am confused by this choice, can you go into more details?

> TE API access tokens are stored as <TE-APP-ID> + | + <TE-APP-SECERT>. These should be input as the user separately to limit confusion 

All the places that give you the token include the bar version. Seems like it might be more confusing to ask users to strip into separate parts.

> because the app ID might be use separately later (displayed in UI?)

The way the CLI handles this is by just doing str.split("|") on the secret.

hasher-matcher-actioner/terraform/variables.tf
```diff
@@ -46,3 +46,14 @@ variable "include_cloudfront_distribution" {
...
   default     = false
 }
+
+variable "te_app_id" {
+  description = "The App ID used to access to ThreatExchange"
+  type = string
+}
+
+variable "te_app_secret" {
+  description = "The secret token used to authenticate your access to the ThreatExchange app. You can find this by navigating to https://developers.facebook.com/apps/<YOUR-APP-ID>/settings/basic/ "
```
Dcallies, line 56: I'd suggest https://developers.facebook.com/tools/accesstoken/ instead, which includes all of your apps (and matches what the CLI does: https://github.com/facebook/ThreatExchange/blob/master/python-threatexchange/threatexchange/cli/main.py#L142 )

Dcallies (changes requested):
Whoops, accidentally fragmented review in a bunch of comments

hasher-matcher-actioner/terraform/variables.tf
jeberl: Didn't know about this link! Makes sense to me!

Dcallies (approved):
shipit

PR #425, opened 2021-03-15
PDQ Hash photos without temporary files

schatten:
Summary
---
At 1MB, some images were not getting hashed because the buffer wasn't getting automatically flushed. I did some digging around.

It appears that we need not write to a file at all. We can manage everything in memory. This depends on another merged PR in threatexchange #420  and accordingly updates the threatexchange version number from #421.

Test Plan
---
Used a docker image built from these sources to run perf tests on ~300 distinct 1 and 10 MB images. The runs were succesful over 100k hashing requests.


BarrettOlson (approved):
Nice!

PR #426, opened 2021-03-16
Webapp wireframe prototype v0.

eriksendc:
Summary
---------

Check in work in progress UI, which is mostly shimmed/non-functional. We'll add functionality out over the next few weeks.

Test Plan
---------

Start webserver locally, visit in browser


Dcallies (approved):
I wish I could use image macros on github!

PR #430, opened 2021-03-17
add Api.js and connect upload image to UI

BarrettOlson:
Summary
---------

You can now use the uploads page on the UI to upload to the images 'folder' in the s3 bucket. I also refactored API requests to their own file.

Test Plan
---------
`terraform apply` and update `API_ROOT` + `API_TOKEN` in `Api.js` to match your deployed instance.
then navigate to the /uploads page either from aws or by spinning up the webapp locally. 
You should be able to submit a file and have it appear in the s3 bucket (for the hma pipeline to try and match).

### DIFF
--- a/hasher-matcher-actioner/webapp/src/Api.js
+++ b/hasher-matcher-actioner/webapp/src/Api.js
@@ -5,41 +5,54 @@
 // TODO figure out root path. For now copy your own URL gateway here
 const API_ROOT = 'https://<APPID>.execute-api.us-east-1.amazonaws.com'
 // TODO replace with real auth someday
-const API_TOKEN = '<TOKEN>'
+const API_TOKEN_KEY = 'access_token';
+const API_TOKEN = 'asupersecrettoken';
 
-export async function getAPI(
-    route,
-    params = {},
-  ) {
-    const urlWithParams = new URL(`${API_ROOT}/${route}`);
-    urlWithParams.searchParams.append("access_token", API_TOKEN);
-    for (const [key, value] of Object.entries(params)) {
-      urlWithParams.searchParams.append(key, value);
-    }
+export async function fetchMatches() {
+  return getAPI("matches")
+    .then((data) => {
+      return (data["matches"])
+    })
+}
 
-    const response = await fetch(urlWithParams.toString(), {
-        method: 'GET',
-      });
-    return response.json();
+export async function uploadImage(fileObject) {
+  const formData = new FormData();
+  formData.append("image", fileObject);
+  return postAPI('upload', formData);
+}
+
+async function getAPI(
+  route,
+  params = {},
+) {
+  const urlWithParams = new URL(`${API_ROOT}/${route}`);
+  urlWithParams.searchParams.append(API_TOKEN_KEY, API_TOKEN);
+  for (const [key, value] of Object.entries(params)) {
+    urlWithParams.searchParams.append(key, value);
   }
 
-export async function postAPI(
-    route,
-    body,
-    params = {},
-  ) {
-    const urlWithParams = new URL(`${API_ROOT}/${route}`);
-    urlWithParams.searchParams.append('access_token', API_TOKEN);
+  const response = await fetch(urlWithParams.toString(), {
+    method: 'GET',
+  });
+  return response.json();
+}
 
-    for (const [key, value] of Object.entries(params)) {
-      urlWithParams.searchParams.append(key, value);
-    }
-    
-    const requestOptions = {
-      method: 'POST',
-      body: body
-    }
+async function postAPI(
+  route,
+  body,
+  params = {},
+) {
+  const urlWithParams = new URL(`${API_ROOT}/${route}`);
+  urlWithParams.searchParams.append(API_TOKEN_KEY, API_TOKEN);
 
-    const response = await fetch(urlWithParams.toString(), requestOptions);
-    return response.json();
+  for (const [key, value] of Object.entries(params)) {
+    urlWithParams.searchParams.append(key, value);
   }
+
+  const requestOptions = {
+    method: 'POST',
+    body: body
+  }
+
+  return await fetch(urlWithParams.toString(), requestOptions);
+}
--- a/hasher-matcher-actioner/webapp/src/Api.js
+++ b/hasher-matcher-actioner/webapp/src/Api.js
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+// TODO figure out root path. For now copy your own URL gateway here
+const API_ROOT = 'https://<APPID>.execute-api.us-east-1.amazonaws.com'
+// TODO replace with real auth someday
+const API_TOKEN = '<TOKEN>'
+
+export async function getAPI(
+    route,
+    params = {},
+  ) {
+    const urlWithParams = new URL(`${API_ROOT}/${route}`);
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/webapp/src/Upload.js
```diff
@@ -7,41 +7,57 @@ import { Link } from 'react-router-dom';
...
-                  setSubmitting(false);
-                }, 3000);
-              }}>Submit</Button>
-            </div>
+              <div className="custom-file">
+                <input type="file" className="custom-file-input" id="customFile" onChange={(e) => {
+                  setFileName(e.nativeEvent.path[0].files[0].name);
+                  setImage({
+                    preview: URL.createObjectURL(e.nativeEvent.path[0].files[0]),
+                    raw: e.nativeEvent.path[0].files[0]
```
eriksendc, line 43: Mini nit / preference: dereference `e.nativeEvent.path[0].files[0]` to something like `file` and then use `file` at 40, 42 and 43.

hasher-matcher-actioner/webapp/src/Api.js
```diff
@@ -0,0 +1,45 @@
+/**
```
Dcallies, line 1: Fancy!

hasher-matcher-actioner/webapp/src/Api.js
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+// TODO figure out root path. For now copy your own URL gateway here
+const API_ROOT = 'https://<APPID>.execute-api.us-east-1.amazonaws.com'
+// TODO replace with real auth someday
+const API_TOKEN = '<TOKEN>'
```
Dcallies, line 8: For now, you can hardcode it with the placeholder, unless this is the new placeholder

hasher-matcher-actioner/webapp/src/Api.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+// TODO figure out root path. For now copy your own URL gateway here
+const API_ROOT = 'https://<APPID>.execute-api.us-east-1.amazonaws.com'
```
Dcallies, line 6: (Question for Brian maybe): Any way to force this to require someone to type in something here?

Dcallies (commented):
As discussed with @eriksendc, want him to sign off on all the JS stuff until more of us are familiar with the code. 

Can you add a photo/video of the flow in the test plan?

hasher-matcher-actioner/webapp/src/Upload.js
```diff
@@ -50,18 +66,15 @@ export default function Upload() {
           </Collapse>
           <Collapse in={submitted}>
             <div>
-              <p>Hash created for {fileName}: acecf3355e3125c8e24e2f30e0d4ec4f8482b878b3c34cdbdf063278db275992</p>
-              <p>...either this...</p>
-              <p>{fileName} does not match any signals.</p>
-              <p>...or this...</p>
-              <p>{fileName} matches one or more signals.</p>
+              <p>Uploaded {fileName}: hash should be generated shortly... </p>
```
eriksendc, line 69: Just calling out this bit needs some kind of polling to check for the hash and match(es). What do you think about two subsequent APIs here, one to check for a hash and one for matches? As they respond, the UI's content would change as follows:

1. first, only call the api looking up the hash; show the message at 69 as is until that api returns info. Assume polling every few seconds is OK for this.
2. next, after we have the hash, start polling for matches. Initially show "Checking for matches..." as the message. Once the api returns something, show information about the matches. If we poll a number of times such that the polling has happened for 60s (120s?) show "No match(es) found."

hasher-matcher-actioner/webapp/src/Upload.js
```diff
@@ -7,41 +7,57 @@ import { Link } from 'react-router-dom';
...
 export default function Upload() {
-  const [ fileName, setFileName ] = useState('Select Browse to choose a file.');
-  const [ submitting, setSubmitting ] = useState(false);
-  const [ submitted, setSubmitted ] = useState(false);
+  const [fileName, setFileName] = useState('Select Browse to choose a file.');
+  const [submitting, setSubmitting] = useState(false);
+  const [submitted, setSubmitted] = useState(false);
+  const [image, setImage] = useState({ preview: "", raw: "" });
+
+  function handleUpload() {
```
schatten, line 18: nit: I like my './Api' equivalent to have more function oriented surfaces. 

eg.
```
import { uploadToHashSourceBucket } from './Api';

...
  uploadToHashSourceBucket(fileObject).then(...);
...

The present setup 'leaks' the internal details to the UI component and is nastier to reuse.

File: hasher-matcher-actioner/webapp/src/Upload.js [hunk 7]
```diff
@@ -7,41 +7,57 @@ import { Link } from 'react-router-dom';
 import { Button, Col, Collapse, Image, Row } from 'react-bootstrap';
 import Spinner from 'react-bootstrap/Spinner';
 
+import { postAPI } from './Api'
+
 export default function Upload() {
-  const [ fileName, setFileName ] = useState('Select Browse to choose a file.');
-  const [ submitting, setSubmitting ] = useState(false);
-  const [ submitted, setSubmitted ] = useState(false);
+  const [fileName, setFileName] = useState('Select Browse to choose a file.');
```
schatten, line 13: Since this is a react component, would love a more UI oriented name. eg. PreviewableImageUploadPane().

hasher-matcher-actioner/webapp/src/Upload.js
```diff
@@ -50,18 +66,15 @@ export default function Upload() {
           </Collapse>
           <Collapse in={submitted}>
             <div>
-              <p>Hash created for {fileName}: acecf3355e3125c8e24e2f30e0d4ec4f8482b878b3c34cdbdf063278db275992</p>
-              <p>...either this...</p>
-              <p>{fileName} does not match any signals.</p>
-              <p>...or this...</p>
-              <p>{fileName} matches one or more signals.</p>
+              <p>Uploaded {fileName}: hash should be generated shortly... </p>
+              <p> if a match is found it will be visable here: </p>
```
schatten, line 70: spolling! visible.

hasher-matcher-actioner/webapp/src/Api.js
```diff
@@ -0,0 +1,45 @@
...
+      urlWithParams.searchParams.append(key, value);
+    }
+
+    const response = await fetch(urlWithParams.toString(), {
+        method: 'GET',
+      });
+    return response.json();
+  }
+
+export async function postAPI(
```
schatten, line 26: See nit on 'consumer' oriented APIs below.

postAPI and getAPI is too low level. You could make those private and expose a surface of high-level functions that do specific things.

Advantages:
* changing the API contract becomes local to this file
* you can make the APIs typed strictly which further makes it reuseable.

schatten (changes requested):
Functionally, no issues. Love that we are thinking about previews and general UX.

However, from a code-structure and separations-of-concerns POV, we could do better. Outlined suggestions in the comments below.

schatten (approved):
LGTM! #shipit

BarrettOlson:
Thanks @schatten!
Merging in advance of a linting refactor but happy to incorporate any additional feedback you have @eriksendc

PR #431, opened 2021-03-17
Update aws_secretsmanager_secret.recovery_window_in_days to 0 for easy apply/destroy

BarrettOlson:
Summary
---------

aws secret defaults to recovery_window_in_days of 7 and while a secret is 'scheduled for deletion' it can't be created. If we set this to 0 we avoid this. 

Test Plan
---------

`terraform -chdir=terraform apply`
`terraform -chdir=terraform destroy`
`terraform -chdir=terraform apply`


schatten (approved):
Noice!

PR #433, opened 2021-03-17
[HMA-UI] add eslint and prettier ci action for webapp

BarrettOlson:
Summary
---------

lint is love. This looks like a major refactor but the code logic shouldn't have changed much (added some props here and there for `eslint-plugin-jsx-a11y` etc and a few minor refactors to make eslint happy). 

Warning: you won't be able to `npm run build` (and `npm start` might be flaky) if you have eslint errors. If you _really_ need to build and don't have time to suppress the rule in the config of inline temporary you can use.
`DISABLE_ESLINT_PLUGIN=true npm run build`  VSCode plugins will be your friends as well.

I plan on updating `webapp/README.md` once I get some initial feedback.

Test Plan
---------
Tested that existing match results and photo upload still work. 
now we can run:
```
npx eslint src --ext .js,.jsx  
npx prettier --check src 
```


.github/workflows/hma-webapp-ci.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: yarn
+      - name: Run ESLint
+        run: eslint src
```
schatten, line 28: Maybe related to the failure we are seeing, but perhaps `./node_modules/.bin/eslint` would be more appropriate.

eriksendc (dismissed):

schatten (approved):
Approving to unblock. I'm seeing that the workflow can be tested while a PR. 

But this seems to be a Draft PR, so dunno why I'm even reviewing.

.github/workflows/hma-webapp-ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm install
+      - name: Run ESLint
+        run: |
+          npx eslint src --ext .js,.jsx
```
schatten, line 29: <3

hasher-matcher-actioner/webapp/package.json
```diff
@@ -5,8 +5,12 @@
   "dependencies": {
     "@testing-library/jest-dom": "^5.11.9",
     "@testing-library/react": "^11.2.5",
-    "@testing-library/user-event": "^12.8.1",
+    "@testing-library/user-event": "^13.0.1",
     "bootstrap": "^4.6.0",
+    "eslint": "^7.22.0",
```
schatten, line 10: nit: Is it customary for things to be dependencies AND devDependencies?

While it is unlikely someone does `npm install <hma>`, this seems weird.

schatten (approved):
This looks great!

Great that we renamed the files as `.jsx`. It's more indicative, perhaps vscode will behave better now.

This is good work. Linters and general tooling is hard to miss these in the first flurry of pull requests, but I'm glad you took the time! This provides necessary guard rails. :) 



---
nit: When you are doing rename and loads of changes, best to do `git mv`. That way git recognizes that you are renaming and doesn't show full delete + full add instead.

AFAIK, there is no way to make this happen after you've already renamed. Git sucks in some ways.

hasher-matcher-actioner/webapp/package.json
BarrettOlson: good catch, these don't need to be in dependencies; moved them to devDependencies.

eriksendc (approved):

PR #435, opened 2021-03-17
[hma][ui] Add reference to actioner action in mock

Dcallies:
Summary
---------

I figured action taken is an interesting bit of result, and it's not on any page. Seems like we might want to see it here.

Test Plan
---------

image


schatten (approved):
<3

BarrettOlson (approved):

PR #436, opened 2021-03-18
[HMA] mypy fixes; black, py.test, mypy in CI; rename dtos -> models

schatten:
Typing fixes; rename dtos to models
    
Summary
---
* Ran mypy along black and realised we've been neglecting types for a while now.
* Made the typing fixes and added lint, type-checks, tests for HMA to CI
* Because the DTOs now specify how they are stored and how they are queried, they are models. eg. [Django's models](https://docs.djangoproject.com/en/3.1/topics/db/models/), [Flask-SQLAlchemy's models](https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/), [Tornado's models](https://github.com/BSTester/tornado-models). I promise this is the last rename I do. 🙏 

Test Plan
---
```
$ cd hasher-matcher-actioner
$ mypy hmalib
```

Screen Shot 2021-03-17 at 20 30 14


hasher-matcher-actioner/hmalib/metrics/__init__.py
```diff
@@ -75,7 +75,7 @@ def _no_op_timer(name):
     yield
 
 
-def _no_op_flush():
+def _no_op_flush(namespace: str = "does not matter"):
```
BarrettOlson, line 78: nit: ` = "no_po"` is it possible spaces in the namespace might mess something up downstream?

BarrettOlson (approved):
LGTM! Thanks for doing this! 
Did you see I added: https://github.com/facebook/ThreatExchange/blob/master/.github/workflows/hmalib-ci.yaml ?
I think we may have some overlap we might want to consolidate

schatten:
Yes! You're correct. Some consolidation is in order. I'll raise a subsequent one after merging this in.

hasher-matcher-actioner/hmalib/metrics/__init__.py
schatten: Extremely unlikely. This is just to get mypy to stop crying. This string is not published anywhere.

PR #437, opened 2021-03-18
Fetcher syncer

Dcallies:
Summary
---------

Now connect to s3 and store state there as though they are files.

Test Plan
---------

Didn't test end-to-end, only by running locally and using the clowny path in the file.

```
$ python  -m hmalib.lambdas.fetcher
INFO:__main__:Triggered at time 15:55:58, found 2 collabs: HMA Test Collab 1, HMA Test Collab 2
INFO:__main__:Processing updates for collaboration HMA Test Collab 1
WARNING:__main__:No s3 checkpoint for 303636684709969. First run?
WARNING:__main__:Store for HMA Test Collab 1 - 303636684709969 stale! Resetting.
INFO:root:Fetch complete, applying 138 updates
INFO:__main__:138 rows stored for 303636684709969
INFO:__main__:Stored checkpoint for 303636684709969. last_fetch_time=1616108159 fetch_checkpoint=1616108158
INFO:__main__:Processing updates for collaboration HMA Test Collab 2
WARNING:__main__:No s3 checkpoint for 258601789084078. First run?
WARNING:__main__:Store for HMA Test Collab 2 - 258601789084078 stale! Resetting.
INFO:root:Fetch complete, applying 77 updates
INFO:__main__:77 rows stored for 258601789084078
INFO:__main__:Stored checkpoint for 258601789084078. last_fetch_time=1616108160 fetch_checkpoint=1616108160

$ python  -m hmalib.lambdas.fetcher
INFO:__main__:Triggered at time 16:25:53, found 2 collabs: HMA Test Collab 1, HMA Test Collab 2
INFO:__main__:Processing updates for collaboration HMA Test Collab 1
INFO:__main__:Loaded checkpoint for 303636684709969. last_fetch_time=1616108159 fetch_checkpoint=1616108158
INFO:root:Fetch complete, applying 0 updates
INFO:__main__:Stored checkpoint for 303636684709969. last_fetch_time=1616109954 fetch_checkpoint=1616109954
INFO:__main__:Processing updates for collaboration HMA Test Collab 2
INFO:__main__:Loaded checkpoint for 258601789084078. last_fetch_time=1616108160 fetch_checkpoint=1616108160
INFO:root:Fetch complete, applying 0 updates
INFO:__main__:Stored checkpoint for 258601789084078. last_fetch_time=1616109954 fetch_checkpoint=1616109954
```

PR #434, opened 2021-03-17
Make fetch pull from te and save locally

jeberl:
Summary
---------

Make a fetch command that reads API tokens and collab configs from TE, fetches from ThreatExchange, and then stores the result locally.

Also add a type argument to the aws command threatupdates command. To push these changes to pypi we bump the version

A future PR will have this write the result to AWS. Good amount of hard coded stuff in here now that will be removed such as table names and secret keys

Test Plan
---------
1. upload collab configs to DynamoDB
2. upload TE app token to AWS Secrets Manager
3. Before the python-threatexchange version bump we have to direct python to our local copy of python-threatexchange with:
```python3 -m pip install python-threatexchange/```
4. Run fetcher:
```
python3 -m hmalib.lambdas.fetcher
```
5. See if the correct collab data is read correctly by looking for subdirectories under `temp_te_data` directory

PR #438, opened 2021-03-19
Bump Pyte version after changes

jeberl:
Summary
---------

After changes in PR #437 and #434 bump the python-threatexchange module in pypi to publish the changes


schatten (approved):
LGTM. 

For posterity, though, it is worthwhile to call out that the `dataclasses` package is only compatible with python > 3.6. So, in effect, python-threatexchange requires python:3.6+

Issue #440, opened 2021-03-19
dataclasses library breaks docker build


jeberl:
To reproduce add `RUN python3.8 -m hmalib.lambdas.pdq.pdq_hasher` to bottom of Dockerfile and run `make docker`

stack trace:
```
 > [prod 4/4] RUN python3.8 -m hmalib.lambdas.pdq.pdq_hasher:
#11 1.001 /hma-deps/pdqhash/__init__.py:3: UserWarning: Hash vector order changed between version 0.1.8 and 0.2.0. See https://github.com/faustomorales/pdqhash-python/issues/1 for more details.
#11 1.001   warnings.warn(
#11 1.001 Traceback (most recent call last):
#11 1.001   File "/var/lang/lib/python3.8/runpy.py", line 194, in _run_module_as_main
#11 1.001     return _run_code(code, main_globals, None,
#11 1.001   File "/var/lang/lib/python3.8/runpy.py", line 87, in _run_code
#11 1.001     exec(code, run_globals)
#11 1.001   File "/var/task/hmalib/lambdas/pdq/pdq_hasher.py", line 15, in <module>
#11 1.001     from hmalib.models import PipelinePDQHashRecord
#11 1.001   File "/var/task/hmalib/models.py", line 24, in <module>
#11 1.001     class PDQRecordBase(DynamoDBItem):
#11 1.001   File "/hma-deps/dataclasses.py", line 958, in dataclass
#11 1.001     return wrap(_cls)
#11 1.001   File "/hma-deps/dataclasses.py", line 950, in wrap
#11 1.001     return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
#11 1.001   File "/hma-deps/dataclasses.py", line 800, in _process_class
#11 1.001     cls_fields = [_get_field(cls, name, type)
#11 1.001   File "/hma-deps/dataclasses.py", line 800, in <listcomp>
#11 1.001     cls_fields = [_get_field(cls, name, type)
#11 1.001   File "/hma-deps/dataclasses.py", line 659, in _get_field
#11 1.001     if (_is_classvar(a_type, typing)
#11 1.001   File "/hma-deps/dataclasses.py", line 550, in _is_classvar
#11 1.001     return type(a_type) is typing._ClassVar
#11 1.001 AttributeError: module 'typing' has no attribute '_ClassVar'
------
executor failed running [/bin/sh -c python3.8 -m hmalib.lambdas.pdq.pdq_hasher]: exit code: 1
[...]

PR #441, opened 2021-03-19
Revert "Bump Pyte version after changes"

jeberl:
Reverts facebook/ThreatExchange#438

This should fix issue 440 
https://github.com/facebook/ThreatExchange/issues/440


python-threatexchange/version.txt
```diff
@@ -1 +1 @@
-0.0.16
```
Dcallies, line 1: I don't think this does anything. I don't you can unpush a version.

Dcallies: Also, a good example of "Don't bump the version in the same diff that makes changes!"

Dcallies (changes requested):

python-threatexchange/version.txt
```diff
@@ -1 +1 @@
...
+0.0.15
```
Dcallies, line 1:
```suggestion
0.0.16
```

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -76,7 +76,9 @@ def __init__(
         self.processed = 0
         self.counts = collections.Counter()
 
-    def execute(self, api: ThreatExchangeAPI, dataset: Dataset) -> None:
+    def execute(
+        self, api: ThreatExchangeAPI, dataset: Dataset
+    ) -> None:
         privacy_groups = dataset.config.privacy_groups
```
Dcallies, line 82:
```suggestion
    def execute(self, api: ThreatExchangeAPI, dataset: Dataset) -> None:
        privacy_groups = dataset.config.privacy_groups
```

python-threatexchange/threatexchange/cli/experimental_fetch.py
```diff
@@ -76,7 +76,9 @@ def __init__(
         self.processed = 0
         self.counts = collections.Counter()
 
-    def execute(self, api: ThreatExchangeAPI, dataset: Dataset) -> None:
+    def execute(
+        self, api: ThreatExchangeAPI, dataset: Dataset
+    ) -> None:
```
Dcallies, line 81:
```suggestion
```suggestion
    def execute(self, api: ThreatExchangeAPI, dataset: Dataset) -> None:
```
```

Dcallies (approved):

PR #442, opened 2021-03-19
[HMA] Make te default source and generalize dynamo key value for match

BarrettOlson:
Summary
---------

This replaces 'te#' for 's#' so the matches aren't assumed to be exclusively from a te source. 
e.g. S.K. for matches will now be `s#te#555555555555` instead of just `te#555555555555`

(also renamed other test script)

Test Plan
---------
tf destroy and apply then, fixed a bug and updated in `hasher-matcher-actioner/tests/scripts/check_deployed_instance.py` and ran `make dev_test_instance`


schatten (approved):
LGTM.

PR #445, opened 2021-03-19
[hma] Wrap hmalib with setup, install with pip

Dcallies:
Summary
---------

dataclass is forwards compatible, but only if properly shadowed by the real library. @schatten theorized this was because we were messing with the python path, which seems to be correct.

Fixes  #440

Test Plan
---------

```
$ cd hasher-matcher-actioner
$ docker build . - testington
$ docker run -p 9000:8080 testington hmalib.lambdas.fetcher.lambda_handler
# Trigger CLI from the docker desktop
$ sh-4.2# python3.8 -m hmalib.lambdas.pdq.pdq_hasher
# Get another environment variable error, but not the dataclass one
```

PR #446, opened 2021-03-19
Revert 438 pyte v0 0 16

jeberl:
Summary
---------

This PR resolves Issue 440 which dealt with conflicting versions of the [dataclasses module](https://docs.python.org/3/library/dataclasses.html). This module only exists for python versions >=3.7 however there is a [backport module by the same name](https://pypi.org/project/dataclasses/) also available which makes it compatible with version 3.6. However using this backport module breaks later versions and so we decided not to use it and instead have a soft requirement that HMA requires python versions >=3.7.

We choose to make >=3.7 a soft requirement rather than a hard requirement because it as actually quite easy to run with version 3.6, all one must do is install the dataclasses backport

This PR makes a number of changes:
- updates CONTRIBUTING.md to note the above
- stops testing HMA on python 3.6
- removes the one usage of `dataclasses` from the `python-threatexchange` module (`threat_updates.py`) because it is not compatible with python3.6 (something we have want to continue to allow)
- removes the previously added dataclasses backport from the `python-threatexchange` requirements
- bumps version on `python-threatexchange` to correct this change

The current state of the world is that HMA will require >=python3.7 while `python-threatexchange` will still be able to handle 3.6

Test Plan
---------

- Tested threat_updates.py locally
- if CI pass, HMA should be comaptable with python3.7 and 3.8
- Docker successfully builds and files can both `import threatexchange` and `import hmalib` work


.github/workflows/python-threatexchange-ci.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
```
schatten, line 39: If we are removing dataclass usage, why NOT test against 3.6?

jeberl: Ohh. I had thought this was testing for hmalib. I see now for pyte. reverting

jeberl:
Did some wonky things with my rebase. may have to force push after acceptance

schatten:
> Did some wonky things with my rebase. may have to force push after acceptance

If there are conflicts, we typically merge in master into this branch, so history doesn't change, but the branch becomes mergeable. Prefer you'd do this after acceptance.

python-threatexchange/threatexchange/threat_updates.py
```diff
@@ -10,7 +10,6 @@
 import pathlib
 import time
 import typing as t
-from dataclasses import dataclass
```
Dcallies, line 13: I fixed this, you can leave it in - #445

Dcallies (changes requested):
I'm not sure that there is such thing as "soft requirements" or "soft support" for 3.6. Either it's tested and it works, or we don't support it. 

I believe my fix from #445 should work here just as well as it did for py-threatexchange if we also want to make HMA 3.6 compatible, though if part of the problem is that it doesn't play well with typing, then I'm fine with it just being >3.6

jeberl:
Fixes were already merged. Moving lint issues to #453 and closing this PR

PR #448, opened 2021-03-23
[HMA] rename Timestamp to UpdatedAt and expand models for querying 

BarrettOlson:
Summary
---------

Timestamp is a reserved word, there are ways around it but UpdateAt is clearer anyway.
Add query interface to be used in API and possibly actioners.

 
Test Plan
---------
`make dev_test_instance` passes 
also added `hmalib/tests/test_models.py` which the CI runs.

### DIFF
--- a/hasher-matcher-actioner/hmalib/models.py
+++ b/hasher-matcher-actioner/hmalib/models.py
@@ -25,6 +25,18 @@ class DynamoDBItem:
     def to_dynamodb_item(self) -> t.Dict:
         raise NotImplementedError
 
+    @staticmethod
+    def get_dynamodb_content_key(c_id: str) -> str:
+        return f"{DynamoDBItem.CONTENT_KEY_PREFIX}{c_id}"
+
+    @staticmethod
+    def get_dynamodb_type_key(type: str) -> str:
+        return f"{DynamoDBItem.TYPE_PREFIX}{type}"
+
+    @staticmethod
+    def remove_content_key_prefix(key: str) -> str:
+        return key[len(DynamoDBItem.CONTENT_KEY_PREFIX) :]
+
 
 @dataclass
 class PDQRecordBase(DynamoDBItem):
@@ -34,17 +46,9 @@ class PDQRecordBase(DynamoDBItem):
 
     SIGNAL_TYPE = "pdq"
 
-    content_key: str
+    content_id: str
     content_hash: str
     updated_at: datetime.datetime  # ISO-8601 formatted
-
-    @staticmethod
-    def get_dynamodb_content_key(key: str) -> str:
-        return f"{PDQRecordBase.CONTENT_KEY_PREFIX}{key}"
-
-    @staticmethod
-    def get_dynamodb_type_key(key: str) -> str:
-        return f"{PDQRecordBase.TYPE_PREFIX}{key}"
 
     def to_dynamodb_item(self) -> dict:
         raise NotImplementedError
@@ -63,7 +67,7 @@ class PipelinePDQHashRecord(PDQRecordBase):
 
     def to_dynamodb_item(self) -> dict:
         return {
-            "PK": self.get_dynamodb_content_key(self.content_key),
+            "PK": self.get_dynamodb_content_key(self.content_id),
             "SK": self.get_dynamodb_type_key(self.SIGNAL_TYPE),
             "ContentHash": self.content_hash,
             "Quality": self.quality,
@@ -75,29 +79,28 @@ class PipelinePDQHashRecord(PDQRecordBase):
         return {
             "hash": self.content_hash,
             "type": self.SIGNAL_TYPE,
-            "key": self.content_key,
+            "key": self.content_id,
         }
 
     @classmethod
-    def get_from_content_key(
+    def get_from_content_id(
         cls, table: Table, content_key: str
-    ) -> t.Union["PipelinePDQHashRecord", None]:
+    ) -> t.Optional["PipelinePDQHashRecord"]:
         items = HashRecordQuery.from_content_key(
             table,
             cls.get_dynamodb_content_key(content_key),
             cls.get_dynamodb_type_key(cls.SIGNAL_TYPE),
         )
-        records = []
-        for item in items():
-            records.append(
-                PipelinePDQHashRecord(
-                    item["PK"],
-                    item["ContentHash"],
-                    item["UpdatedAt"],
-                    item["Quality"],
-                )
+        records = [
+            PipelinePDQHashRecord(
+                item["PK"][len(DynamoDBItem.CONTENT_KEY_PREFIX) :],
+                item["ContentHash"],
+                datetime.datetime.fromisoformat(item["UpdatedAt"]),
+                item["Quality"],
             )
-        return None if not len(records) else records[0]
+            for item in items
+        ]
+        return None if not records else records[0]
 
 
 @dataclass
@@ -106,7 +109,7 @@ class PDQMatchRecord(PDQRecordBase):
     Successful execution at the matcher produces this record.
     """
 
-    signal_id: int
+    signal_id: t.Union[str, int]
     signal_source: str
     signal_hash: str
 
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/hmalib/models.py
```diff
@@ -74,14 +78,34 @@ def to_sqs_message(self) -> dict:
             "key": self.content_key,
         }
 
+    @classmethod
+    def get_from_content_key(
+        cls, table: Table, content_key: str
+    ) -> t.Union["PipelinePDQHashRecord", None]:
```
schatten, line 84: Idiomatic type of Union[X, None] is Optional[X]

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -107,3 +131,144 @@ def to_dynamodb_item(self) -> dict:
...
+        Given a hash type and time range, give me all the matches found for that type and time range
+        """
+        if start_time is None:
+            start_time = datetime.datetime.min.isoformat()
+        if end_time is None:
+            end_time = datetime.datetime.max.isoformat()
+        return table.query(
+            IndexName="GSI-2",
+            KeyConditionExpression=Key("GSI2-PK").eq(hash_type)
+            & Key("UpdatedAt").between(start_time, end_time),
```
schatten, line 272: Why must we provide a time constraint at all?

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -74,14 +78,34 @@ def to_sqs_message(self) -> dict:
...
+        for item in items():
+            records.append(
+                PipelinePDQHashRecord(
+                    item["PK"],
+                    item["ContentHash"],
+                    item["UpdatedAt"],
+                    item["Quality"],
+                )
+            )
+        return None if not len(records) else records[0]
```
Dcallies, line 100:
```suggestion
        return None if not records else records[0]
```

If you want to be extra silly you can do:
```
return (records or [None])[0]
```

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -107,3 +131,144 @@ def to_dynamodb_item(self) -> dict:
...
+                PDQMatchRecord(
+                    item["PK"],
+                    item["ContentHash"],
+                    item["UpdatedAt"],
+                    item["SK"],
+                    item["SignalSource"],
+                    item["SignalHash"],
+                )
+            )
+        return records
```
Dcallies, line 181: alt, ignorable - This might be a good candidate for a list comprehension

  return [PDQMatchRecord(...) for item in items]

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -59,51 +67,217 @@ class PipelinePDQHashRecord(PDQRecordBase):
...
+            cls.get_dynamodb_signal_key(signal_source, signal_id),
+            cls.SIGNAL_TYPE,
+        )
+        return cls._result_items_to_records(items)
+
+    @classmethod
+    def get_from_time_range(
+        cls, table: Table, start_time: str = None, end_time: str = None
+    ) -> t.List["PDQMatchRecord"]:
+        items = MatchRecordQuery.from_time_range(
```
schatten, line 169: If we were to create a second MatchRecord subclass, do you believe that would have the same Projection String? We might have discussed this before, but would like to get your present understanding.

schatten: Because if not, you might want to retain the ProjectionString as an attribute of this class rather than delegate to MatchRecordQuery.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -59,51 +67,217 @@ class PipelinePDQHashRecord(PDQRecordBase):
...
+            ProjectionExpression=cls.DEFAULT_PROJ_EXP,
+            FilterExpression=filter_exp,
+        ).get("Items", [])
+
+    @classmethod
+    def from_time_range(
+        cls, table: Table, hash_type: str, start_time: str = None, end_time: str = None
+    ) -> t.List[t.Dict]:
+        """
+        Given a hash type and time range, give me all the matches found for that type and time range
```
schatten, line 272: Can you be more specific here? Does between include both bounds or only one, or none? Would make a huge difference when we implement pagination.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -59,51 +67,217 @@ class PipelinePDQHashRecord(PDQRecordBase):
...
+    ) -> t.List[t.Dict]:
+        """
+        Given a Signal ID/Key (and optional hash type), give me any content matches found
+        """
+        filter_exp = None
+        if not hash_type is None:
+            filter_exp = Attr("HashType").eq(hash_type)
+
+        return table.query(
+            IndexName="GSI-1",
```
schatten, line 261: One easy/convenient way to document within code would be to create class constants that describe the index and then use them.

eg.
```python
class PDQMatchRecord:
  SIGNAL_TO_CONTENT_INDEX = "GSI-1"
...
  return table.query(IndexName=SIGNAL_TO_CONTENT_INDEX)
...
```

hasher-matcher-actioner/hmalib/tests/test_models.py
```diff
@@ -0,0 +1,206 @@
...
+            BillingMode="PAY_PER_REQUEST",
+            KeySchema=[
+                {"AttributeName": "PK", "KeyType": "HASH"},
+                {"AttributeName": "SK", "KeyType": "RANGE"},
+            ],
+            GlobalSecondaryIndexes=[
+                {
+                    "IndexName": "GSI-1",
+                    "KeySchema": [
+                        {"AttributeName": "GSI1-PK", "KeyType": "HASH"},
```
schatten, line 62: potential point of maintenance in addition to terraform. I see your point now.

hasher-matcher-actioner/mypy.ini
```diff
@@ -1,3 +1,8 @@
+[mypy]
+
+[mypy-threatexchange.*]
```
schatten, line 3: Sad! I thought most of threatexchange was typed.

hasher-matcher-actioner/requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 -r requirements.txt
 pytest==6.2.1
 mypy==0.812
+moto
```
schatten, line 4: You will have to figure out how to do this once you try to merge into master. David's setup.py changes will need to be accounted for. But you already knew that. :)

schatten (approved):
LGTM. Have some comments, but even in present shape good to merge!

Go ahead.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -59,51 +67,217 @@ class PipelinePDQHashRecord(PDQRecordBase):
...
+        cls, table: Table, signal_id: t.Union[str, int], signal_source: str
+    ) -> t.List["PDQMatchRecord"]:
+        items = MatchRecordQuery.from_signal_key(
+            table,
+            cls.get_dynamodb_signal_key(signal_source, signal_id),
+            cls.SIGNAL_TYPE,
+        )
+        return cls._result_items_to_records(items)
+
+    @classmethod
```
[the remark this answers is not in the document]
BarrettOlson: The distinction I made in my head (perhaps incorrectly) was to make things class methods if they are more involved functions that called other methods in the class hierarchy (e.g. here this method calls _result_items_to_records)

hasher-matcher-actioner/hmalib/models.py
BarrettOlson: Good question, I wasn't sure. I think you're likely right but after going back and forth with where I wanted thas string to live I decided on in MatchRecordQuery for now, if we end up not being able to/wanting to reuse it I will definitely move it to the data class.

hasher-matcher-actioner/mypy.ini
BarrettOlson: yeah... brought that up in #453 we should probably revisit

hasher-matcher-actioner/requirements-dev.txt
BarrettOlson: I think we only need to worry about setup.py for deployed images, if you run the commands that are in the CI locally this seems to be enough but I think it is probably work consolidating to just using setup.py eventually

PR #449, opened 2021-03-23
Webapp make a clean "prod" terraform work + webapp minor fixes

eriksendc:
Summary
---------

When trying to do terraform apply with a prefix of "prod" you get errors (prod-webapp is not unique enough of a name for an s3 bucket). Added hma into the static website bucket, cloudfront distribution and user pool names.

Changed the sign in form so that the create account option is hidden.

Minor tweak to the webapp readme.

Test Plan
---------

1. terraform destroy
2. terraform apply
3. npm start


BarrettOlson (approved):
LGTM

PR #450, opened 2021-03-23
[hma][ui] Settings mock

Dcallies:
Summary
---------

Mock out a settings page based on discussion offline.

Test Plan
---------

View page:
image


hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,150 @@
...
 
 export default function Settings() {
   return (
     <>
-      <h1>Settings</h1>
+      <Tabs defaultActiveKey="signals" id="setting-tabs">
+        <Tab eventKey="signals" title="Signals">
+          <h2>Datasets</h2>
+          <div className="row mt-3">
+            <div className="col-xl-4 col-lg-4 col-md-6 col-sm-6 col-xs-12 mb-4">
```
schatten, line 16: Usually, you just note the breakpoints. eg. 'col-lg-4, col-md-6', bootstrap figures the rest out on its own.

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,150 @@
...
+                </h6>
+                <div className="card-body text-left">
+                  <p className="card-text">
+                    <h5>Privacy Group</h5>
+                    <ul>
+                      <li>Cat Image Collaboration</li>
+                    </ul>
+                    <h5>Settings</h5>
+                    <ul>
+                      <li>✔️ SEEN status</li>
```
schatten, line 133: This works, but the more idiomatic way is to create a component and use it.
```
<ul>
  <ListItemWithIcon icon="check">Seen status</ListItemWithIcon>
...
</ul>
```

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,150 @@
...
 import React from 'react';
+import Tabs from 'react-bootstrap/Tabs';
+import Tab from 'react-bootstrap/Tab';
 
 export default function Settings() {
   return (
     <>
-      <h1>Settings</h1>
+      <Tabs defaultActiveKey="signals" id="setting-tabs">
+        <Tab eventKey="signals" title="Signals">
```
schatten, line 13: Best if you create another Component called `<SignalSettingsTab>` which extends from `<Tab>`. That way, the individual files do not get super long.

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,150 @@
...
+          <h2>Datasets</h2>
+          <div className="row mt-3">
+            <div className="col-xl-4 col-lg-4 col-md-6 col-sm-6 col-xs-12 mb-4">
+              <div className="card text-center">
```
schatten, line 17: I'm assuming this was just a first stab at UI. Commonly (and this helps with the final version), you'd create a few javascript types to contain a Dataset.

Then you would create a mock array of such dataset instances. 

Then you would iterate through the array and render the individual cards.

That speeds up development and is closer to what your final solution is.

schatten: Because react allows strong typing of props and stronger typing of components, this will become second nature over time. Imagine if you wanted to change the color of the heading to orange, you'd have to change it in 4 places now, but if you had rendered from a javascript list, you'd change it only in one place.

hasher-matcher-actioner/webapp/src/Settings.jsx
schatten, line 16: This is a `<CollaborationCard` component.

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,150 @@
...
+import Tabs from 'react-bootstrap/Tabs';
+import Tab from 'react-bootstrap/Tab';
 
 export default function Settings() {
   return (
     <>
-      <h1>Settings</h1>
+      <Tabs defaultActiveKey="signals" id="setting-tabs">
+        <Tab eventKey="signals" title="Signals">
+          <h2>Datasets</h2>
```
schatten, line 14: Having looked at the UI, you probably want more vertical margin on this one. Or Padding on your tabbed cards.

hasher-matcher-actioner/webapp/src/Settings.jsx
Dcallies: I wasn't able to get this to work - ListWithIcon only seems to be in libraries that we aren't including

hasher-matcher-actioner/webapp/src/Settings.jsx
Dcallies: Because each my components has a slightly different number of internal components, I couldn't figure out a way to easily do this until I figure out control flow in assembling components.

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,189 @@
...
+              <h4 className="mb-0">Verified Cat Imagery</h4>
+            </Card.Header>
+            <Card.Subtitle className="mt-0 mb-0 text-muted">
+              Shared
+            </Card.Subtitle>
+            <Card.Body className="text-left">
+              <Card.Text>
+                <h5>Labels</h5>
+                <ul>
+                  <li>cat</li>
```
eriksendc, line 55: Using <Badge>'s here instead of ul/li might look nice... https://react-bootstrap.netlify.app/components/badge/#badges

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,189 @@
...
+            <ListGroup variant="flush" className="text-left">
+              <ListGroup.Item>
+                <h5>Share To: Cat Dataset</h5>
+                <h6>Labels</h6>
+                <ul>
+                  <li>cat</li>
+                </ul>
+              </ListGroup.Item>
+            </ListGroup>
+            <a href="#todo" className="btn btn-primary">
```
eriksendc, line 73: Maybe use `<Button>` here. Can still be rendered as an anchor tag if that's necessary / desired using the as attribute: `<Button as="a">Edit</Button>`

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,189 @@
...
+              Edit
+            </a>
+          </Card>
+        </Col>
+        <div className="col-xl-4 col-lg-4 col-md-6 col-sm-6 col-xs-12 mb-4">
```
eriksendc, line 78: assuming component pattern continues here. Looks super great up above!! :)

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -3,11 +3,189 @@
...
+            </Card.Subtitle>
+            <Card.Body className="text-left">
+              <Card.Text>
+                <div>
+                  <h5 className="mb-0">Privacy Group</h5>
+                  <div>Cat Image Collaboration</div>
+                </div>
+                <div className="mt-4">
+                  <h5 className="mb-0">Settings</h5>
+                  <div>✔️ Seen status</div>
```
eriksendc, line 177: In addition to iconography, another additional way to indicate on / off or selected / not selected is bold vs not. These days bold is done with `<strong>this will be bold</strong>`.

hasher-matcher-actioner/webapp/src/Settings.jsx
Dcallies: Whoops, missed

hasher-matcher-actioner/webapp/src/Settings.jsx
Dcallies: Tried it out, and thought that it might generate confusion, think icon or bold is enough, and think icon is clearer. 
image

eriksendc (approved):

PR #451, opened 2021-03-24
Make fetcher work

jeberl:
Summary
---------

- Edit permissions to allow fetcher to read+write s3 checkpoint and PDQ file
- Edit permissions to allow fetcher to get threat exchange api key secret
- create THREAT_EXCHANGE_PDQ_DATA_KEY environment variable

Test Plan
---------

`terraform apply`
Look at fetcher logs and see no errors
Delete privacy group checkpoints and pdq files using AWS UI and see that they are regenerated

### DIFF
--- a/hasher-matcher-actioner/terraform/fetcher/main.tf
+++ b/hasher-matcher-actioner/terraform/fetcher/main.tf
@@ -83,13 +83,12 @@ resource "aws_iam_role" "fetcher" {
 data "aws_iam_policy_document" "fetcher" {
   statement {
     effect    = "Allow"
-    actions   = ["s3:GetObject"]
-    resources = ["arn:aws:s3:::${var.threat_exchange_data.bucket_name}/${var.threat_exchange_data.pdq_data_file_key}"]
-  }
-  statement {
-    effect    = "Allow"
-    actions   = ["s3:PutObject"]
-    resources = ["arn:aws:s3:::${var.threat_exchange_data.bucket_name}/${var.threat_exchange_data.pdq_data_file_key}"]
+    actions   = [
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:ListBucket",
+    ]
+    resources = ["*"]
   }
   statement {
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/terraform/fetcher/main.tf
```diff
@@ -81,13 +83,12 @@ resource "aws_iam_role" "fetcher" {
...
-  statement {
-    effect    = "Allow"
-    actions   = ["s3:PutObject"]
-    resources = ["arn:aws:s3:::${var.threat_exchange_data.bucket_name}/${var.threat_exchange_data.pdq_data_file_key}"]
+    actions   = [
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:ListBucket",
+    ]
+    resources = ["*"]
```
eriksendc, line 91: I'm not an expert, but is `resources = ["*"]` an OK practice?

schatten: True, this could be scoped more narrowly.

schatten (approved):
Approving to unblock, but would love to see the s3 permissions constrained to (atleast) the bucket.

BarrettOlson:
Now that we are connecting the fetcher e2e, similar to what @eriksendc, did with 'include_cloudfront_distribution' in `variables.tf` do you think we could add 'include_fetcher_trigger' variables? Could you also document somewhere what the fetcher does once deployed?

hasher-matcher-actioner/terraform/fetcher/main.tf
eriksendc: fwiw I searched the existing terraform scripts and only found `resources = ["*"]` in things relating to logging in cloudfront.

jeberl: I tried a few different ways but the most specific I could get without failing seems to be to limit to s3. I've added this

PR #452, opened 2021-03-24
[hma] add upload test data make command

BarrettOlson:
Summary
---------

This adds make commands to populate the pdq data file and uploads a test images for the hasher -> matcher. I also added this test photo's hash to `sample_data/pdq.csv` to a match will should be found.

Commands added:
- `dev_upload_test_signal` : uploads `sample_data/pdq.csv` to `<bucket>/threat_exchange_data/pdq.te`
- `dev_upload_test_content` : uploads `tests/data/b.jpg` to `<bucket>/images/`
- `dev_upload_test_data`: runs both commands

Test Plan
---------

I may never fully understand make...
```
$ terraform -chdir=terraform apply 
...
$ make dev_upload_test_data
Upload sample signals? [y/N] This will copy over threat_exchange_data/pdq.te in <bucket>
y
upload: sample_data/pdq.csv to s3://<bucket>/threat_exchange_data/pdq.te
Upload test photo? [y/N]  This will upload an images to images/ in s3 and will likely trigger the hasher (and matcher)
y
upload: tests/data/b.jpg to s3://<bucket>/images/b.jpg
```


jeberl:
Two questions:
1. Why do we want this as a make command and not just always present?
2. This make command uploads test data directly to AWS. Since we have the fetcher working and we have test collaborations set up, why not read the test data from TE directly. This seems like it would be a more e2e test. Instead of uploading these files to S3 we could add config files to Dynamo. You can look here for examples: https://console.aws.amazon.com/dynamodb/home?region=us-east-1#tables:selected=jeberl-ThreatExchangeConfig;tab=items

hasher-matcher-actioner/Makefile
```diff
@@ -42,6 +42,21 @@ dev_test_instance:
 	TF_PREFIX=$(call shell-or-die, cd terraform/ &&  terraform output -json prefix) \
 	./tests/scripts/check_deployed_instance.py
 
+dev_upload_test_signal:
+	@TF_BUCKET_NAME=$(call shell-or-die, set -o pipefail && cd terraform/ &&  terraform output -json bucket_name | tr -d '"') && \
+	TF_PDQ_DATA_FILE_KEY=$(call shell-or-die, set -o pipefail && cd terraform/ &&  terraform output -json pdq_data_file_key | tr -d '"') && \
+	echo "Upload sample signals? [y/N] This will copy over $${TF_PDQ_DATA_FILE_KEY} in $${TF_BUCKET_NAME}" && read ans && [ $${ans:-N} = y ] && \
+	aws s3 cp sample_data/pdq.csv s3://$${TF_BUCKET_NAME}/$${TF_PDQ_DATA_FILE_KEY}
```
jeberl, line 49: The indexer needs to be able read from multiple pdq files (eg files for 2 collaborations plus one local file). I think ti would be useful to have more descriptive file names so that indexer knows where the index came from. Currently, for pdq hashes from TE were using the following file name format:
```<bucket>/threat_exchange_data/<privacy-group-id>.pdq.te```

Perhaps for local index banks we could use 
```<bucket>/threat_exchange_data/local.<bank-name>.pdq.te```

So this test index could be..

```<bucket>/threat_exchange_data/local.test.pdq.te```

jeberl (changes requested):

BarrettOlson:
> Two questions:
> 
> 1. Why do we want this as a make command and not just always present?
> 2. This make command uploads test data directly to AWS. Since we have the fetcher working and we have test collaborations set up, why not read the test data from TE directly. This seems like it would be a more e2e test. Instead of uploading these files to S3 we could add config files to Dynamo. You can look here for examples: https://console.aws.amazon.com/dynamodb/home?region=us-east-1#tables:selected=jeberl-ThreatExchangeConfig;tab=items

I think it is worth providing a way to demo that doesn't require TE access. I also think some separation between the 'turnkey' of deploying this system compared to 'demoing' is useful instead of always having the test data present but it's a good point and something that uses the fetcher e2e will likely be a better go to in most cases.

hasher-matcher-actioner/Makefile
BarrettOlson: This change makes sense eventually but requires changes to terraform and pdq_indexer.py so I will leave it as is for now if that is okay.

jeberl (approved):
lgtm!

PR #453, opened 2021-03-24
Linters

jeberl:
Summary
---------

Test Plan
---------
from `hmalib
`black .`


hasher-matcher-actioner/mypy.ini
```diff
@@ -1,26 +1,6 @@
-[mypy]
```
BarrettOlson, line 1: This was just to suppress a warning from `python -m mypy hmalib` if we don't have it

hasher-matcher-actioner/mypy.ini
```diff
@@ -1,26 +1,6 @@
...
-
 [mypy-threatexchange.*]
```
BarrettOlson, line 1: I added this to suppress the CI issues but it feels too general for the long term which is why I kept the  `[mypy-threatexchange.hashing.*]` etc because I was hoping we'd be able to remove the root ignore it eventually...

hasher-matcher-actioner/mypy.ini
jeberl: Got it. Will add back

hasher-matcher-actioner/mypy.ini
```diff
@@ -30,5 +12,8 @@ ignore_missing_imports = True
 [mypy-moto.*]
 ignore_missing_imports = True
 
+[mypy-botocore.errorfactory]
```
BarrettOlson, line 15: nit: is this redundant to the lines below it?

BarrettOlson (approved):
Accepting to unblock, we can make other changes to `mpyi.ini` as needed

PR #456, opened 2021-03-25
Upload test configs

jeberl:
Summary
---------

When creating the ThreatExchangeConfig dynamo table in terraform, we now search for a file with configs that we should auto populate with. We do this with a new bash script that we execute from terraform whenever the table is created.

Note: terraform will only run the bash script if it is creating the table from new. To populate an existing table using the bash script you will need tot delete the table first 

Also add a config file to git that links to test configs

Test Plan
---------

Run `terraform apply` with...
- no test file variable and no config file
- test file variable and a gibberish config file
- test file variable and a valid config file


hasher-matcher-actioner/terraform/fetcher/main.tf
```diff
@@ -201,7 +201,7 @@ resource "aws_iam_role_policy_attachment" "fetcher_trigger" {
 ### Config storage ###
 
 resource "aws_dynamodb_table" "threatexchange_config" {
-  name         = "${var.prefix}-ThreatExchangeConfig"
+  name         = "${var.prefix}ThreatExchangeConfig"
```
Dcallies, line 204: Why this change?

Dcallies (commented):
Seems reasonable. @BarrettOlson - you ended up doing something to directly insert into the table after creation. Was this a route?

Dcallies (approved):
I'm fine with this, but have @schatten  or @BarrettOlson  take a quick look before landing

jeberl:
re @Dcallies - There is a aws_dynamodb_insert command but I wasn't sure how to use it to insert an unknown known number of objects so this method seemed easier

hasher-matcher-actioner/terraform/fetcher/main.tf
jeberl: `jq` was doing weird things with JSON keys that had `-` in them. COuld figure out how to escape correctly so this seemed easiest

schatten (approved):
Looks fine, but wouldn't this conflict with the other PR you have open #457?

jeberl:
@schatten Will resolve conflicts

BarrettOlson:
> Seems reasonable. @BarrettOlson - you ended up doing something to directly insert into the table after creation. Was this a route?

I used boto3 in a test script which was different as it required another command.

PR #457, opened 2021-03-25
Move Fetcher frequency variable to top level

jeberl:
Summary
---------
Meant to do this before. Move the fetcher frequency variable from just the fetcher module to the to a top level tfvars variable

Test Plan
---------

`terraform validate && terraform apply`

### DIFF
--- a/hasher-matcher-actioner/terraform/fetcher/variables.tf
+++ b/hasher-matcher-actioner/terraform/fetcher/variables.tf
@@ -50,7 +50,6 @@ variable "log_retention_in_days" {
 variable "fetch_frequency_min" {
   description = "How many minutes to wait between queries to ThreatExchange for updates"
   type        = number
-  default     = 15
 }
 
 variable "te_api_token" {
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/terraform/fetcher/variables.tf
```diff
@@ -47,10 +47,9 @@ variable "log_retention_in_days" {
   type        = number
 }
 
-variable "fetch_frequency_min" {
-  description = "How many minutes to wait between queries to ThreatExchange for updates"
-  type        = number
-  default     = 15
+variable "fetch_frequency" {
+  description = "How long to wait between calls to ThreatExcahnge. Must be an AWS Rate Expression. See here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  type        = string
```
Dcallies, line 52: I ran into problems with this a few times, thanks for changing!

Dcallies (approved):
Shipit

schatten (approved):
LGTM!

hasher-matcher-actioner/terraform/fetcher/variables.tf
jeberl: Seems like a terrible AWS design decision that you have to say `1 minute` but `5 minute**s**` lol

PR #458, opened 2021-03-26
Cognito: fix callback & sign out urls; add domain

eriksendc:
Summary
---------

This PR changes terraforming as follows:

1. If a CloudFront distribution is created, the Cognito callback and sign in urls are updated with a url formulated based on the CloudFront distribution domain name.
2. The Cognito Domain Name is now set deterministically using a new organization variable in terraform.tfvars.

Test Plan
---------

1. `terraform destroy`
2. Edit terraform.tfvars: `include_cloudfront_distribution = true`
3. `terraform apply`
4. Go to Cognito and add a user to the newly created user pool.
5. Go to CloudFront and locate the domain name
6. Open a browser using the CloudFront domain name
7. Sign in and test the web app
8. `terraform destroy`
9. Edit terraform.tfvars: `include_cloudfront_distribution = false`
10. `terraform apply`
11. Repeat 4
12. Go to S3 and locate the url of the static website on the Properties tab of the newly created bucket for the web app
13. Repeat 7


hasher-matcher-actioner/terraform/webapp/main.tf
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 resource "aws_s3_bucket" "webapp" {
-  bucket = "${var.prefix}-hma-webapp"
+  bucket = "${var.organization}-${var.prefix}-hma-webapp"
```
Dcallies, line 4: Why are both organization and prefix needed here? Wouldn't prefix be unique already? If wanting to share between prefixes, would need to remove prefix?

Dcallies (approved):
I don't claim to understand this whole thing, and am surprised it ended up coming to this, but I don't want to block on this, since we can always keep iterating in the background assuming this is reasonably stable.

The only resource I thought to check was to look for usages of cognito in public terraform recipes, to see if someone else came up with a solution (or even a complete recipe for a solution).

cc @schatten for more thoughts.

BarrettOlson:
> I don't claim to understand this whole thing, and am surprised it ended up coming to this, but I don't want to block on this, since we can always keep iterating in the background assuming this is reasonably stable.
> 
> The only resource I thought to check was to look for usages of cognito in public terraform recipes, to see if someone else came up with a solution (or even a complete recipe for a solution).

FYI: Followed up with @eriksendc offline on a possible refactor. From my experience with looking at public terraform recipes with API Gateway, I always seemed to find one that fit well but fell short in a minor but critical way for our use case. Not a bad idea to do some more digging though!

schatten (approved):
I was a bit concerned with the overhead of tearing down and setting up the cognito pool each time we made changes. 

Discussed offline with Brian
1. We don't expect partners to be using cognito from day 1
2. Developers can reduce the overhead by doing an apply without a destroy
3. If we run into dev-experience issues, we can tackle them.

Let's merge this in! Thanks @eriksendc!

hasher-matcher-actioner/terraform/webapp/main.tf
eriksendc: @Dcallies From the docs:

An Amazon S3 bucket name is globally unique, and the namespace is shared by all AWS accounts. This means that after a bucket is created, the name of that bucket cannot be used by another AWS account in any AWS Region until the bucket is deleted.

So, similar to the cognito domain name, organization is introduced so we can have fb-prod-hma-webapp and snap-prod-hma-webapp.

BarrettOlson (approved):
LGTM!

eriksendc:
@BarrettOlson @schatten @Dcallies For whatever reason I'm seeing comments in 

> I don't claim to understand this whole thing, and am surprised it ended up coming to this, but I don't want to block on this, since we can always keep iterating in the background assuming this is reasonably stable.
> 
> The only resource I thought to check was to look for usages of cognito in public terraform recipes, to see if someone else came up with a solution (or even a complete recipe for a solution).
> 
> cc @schatten for more thoughts.

@Dcallies You wrote "...am surprised it ended up coming to this, ..." Having trouble knowing what "it" refers to, as there are a couple of different topics being covered. I think you made a comment in the conversation tab, so I'm not sure about what "it" is. Assuming "it" is using the aws cli. If that's "It" then we're good... reworked things. You can take another look. If "It is something about having a cognito instance per developer (which is what Dipanjan's "it" is) then let me know.

PR #460, opened 2021-03-26
Index multi reader

jeberl:
Summary
---------

Make the PDQ indexer read from all `.pdq.te` files and merge them before building the PDQ index

Test Plan
---------

To test AWS reading writing works correctly:
1. `terraform apply`
2. Push new docker image to aws (refresh `fetcher` and `pdq_indexer` functions)
3. Delete the `threat_exchange_data/` and `index/` folders 
4. Watch them repopulate after fetcher runs

To test merging works correctly
1. Make sure `threat_exchange_data/` has files for the two test configs. One file should be a superset of the other
2. locally run the pdq_indexer (`python3 -m hmalib.lambdas.pdq.pdq_indexer`) and look (using `print`) to see that the merged index has the right number of hashes and that the privacy_groups were correctly merged


hasher-matcher-actioner/hmalib/common.py
```diff
@@ -3,11 +3,24 @@
...
 
+class PrintLogger:
+    """
+    Custom logger that just prints
+    """
+
+    def info(self, log_string):
+        print(log_string)
+
+
```
Dcallies, line 14: (blocking): What's the intent of this? The logging module should be able to do this for you with the right settings.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -102,7 +98,7 @@ def lambda_handler(event, context):
             privacy_group,
             api.app_id,
             te_data_bucket,
-            config.s3_key_prefix,
+            config.s3_te_data_folder,
```
Dcallies, line 101: YMMV: I nearly did this in the original diff, but I learned that s3 actually doesn't support folders, and any `ls` option actually lists you the entire set. The amazon UI shows you folders, but only because it has a convention on... key prefixes.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -294,14 +290,6 @@ def write_s3_text(txt_content: io.StringIO, bucket, key: str) -> None:
...
     logging.basicConfig(level=logging.INFO)
     FetcherConfig.get.cache_clear()  # Just in case
-    os.environ.setdefault(
-        "THREAT_EXCHANGE_DATA_BUCKET_NAME",
-        "jeberl-hashing-data20210324205948477200000003",
-    )
-    os.environ.setdefault("THREAT_EXCHANGE_PDQ_DATA_KEY", "threat_exchange_data/pdq.te")
-    os.environ.setdefault(
-        "THREAT_EXCHANGE_CONFIG_DYNAMODB", "jeberl-ThreatExchangeConfig"
-    )
```
Dcallies, line 304: Why this change?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -19,11 +22,14 @@
...
 
 THREAT_EXCHANGE_DATA_BUCKET_NAME = os.environ["THREAT_EXCHANGE_DATA_BUCKET_NAME"]
-THREAT_EXCHANGE_PDQ_DATA_KEY = os.environ["THREAT_EXCHANGE_PDQ_DATA_KEY"]
-
+THREAT_EXCHANGE_DATA_FOLDER = os.environ["THREAT_EXCHANGE_DATA_FOLDER"]
+THREAT_EXCHANGE_PDQ_KEY_SUFFIX = os.environ["THREAT_EXCHANGE_PDQ_KEY_SUFFIX"]
 INDEXES_BUCKET_NAME = os.environ["INDEXES_BUCKET_NAME"]
 PDQ_INDEX_KEY = os.environ["PDQ_INDEX_KEY"]
 
+HashRowType = t.Tuple[str, t.Dict[str, t.Any]]
```
Dcallies, line 30: (ignorable): I've also seen just 'T' used as a suffix/prefix

```
HashRowT
S3FileTypeT
```

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -75,33 +126,33 @@ def lambda_handler(event, context):
...
+        s3_bucket_files = s3_client.list_objects_v2(
+            Bucket=THREAT_EXCHANGE_DATA_BUCKET_NAME,
+            Prefix=THREAT_EXCHANGE_DATA_FOLDER,
+        )["Contents"]
+        logger.info("Found " + str(len(s3_bucket_files)) + " Files")
+
+        pdq_data_files = [
+            get_pdq_file(file["Key"])
+            for file in s3_bucket_files
+            if file["Key"].endswith(THREAT_EXCHANGE_PDQ_KEY_SUFFIX)
```
Dcallies, line 139: Hmm, so we're ending up in a case where we are still using S3 triggers, but now we're triggering it multiple times in a row (since there are now multiple files). Each of those triggers will use ls to compute on all of the files that it's triggering for, thus generating n^2 behavior.

This seems non-ideal in the long term. Is there anything that can be done about this?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -75,33 +126,33 @@ def lambda_handler(event, context):
...
-                    "source": "te",  # default for now to make downstream easier to generalize
-                },
-            )
-            for row in pdq_data_reader
+    with metrics.timer(metrics.names.pdq_indexer_lambda.download_datafiles):
+        s3_bucket_files = s3_client.list_objects_v2(
+            Bucket=THREAT_EXCHANGE_DATA_BUCKET_NAME,
+            Prefix=THREAT_EXCHANGE_DATA_FOLDER,
+        )["Contents"]
+        logger.info("Found " + str(len(s3_bucket_files)) + " Files")
```
Dcallies, line 134: logger.info("Found %d files", len(s3_bucket_files))

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -75,33 +126,33 @@ def lambda_handler(event, context):
...
+
+        pdq_data_files = [
+            get_pdq_file(file["Key"])
+            for file in s3_bucket_files
+            if file["Key"].endswith(THREAT_EXCHANGE_PDQ_KEY_SUFFIX)
         ]
+        logger.info("Found " + str(len(pdq_data_files)) + " PDQ Files")
```
Dcallies, line 141: see above on formatting

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -113,3 +164,27 @@ def lambda_handler(event, context):
 
     logger.info("Index update complete")
     metrics.flush()
+
+
+if __name__ == "__main__":
```
Dcallies, line 169: Comment about how to run this and that it's for testing only

hasher-matcher-actioner/requirements-dev.txt
```diff
@@ -1,4 +1,5 @@
 -r requirements.txt
 pytest==6.2.1
 mypy==0.812
-moto
\ No newline at end of file
+black
```
Dcallies, line 4: What is this change for?

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -51,7 +51,8 @@ module "pdq_signals" {
   }
   threat_exchange_data = {
     bucket_name        = module.hashing_data.threat_exchange_data_folder_info.bucket_name
-    pdq_data_file_key  = local.pdq_data_file_key
+    pdq_key_suffix     = local.pdq_key_suffix
+    data_folder        = module.hashing_data.threat_exchange_data_folder_info.key
```
Dcallies, line 55: See note about folder being a misnomer

hasher-matcher-actioner/terraform/variables.tf
```diff
@@ -52,14 +52,14 @@ variable "te_api_token" {
...
-  default = "collabs_example.json"
-}
-
 variable "fetch_frequency" {
   description = "How long to wait between calls to ThreatExcahnge. Must be an AWS Rate Expression. See here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
   type        = string
   default     = "15 minutes"
 }
+
+variable "collab_file" {
```
Dcallies, line 61: (Question that I didn't ask on the original diff): How do optional values work? It looks like its used unconditionally, where does the optional part come in?

Dcallies (changes requested):
Main blocking comment: 
* Folder isn't actually a concept in s3, and I think that adding one in can generate confusion. Wdyt?
* What's the plan with N^2 behavior on indexer runs?

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
jeberl: I personally found most of the code easier to read when I treated it as a folder. Don't feel particularly strongly but I think it makes the code more understandable everywhere except the prefix search. AWS also clearly endorses thinking of s3 as a file system so I think it's alright for us to treat it as one in code.

Could add a comment to the one location where we use the folder name as a key prefix

hasher-matcher-actioner/hmalib/common.py
jeberl: This was helpful for debugging as I ran locally. Instead of replacing all the `logger.info` instances with `print` I could just change the logger creation line from `logger = get_logger(__name__)` to `logger = get_logger("print")` during development

If no one else will use this I can remove. If your worried about the other logger functions could implement those as well

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
jeberl: @BarrettOlson and I discussed this but if feels weird to have hard coded data specific to my instance in the code. Dont necessarily think it's a security risk just seems off.

I will admit its very helpful for testing though. Wish we has some way to set these variables locally after terraform runs

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
jeberl: Agreed this could become an issue. `n`  here is number of collaborations not number of indicators correct? So with <10 collabs I don't anticipate it becoming a huge deal.

The solution I see here is to no longer rely on SNS but have a timed job (like the fetcher) that reads the update times of all the pdq files to see if the index needs to be recalculated. Update time is also available in the prefix query

I dont necessarily think its worth implementing this until it becomes an issue

hasher-matcher-actioner/requirements-dev.txt
jeberl: If we run `black` formatting on PRs it seems like we should have it as a requirement for developers

hasher-matcher-actioner/terraform/variables.tf
jeberl: I wasn't sure (and honestly didn't look into) whether terraform supported optional variables. The script first checks if the file given exists and will only populate aws if it does. Thus the optional part come in by having a default value for a file that doesn't exists.

I'm realizing now that having a default value that looks like a file is probably confusing. Perhaps I should change the description to "A file name of ThreatExchange Collaborations objects. If the file exists, the configs in it will be pre-populated into AWS. See collabs_example.json for the correct formatting"

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: We could have local environment variables that was given as an optional override. Probably not that tricky to do with a shell or python script.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
Dcallies: I agree that the concept of folders makes this easier to describe. What I'm worried about is that the concept doesn't describe the underlying system.

Here's amazon's documentation: https://docs.aws.amazon.com/AmazonS3/latest/userguide/organizing-objects.html

In the subpage on https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html, we can see that the API doesn't use folder, and instead uses prefix. In fact, no API (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) mentions folder that I can find.

If you absolutely want to follow through with this change, please do it as a separate PR and grep and replace every usage of "prefix". I'm okay with folder if we're internally consistent.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
Dcallies: It's c*s where c is the number of collabs and s is the number of signal types.

I'm fine punting, please add a TODO somewhere

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
Dcallies: Ping

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
Dcallies: .

hasher-matcher-actioner/requirements-dev.txt
Dcallies: I thought requirements-dev is only consumed by github actions. 

@schatten is it generally understood that this is the requirements to install if you are a developer?

hasher-matcher-actioner/terraform/fetcher/variables.tf
```diff
@@ -31,7 +31,7 @@ variable "threat_exchange_data" {
   description = "Configuration information for the S3 Bucket that will hold ThreatExchange Data"
   type = object({
     bucket_name       = string
-    pdq_data_file_key = string
+    data_folder       = string
```
Dcallies, line 34: Update the comment here so that it mentions that the folder is a key prefix.

hasher-matcher-actioner/terraform/main.tf
Dcallies: Does the key variable need to be renamed? These are the inconsistencies I was flagging above.

Also should pdq_key_suffix be file_extension? Folder, files or key prefix, key suffix?

Dcallies (changes requested):
Toplevel comments:
* Let me know if you need help debugging logging
* folder name okay if you chase down the remaining "key" variables and names. Would accept a pinkie swear to do it in the next PR.

hasher-matcher-actioner/hmalib/common.py
jeberl: Hmm. I think the issue is that i've been using `logger.info` which according to your link does not print to the console. I can switch those instances to `logger.warning` during my debug sessions. Will remove this

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
jeberl: So I supposed the question is wether we expect partners to interact with the AWS UI or CLI/API more. I would personally expect UI or else API through our code which I will comment on.

Code grep shows no other uses of `prefix` in `hmalib`

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
jeberl: Hmm my math is different (or were solving different problems). I calculate the number of files parsed sa `c^2 * s` currently and this could be optimized to `c * s`. So the difference is a factor of `c < 10` Maybe we're saying the same thing?

Either way, added a TODO on the `was_pdq_data_updated` function

hasher-matcher-actioner/requirements-dev.txt
jeberl: I would have though github actions needs this anyway to run checks?

hasher-matcher-actioner/terraform/main.tf
jeberl: I think `pdq_file_extention` makes sense along with `data_folder`. Will change

hasher-matcher-actioner/terraform/variables.tf
eriksendc: A pattern I used once was to have a bool to trigger processing, and optionally required variables used by the processing gated on the bool. See the variable `use_shared_user_pool` and the related variables  `webapp_and_api_shared_user_pool_id` and `webapp_and_api_shared_user_pool_client_id`.

hasher-matcher-actioner/hmalib/common.py
```diff
@@ -8,6 +8,10 @@ def get_logger(name=__name__, level=logging.INFO):
     This pattern prevents creates implicitly creating a root logger by creating the sub-logger named __name__
     Also by default sets level to INFO
     """
+
+    if name == "print":
+        return PrintLogger()
```
Dcallies, line 13: Remove

Dcallies (approved):
Chatted offline, punting the N^2 index triggers into the future.

PR #461, opened 2021-03-26
[HMA-UI] Connect Match Details Page to API

BarrettOlson:
Summary
---------

Connected some of the match details page field to the API, anything not connected should not say '(Mocked)' next to it.
This also adds the following (still rough/WIP, but see test video):
- Click to copy tooltip and functionality for some ids/hashes
- Image component that blurs until hovered over
- Endpoint to actually render image that is in the s3 bucket based on content key

I haven't worked with this kind of react/js stack much so feel free to rip this apart and request a ton of changes :)

Test Plan
---------

```

terraform -chdir=terraform apply
...
make dev_upload_test_data
...
# create a test user on aws console
...
cd webapp/ && npm run
...
```
After logging in you can go to the matches page and click details on the row of the match (not the table header) to navigate to the details page. e.g.: 

https://user-images.githubusercontent.com/7664526/112700880-876f6e00-8e65-11eb-9239-8fb333894ede.mov


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -85,6 +83,22 @@ def upload():
     }
 
 
+@app.get("/image/<key>")
+def image(key=None):
+    """
+    image endpoint
```
jeberl, line 89: Nit: comment seems a bit redundant. Could we either remove or explain what this endpoint is used for (matches page?)

hasher-matcher-actioner/webapp/src/utils/ImageUtils.jsx
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+import React, {useState} from 'react';
+import {Image} from 'react-bootstrap';
+
+export const BlurUntilHoverImage =
```
jeberl, line 8: blocking: It looks like from the video in the test plan that when loading the image theres a split second where it's not blurred (Am I seeing things?). This seems to somewhat defeat the purpose of blurring. Any idea why this is happening?

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+    return '';
+  }
+  if (input.length <= maxLength) {
+    return input;
+  }
+  return `${input.substring(0, maxLength)}...`;
+}
+
+// eslint-disable-next-line react/prop-types
+export function CopyableHashField({text, tooltip}) {
```
jeberl, line 20: This seems like a great feature!

ymmv:
```suggestion
export function CopyableHashField({text, tooltip=''}) {
```

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+      <span style={{overflow: 'hidden'}} />
+      <CopyableTextField
+        text={text}
+        tooltip={tooltip ?? 'Copy hash to clipboard?'}
+      />
+    </td>
+  );
+}
+
+const UpdatingPopover = React.forwardRef(
```
jeberl, line 32: Whats a popover? Is this in the demo video?

hasher-matcher-actioner/webapp/src/MatchDetails.jsx
```diff
@@ -69,48 +104,71 @@ export default function MatchDetails() {
...
+  const {contentKey} = props;
+  useEffect(() => {
+    fetchMatchDetails(contentKey).then(matches => {
+      setMatchDetails(matches.match_details);
+    });
+  }, []);
+
+  return (
+    <>
+      <Spinner hidden={matchDetails !== null} animation="border" role="status">
```
jeberl, line 127: ymmv: Wonder if a header here highlighting these as matches could be useful. Maybe just "Matches".

Goal would be to distinguish what is details of the content vs details of the matches. This might just be something we iterate on once we get partner feedback. Absolutely great first pass

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+      <button
+        type="button"
+        role="link"
+        style={linkButtonStyle}
+        onClick={copyText}
+        onKeyDown={e => {
+          if (e.code === 'Enter') {
+            copyText();
+          }
+        }}
```
jeberl, line 88: So when you press enter it will also copy? Dont think I've seen this pattern before

jeberl (changes requested):
Love the video a few small things in line most just for my own knowledge. 

Only blocking change is the blur

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+        tabIndex={0}>
+        {truncateString(text)}
```
eriksendc, line 90: Is there a way to use max length and overflow hidden instead of truncating the string here? There's a way to get an ellipsis to show, too (https://www.benmarshall.me/max-character-length-ellipsis-css/). I saw the max length / overflow hidden pattern in use above, but not here and wondered about that.

eriksendc (approved):

hasher-matcher-actioner/webapp/src/utils/ImageUtils.jsx
BarrettOlson: It has to do with the page layout, when you click 'details' your mouse is over the image when it loads... looked at a bunch of different solutions to avoid this but most are unintuitive unfortunately.

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
BarrettOlson: why blank string?

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
BarrettOlson: popover is what react-boostrap uses for dynamic content tool tips

hasher-matcher-actioner/webapp/src/MatchDetails.jsx
BarrettOlson: yeah, we're having a lot of discussions around this page, I agree with you and none of the headings are set in stone right now. Leaving it for now.

hasher-matcher-actioner/webapp/src/Api.jsx
```diff
@@ -35,6 +36,18 @@ export function fetchMatches() {
...
+export function fetchMatchDetails(key) {
+  return apiGet(`/match/${key}`);
+}
+
+export function fetchHash(key) {
+  return apiGet(`/hash/${key}`);
+}
+
+export function fetchImage(key) {
+  return apiGet(`/image/${key}`, {}, 'blob');
```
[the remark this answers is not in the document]
BarrettOlson: you mean like linking to: https://developer.mozilla.org/en-US/docs/Web/API/Blob

For reference, I did some searching and found the 'blob' is how you can use a byte stream as a value for `src` or `img`

hasher-matcher-actioner/webapp/src/utils/TextFieldsUtils.jsx
BarrettOlson: Thanks, I was just trying to get rid of the scroll box. Figured out how to make it work with css

hasher-matcher-actioner/webapp/src/utils/ImageUtils.jsx
BarrettOlson: Leaving this for now, fixes like 'delaying mouse events' or having another state variable made for a less than ideal UX imo, possible solutions to consider in future
- reorganization of the page to prevent the button from the last page from falling right on the image
- Adding button to blur and unblur (will likely end up adding anyway)

PR #462, opened 2021-03-29
Terraform Shared Cognito User Pool

eriksendc:
Summary
---------

Added terraforming of a shared Cognito user pool and the ability to use it in the main terraform flow. Also updated some documentation.

Test Plan
---------

1. In a terminal go to `/terraform/authentication-standalone`
2. Set up terraform.tfvars (set organization to fb)
3. Run `terraform apply`
4. Use the outputs (`webapp_and_api_shared_user_pool_id` and `webapp_and_api_shared_user_pool_client_id`) in `/terraform/terraform.tfvars` and set `use_shared_user_pool` to `true` (see `/terraform/terraform.tfvars.example`).
5. Run `terraform apply` from `/terraform`
6. Create a user in the shared user pool
7. Go to the webapp s3 bucket and open the static website url from the Properties tab
8. Test signing in to the webapp
9. Regression testing (test with `use_shared_user_pool` = `false`, both with and without a cloudfront distribution)


hasher-matcher-actioner/terraform/authentication-shared/main.tf
```diff
@@ -0,0 +1,80 @@
...
+      version = "~> 3.0"
+    }
+  }
+}
+
+provider "aws" {
+  region = "us-east-1"
+}
+
+resource "aws_cognito_user_pool" "webapp_and_api_shared_user_pool" {
```
BarrettOlson, line 16: do you know a way we can reduce the copy paste from not shared?

hasher-matcher-actioner/terraform/authentication-shared/main.tf
```diff
@@ -0,0 +1,80 @@
...
+
+resource "aws_cognito_user_pool_client" "webapp_and_api_shared_user_pool_client" {
+  name                                 = "shared-hma-user-pool-client"
+  user_pool_id                         = aws_cognito_user_pool.webapp_and_api_shared_user_pool.id
+  generate_secret                      = false
+  allowed_oauth_flows_user_pool_client = true
+  explicit_auth_flows                  = ["ALLOW_USER_SRP_AUTH", "ALLOW_REFRESH_TOKEN_AUTH"]
+  allowed_oauth_scopes                 = ["openid"]
+  allowed_oauth_flows                  = ["code"]
+  callback_urls                        = ["https://localhost:3000"]
```
BarrettOlson, line 68: I don't know about in browser usage but I haven't been able to auth through postman without this being a non-local field. Is there a way we can update this?

hasher-matcher-actioner/terraform/terraform.tfvars.example
```diff
@@ -7,3 +7,6 @@ log_retention_in_days = # DAYS to retain logs
 include_cloudfront_distribution = true
 te_api_token = "<threat-exchange-api-token>"
 collab_file = "collabs_example.json"
+use_shared_user_pool = false # if true, run authentication-shared terraform, and use its outputs for webapp_and_api_shared_user_pool_id and webapp_and_api_shared_user_pool_client_id
+webapp_and_api_shared_user_pool_id = "us-east-1_T11ZBU6gM"
```
BarrettOlson, line 11: did you mean to hardcode these are they at all sensitive?

hasher-matcher-actioner/terraform/authentication-shared/main.tf
eriksendc: No, sorry. I figured this comment would come up - and it's a good one. I tried to think of something that would help with that. The issue is that I wanted to keep the names of the user pool resources different so terraform state wouldn't get messed up in some way. I searched terraform docs for the concept of re-use at the resource level and couldn't find anything (like using a var in the resource name label).

hasher-matcher-actioner/terraform/terraform.tfvars.example
eriksendc: I think I thought about it and thought why not leave them. They're randomly generated and not in use by us... but others haven't been left in that way so I'll change.

hasher-matcher-actioner/terraform/authentication-shared/main.tf
BarrettOlson: can we have name labels uniqueness from the module level? e.g. we could have a wrapper module here that way it would be `shared_pool.webapp_and_api_shared_user_pool` vs `authentication.webapp_and_api_shared_user_pool`

hasher-matcher-actioner/terraform/authentication-shared/main.tf
BarrettOlson: not blocking: but can you add a comment that this value (and therefore the pool itself) is just for testing.

hasher-matcher-actioner/terraform/terraform.tfvars.example
BarrettOlson: Yeah, I think it probably makes sense to remove because of like you said, not known what is in use vs an example, if you wanted to add a comment on that line with an "e.g. ..." that could work.

BarrettOlson (approved):
LGTM, added two nits in-line about documenting and code reuse but just adding a quick comment is more than fine for now imo.

hasher-matcher-actioner/terraform/authentication-shared/main.tf
eriksendc: Done

PR #463, opened 2021-03-30
 [WIP][actions] Rudimentary lambda to 'act' on match messages

schatten:
Summary
---
Add lambda handler to receive messages from SQS in a specific serialization format.
Add terraform module to deploy an actioning lambda.

Test Summary
---
Long time ago, dunno whether still true.

Had deployed using `terraform apply` and verified that messages received could be printed.


hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -0,0 +1,25 @@
...
+    The output SNS topics from matchers have SQS instances subscribed to them.
+    This indirection allows for fanout. Multiple SQS queues can be tagged to the
+    same SNS topic and SNS will guarantee delivery to the queues.
+
+    Architecture aside, here's what you need to know. The actions layer will be
+    listening on an SQS queue. For every record,
+    - it will construct the MatchMessage object and identify the actions it
+      needs to invoke.
+    - for now, rather than fanning out to individual specific lambdas, it will
+      call all specific functions serially!
```
Dcallies, line 22: I can't tell if this is boilerplate/copypaste from an example or specific to actioner.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -27,6 +29,21 @@
 DYNAMODB_TABLE = os.environ["DYNAMODB_TABLE"]
 
 
+
+def get_default_labels() -> t.List[Label]:
+    """
+    As a stop gap measure, the matcher will add default labels that instruct the
+    actioner. In the long-term, we expect labels to be the message-passing infra
+    between matched records and post-match phases.
+    """
```
Dcallies, line 37: I like this approach and clear instructions on future work

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -88,11 +112,19 @@ def lambda_handler(event, context):
...
+                    banked_indicator_id=signal_id
+                )
+
+                # Publish one message per match. This might be a perf penalty if
+                # we get > 1 match per index query. But I'm assuming that's
+                # rare. Unfortunately, there is no batch publish in SNS.
+                sns_client.publish(
+                    TopicArn=OUTPUT_TOPIC_ARN,
+                    Message=message.to_sns_message()
+                )
```
Dcallies: (Question): Which component will handle combining actions from multiple sources and taking only the highest priority one? I.e. if one action is delete the content, and another is send the content for review, the delete should win

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -281,3 +283,56 @@ def from_time_range(
...
+    - `bank_source`: [optional][Defaults to 'api/threatexchange'] This is
+      forward looking, but potentially, we could have this be 'local', or
+      'api/some-other-api'
+    """
+    content_key: str
+    content_hash: str
+    banked_indicator_id: str
+
+    # source information, for now, it's okay to be hardcoded
+    # to threatexchange
```
Dcallies, line 318: 👍

hasher-matcher-actioner/requirements.txt
```diff
@@ -1 +1 @@
-.
\ No newline at end of file
+.
```
Dcallies, line 1: What change happened here?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -88,11 +112,19 @@ def lambda_handler(event, context):
...
                 match_ids.append(signal_id)
-            sns_client.publish(
-                TopicArn=OUTPUT_TOPIC_ARN,
-                Subject="Match found in pdq_matcher lambda",
-                Message=f"Match found for key: {key}, hash: {hash_str}, for IDs: {match_ids}",
-            )
+                message = MatchMessage(
+                    content_key=key,
+                    content_hash=hash_str,
+                    banked_indicator_id=signal_id
```
Dcallies: Would expect the default labels to be added here. You can also add a TODO to add labels from the ThreatExchange data, if that happens here (it should happen wherever the bank metadata is being looked up).

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -281,3 +283,56 @@ def from_time_range(
...
+    @staticmethod
+    def from_dynamodb_string(string: str) -> "Label":
+        return Label(*string.split(":"))
+
+
+@dataclass
+class MatchMessage(SNSMessage):
+    """
+    Captures a match that will need to be processed. Joins the dataset and the
+    pipeline together.
```
Dcallies, line 296: Are dataset and pipeline defined somewhere? Worried they are super generic terms (especially pipeline)

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -281,3 +283,56 @@ def from_time_range(
...
+
+    Pipeline fields:
+    - `content_key`: A way for partners to refer uniquely to content on their
+      site
+    - `content_hash`: The pipeline generated hash for the content_key
```
Dcallies, line 301: Wait, if we are sending this to the pipeline how did it get a pipeline hash?

Dcallies (commented):
Are you adding a new stage between matcher and actioner that is adding labels? Or is that happening at the actioner stage?

If that's happening at the actioner stage, why emit one event per match vs emitting one event with a list of matches?

hasher-matcher-actioner/requirements.txt
schatten: Probably a line ending thing. Can be ignored, I suppose.

hasher-matcher-actioner/hmalib/models.py
schatten: @BarrettOlson Just to be clear, these models pre-date your work on signal_source. I'd be happy to revisit the name for these variables.

hasher-matcher-actioner/hmalib/models.py
schatten: We're not sending this to the pipeline. One phase (matcher) of the pipeline generates messages for the subsequent phase (actions). But they all need some 'pipeline-unique-ids' to make sense of the information. These fields provide those.

hasher-matcher-actioner/hmalib/models.py
schatten: No they are not. I haven't updated this since we discussed. Completely open to changing the names of the variables as we get clearer on the terminology.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
schatten: I wanted it to be specific to the 'action_performer'. Can you tell me what feels vague?

Dcallies: It's hard to put my finger on, but I think a description of what the module does (the actioner converts matches into results from actions) rather than how it does it (sns queues with subscriptions) might help.

hasher-matcher-actioner/hmalib/models.py
Dcallies: Is "matched hash"/"matched_signal" a synonym for "pipeline hash"?

What's the difference between content id/key and pipeline id?

Dcallies (changes requested):
Just putting back to your queue.

Toplevel questions:
1. I'm still confused about the multiplexing with the queues. Can I catch you offline to discuss?
2. How do you want to resolve the naming thing/term definition stuff?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
schatten: In my head, that is part of the decisions component. Which is TBH, not making it to V0.

When decisions component exists, the actions layer will become a labeler. It will recommend certain actions based on the match data and metadata. A decisions component will receive all labels and decide which action is 'applicable'.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -44,12 +59,19 @@ def get_index(bucket_name, key):
...
     As per the default configuration
     - the index data bucket is INDEXES_BUCKET_NAME
     - the key name must be PDQ_INDEX_KEY
+
+    When matched, publishes a notification to an SNS endpoint. Note this is in
+    contrast with hasher and indexer. They publish to SQS directly. Publishing
+    to SQS implies there can be only one consumer.
+
+    Because, here, in the matcher, we publish to SNS, we can plug multiple
+    queues behind it and profit!
```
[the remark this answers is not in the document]
schatten: An SNS topic can have any number of subscriptions. Right now, the only subscription is an SQS queue, but partners could (using the console) configure other subscriptions (email, firehose, http).

You wouldn't route to a specific queue, more like double process. eg. Match happened, one queue plugs in to actioning lambdas, the other queue plugs in to record-keeping / enrichment lambdas. or drops it to a review queue.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -281,3 +283,60 @@ def from_time_range(
             & Key("UpdatedAt").between(start_time, end_time),
             ProjectionExpression=cls.DEFAULT_PROJ_EXP,
         ).get("Items", [])
+
+    @staticmethod
+    def from_dynamodb_string(string: str) -> "Label":
+        return Label(*string.split(":"))
```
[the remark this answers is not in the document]
schatten: I think I messed up on resolving a merge conflict. The labels class is missing and with it all the validation I was doing. Will fix.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -78,6 +100,7 @@ def lambda_handler(event, context):
                 )
```
jeberl, line 100: I'm realizing now the this version of the matcher wont work with them multi-indexer in PR #460. That version merges multiple lists of (hash, metadata) tuples into a unique set of hashes with merged metatdata. Metadata has a list of privacy groups indicating which collaboration it came from. 

I'll put up a fix. I think it could be either here or in pdq_indexer and I'm not sure which will be more efficient. Do you feel strongly?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -78,6 +100,7 @@ def lambda_handler(event, context):
...
                 signal_id = metadata["id"]
 
+                # TODO: Add source (threatexchange) tags to match record
```
jeberl, line 103: Is this still a TODO? Looks like source is included in the match record on line 109

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -38,6 +39,33 @@ def remove_content_key_prefix(key: str) -> str:
...
+
+@dataclass
+class Label:
+    key: str
+    value: str
+
+    def __post_init__(self):
+        """
+        Validate that the key name does not have colons.
+        """
```
jeberl, line 59: Is this a TODO or is the function meant to be blank? If so, i've often seen `pass` used in python

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -38,6 +39,33 @@ def remove_content_key_prefix(key: str) -> str:
...
+    def to_sns_message(self) -> str:
+        raise NotImplementedError
+
+    @classmethod
+    def from_sns_message(cls, message: str) -> "SNSMessage":
+        raise NotImplementedError
+
+
+@dataclass
+class Label:
```
jeberl, line 52: TODO make name more descriptive? What's this used for?

jeberl (commented):
Some comments, still reviewing the rest of the file. Not sure if the plan is to merge as is, address the comments, or commander the branch?

hasher-matcher-actioner/hmalib/models.py
schatten: That's an oopsie! Wonder how it got through mypy. Will fix.

hasher-matcher-actioner/hmalib/models.py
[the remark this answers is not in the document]
schatten: Have fixed this. Label keys and values can be anything now. No restrictions.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -281,3 +309,75 @@ def from_time_range(
...
+    - `banked_content_id`: Inside the bank, what's a unique way to refer to what
+      was matched against?
+    - `bank_id`: [optional][Defaults to 'threatexchange_all_collabs'] Which bank
+      did we fetch this banked_content from?
+    - `bank_source`: [optional][Defaults to 'api/threatexchange'] This is
+      forward looking, but potentially, we could have this be 'local', or
+      'api/some-other-api'
+    """
+
+    banked_indicator_id: str
```
BarrettOlson, line 363: Don't feel strongly about it but indicator is pretty source specific. I don't love the sound of`banked_signal_id` but if we wanted to match the signal_source pattern in PDQMatchRecord that's what I'd go with.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import json
+
+from hmalib.common import get_logger
+from hmalib.models import MatchMessage
+
+logger = get_logger(__name__)
```
BarrettOlson, line 8: doesn't look like we use this yet? maybe add an event received log for now?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
Dcallies: Hmm, in my mind these were compatible, but I have to read more to figure out what the break is. See the comment about the division of labor above.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
Dcallies: Looks like that's the threat exchange ID and not the tags. One more lookup is needed to get the tags.

hasher-matcher-actioner/hmalib/models.py
Dcallies: We can do what mms does and just call everything under the sun cluster! (please don't do this).

We will eventually go content-centric even on banking, and so submit `banked_content_id`, which will parallel `content_id` on the pipeline.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
Dcallies: I think this will probably get used in future, so fine leaving it in, and impl below is definitely not filled in.

Dcallies (approved):
Approving based mostly on offline discussion. Please sync with @jeberl  for potential conflicts.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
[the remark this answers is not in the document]
schatten: > @schatten - I still think we potentially need this topic, because the SEEN writeback might be determined at the matcher level (though it could also work from the decision making level, which just then becomes an implicit rule of (level for bank with match) -> Action for seen.

In an offline discussion, @jeberl @eriksendc and I came to the same conclusion. The SEEN writeback would be the first case of branching out from the match SNS notification. As Jesse has hinted, 

```
--> matcher λ --> SNS Topic --> actioner queue --> actioner λ
                          \--> logging queue --> seen writeback λ
```

hasher-matcher-actioner/hmalib/models.py
schatten: @jeberl Label are a first class entity and I expect them to be used throughout the 'pipelines' to attach more information to a match. Attaching labels decouples subsequent components from any data models specific to preceeding components. Any qualifier more descriptive here would mean we limit what Labels are used for.

I am strongly for retaining the vanilla 'Label' name. Let's discuss if you have concerns with this.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -38,6 +39,28 @@ def remove_content_key_prefix(key: str) -> str:
...
+
+    @classmethod
+    def from_sns_message(cls, message: str) -> "SNSMessage":
+        raise NotImplementedError
+
+
+@dataclass
+class Label:
+    key: str
+    value: str
```
Dcallies, line 54: What are example keys and values?

schatten: eg. 
- `CallWebhook:6734023402834`
- `ApplyAction:WEBHOOK-898`
- `BankTag:cat-pictures`, `BankTag:cat-on-beach-pictures`

jeberl:
@eriksendc and I discussed that this will be best to merge and build off of so doing that now

PR #464, opened 2021-03-31
[py-tx] fix hash from file bug and add test

BarrettOlson:
Summary
---------

Something between the testing on #389 and now seems to have broken `threatexchange match photo <filepath>` and `threatexchange hash <filepath>` this PR makes the fix and creates a test to hopefully better detect this in the future. 

Test Plan
---------

added `python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py`

Before
```
$ threatexchange hash photo .../tests/data/b.jpg                  
  File "/Users/barrett/.venv/hma/bin/threatexchange", line 33, in <module>
...
    all_signal_types = dataset.load_cache(
  File "/Users/barrett/dev/github/ThreatExchange/python-threatexchange/threatexchange/dataset.py", line 110, in load_cache
    signal_type.load(signal_state_file)
TypeError: load() missing 1 required positional argument: 'path'

$ threatexchange match photo .../tests/data/b.jpg 
  File "/Users/barrett/.venv/hma/bin/threatexchange", line 33, in <module>
...
  File "/Users/barrett/dev/github/ThreatExchange/python-threatexchange/threatexchange/signal_type/signal_base.py", line 179, in hash_from_file
    raise NotImplementedError
NotImplementedError
```
After
```
$ threatexchange hash photo tests/data/b.jpg 2>/dev/null 
photo_md5 d35c785545392755e7e4164457657269                                 
pdq f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22          
pdq_ocr f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22,     

$ threatexchange match photo tests/data/b.jpg 2>/dev/null
<ID> pdq disputed media_priority_samples     
```


python-threatexchange/threatexchange/signal_type/md5.py
```diff
@@ -29,7 +29,7 @@ class VideoMD5Signal(signal_base.SimpleSignalType, signal_base.FileHasher):
     TYPE_TAG = "media_type_video"
 
     @classmethod
-    def hash_file(cls, file: pathlib.Path) -> str:
+    def hash_from_file(cls, file: pathlib.Path) -> str:
```
Dcallies, line 32: Why this change?

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -93,16 +93,16 @@ def _parse_input(
     def execute(self, api: ThreatExchangeAPI, dataset: Dataset) -> None:
 
         all_signal_types = dataset.load_cache(
-            s
+            s()
```
Dcallies, line 96: This shouldn't be required if the target function is a classmethod. Why this change?

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -93,16 +93,16 @@ def _parse_input(
...
             for s in self.content_type.get_signal_types()
             if self.signal_type in (None, s.get_name())
         )
 
-        file_hashers = [s for s in all_signal_types if issubclass(s, FileHasher)]
-        str_hashers = [s for s in all_signal_types if issubclass(s, StrHasher)]
+        file_hashers = [s for s in all_signal_types if issubclass(type(s), FileHasher)]
+        str_hashers = [s for s in all_signal_types if issubclass(type(s), StrHasher)]
```
Dcallies, line 102: if sticking with this, it should be changed to isinstance()

Dcallies (changes requested):
blocking: I'd like to understand this change more. I'm think that the fix is to swap an inheritence order in one of the classes rather than this change, but want to try and convert "weird inheritance structure" into a more specific reason, of which the answer is probably something related to https://www.python.org/download/releases/2.3/mro/

Since this is my bug, and I'm the one blocking you, I can try and figure this out, and if I can't we'll just go with this.

python-threatexchange/threatexchange/signal_type/md5.py
BarrettOlson: in the current master `hash_file` is only referenced `hash_cmd.py`everywhere else looks like it expects `hash_from_file` see https://github.com/facebook/ThreatExchange/blob/5acf1bf5fdb47873c83f71bb785b14f3583805c2/python-threatexchange/threatexchange/signal_type/signal_base.py#L177

python-threatexchange/threatexchange/cli/hash_cmd.py
BarrettOlson: SignalType::load called in dataset is not a classmethod https://github.com/facebook/ThreatExchange/blob/5acf1bf5fdb47873c83f71bb785b14f3583805c2/python-threatexchange/threatexchange/signal_type/signal_base.py#L114

eriksendc (approved):

python-threatexchange/threatexchange/signal_type/md5.py
Dcallies: Whoops, that's definitely a refactor fail by me then.

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -93,16 +93,16 @@ def _parse_input(
     def execute(self, api: ThreatExchangeAPI, dataset: Dataset) -> None:
 
         all_signal_types = dataset.load_cache(
```
Dcallies, line 95: Discussed offline the real problem is this line right here - dataset.load_cache() should not happen, because we should only be using stateless class methods.

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
```diff
@@ -0,0 +1,30 @@
...
+from threatexchange.content_type import meta
+from threatexchange.signal_type.signal_base import FileHasher, StrHasher
+from pathlib import Path
+
+
+class TestHashFrom(unittest.TestCase):
+    """
+    Signal Type has a weird inheritance structure the error around which are easy to
+    miss before run time. If a SignalType is a FileHasher/StrHasher it should be able to
+    take the corresponding content medium without throwing.
```
Dcallies, line 15: Update this comment - it's not an inheritance issue, just a pair of bugs:
* Refactor missed a method rename
* copy-paste fail from match to hash cmd accidentally attempting to call load(), which is not needed for the command

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+
+    def test_file_hashers_have_impl(self):
+        signal_types = [s() for s in meta.get_all_signal_types()]
+        file_hashers = [s for s in signal_types if issubclass(type(s), FileHasher)]
```
Dcallies, line 20:
```suggestion
        file_hashers = [s for s in signal_types if isinstance(s, FileHasher)]
```

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
```diff
@@ -0,0 +1,30 @@
...
+        for file_hasher in file_hashers:
+            file_hasher.hash_from_file(
+                Path(os.path.dirname(__file__) + "/../../../tests/data/b.jpg")
```
Dcallies, line 23: I think this might have issues if run from other directories, or if we move this code later, but I don't know how to fix it

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
```diff
@@ -0,0 +1,30 @@
...
+
+import unittest
+import os
+
+from threatexchange.content_type import meta
+from threatexchange.signal_type.signal_base import FileHasher, StrHasher
+from pathlib import Path
+
+
+class TestHashFrom(unittest.TestCase):
```
Dcallies, line 11: Nit, ignorable: SignalTypeHashTest or TestHash or something similar, TestHashFrom (and having the reader put together that this is a prefix of methods on SignalType might be tough.

Dcallies (changes requested):
Discussed offline!

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
BarrettOlson: that is why I take the dirname of the file, it should only break if either the test or the image file move

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
```diff
@@ -0,0 +1,30 @@
...
+        signal_types = [s() for s in meta.get_all_signal_types()]
+        file_hashers = [s for s in signal_types if issubclass(type(s), FileHasher)]
+        for file_hasher in file_hashers:
+            file_hasher.hash_from_file(
+                Path(os.path.dirname(__file__) + "/../../../tests/data/b.jpg")
+            )
+
+    def test_str_hashers_have_impl(self):
+        signal_types = [s() for s in meta.get_all_signal_types()]
+        str_hashers = [s for s in signal_types if issubclass(type(s), StrHasher)]
```
Dcallies, line 28: And this guy

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
Dcallies: If you only just need a file to exist (since you aren't testing a specific file), check out the tmpfile: https://docs.python.org/3/library/tempfile.html

Dcallies (approved):
Thanks!

Issue #455, opened 2021-03-25
pdq_hasher - Buffer has wrong number of dimensions (expected 3, got 2)


tgthorley:
If images only have two dimensions (e.g. QR codes) pdqhash fails because it is expecting an array with 3 dimension

```
/opt/conda/lib/python3.8/site-packages/threatexchange/hashing/pdq_hasher.py in pdq_from_file(path)
     22     img_pil = Image.open(path)
     23     image = numpy.asarray(img_pil)
---> 24     return _pdq_from_numpy_array(image)
     25 
     26 

/opt/conda/lib/python3.8/site-packages/threatexchange/hashing/pdq_hasher.py in _pdq_from_numpy_array(array)
     34 
     35 def _pdq_from_numpy_array(array: numpy.ndarray) -> PDQOutput:
---> 36     hash_vector, quality = pdqhash.compute(array)
     37 
     38     bin_str = "".join([str(x) for x in hash_vector])

/opt/conda/lib/python3.8/site-packages/pdqhash/bindings.pyx in pdqhash.bindings.compute()

ValueError: Buffer has wrong number of dimensions (expected 3, got 2)
```

Example image is attached: 

![A Brief Introduction to Neural Networks (neuronalenetze-en-zeta2-2col-dkrieselcom) pdf_897](https://user-images.githubusercontent.com/3315649/112519685-fcad4700-8d70-11eb-8527-aa38bf18e86d.png)

tgthorley:
Update - our of a corpus of 63950 images (extracted from pdfs) I found 544 images that caused this error all of them are black and white (though there is a variety of scanned text, QR codes and hand-drawn diagrams etc.)

BarrettOlson:
Thanks so much for flagging this! Based on my reading of our c++ pdq code there is some nuance on how exactly to handle pdq for grayscale/black and white photos. I'm not convinced #465 is a complete fix but it should address the error and it does seem to generate the correct hashes based on my initial testing. 

I will follow up once it and a dependent PR are reviewed and merged. Thanks again!

PR #465, opened 2021-03-31
[py-tx] update pdq_hasher to handle grayscale

BarrettOlson:
Summary
---------

#455 called out a bug in 'python-threatexchange/threatexchange/hashing/pdq_hasher.py' Did some digging and found what I think should be a one line fix. 

Using the photo attached to the issue I compared the output with the changes to `threatexchange.hashing.pdq_hasher.pdq_from_file`with our cpp version of pdq (`./pdq/cpp/pdq-photo-hasher` after running make) and found the hashes matched.

Test Plan
---------

See methodology above in summary, unfortunately testing is hard until the fixes in #464 are merged.

### DIFF
--- a/python-threatexchange/threatexchange/hashing/pdq_hasher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_hasher.py
@@ -19,7 +19,9 @@ def pdq_from_file(path: pathlib.Path) -> PDQOutput:
     Given a path to a file return the PDQ Hash string in hex.
     Current tested against: jpg
     """
-    img_pil = Image.open(path)
+    # Convert to RGB makes sure the resulting array has the number of dimensions
+    # expected by 'pdqhash' (without this black and white images result in errors).
+    img_pil = Image.open(path).convert("RGB")
     image = numpy.asarray(img_pil)
     return _pdq_from_numpy_array(image)
 


Dcallies (approved):
Add "Fixes #455 " in the summary to automatically close the issue when you land!

python-threatexchange/threatexchange/hashing/pdq_hasher.py
```diff
@@ -19,7 +19,9 @@ def pdq_from_file(path: pathlib.Path) -> PDQOutput:
     Given a path to a file return the PDQ Hash string in hex.
     Current tested against: jpg
     """
-    img_pil = Image.open(path)
+    # Convert to RGB makes sure the resulting array has the number of dimensions
+    # expected by 'pdqhash' (without this black and white images result in errors).
+    img_pil = Image.open(path).convert("RGB")
```
eriksendc, line 24: I'm wondering if there's a way to do this conversion only when needed? Dunno if the conversion could ever have a side effect (either now or in the future). Looking at https://pillow.readthedocs.io/en/stable/_modules/PIL/Image.html?highlight=grayscale# there seems to be a way to detect the current state of affairs... if you get L then you'd do this conversion? What do you think?

eriksendc (approved):

python-threatexchange/threatexchange/hashing/pdq_hasher.py
BarrettOlson: yeah, @eriksendc I went back and forth on that, you can even manually detect the dimensions of the array and add the 3rd one manually. My first solution was adding the following helper:
```
def _check_dimension_and_expand_if_needed(array: np.ndarray) -> np.ndarray:
    if array.ndim == 2:
        array = np.concatenate([array[..., np.newaxis]] * 3, axis=2)
    return array
``` 

The reason I went this route is 1) the output was identical for all the images I tested 2) covert("RGB") will use the library to help us be more error tolerant as the python pdq lib we currently use expects 3 dimensions and I would rather have the images be converted to expected by default.  This way we should handle most cases not just the narrow one of the 'L' check. I'd prefer to avoid similar failures of just not working for some images. Happy to change it back if you feel strongly though :)

eriksendc: @BarrettOlson The thing I wonder about is the potential side effect of "converting" something that doesn't really need to be converted in the first place. I understand you tested a particular corpus of images... but without that being a super giant number, what would we really know? Also, the code for this conversion might not have a side effect now, but it could some day. I would recommend a hybrid by the way, detect if array.ndim is 2 and then use `convert()` to do the conversion. I'd say I feel "a little past 50-50" strong about it, but totally good with going with a tie-breaker. @schatten what do you think?

BarrettOlson: @schatten brought up perf and is completely right, this is a bad idea, putting up a change soon :)

PR #469, opened 2021-04-01
[py-tx] pdq hash revert automatic convert and check dim of image array for better perf

BarrettOlson:
Summary
---------

#465 worked but upon review had a significant performance hit. The core issue in #455 was only that images with 2 dimensions were resulting in an error. Instead of convert RGB this does the array manipulation in the case only where it is needed (a possible future optimization would be to also down scale the image).

Test Plan
---------

Again tested by using the photo attached to the issue and compared the output with the changes to threatexchange.hashing.pdq_hasher.pdq_from_filewith our cpp version of pdq (./pdq/cpp/pdq-photo-hasher after running make) and found the hashes matched.

### DIFF
--- a/python-threatexchange/threatexchange/hashing/pdq_hasher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_hasher.py
@@ -49,7 +49,10 @@ def _pdq_from_numpy_array(array: np.ndarray) -> PDQOutput:
 
 
 def _check_dimension_and_expand_if_needed(array: np.ndarray) -> np.ndarray:
-    # 'pdqhash' expected 3 dim so (without this black and white images result in errors).
+    """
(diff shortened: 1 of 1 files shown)


eriksendc (approved):

python-threatexchange/threatexchange/hashing/pdq_hasher.py
```diff
@@ -46,3 +46,10 @@ def _pdq_from_numpy_array(array: numpy.ndarray) -> PDQOutput:
     hex_str = hex_str.lower()
 
     return hex_str, quality
+
+
+def _check_dimension_and_expand_if_needed(array: np.ndarray) -> np.ndarray:
+    # 'pdqhash' expected 3 dim so (without this black and white images result in errors).
```
Dcallies, line 52: Good candidate for a docstring instead of a comment

PR #470, opened 2021-04-06
[HMA] Move mock to API level

BarrettOlson:
Summary
---------

Connects more of the UI mocks to the python lambda API. The changes aren't easily apparent on the UI apart from where it nows says 'mocked' on the Dashboard and Signals Page. Most changes are likely to be refactored, consolidated, or extended once hooked up to real data.

Test Plan
---------

Tested new endpoints with UI and Postman. 
Screen Shot 2021-04-08 at 11 53 04 AM
Screen Shot 2021-04-08 at 11 52 58 AM

PR #472, opened 2021-04-07
Refactor pdq indexer

jeberl:
Summary
---------

The way pdq indexer reads all PDQ files in S3 is functionality that can be used elsewhere. Specifically its needed to provide a list of hash counts. This PR does some refactoring and creates a first pass of how the new functionality can be used to count hashes

Test Plan
---------

ensure function still runs and compiles correctly: `python3 -m hmalib.lambdas.pdq.pdq_indexer`

ensure it works correctly in s3:
1.`make upload_docker`
2. delete pdq files and pdq index from s3 using UI
3. Wait for fetcher to rebuild pdq files from TE
4. Ensure pdq_indexer correctly builds index


hasher-matcher-actioner/hmalib/common.py
```diff
@@ -12,3 +23,132 @@ def get_logger(name=__name__, level=logging.INFO):
...
     logger.setLevel(level)
     return logger
+
+
+logger = get_logger(__name__)
+s3_client = boto3.client("s3")
+
+THREAT_EXCHANGE_DATA_BUCKET_NAME = os.environ["THREAT_EXCHANGE_DATA_BUCKET_NAME"]
+THREAT_EXCHANGE_DATA_FOLDER = os.environ["THREAT_EXCHANGE_DATA_FOLDER"]
+THREAT_EXCHANGE_PDQ_FILE_EXTENSION = os.environ["THREAT_EXCHANGE_PDQ_FILE_EXTENSION"]
```
Dcallies, line 33: Nit: Put all the module level attributes at the to level (except logger).

Maybe blocking: Overall, this all feels like it belongs in another module called s3. Maybe common should be a folder and not a file? common/logging.py ?

Nit: I'm partial to using a wrapping class for this (https://github.com/Dcallies/ThreatExchange/blob/a5df357bd9d86243b562984d9030915a30f2980b/hasher-matcher-actioner/hmalib/lambdas/fetcher.py#L13), because it makes it easier to override the functionality for testing - relying entirely on environment variables can be very error prone.

hasher-matcher-actioner/hmalib/common.py
```diff
@@ -12,3 +23,132 @@ def get_logger(name=__name__, level=logging.INFO):
...
+
+
+@dataclass
+class ThreatExchangeS3Adapter:
```
Dcallies, line 37: blocking: Yea, this is actually pretty beefy, should be it's own file. 

This also pre-supposes that the file format is csv. May want to mention that in the title.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -133,6 +133,16 @@ def hashes(key=None):
     return results if results else {}
 
 
+@app.get("/hash_count")
+def hash_count():
+    """
+    how many hashes exist in HMA
+    """
+    results = hash_count()
+    logger.debug(results)
```
Dcallies, line 142: Good use of debug

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -133,6 +133,16 @@ def hashes(key=None):
     return results if results else {}
 
 
+@app.get("/hash_count")
+def hash_count():
+    """
+    how many hashes exist in HMA
+    """
+    results = hash_count()
```
Dcallies, line 141: blocking: Doesn't this infinite loop?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -216,3 +226,12 @@ def gen_hash(content_id: str) -> t.Optional[HashResult]:
         "content_hash": record.content_hash,
         "updated_at": record.updated_at.isoformat(),
     }
+
+
+def gen_hash_count() -> t.Dict[str, int]:
```
Dcallies, line 231: nit: gen_ usually indicates that this is an asyncronous our yielding function, suggest renaming

Dcallies (changes requested):
I love, love, love this PR. Yea using this approach to count hashes is not the most optimized, yea we might change it some day to be more specialized, but it gets the job done, and the TODOs are easy to spot.

Blocking: Move this to a new file. Consider making a common directory rather than a common.py

hasher-matcher-actioner/hmalib/common.py
BarrettOlson: yeah, this is partially my mistake, I recommended putting this in common for now without seeing the full scope of the class. CSV has certain advantages (e.g. s3 has built in parsing) agree that calling it out as a requirement is good and might be something to call out at the python-threatexchange level as well.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: following a bad pattern of mine, removed the use of gen in #470

jeberl:
I like the idea of making a common module. I think it should be a separate PR through so that all the current comments are preserved

hasher-matcher-actioner/hmalib/common.py
jeberl: I like the class idea. Will add. 

Also agree about moving but think that should happen in another PR to keep comments

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
jeberl: yup. Should be get_hash_count

hasher-matcher-actioner/hmalib/common/logging.py
```diff
@@ -1,5 +1,3 @@
-# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
BarrettOlson, line 1: add back please :)

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
Dcallies: How did you test the PR?

hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -0,0 +1,161 @@
...
+import os
+import boto3
+import csv
+import codecs
+
+import typing as t
+
+from dataclasses import dataclass
+from hmalib import metrics
+from hmalib.common.logging import get_logger
```
Dcallies, line 12: (ignorable/ymmv/there are actually reasons not to do this too): You can also do from .logging import get_logger

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -13,7 +13,7 @@
 
 from hmalib import metrics
 from hmalib.models import PipelinePDQHashRecord
-from hmalib.common import get_logger
+from hmalib.common.logging import get_logger
```
Dcallies, line 16: Make sure to do a grep or something similar to make sure you get all the occurences.

hasher-matcher-actioner/hmalib/metrics/__init__.py
```diff
@@ -22,6 +22,19 @@
 logger = logging.getLogger(__name__)
 
 
+class lambda_with_datafiles:
```
Dcallies, line 25: Nit: Class names are always UpperCamelCase

https://www.python.org/dev/peps/pep-0008/

LambdaWithDatafiles

All of the implementations just have a constant. What about this?

class LambdaWithDatafiles
  PREFIX = ""

  @property
    def download_datafiles(self):
        return f"{self.PREFIX}.download_datafiles"
  
class PDQIndexerLambda(LambdaWithDatafiles):
   PREFIX = "lambdas.pdqhasher"

Alternatively, just use dataclass and prefix: str

or alternatively this:
```
@dataclass
class LambdaName:
  prefix: str
  
  @property
  ...

pdq_indexer_lambda = LambdaName("prefix1")
whatever_lambda = LambdaName("prefix2")
```

hasher-matcher-actioner/hmalib/metrics/__init__.py
```diff
@@ -45,11 +58,12 @@ class pdq_hasher_lambda:
         download_file = f"{_prefix}.download_file"
         hash = f"{_prefix}.hash"
 
-    class pdq_indexer_lambda:
+    class pdq_indexer_lambda(lambda_with_datafiles):
```
Dcallies, line 61: .

hasher-matcher-actioner/hmalib/metrics/__init__.py
```diff
@@ -45,11 +58,12 @@ class pdq_hasher_lambda:
...
         _prefix = "lambdas.pdqindexer"
```
Dcallies, line 62: Use ALL_UPPER for class constants

hasher-matcher-actioner/hmalib/metrics/__init__.py
```diff
@@ -22,6 +22,19 @@
...
+    @property
+    def download_datafiles(self):
+        return f"{self.prefix_impl()}.download_datafiles"
+
+    @property
+    def parse_datafiles(self):
+        return f"{self.prefix_impl()}.parse_datafiles"
+
+
 class names:
```
Dcallies, line 38: .

hasher-matcher-actioner/hmalib/metrics/__init__.py
```diff
@@ -61,6 +75,12 @@ class pdq_matcher_lambda:
         parse_index = f"{_prefix}.parse_index"
         search_index = f"{_prefix}.search_index"
 
+    class api_hash_count(lambda_with_datafiles):
```
Dcallies, line 78: .

Dcallies (changes requested):
Will sync with you offline

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
jeberl: mypy was catching imports correctly

Dcallies (approved):
Discussed offline, will cleanup the file in a future PR with requested changes

jeberl:
TODO: fix davids comments about python titling

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -133,6 +134,16 @@ def hashes(key=None):
     return results if results else {}
 
 
+@app.get("/hash_count")
```
eriksendc, line 137: Sorry, late to the party, @jeberl, but the route should be `/hash-count` if left this way, but probably a better route would be `/hash/count` since `/hash` is the main "concept". @BarrettOlson since you might be the consumer of this at some point, would you fix at that time?

eriksendc: @jeberl and @BarrettOlson I know I missed making this comment when the PR was open. Let me know if making this comment here at this time is confusing / not a best practice. If there's another way to communicate this better, let me know.

hasher-matcher-actioner/hmalib/metrics/__init__.py
eriksendc: @jeberl the code looks like it stiil has `lambda_with_datafiles` instead of `LambdaWithDatafiles`. I know this PR is already merged at this point. What's the right process to take from here?

PR #473, opened 2021-04-08
[hma] update test commands and script

BarrettOlson:
Summary
---------

#460 unintentionally broke our test commands in the Makefile.
This makes the necessary changes to provide test signals and run the smoke test script again.

Test Plan
---------
Ran the following commands and confirmed output + checked the cloudwatch logs of the indexer.
`make dev_test_instance`
`make dev_upload_test_signal`

### DIFF
--- a/hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
+++ b/hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
@@ -15,16 +15,20 @@ import sys
 s3_client = boto3.client("s3")
 dynamodb = boto3.resource("dynamodb")
 
+PREFIX = os.environ["TF_PREFIX"]
+BUCKET_NAME = os.environ["TF_BUCKET_NAME"]
+DYNAMODB_TABLE = os.environ["TF_DYNAMODB_TABLE"]
+
 # Copy of what is configured in tf files (not worth building a parser for them for single point of mataince imo)
-THREAT_EXCHANGE_PDQ_DATA_KEY = "threat_exchange_data/pdq.te"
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
```diff
@@ -15,16 +15,20 @@
...
 PDQ_INDEX_KEY = "index/pdq_hashes.index"
 
 # Test uses hasher-matcher-actioner/tests/data/b.jpg
-TEST_PHOTO_KEY = "images/test_photo.jpg"
+TEST_PHOTO_KEY = f"""{os.environ["TF_IMAGE_FOLDER_KEY"]}test_photo_from_script.jpg"""
 TEST_PHOTO_EXPECTED_HASH = (
     "f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22"
 )
-TEST_PHOTO_EXPECTED_ID = "5555555555555555"
+TEST_PHOTO_EXPECTED_ID = "4109214869142910"
```
schatten, line 31: This looks like a real indicator. Verified in the id tool. Does this make the test flaky?

schatten (approved):
LGTM. Approving to unblock. Have a question about the flakiness of the test because a 'specific' id is getting used.

hasher-matcher-actioner/tests/scripts/check_deployed_instance.py
BarrettOlson: not really, this is to cover the possible overlap of existing values in the index before you run this test. There are no external dependencies because the test itself updates the signal source [here](https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/tests/scripts/check_deployed_instance.py#L35). 

The most common error this is avoiding is running:
`make dev_upload_test_signal` (which has the correct id) and then
`make dev_test_instance` (which until this change was using the fake id and would error)

(This could also be solved by additional test images but for now I think it is okay)

PR #477, opened 2021-04-09
Make primary nav vertical and move it to its own file

schatten:
Summary
---
Changes the primary navigation from a top nav bar to a sidebar based nav. This allows larger primary nav lisss and allows you to use tabs as secondary nav to implement individual screens. Was already being done in the settings screen.

**Other Niceties:**
1. Current route is highlighted in the sidebar.
2. First usage of a css file. Hoping this will become more of a pattern
3. First usage of prop-types. Hoping this too will become a pattern.
4. Adjusted padding and spacing to make things a little more pleasant (subjective)

Test Plan
---
Screenshot, before and after. 

hma-vertical-navbar

### DIFF
--- a/hasher-matcher-actioner/webapp/public/index.html
+++ b/hasher-matcher-actioner/webapp/public/index.html
@@ -11,9 +11,22 @@
       .switch-wrapper > div {
         position: absolute;
       }
+
+      div#root {
+        display: flex;
+        flex-wrap: nowrap;
+        height: 100vh;
+        height: -webkit-fill-available;
+
+        /* In case something goes wrong and horizontal scrolling is necessary. */
+        overflow-x: auto;
+
+        /* The sidebar and the main area will have their own vertical scrolls */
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/webapp/public/index.html
```diff
@@ -11,9 +11,22 @@
       .switch-wrapper > div {
         position: absolute;
       }
+
```
eriksendc, line 14: Having this `<style>` element here inline in our main HTML entry point is a turd left over from my rushing through a first hack at things. Can you create a main .css file, reference it here and move the css out to there?

eriksendc (changes requested):

hasher-matcher-actioner/webapp/public/index.html
schatten: Sure!

eriksendc (approved):
LURV IT

PR #478, opened 2021-04-09
python-threatexchange supports fetching list of privacy groups an app is a member of

schatten:
Summary
---
Add a method to ThreatExchangeAPI

**Addons:**
- An integration test for the API was missing, so added it. Only executed when appropriate env variables are provided.
- A utility method to create threatexchange graph API urls was missing, so added it. A future PR will convert all handwritten URL mangling to calls to this method.

TestPlan
---
```
$ THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN=$(cat ~/.txtoken) pytest -v
...
==== 58 passed, 3 warnings in 1.01s ====
....
```

Lint
```
$ black .
$ mypy threatexchange # produced no new errors :(
```

### DIFF
--- a/python-threatexchange/tests/test_api.py
+++ b/python-threatexchange/tests/test_api.py
@@ -19,7 +19,7 @@ class APIIntegrationTest(unittest.TestCase):
 
     def test_get_threat_privacy_groups_member(self):
         """
-        Assumes that the app (if token is provided) will have atleast one
+        Assumes that the app (if token is provided) will have at least one
         privacy group.
         """
         response = self.api.get_threat_privacy_groups_member()
--- a/python-threatexchange/tests/test_api.py
+++ b/python-threatexchange/tests/test_api.py
@@ -1,0 +1,32 @@
+import os
+import unittest
+import collections.abc
+
+from threatexchange.api import ThreatExchangeAPI, ThreatPrivacyGroup
+
+THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN = os.getenv(
+    "THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN"
+)
+
+
+@unittest.skipUnless(
+    THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN,
+    "Integration Test requires tokens. Use THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN environment variable.",
+)
+class APIIntegrationTest(unittest.TestCase):
+    def setUp(self):
+        self.api = ThreatExchangeAPI(THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN)
+
+    def test_get_threat_privacy_groups_member(self):
+        """
+        Assumes that the app (if token is provided) will have atleast one
+        privacy group.
+        """
+        response = self.api.get_threat_privacy_groups_member()
+        self.assertTrue(
+            isinstance(response, collections.abc.Sequence)
+            and not isinstance(response, staticmethod),
+            "API returned something that's not a list!",
+        )
+
+        self.assertTrue(isinstance(response[0], ThreatPrivacyGroup))
--- a/python-threatexchange/threatexchange/api.py
+++ b/python-threatexchange/threatexchange/api.py
@@ -18,6 +18,8 @@ import requests
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
+from .api_models import ThreatPrivacyGroup
+
 
 class TimeoutHTTPAdapter(HTTPAdapter):
     """
@@ -284,6 +286,41 @@ class ThreatExchangeAPI:
 
         url = f"{self._base_url}/{privacy_group}/threat_updates/"
         return _CursoredResponse(self, url, params, decode_fn=decode_fn)
+
+    def get_threat_privacy_groups_member(
+        self,
+    ) -> t.List[ThreatPrivacyGroup]:
+        """
+        Returns a non-paginated list of all privacy groups the current app is a
+        member of.
+        """
+        url = self._get_graph_api_url(f"{self.app_id}/threat_privacy_groups_member")
+        response = self.get_json_from_url(url)
+        return [ThreatPrivacyGroup.from_graph_api_dict(d) for d in response["data"]]
+
+    def _get_graph_api_url(
+        self, sub_path: t.Optional[str], query_dict: t.Dict = {}
+    ) -> str:
+        """
+        Returns a threatexchange URL for a sub-path and a dictionary of query
+        parameters. Automatically adds access_token to the query dictionary.
+        """
+        if "access_token" not in query_dict:
+            query_dict["access_token"] = self.api_token
+
+        query = urllib.parse.urlencode(query_dict)
+
+        base_url_parts = urllib.parse.urlparse(self._base_url)
+        url_parts = urllib.parse.ParseResult(
+            base_url_parts.scheme,
+            base_url_parts.netloc,
+            f"{base_url_parts.path}/{sub_path}",
+            base_url_parts.params,
+            query,
+            base_url_parts.fragment,
+        )
+
+        return urllib.parse.urlunparse(url_parts)
 
     def _validate_post_params_for_submit(self, postParams):
         """
--- a/python-threatexchange/threatexchange/api.py
+++ b/python-threatexchange/threatexchange/api.py
@@ -286,7 +286,7 @@ class ThreatExchangeAPI:
 
         url = f"{self._base_url}/{privacy_group}/threat_updates/"
         return _CursoredResponse(self, url, params, decode_fn=decode_fn)
-
+        
     def get_threat_privacy_groups_member(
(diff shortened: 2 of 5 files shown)


eriksendc (approved):
Looks good to me but I'm looking forward to learning more when others comment.

python-threatexchange/tests/test_api.py
```diff
@@ -0,0 +1,32 @@
...
+THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN = os.getenv(
+    "THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN"
+)
+
+
+@unittest.skipUnless(
+    THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN,
+    "Integration Test requires tokens. Use THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN environment variable.",
+)
+class APIIntegrationTest(unittest.TestCase):
```
Dcallies, line 16: Excellent use of skip!

python-threatexchange/threatexchange/api.py
```diff
@@ -285,6 +287,41 @@ def get_threat_updates(
...
+        return [ThreatPrivacyGroup.from_graph_api_dict(d) for d in response["data"]]
+
+    def _get_graph_api_url(
+        self, sub_path: t.Optional[str], query_dict: t.Dict = {}
+    ) -> str:
+        """
+        Returns a threatexchange URL for a sub-path and a dictionary of query
+        parameters. Automatically adds access_token to the query dictionary.
+        """
+        if "access_token" not in query_dict:
```
Dcallies, line 308: (ignorable): Hmm, I'm super apprehensive about a method that breaks the convention that an API instance only belongs to a single app (single token).

I'd rather unconditionally rewrite it every time

python-threatexchange/threatexchange/api_models.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+from datetime import datetime
+from dateutil.parser import parse
+from dataclasses import dataclass
+
+"""
+Models (dataclasses only) for interfacing with the threatexchange API.
+"""
```
Dcallies, line 9: Nit: Docstring goes before modules https://www.python.org/dev/peps/pep-0008/#imports

python-threatexchange/threatexchange/api_models.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+from datetime import datetime
+from dateutil.parser import parse
+from dataclasses import dataclass
+
+"""
+Models (dataclasses only) for interfacing with the threatexchange API.
```
Dcallies, line 8: (ignroable): I have never seen a specific section of libraries split out and called "models" before, and generally find the convention confusing. If you had a models.py class in a directory and asked me what I expected to be in there, I would guess simulation or ML before I would guess MVC.

This is probably due to me never explicitly building something with strict MVC in mind, and it's unclear to me still what following the conventions get me.

python-threatexchange/threatexchange/api_models.py
```diff
@@ -0,0 +1,38 @@
...
+def _parse_datetime_from_iso_8601(datestr: str) -> datetime:
+    """
+    Parses strings representing date like 2019-05-20T16:44:47+0000 from the
+    graph api into datetime objects.
+    """
+    return parse(datestr)
+
+
+@dataclass
+class ThreatPrivacyGroup:
```
Dcallies, line 21: Check out the pytx library (the original pytx, which I stole the name of for pytx3) for more ideas: https://github.com/facebook/ThreatExchange/blob/master/api-reference-examples/python/pytx/pytx/threat_privacy_group.py

python-threatexchange/tests/test_api.py
```diff
@@ -0,0 +1,32 @@
+import os
```
Dcallies, line 1: yaytests

python-threatexchange/threatexchange/api_models.py
```diff
@@ -0,0 +1,38 @@
...
+    id: str
+    name: str
+    description: str
+    members_can_see: bool
+    members_can_use: bool
+    last_updated: datetime
+
+    @classmethod
+    def from_graph_api_dict(cls, d: dict) -> "ThreatPrivacyGroup":
+        return cls(
```
Dcallies, line 31: Be super super careful here. This isn't actually getting you any protection than using the dict directly (dict is erasing you all your types). This function isn't too too different than cls(**d).

Part of the reason why I didn't go the pytx/class route when I started on pytx3 is the original classes I wrote were all just giving me class method access to the same untyped json under the hood, and I didn't know enough about open source to figure out how to do typechecking.

Dcallies (approved):
non-blocking, just some food for thought

python-threatexchange/tests/test_api.py
Dcallies, line 1: yaytests!

python-threatexchange/threatexchange/api.py
schatten: I'm not seeing it, @Dcallies.

You'd rather I skip the 'if' and always overwrite the "access_token" value?

python-threatexchange/threatexchange/api_models.py
schatten: These are DTO objects for interfacing with the Graph APIs. You are correct, they are not models. I'll figure out a better name. I've used entities.py, representations.py etc before. Perhaps shapes.py?

eriksendc: Besides Hack, do other languages use something known specifically as a `shape`? I think we should stick to a name that's a bit more generic. It seems very natural to me to have something I might call a domain model object like `ThreatPrivacyGroup` in a module with models in the name. Whether it's an MVC model or a domain model, these seem to be models to me. Either that or entities. I get it that there's a very ML-specific thing called a model, too, but given all our other code (and the lack of anything ML-ish) I'm not thinking there's really any confusion here?

python-threatexchange/threatexchange/api.py
```diff
@@ -285,6 +287,41 @@ def get_threat_updates(
         url = f"{self._base_url}/{privacy_group}/threat_updates/"
         return _CursoredResponse(self, url, params, decode_fn=decode_fn)
 
+    def get_threat_privacy_groups_member(
```
jeberl, line 290: For HMA we will need to further filter this list to privacy groups that have collaboration functionality enabled (can use `\threat_updates` endpoint). Do you envision this functionality going here? separate function? 

Could be in a later pr

jeberl (approved):

jeberl (changes requested):
see blocking comment about quering owners endpoint as well

python-threatexchange/threatexchange/api.py
schatten: How would one go about determining whether a PG can use `/threat_updates`? @jeberl

python-threatexchange/threatexchange/api.py
```diff
@@ -285,6 +287,41 @@ def get_threat_updates(
...
+        self,
+    ) -> t.List[ThreatPrivacyGroup]:
+        """
+        Returns a non-paginated list of all privacy groups the current app is a
+        member of.
+        """
+        url = self._get_graph_api_url(f"{self.app_id}/threat_privacy_groups_member")
+        response = self.get_json_from_url(url)
+        return [ThreatPrivacyGroup.from_graph_api_dict(d) for d in response["data"]]
```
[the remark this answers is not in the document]
schatten: Discussed offline with: @mengyangwang.

For HMA, we need member and owner lists, but python-threatexchange should not do the join. Instead, we'll expose both owners and members lists as separate threatexchange APIs and join at the HMA layer.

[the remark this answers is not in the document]
schatten: Will be done as a subsequent PR.

python-threatexchange/threatexchange/api.py
schatten: This is a private method, so only internal users 'should' call it. I'll remove the `if` statement.

schatten:
@jeberl Your callout about the owners endpoint is valid and we'll address that in a subsequent PR. Per my understanding, that's what @mengyangwang is going to work on next.

python-threatexchange/threatexchange/api.py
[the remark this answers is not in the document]
jeberl: I'm not convinced these should be separate ever. In what situation would we want a list of members that may or may not include the owner?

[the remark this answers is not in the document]
schatten: python-threatexchange should have a 1:1 mapping with the ThreatExchange graph API.  These are separate APIs on the Graph API so I propose we keep them separate here. 

If the meaning or requirement were to change, we should change it in python-threatexchange AND in ThreatExchange.

Does that make sense?

PR #479, opened 2021-04-11
Provide a devcontainer for HMA

schatten:
Summary
---------
Think of a devcontainer as a devserver that runs locally on your docker. This has (at least) the following benefits:
1. Gets everyone to use the same standard tooling
2. No need to manage any tooling on your local macbook/linuxbox/what have you
3. Standardizes vscode extensions so lint/type errors are a thing of the past

Test Plan
---------
1. Install [Visual Studio Code](https://code.visualstudio.com/). VS Code @ FB will not work. :(
2. Install [Remote Container](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension
3. In `devcontainer.json`, replace all occurrences of `dipanjanm` with your unixname. TODO (automate)
3. Have docker desktop running
3. With visual studio open, try `[F1]` or `[Cmd+Shift+P]` and select "Remote-Containers: Open Folder in Container...", open the hasher-matcher-actioner directory

Note: The first build may take a while. Couple of minutes. Subsequent bootups are quite fast. Bash history is not yet persisted, but we can figure that out.

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/Dockerfile
+++ b/hasher-matcher-actioner/.devcontainer/Dockerfile
@@ -1,0 +1,104 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.163.1/containers/javascript-node/.devcontainer/base.Dockerfile
+
+# [Install python]
+# TODO: pin version later. Otherwise starting up will be too slow.
+FROM python:3.8-alpine
+
+# [Install ZSH]
+RUN apk add --update zsh git
+
+# [Install node] Node.js version: 15 only. Stolen from: https://github.com/nodejs/docker-node/blob/main/15/alpine3.10/Dockerfile
+ENV NODE_VERSION 15.14.0
+
+RUN addgroup -g 1000 node \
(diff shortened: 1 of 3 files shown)


BarrettOlson (commented):
I really love this and the reasoning behind it. However in is current state I think it has the potential to introduce the thrash in folks development flows we are trying to avoid by standardizing. 

For us to merge a version to have everyone use I think we
need:
- a clear and concise readme about this (and updates to CONTRIBUTING.md) 
- consistent use of yarn vs npm

would like to have:
- shell support other than bash (and persistent history + maybe a way to pass along dotfiles) 
- docs on how to add to this container and how to test updates and additions (and then have others update locally)

stretch: 
- a way to have the rest of the TE repo in the workspace for jumping between hma and py-te 
- the ability to use a terminal outside of vscode window (e.g ~ssh)
- tmux support

BarrettOlson (changes requested):
RC for visibility, the only thing really 'blocking' is that this creates a yarn.lock file which we should either avoid or acknowledge in the docs. All the other things listed above are stuff we should talk about before adopting this but not in the way of merge. Thanks!

PR #480, opened 2021-04-12
[hma] get populate_config_db to work on windows

### LINKED ISSUE #475: [HMA] Cannot apply with windows due to shell quote issue with populate_config_db.sh
```
$ ./terraform/populate_config_db.sh
jq: error: syntax error, unexpected ':' (Windows cmd shell quoting issues?) at <top-level>, line 1:
{  : . }
jq: error: May need parentheses around object key expression at <top-level>, line 1:
{  : . }
jq: 2 compile errors
```

This script attempts to be run unconditionally by the terraform file

Dcallies:
Summary
---------

Turns out shell quoting in windows (basically required for jq) is difficult. Instead rewrite script in python which is more platform agnostic. 

Fixes  #475

Test Plan
---------

use terraform taint to force recreation of the fetcher configs.

Then apply, see that the new table has the two sample rows.

### DIFF
--- a/hasher-matcher-actioner/scripts/populate_config_db
+++ b/hasher-matcher-actioner/scripts/populate_config_db
@@ -1,0 +1,40 @@
+#! /usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Write simple configs to the HMA config database
+"""
+
+import argparse
+import json
+import tempfile
+import subprocess
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "in_file",
+        type=argparse.FileType(),
+        help="file to read items from",
+    )
+    parser.add_argument(
+        "table_name",
+        help="Name of the config table to add to",
+    )
+
+    args = parser.parse_args()
+    rows = json.load(args.in_file)
+    with tempfile.NamedTemporaryFile("w", delete=False) as tmp:
+        # Rewrite so the table name is the key of dicts with these as the results
+        json.dump({args.table_name: rows}, tmp)
+    # Windows environments require file not be held
+    subprocess.check_call(
+        [
+            "aws",
+            "dynamodb",
+            "batch-write-item",
+            "--request-items",
+            f"file://{tmp.name}",
+        ]
+    )
--- a/hasher-matcher-actioner/scripts/storm_images_queue
+++ b/hasher-matcher-actioner/scripts/storm_images_queue
@@ -108,19 +108,16 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "queue_url",
-        type=str,
         help="SQS Queue URL. Must include account id. eg. https://queue.amazonaws.com/<acid>/<queue_name>",
     )
 
     parser.add_argument(
         "s3_bucket",
-        type=str,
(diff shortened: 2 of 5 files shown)


hasher-matcher-actioner/scripts/populate_config_db
```diff
@@ -0,0 +1,40 @@
+#! /usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Write simple configs to the HMA config database
```
jeberl, line 5: "Sample configs"?

hasher-matcher-actioner/scripts/storm_images_queue
```diff
@@ -108,19 +108,16 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "queue_url",
-        type=str,
```
jeberl, line 111: What are these changes for?

hasher-matcher-actioner/terraform/fetcher/main.tf
```diff
@@ -205,7 +205,7 @@ resource "aws_iam_role_policy_attachment" "fetcher_trigger" {
 ### Config storage ###
 
 resource "aws_dynamodb_table" "threatexchange_config" {
-  name         = "${var.prefix}ThreatExchangeConfig"
```
jeberl, line 208: This is tested and works? I had though json keys cant include `-`

jeberl (approved):
Looks great. Agree python is the best way moving forward. Thanks for this refactor

hasher-matcher-actioner/terraform/.terraform.lock.hcl
```diff
@@ -43,6 +43,7 @@ provider "registry.terraform.io/hashicorp/local" {
 provider "registry.terraform.io/hashicorp/null" {
   version = "3.1.0"
   hashes = [
+    "h1:SFT7X3zY18CLWjoH2GfQyapxsRv6GDKsy9cF1aRwncc=",
```
jeberl, line 46: Also not sure what this change is?

hasher-matcher-actioner/scripts/storm_images_queue
Dcallies: Was in the area and they are nitty changes (str is the default type if unspecified). If you want to be ultra strict I can move them to another commit

hasher-matcher-actioner/terraform/fetcher/main.tf
Dcallies: ? Not heard that, "-" appears to be valid: https://stackoverflow.com/questions/8676011/which-characters-are-valid-invalid-in-a-json-key-name

. I did indeed test it end to end, and records are in by db
image

hasher-matcher-actioner/terraform/.terraform.lock.hcl
Dcallies: This is an auto-generated file from terraform. I have no idea why it's edited. @schatten might know

hasher-matcher-actioner/scripts/populate_config_db
Dcallies: Simple is intended, but perhaps not clear. Decided not to fix up the description more because I suspect that will be making another pass on this assuming the config proposal I had this morning works.

PR #481, opened 2021-04-12
Update DynamoDB Config table and fetcher Lambda to allow configs to b…

mengyangwang:
Summary
---------
1. added active config in collabs_example.json, default to N, which means should not be updated
2. modified fetcher.py to update collabarations if the active config is N

Test Plan
---------
tested locally
python3 -m hmalib.lambdas.fetcher
black hmalib
mypy hmalib

tested in aws:
go to dynamodb, delete table wangmengyangThreatExchangeConfig
terraform apply
make docker
make upload_docker
go to dynamodb  table wangmengyangThreatExchangeConfig, saw 'active' is 'N'


hasher-matcher-actioner/terraform/collabs_example.json
```diff
@@ -9,6 +9,9 @@
             },
             "tags" : {
                 "L" : []
+            },
+            "active" : {
+                "S": "N"
```
jeberl, line 14: blocking: Can we make this type an AWS boolean value?https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.DataTypes.html

hasher-matcher-actioner/terraform/collabs_example.json
```diff
@@ -9,6 +9,9 @@
             },
             "tags" : {
                 "L" : []
+            },
+            "active" : {
```
jeberl, line 13: Can we specify what `active` means? Maybe `fetcher_active` or `keep_in_sync`

jeberl (changes requested):

hasher-matcher-actioner/terraform/collabs_example.json
mengyangwang: sure will do

hasher-matcher-actioner/terraform/collabs_example.json
mengyangwang: yeah I think fetcher_active makes sense

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -66,14 +66,17 @@ def lambda_handler(event, context):
...
+            ("#Name", "privacy_group", "tags", "fetcher_active")
+        ),
         ExpressionAttributeNames={"#Name": "Name"},
     )
 
     collabs = []
     for page in response_iterator:
         for item in page["Items"]:
-            collabs.append((item["Name"], item["privacy_group"]))
+            if item["fetcher_active"] == True:
```
jeberl, line 78: nit: `if item["fetcher_active"] :`

jeberl (approved):
Looks good! Ship it

PR #482, opened 2021-04-12
[hma] fix missing lambda env values and refactor to assign in the handler

BarrettOlson:
Summary
---------

#472 unintentionally broke the api_root lambda as the use of ThreatExchangeS3PDQAdapter depended on environment variables not passed to the api lambda in terraform. 

This adds the environment vars and refactors s3_adapters.py to require the config be created by pulling from the environment var in the same file as the lambda handler.

Test Plan
---------

before all endpoints were returning a 500 error. With these fixes the UI page loads values again and the \hash_count endpoint actually returns a result to Postman (which required a small fix in metrics/__init__.py).


schatten (approved):
Looks good. Thanks!

eriksendc (approved):

jeberl:
Looks good! Thanks for finding this

PR #483, opened 2021-04-12
Initial skeleton of action evaluation.

eriksendc:
Summary
---------

This is the basic skeleton of action evaluation / orchestration. The entry point is `lambda_handler` that receives a match message (sent by the matcher).

The processing begins with evaluating what action rules apply to the match message (each action rule implies an action via its action labels). Actions are deduplicated along the way. From there, `perform_action()` is called to perform each action.

After that the applicable ThreatExchange reactions are generated, and then `react_to_threat_exchange()` is called for each reaction.

The next PR will be implementing labeling a match message.

Test Plan
---------

`terraform destroy`
`terraform apply`
upload an image
review logs
also, python -m hmalib.lambdas.actions.action_performer

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
@@ -3,6 +3,7 @@
 import json
 import typing as t
 
+from dataclasses import dataclass, field
 from hmalib.common.logging import get_logger
 from hmalib.models import MatchMessage, Label
 
@@ -19,30 +20,35 @@ def lambda_handler(event, context):
     """
     for sqs_record in event[
         "Records"
-    ]:  # confused about this... are we popping off more than one thing at a time from the "matches" queue?
+    ]:  # TODO research max # sqs records / lambda_handler invocation
         sns_notification = json.loads(sqs_record["body"])
         match_message: MatchMessage = MatchMessage.from_sns_message(
             sns_notification["Message"]
         )
         action_labels = get_action_labels(match_message)
         for action_label in action_labels:
+            # TODO create a new action execution queue and enqueue the
+            # match message and action label (or, possibly, add the
+            # action label to the match message and enqueue the match
+            # message by itself)
             perform_action(match_message, action_label)
 
-        # the more I write code here about reacting to ThreatExchange, the more I wish it was somewhere else entirely. Implementing
-        # here forever ties up action-related code with ThreatExchange-specific code / functionality. I prefer putting
-        # ThreatExchange-related specifics for reacting into another lambda that also subscribes to the matcher's notifications.
-        if threat_exchange_reacting_is_enabled():
+        if threat_exchange_reacting_is_enabled(match_message):
             threat_exchange_reaction_labels = get_threat_exchange_reaction_labels(
-                action_labels
+                match_message, action_labels
             )
-            if threat_exchange_reaction_labels is t.List[Label]:
+            if threat_exchange_reaction_labels:
                 for threat_exchange_reaction_label in threat_exchange_reaction_labels:
+                    # TODO create a new ThreatExchange reaction queue and enqueue
+                    # the match message and threat exchange reaction label (or, possibly,
+                    # add the threat exchange reaction label to the match message
+                    # and enqueue the match message by itself)
                     react_to_threat_exchange(
                         match_message, threat_exchange_reaction_label
                     )
 
 
-def get_action_labels(match_message: MatchMessage) -> t.List[ActionLabel]:
+def get_action_labels(match_message: MatchMessage) -> t.List["ActionLabel"]:
     """
     TODO finish implementation
     Returns an ActionLabel for each ActionRule that applies to a MatchMessage.
@@ -58,11 +64,11 @@ def get_action_labels(match_message: MatchMessage) -> t.List[ActionLabel]:
     return action_labels
 
 
-def get_action_rules() -> t.List[ActionRule]:
+def get_action_rules() -> t.List["ActionRule"]:
     """
     TODO implement
-    Returns a collection of ActionRule objects. An ActionRule will have the following attributes:
-    MustHaveLabels, MustNotHaveLabels, ActionLabel
+    Returns the ActionRule objects stored in the config repository. Each ActionRule
+    will have the following attributes: MustHaveLabels, MustNotHaveLabels, ActionLabel
     """
     return [
         ActionRule(
@@ -73,24 +79,24 @@ def get_action_rules() -> t.List[ActionRule]:
     ]
 
 
-def get_actions() -> t.List[Action]:
+def get_actions() -> t.List["Action"]:
     """
     TODO implement
-    Returns a collection of Action objects. An Action will have the following attributes:
-    ActionLabel, Priority, SupersededByActionLabel
+    Returns the Action objects stored in the config repository. Each Action will have
+    the following attributes: ActionLabel, Priority, SupersededByActionLabel
     """
     return [
         Action(
-            ActionLabel("Action", "EnqueueForReview"),
+            ActionLabel("ENQUEUE_FOR_REVIEW"),
             1,
-            [ActionLabel("Action", "SomeMoreImortantAction")],
+            [ActionLabel("A_MORE_IMPORTANT_ACTION")],
         )
     ]
 
 
 def remove_superseded_actions(
-    action_labels: t.List[ActionLabel],
-) -> t.List[ActionLabel]:
+    action_labels: t.List["ActionLabel"],
+) -> t.List["ActionLabel"]:
     """
     TODO implement
     Evaluates a collection of ActionLabels against the configured Action objects, removing
@@ -99,23 +105,29 @@ def remove_superseded_actions(
     return action_labels
 
 
-def threat_exchange_reacting_is_enabled() -> bool:
+def threat_exchange_reacting_is_enabled(match_message: MatchMessage) -> bool:
     """
     TODO implement
-    Looks up whether ThreatExchange reactions are enabled.
+    Looks up from a config whether ThreatExchange reacting is enabled. Initially this will be a global
+    config, and this method will return True if reacting is enabled, False otherwise. At some point the
+    config for reacting to ThreatExchange may be on a per collaboration basis. In that case, the config
+    will be referenced for each collaboration involved (implied by the match message). If reacting
+    is enabled for a given collaboration, a label will be added to the match message
+    (e.g. "ThreatExchangeReactingEnabled:<collaboration-id>").
     """
     return True
 
 
 def get_threat_exchange_reaction_labels(
-    action_labels: t.List[ActionLabel],
-) -> t.List[Label]:
+    match_message: MatchMessage,
+    action_labels: t.List["ActionLabel"],
+) -> t.List["Label"]:
     """
     TODO implement
     Evaluates a collection of action_labels against some yet to be defined configuration
     (and possible business login) to produce
     """
-    return [Label("ThreatExchangeReaction", "SawThisToo")]
+    return [ThreatExchangeReactionLabel("SAW_THIS_TOO")]
 
 
 def perform_writeback_in_review(match_message: MatchMessage):
@@ -128,59 +140,38 @@ def perform_enque_for_review(match_message: MatchMessage):
     logger.debug("enqued for review")
 
 
-possible_actions = {
-    "WRITEBACK_IN_REVIEW": perform_writeback_in_review,
-    "ENQUE_FOR_REVIEW": perform_enque_for_review,
-}
+class LabelWithConstraints(Label):
+    _KEY_CONSTRAINT = "KeyConstraint"
+
+    def __init__(self, value: str):
+        super(LabelWithConstraints, self).__init__(self._KEY_CONSTRAINT, value)
 
 
-class ActionLabel(Label):
-    def __init__(self, key: str, value: str):
-        if key != "Action":
-            raise Exception("ActionLabels must have a key Action")
-
-        if value not in possible_actions.keys():
-            raise Exception(f"'%s' is not a valid Action" % value)
-
-        super(self.__class__, self).__init__(key, value)
+class ActionLabel(LabelWithConstraints):
+    _KEY_CONSTRAINT = "Action"
 
 
+class ThreatExchangeReactionLabel(LabelWithConstraints):
+    _KEY_CONSTRAINT = "ThreatExchangeReaction"
+
+
+@dataclass
 class Action:
     action_label: ActionLabel
     priority: int
     superseded_by: t.List[ActionLabel]
 
-    def __init__(
-        self,
-        action_label: ActionLabel,
-        priority: int,
-        superseded_by: t.List[ActionLabel],
-    ):
-        self.action_label = action_label
-        self.priority = priority
-        self.superseded_by = superseded_by
 
-
+@dataclass
 class ActionRule:
     action_label: ActionLabel
     must_have_labels: t.List[Label]
     must_not_have_labels: t.List[Label]
 
-    def __init__(
-        self,
-        action_label: ActionLabel,
-        must_have_labels: t.List[Label],
-        must_not_have_labels: t.List[Label],
-    ):
-        self.action_label = action_label
-        self.must_have_labels = must_have_labels
-        self.must_not_have_labels = must_not_have_labels
-
 
 def perform_action(match_message: MatchMessage, action_label: ActionLabel):
-    for action, action_performer in possible_actions.items():
-        if action is action_label.value:
-            action_performer(match_message)
+    # TODO implement
+    logger.debug("perform action")
 
 
 def react_to_threat_exchange(match_message: MatchMessage, reaction_label: Label):
@@ -188,9 +179,9 @@ def react_to_threat_exchange(match_message: MatchMessage, reaction_label: Label)
     logger.debug("react to threat exchange")
 
 
-# if __name__ == "__main__":
-#    # For basic debugging
-#    match_message = MatchMessage("key", "hash", [])
-#    action_label = ActionLabel("Action", "ENQUE_FOR_REVIEW")
+if __name__ == "__main__":
+    # For basic debugging
+    match_message = MatchMessage("key", "hash", [])
+    action_label = ActionLabel("ENQUE_FOR_REVIEW")
 
-#    perform_action(match_message, action_label)
+    perform_action(match_message, action_label)
--- a/hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
@@ -1,33 +1,96 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 import json
+import typing as t
 
 from hmalib.common.logging import get_logger
 from hmalib.models import MatchMessage, Label
 
 logger = get_logger(__name__)
 
-
 def lambda_handler(event, context):
     """
-    This lambda is called once per match per dataset. If a single hash matches
-    multiple datasets, this will be called multiple times.
+    This lambda is called when one or more matches are found. If a single hash matches
+    multiple datasets, this will be called only once.
 
-    Eventually, this will be just an action labeller. It will label a match
-    record with the action it recommends. A separate system will be stood up
-    that aggregates labels and 'decides' which action would be taken.
-
-    For now, this method will,
-    - it will construct the MatchMessage object and identify the actions it
-      needs to invoke.
-    - for now, rather than fanning out to individual specific lambdas, it will
-      call all specific functions serially!
+    Action labels are generated for each match message, then an action is performed
+    corresponding to each action label.
     """
-    for sqs_record in event["Records"]:
+    for sqs_record in event["Records"]: #confused about this... are we popping off more than one thing at a time from the "matches" queue?
         sns_notification = json.loads(sqs_record["body"])
         match_message: MatchMessage = MatchMessage.from_sns_message(
             sns_notification["Message"]
         )
+        action_labels = get_action_labels(match_message)
+        for action_label in action_labels:
+            perform_action(match_message, action_label)
+
+        # the more I write code here about reacting to ThreatExchange, the more I wish it was somewhere else entirely. Implementing
+        # here forever ties up action-related code with ThreatExchange-specific code / functionality. I prefer putting
+        # ThreatExchange-related specifics for reacting into another lambda that also subscribes to the matcher's notifications.
+        if threat_exchange_reacting_is_enabled():
+            threat_exchange_reaction_labels = get_threat_exchange_reaction_labels(action_labels)
+            if threat_exchange_reaction_labels is t.List[Label]:
+                for threat_exchange_reaction_label in threat_exchange_reaction_labels:
+                    react_to_threat_exchange(match_message, threat_exchange_reaction_label)
+
+
+def get_action_labels(match_message: MatchMessage) -> t.List[ActionLabel]:
+    """
+    TODO finish implementation
+    Returns an ActionLabel for each ActionRule that applies to a MatchMessage. 
+    """
+    action_rules = get_action_rules()
+    action_labels = [] # use Set here
+    for action_rule in action_rules:
+        if action_rule.applies(match_message) and not action_labels.contains(action_rule.action_label):
+            action_labels.append(action_rule.action_label)
+    action_labels = remove_superseded_actions(action_labels) # maybe not needed for v0
+    return action_labels
+
+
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
-    - for now, rather than fanning out to individual specific lambdas, it will
-      call all specific functions serially!
+    This lambda is called when one or more matches are found. If a single hash matches
+    multiple datasets, this will be called only once.
+
+    Action labels are generated for each match message, then an action is performed
+    corresponding to each action label.
     """
-    for sqs_record in event["Records"]:
+    for sqs_record in event["Records"]: #confused about this... are we popping off more than one thing at a time from the "matches" queue?
```
schatten, line 19: SQS + Lambda batches events together. Max 5 subject to a timeout. [1]

1: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -57,12 +120,39 @@ def __init__(self, key: str, value: str):
...
+    def __init__(self, action_label: ActionLabel, must_have_labels: t.List[Label], must_not_have_labels: t.List[Label]):
+        self.action_label = action_label
+        self.must_have_labels = must_have_labels
+        self.must_not_have_labels = must_not_have_labels
+
+
 def perform_action(match_message: MatchMessage, action_label: ActionLabel):
     for action, action_performer in possible_actions.items():
         if action is action_label.value:
             action_performer(match_message)
```
schatten, line 148: Since possible_actions is a dict, you can do:
```python
possible_actions[action_label.value](match_message)
```

Looping is unnecessary and inefficient.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
         match_message: MatchMessage = MatchMessage.from_sns_message(
             sns_notification["Message"]
         )
+        action_labels = get_action_labels(match_message)
+        for action_label in action_labels:
+            perform_action(match_message, action_label)
+
+        # the more I write code here about reacting to ThreatExchange, the more I wish it was somewhere else entirely. Implementing
+        # here forever ties up action-related code with ThreatExchange-specific code / functionality. I prefer putting
+        # ThreatExchange-related specifics for reacting into another lambda that also subscribes to the matcher's notifications.
```
schatten, line 30: I agree. While for v0, ThreatExchange is the only datasource supported, I like that we're already thinking ahead. One of the ways to decouple, IMO, is to use the match metadata (dataset,collaboration labels) to establish that this is a 'threatexchange' match and then act accordingly.

Additionally, food for thought: Something that @BarrettOlson and I've been noodling with is forwarding 'true positive' and 'false positive' reactions to be state-managed by the actions (queue+lambda). How do we do that?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+        if threat_exchange_reacting_is_enabled():
+            threat_exchange_reaction_labels = get_threat_exchange_reaction_labels(action_labels)
+            if threat_exchange_reaction_labels is t.List[Label]:
```
schatten, line 33: This probably would not work. `python` still does not have types. You can only check whether something is a list, not whether the list has a specific type.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+                for threat_exchange_reaction_label in threat_exchange_reaction_labels:
+                    react_to_threat_exchange(match_message, threat_exchange_reaction_label)
```
schatten, line 35: What would be helpful in the context of this super-drafty review is for you to call out what these functions will eventually do.

eg. am I looking at `react_to_threatexchange`, or `enqueue_a_react_to_threatexchange_job`?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
             sns_notification["Message"]
         )
+        action_labels = get_action_labels(match_message)
+        for action_label in action_labels:
+            perform_action(match_message, action_label)
+
+        # the more I write code here about reacting to ThreatExchange, the more I wish it was somewhere else entirely. Implementing
+        # here forever ties up action-related code with ThreatExchange-specific code / functionality. I prefer putting
+        # ThreatExchange-related specifics for reacting into another lambda that also subscribes to the matcher's notifications.
+        if threat_exchange_reacting_is_enabled():
```
schatten, line 31: This will also depend on the privacy-group of the source match, maybe add to the arguments list?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+    TODO implement
+    Returns a collection of ActionRule objects. An ActionRule will have the following attributes:
+    MustHaveLabels, MustNotHaveLabels, ActionLabel
+    """
+    return [ActionRule(ActionLabel("Action", "EnqueueForReview"), [Label("Collaboration", "12345")], [])]
+
+
+def get_actions() -> t.List[Action]:
+    """
+    TODO implement
```
schatten, line 63: And this will be fetched from the database, right? That's where the rules live?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+    Returns a collection of Action objects. An Action will have the following attributes:
+    ActionLabel, Priority, SupersededByActionLabel
+    """
+    return [Action(ActionLabel("Action", "EnqueueForReview"), 1, [ActionLabel("Action", "SomeMoreImortantAction")])]
+
+
+def remove_superseded_actions(action_labels: t.List[ActionLabel]) -> t.List[ActionLabel]:
+    """
```
schatten, line 71: I'm assuming this will continue to be a no-op for v0?

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -60,6 +60,9 @@ def to_dynamodb_dict(self) -> dict:
     def from_dynamodb_dict(cls, d: dict) -> "Label":
         return cls(d["K"], d["V"])
 
+    def __eq__(self, another_label: object):
```
schatten, line 63: nit: another_label will be of type 'Label', not object.

Because the type is not materialized yet, mypy will allow you to do:
```python
def __eq__(self, other: "Label"):
```

Note how we are using quotes around Label.

schatten (changes requested):
Using RC to put back into your queue.

What would help to review this diff:
1. Call out in comments or in method names how the data will be fetched eg. database, static configuration.
2. For methods that are placeholder for delegation eg. perform_action, react..., specify what that will delegate to. eg. SQS, multiprocessing thread, etc.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
-
-    For now, this method will,
-    - it will construct the MatchMessage object and identify the actions it
-      needs to invoke.
-    - for now, rather than fanning out to individual specific lambdas, it will
-      call all specific functions serially!
+    This lambda is called when one or more matches are found. If a single hash matches
+    multiple datasets, this will be called only once.
+
+    Action labels are generated for each match message, then an action is performed
```
jeberl, line 16: maybe make it clear ", then a single action is performed"

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
jeberl: Does it? Because a partner might want to react back to some privacy groups but not all? This seems like a feature we can punt on

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
jeberl: This is trying to check if all items in the list are Labels? if so the pythonic way would be...
```
if all( label is Label for label in threat_exchange_reaction_labels ):
    ...
```

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+    return True
+
+
+def get_threat_exchange_reaction_labels(action_labels: t.List[ActionLabel]) -> t.List[Label]:
+    """
+    TODO implement
+    Evaluates a collection of action_labels against some yet to be defined configuration
+    (and possible business login) to produce 
+    """
+    return [Label("ThreatExchangeReaction", "SawThisToo")]
```
jeberl, line 93: Maybe we want to create a class `ThreatExchangeReactionLabel`?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+                    react_to_threat_exchange(match_message, threat_exchange_reaction_label)
+
+
+def get_action_labels(match_message: MatchMessage) -> t.List[ActionLabel]:
+    """
+    TODO finish implementation
+    Returns an ActionLabel for each ActionRule that applies to a MatchMessage. 
+    """
+    action_rules = get_action_rules()
+    action_labels = [] # use Set here
```
jeberl, line 44: Python tips:
`[]` is empty List
`{}` is empty Set or empty Dict

This also might be a good place for List comprehension where you could write lines 44-47 as:
```
action_labels = { action_rule.action_label for for action_rule in action_rules if action_rule.applies(match_message) }
```

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -1,33 +1,96 @@
...
+def get_actions() -> t.List[Action]:
+    """
+    TODO implement
+    Returns a collection of Action objects. An Action will have the following attributes:
+    ActionLabel, Priority, SupersededByActionLabel
+    """
+    return [Action(ActionLabel("Action", "EnqueueForReview"), 1, [ActionLabel("Action", "SomeMoreImortantAction")])]
+
+
+def remove_superseded_actions(action_labels: t.List[ActionLabel]) -> t.List[ActionLabel]:
```
jeberl, line 70: The more I think about this the less I like it for v0. @Dcallies brought up a good point that the actioner is likely going to be the most sensitive part of HMA where partners are most nervous about accidentally taking the wrong or an unexpected action. If we can make the outcome as obvious as possible I think that will be best until we need more complex logic like ("dont demote if we're going to delete")

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -57,12 +120,39 @@ def __init__(self, key: str, value: str):
...
+
+    def __init__(self, action_label: ActionLabel, priority: int, superseded_by: t.List[ActionLabel]):
+        self.action_label = action_label
+        self.priority = priority
+        self.superseded_by = superseded_by
+
+
+class ActionRule:
+    action_label: ActionLabel
+    must_have_labels: t.List[Label]
```
jeberl, line 136: Wasn't quite thinking about this before but this implies that tags in ThreatExchange at somepoint become Label objects. That seems right to me but where will that take place? In the actioner or the matcher? (or the fetcher?)

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -57,12 +120,39 @@ def __init__(self, key: str, value: str):
...
+    priority: int
+    superseded_by: t.List[ActionLabel]
+
+    def __init__(self, action_label: ActionLabel, priority: int, superseded_by: t.List[ActionLabel]):
+        self.action_label = action_label
+        self.priority = priority
+        self.superseded_by = superseded_by
+
+
+class ActionRule:
```
jeberl, line 134: same as above

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -57,12 +120,39 @@ def __init__(self, key: str, value: str):
         super(self.__class__, self).__init__(key, value)
 
 
+class Action:
```
jeberl, line 123: I think we can use `@dataclass` as a decorator and remove the `__init__` function

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
jeberl: Me to as I would imagine this functionality in the `get_action_labels` function or a separate function but before line 26. specifically replace lines 24-26 with something like:
```
user_defined_action_labels = get_action_labels(match_message)
user_defined_action_labels_and_business_logic_action_labels = add_business_logic_labels(match_message, user_defined_action_labels)
for action_label in user_defined_action_labels_and_business_logic__action_labels:
     perform_action(match_message, action_label)
```

where `add_business_logic_labels` add reaction labels and in future iterations would add ncmec reporting labels, and maybe other labels

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @schatten @BarrettOlson @jeberl re "Additionally..." once we have action execution and react to threat exchange queues, how about putting a true / false positive reaction onto the the react to threat exchange queue?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @schatten I thought we were doing this at the global level for now, but happy to change to being per collab.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @schatten My need / intent is to validate (in a python-y way) that there's something to iterate over, in case there's a concept of null or undefined in Python... I don't want to do the `for` if there isn't a list... But this is a general thing that I don't know anything about yet in Python. Maybe I can just remove line 33? Or should I leave it the way it is but remove `Label`:

`if threat_exchange_reaction_labels is t.List[]:`

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @schatten what is `enqueue_a_react_to_threatexchange_job` I'm not seeing that in the code anywhere?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @schatten I'm wondering about that. I was going to wait to hear from @Dcallies before making any decision... Seems like if we've got a single place for all our configurations, rules, actions, webhook details, pretty much anything might land in the common config data store that might emerge.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @schatten Yeah I may even remove it completely but wanted it to be present to tell the initial story.

hasher-matcher-actioner/hmalib/models.py
eriksendc: @schatten That was definitely a miss (you'd helped me with types and I didn't carry that learning to this module). Are you saying we _can_ use the quoted type name, or that we _should_?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl As I'm unfamiliar with syntax, is the `for` being repeated a type-o or really something that should be there?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl Yep see other comment below to @schatten

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl Yep was thinking that, too. Was hoping for some kind of comment to push that way.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: We decided that all actions corresponding to all rules that apply will be executed, so you need to read line 17 and 18 together, and I think that as that's currently worded, its meaning is clear, no?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl I planned on that being post match, pre-action, and I planned on putting that in to the matcher module based on the beginnings of what @schatten started I think. @schatten what do you think?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
jeberl: Oh you want to check if the list is empty?

either:
```
if len(threat_exchange_reaction_labels) > 0 :
```

or just :
```
if threat_exchange_reaction_labels :
```

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -10,24 +12,122 @@
...
     """
-    for sqs_record in event["Records"]:
+    for sqs_record in event[
+        "Records"
+    ]:  # TODO research max # sqs records / lambda_handler invocation
         sns_notification = json.loads(sqs_record["body"])
         match_message: MatchMessage = MatchMessage.from_sns_message(
             sns_notification["Message"]
         )
+        action_labels = get_action_labels(match_message)
```
jeberl, line 28: I think we want to get additional labels here for reacting before we pass to the `perform_action` function. We might have a label `Label("Action", "ReactEnquedForReview") These can be calculated here and executed by `perfrom_action` as well

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -10,24 +12,122 @@
...
+            )
+            if threat_exchange_reaction_labels:
+                for threat_exchange_reaction_label in threat_exchange_reaction_labels:
+                    # TODO create a new ThreatExchange reaction queue and enqueue
+                    # the match message and threat exchange reaction label (or, possibly,
+                    # add the threat exchange reaction label to the match message
+                    # and enqueue the match message by itself)
+                    react_to_threat_exchange(
+                        match_message, threat_exchange_reaction_label
+                    )
```
jeberl, line 48: See comment above. I think lines 36-39 should go above line 29 and lines 40-48 should go in the perform_action fucntion (ie. I was planning on handling these in the action executor).

Is this some thing you considered and decided against? Should we discuss?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -40,32 +140,48 @@ def perform_enque_for_review(match_message: MatchMessage):
     logger.debug("enqued for review")
 
 
-possible_actions = {
-    "WRITEBACK_IN_REVIEW": perform_writeback_in_review,
-    "ENQUE_FOR_REVIEW": perform_enque_for_review,
-}
+class LabelWithConstraints(Label):
```
jeberl, line 143: Nice

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -40,32 +140,48 @@ def perform_enque_for_review(match_message: MatchMessage):
...
 
-class ActionLabel(Label):
-    def __init__(self, key: str, value: str):
-        if key != "Action":
-            raise Exception("ActionLabels must have a key Action")
+class ActionLabel(LabelWithConstraints):
+    _KEY_CONSTRAINT = "Action"
 
-        if value not in possible_actions.keys():
-            raise Exception(f"'%s' is not a valid Action" % value)
```
jeberl, line 55: Where did this value constraint go?

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -40,32 +140,48 @@ def perform_enque_for_review(match_message: MatchMessage):
...
 
-        super(self.__class__, self).__init__(key, value)
+class ThreatExchangeReactionLabel(LabelWithConstraints):
+    _KEY_CONSTRAINT = "ThreatExchangeReaction"
+
+
+@dataclass
+class Action:
```
jeberl, line 159: Again, I think we might want to remove this if we're not doing the priority

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -40,32 +140,48 @@ def perform_enque_for_review(match_message: MatchMessage):
...
+class ActionRule:
+    action_label: ActionLabel
+    must_have_labels: t.List[Label]
+    must_not_have_labels: t.List[Label]
 
 
 def perform_action(match_message: MatchMessage, action_label: ActionLabel):
-    for action, action_performer in possible_actions.items():
-        if action is action_label.value:
-            action_performer(match_message)
```
jeberl, line 63: Why was this removed? Did you mean to replace it with 
```
possible_actions[action_label.value](match_message)
```
from @schatten 's comment?

jeberl (changes requested):
Looks good. I think the only thing to get clarity on is the reaction. RC to put back in your queue

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl I can see how what I did with git is getting in the way here. There was a conversation that I thought came to the conclusion that we want to keep ThreatExchange-y processing separate from "core" functionality. I proposed that we have another lambda specifically for ThreatExchange reactions, hence the `ThreatExchangeReactionLabel` concept and its separate processing at 36-48.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: Thx! :) My first stab at something with inheritance in Python!

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl Nope removed on purpose. Actions are meant to be configured. At any time I should be able to go to the UI, add another rule with some new action label, along with corresponding configuration for action execution, and then there's a new rule that should be able to fire. So the concept of a hard-coded list / dict `possible_actions` is contrary to enabling that on-the-fly, configuration-based way of things working. But I might be missing something you were thinking, so let's talk.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: see above.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
eriksendc: @jeberl I have a feeling that the concept of priority and superseded_by is probably v1 (so, "coming soon to a codebase near you"). What's our standard on leaving stubs / reference points in code? I'm OK with deleting, or maybe commenting things out? Let's talk in the AM.

schatten (approved):
I'm good with how things look right now.

Have a question or two. Perhaps noodle on them.

Since we are hardcoding (hardcoding == not configurable via db) the value for the action labels, how would the UI function?

eg. 
in the rules page, user says: If Labels contain (x, y, z): attach the following action.

where/how do I configure the webhook? Can a user only have 2 webhooks with the `possible_actions` as values?

I'm not able to wrap my head around this right now. But doesn't mean we need to address it yet.

jeberl (approved):
Talked offline. Ship it!

PR #484, opened 2021-04-13
[hma] have dashboard and signal page return a few real counts

BarrettOlson:
Summary
---------

There is an (albeit inefficient) way to get a subset of the counts we want and surface them on the UI. why not? 

Test Plan
---------

```
make upload docker
...
# apply with test app token set in `/terraform/terraform.tfvars`
...
make upload_test_data
...
npm start
...
```
http://localhost:3000/ 
login and view dashboard and signal pages

Screen Shot 2021-04-12 at 8 14 32 PM
Screen Shot 2021-04-12 at 8 14 23 PM


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -323,22 +324,22 @@ def get_signals() -> t.List[SignalSourceSummary]:
...
-        name="Mocked Signal Source 2",
-        signals=[mocked_signal_type2a, mocked_signal_type2b],
-        updated_at="mocked datetime2",
-    )
-    return [mocked_signal_source1, mocked_signal_source2]
+    signals = []
+    counts = get_signal_hash_count()
+    for dataset, total in counts.items():
+        if dataset.endswith(THREAT_EXCHANGE_PDQ_FILE_EXTENSION):
+            dataset_name = dataset.replace(
```
jeberl, line 331: So for now dataset_name will always be the collaboration ID?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -354,18 +355,26 @@ class DashboardSystemStatus(t.TypedDict):
...
-    )
+    return get_count_from_record_cls(PipelinePDQHashRecord)
 
 
 def get_dashboard_matches() -> DashboardCount:
+    return get_count_from_record_cls(PDQMatchRecord)
+
+
+def get_count_from_record_cls(
+    record_cls: t.Union[t.Type[PDQMatchRecord], t.Type[PipelinePDQHashRecord]]
```
jeberl, line 366: blocking: Can we make a real class `Record` that these inherit from? 

Then we could make `get_count_from_record_cls` a classmethod on `Record`

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -354,18 +355,26 @@ class DashboardSystemStatus(t.TypedDict):
...
+) -> DashboardCount:
+    # TODO this is an very inefficient way to do this but connecting to real counts > mock
```
jeberl, line 368: Any estimate of how long the page currently takes to load?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -393,7 +405,7 @@ def get_dashboard_system_status() -> DashboardSystemStatus:
     )
 
 
-def get_hash_count() -> t.Dict[str, int]:
+def get_signal_hash_count() -> t.Dict[str, int]:
```
jeberl, line 408: blocking: Seems like we call this function a bunch and its quite expensive. Can we memoize it with `@cache`?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -378,10 +387,13 @@ def get_dashboard_actions() -> DashboardCount:
 
 
 def get_dashboard_signals() -> DashboardCount:
+    """
+    note: SIGNALS COUNT != NUMBER OF UNQUIE HASHES
+    """
```
jeberl, line 392: Can we describe what it is (instead of what its not)

hasher-matcher-actioner/hmalib/tests/test_models.py
```diff
@@ -206,3 +219,15 @@ def test_query_match_record_by_time(self):
         query_record = models.PDQMatchRecord.get_from_time_range(self.table)[0]
 
         assert record == query_record
+
+
+class LabelsTestCase(unittest.TestCase):
+    def test_label_validation(self):
```
jeberl, line 225: nit: still needed with lower test?

hasher-matcher-actioner/hmalib/tests/test_models.py
```diff
@@ -206,3 +219,15 @@ def test_query_match_record_by_time(self):
...
+        l = models.Label("some key", "some value")
+        # Just validate that no error is raised
+
+    def test_label_serde(self):
+        # serde is serialization/deserialization
```
jeberl, line 230: TIL!

jeberl (changes requested):
Wow! This looks so good. 

lmk what you think about creating a record class or if there was some reason you decided against it.

If you really want to get this out today let me know I can can stamp

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: yeah, the <filename> of <filename>.pdq.te is which for now is the pg id

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: The class exists will update.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: I don't think it is a bottleneck right now this is more a warning for when we get a lot of hashes/matches in a system it might become one.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: none of the functions like this one have docstrings because they are pretty straight forward this is just a heads up for someone coming back that might assume.

jeberl: Signals is the only one that seems non obvious to me. "Hashes" and "matches" we elsewhere in the code and documentation "Signals" I haven't seen so I think a docstring here is more necessary

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: agree but punting for now @Dcallies thoughts on @lru_cache in another PR maybe?

hasher-matcher-actioner/hmalib/tests/test_models.py
BarrettOlson: didn't write the test, just consolidating but this might help narrow down if the issue real is serde or not

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: (the mocked data Action card was the last to populate on page load when I tested it just now)

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: I don't disagree but going to leave it for now if that is alright with you. we're approaching the need for a glossary as is and I'd rather be more consistent in clarification around 'Signals' overall.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
jeberl: I'm good with punting to a future PR

jeberl (approved):

hasher-matcher-actioner/webapp/src/Signals.jsx
```diff
@@ -45,41 +45,39 @@ export default function Signals() {
...
+            <thead>
+              <tr>
+                <th>Signal Type</th>
+                <th>Number of Signals</th>
+              </tr>
+            </thead>
+            <tbody>
+              {summary.signals !== null && summary.signals.length ? (
+                summary.signals.map(signal => (
+                  <tr key={signal}>
```
eriksendc, line 64: I took a quick look and it does sound like this should be a string, not an object. Behind the scenes React is probably doing a "to string" on each `signal` as its mapped. From the docs:

The best way to pick a key is to use a string that uniquely identifies a list item among its siblings.

https://reactjs.org/docs/lists-and-keys.html#keys

So it's probably better to pick something that's a bit more predictable and concise if possible. JavaScript toString() creates a big chunk of text? (I don't know what the javascript tostring output on `signal` will be, for example?

eriksendc (approved):

PR #487, opened 2021-04-13
Get privacy groups

mengyangwang:
Summary
---

1. Add function get_threat_privacy_groups_owner to find the list of all privacy groups the current app is a owner of
2. Add function privacy_group_can_use_threat_updates

TestPlan
---
```
$ THREAT_EXCHANGE_INTEGRATION_TEST_TOKEN=$(access_token) pytest -v
==== 60 passed, 5 warnings in 1.37s ====
....
```

Lint
```
$ black .
$ mypy threatexchange
```


python-threatexchange/tests/test_api.py
```diff
@@ -30,3 +32,27 @@ def test_get_threat_privacy_groups_member(self):
...
+
+    def test_get_threat_privacy_groups_owner(self):
+        """
+        Assumes that the app (if token is provided) will have at least one
+        privacy group.
+        """
+        response = self.api.get_threat_privacy_groups_owner()
+        self.assertTrue(
+            isinstance(response, collections.abc.Sequence)
+            and not isinstance(response, staticmethod),
```
jeberl, line 44: Why are we checking if its not a static method? I see its copied from above but I think we should remove

python-threatexchange/threatexchange/api.py
```diff
@@ -302,13 +302,48 @@ def get_threat_privacy_groups_member(
...
+    def privacy_group_can_use_threat_updates(self, privacy_group_id: int) -> bool:
+        """
+        Returns if the privacy group can be used with the threat updates endpoint.
+        """
+        url = self._get_graph_api_url(f"{privacy_group_id}/threat_updates")
+        response = self.get_json_from_url(url)
+        if "error" not in response:
+            return True
+        else:
+            return False
```
jeberl, line 345: blocking: We can use the field `threat_updates_enabled` once it lands (~1 hour). We should add this as a property to `ThreatPrivacyGroup` and check it here instead of doing another api call

python-threatexchange/tests/test_api.py
```diff
@@ -30,3 +32,27 @@ def test_get_threat_privacy_groups_member(self):
...
+
+    def test_privacy_group_can_use_threat_updates(self):
+        """
+        Test with correct privacy_group_id and wrong privacy_group_id
+        """
+        response = self.api.privacy_group_can_use_threat_updates(
+            PRIVACY_GROUP_ID_CORRECT
+        )
+
+        self.assertTrue(response)
```
jeberl, line 58: This feels like more of a canary then a unit test since its asserting with production data which may change. Can we try to find a way to mock the data or else remove?

jeberl (changes requested):

jeberl (approved):
lgtm!

PR #489, opened 2021-04-14
[hma] Propagate tags to index metadata + minor fixes

BarrettOlson:
Summary
---------

This adds the labels/tags to the metadata object in the index. 

Other minor issues I saw from the logs while testing and fixed:
`pdq_matcher.py` imports was missing stuff from models.py and 
`ThreatExchangeS3PDQAdapter.file_type_str_name` was missing a return statement
`webapp/src/Signals.jsx` should handle the case where there are no signals cleaner

Test Plan
---------
made tmp change to `hasher-matcher-actioner/sample_data/pdq.csv` then ran:
`make dev_upload_test_data` 
and used cloudwatch logs to test: 
```
...
Match found for key: images/b.jpg, hash f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22 ->
{'id': 4109214869142910, 'hash': 'f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22', 'source': 'te', 'privacy_groups': {'sample_data'}, 'tags': {'sample_data': ['media_priority_pdq_samples', 'disputed', 'media_priority_samples', 'media_type_photo']}}
```

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/s3_adapters.py
+++ b/hasher-matcher-actioner/hmalib/common/s3_adapters.py
@@ -124,9 +124,9 @@ class ThreatExchangeS3Adapter:
                     "privacy_groups": [
                         file_name.split("/")[-1].split(".")[0]
                     ],  # read privacy group from key
-                    "tags": row[
-                        "tags"
-                    ],  # note: these are the labels assigned by pytx in descriptor.py (NOT a 1-1 with tags on TE)
+                    "tags": row["tags"].split(" ")
+                    if row["tags"]
+                    else None,  # note: these are the labels assigned by pytx in descriptor.py (NOT a 1-1 with tags on TE)
                 },
             )
             for row in data_reader
--- a/hasher-matcher-actioner/hmalib/common/s3_adapters.py
+++ b/hasher-matcher-actioner/hmalib/common/s3_adapters.py
@@ -126,7 +126,7 @@ class ThreatExchangeS3Adapter:
                     ],  # read privacy group from key
                     "tags": row["tags"].split(" ")
                     if row["tags"]
-                    else None,  # note: these are the labels assigned by pytx in descriptor.py (NOT a 1-1 with tags on TE)
+                    else [],  # note: these are the labels assigned by pytx in descriptor.py (NOT a 1-1 with tags on TE)
                 },
             )
             for row in data_reader
--- a/hasher-matcher-actioner/hmalib/common/s3_adapters.py
+++ b/hasher-matcher-actioner/hmalib/common/s3_adapters.py
@@ -73,7 +73,6 @@ class ThreatExchangeS3Adapter:
                 file_name: self._parse_file(**typed_data_file)
                 for file_name, typed_data_file in typed_data_files.items()
             }
-
         return typed_data
 
     @property
@@ -125,6 +124,9 @@ class ThreatExchangeS3Adapter:
                     "privacy_groups": [
                         file_name.split("/")[-1].split(".")[0]
                     ],  # read privacy group from key
+                    "tags": row[
+                        "tags"
+                    ],  # note: these are the labels assigned by pytx in descriptor.py (NOT a 1-1 with tags on TE)
                 },
             )
             for row in data_reader
@@ -146,4 +148,4 @@ class ThreatExchangeS3PDQAdapter(ThreatExchangeS3Adapter):
 
     @property
     def file_type_str_name(self):
-        "PDQ"
+        return "PDQ"
--- a/hasher-matcher-actioner/hmalib/common/s3_adapters.py
+++ b/hasher-matcher-actioner/hmalib/common/s3_adapters.py
@@ -113,6 +113,7 @@ class ThreatExchangeS3Adapter:
             codecs.getreader("utf-8")(data_file["Body"]),
             fieldnames=self.indicator_type_file_columns,
         )
+        privacy_group = file_name.split("/")[-1].split(".")[0]
         return [
             (
                 row["hash"],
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -11,7 +11,7 @@
 from threatexchange.signal_type.pdq_index import PDQIndex
 
 from hmalib import metrics
-from hmalib.models import PDQMatchRecord, Label
+from hmalib.models import PDQMatchRecord, Label, MatchMessage, DatasetMatchDetails
```
schatten, line 14: Wow! How was the matcher working prior to this? 

Makes me wonder. What's `mypy` doing?

hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -125,6 +124,9 @@ def _parse_file(self, file_name: str, data_file: S3FileT) -> t.List[HashRowT]:
                     "privacy_groups": [
                         file_name.split("/")[-1].split(".")[0]
                     ],  # read privacy group from key
+                    "tags": row["tags"].split(" ")
+                    if row["tags"]
+                    else None,  # note: these are the labels assigned by pytx in descriptor.py (NOT a 1-1 with tags on TE)
```
schatten, line 129: nit: Would an empty list be more pertinent here?

schatten (approved):
LGTM. Have a nit about returning empty lists instead of None. Makes downstream guard clauses unnecessary, but I'm good with both.

#shipit!

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
BarrettOlson: because nothing depended on MatchMessage yet and we write to the dydb before trying to use the class

good question...

eriksendc (approved):

hasher-matcher-actioner/webapp/src/Signals.jsx
```diff
@@ -15,7 +15,9 @@ export default function Signals() {
 
   useEffect(() => {
     fetchSignalSummary().then(summaries => {
-      setSignalSummary(summaries.signals);
+      if (summaries.signals.length) {
```
eriksendc, line 18: Dunno if I'm reading this right, but if this check prevents the call to `setSignalSummary()` then the spinner will show forever, even though the API call completed, right? So... introducing this maybe you need to let `setSignalSummary()` unconditionally and then have two `<Collapse>` elements, one for data and one for a "no data returned" kind of message?

eriksendc (changes requested):

hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -114,6 +113,7 @@ def _parse_file(self, file_name: str, data_file: S3FileT) -> t.List[HashRowT]:
             codecs.getreader("utf-8")(data_file["Body"]),
             fieldnames=self.indicator_type_file_columns,
         )
+        privacy_group = file_name.split("/")[-1].split(".")[0]
```
schatten, line 116: Can you add a comment where ever this file_name is created so that future refactors do not accidentally break this?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -72,7 +72,9 @@ def merge_pdq_files(
         accumulator[hash] = hash_row
     else:
         # Add new privacy group to existing row
-        accumulator[hash][1]["privacy_groups"].append(meta_data["privacy_groups"][0])
+        accumulator[hash][1]["privacy_groups"].update(meta_data["privacy_groups"])
+        # Add new labels from privacy group to existing row
+        accumulator[hash][1]["tags"].update(meta_data["tags"])
     return accumulator
```
schatten, line 78: I read this again. This makes sense! Thanks!

hasher-matcher-actioner/webapp/src/Signals.jsx
BarrettOlson: yeah, going from breaking -> spinning with this change, will add TODO comment for now.

hasher-matcher-actioner/hmalib/common/s3_adapters.py
BarrettOlson: Done. (also added a TODO because where it is created has the extension hardcoded for some reason...)

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py
```diff
@@ -72,7 +72,9 @@ def merge_pdq_files(
         accumulator[hash] = hash_row
     else:
         # Add new privacy group to existing row
-        accumulator[hash][1]["privacy_groups"].append(meta_data["privacy_groups"][0])
+        accumulator[hash][1]["privacy_groups"].update(meta_data["privacy_groups"])
+        # Add new labels from privacy group to existing row
```
eriksendc, line 76: complete nit, but from a naming convention, for now, at this point here, this is still te data and so here it's tags not labels, so I'd change Add new labels... to Add new tags...

BarrettOlson: it's in a superposition, schrodinger metadata if you will

PR #491, opened 2021-04-14
[hma] Simple Config Framework

Dcallies:
Summary
---------

We need to create a bunch more configs:
1. Fetcher needs expansion
2. Banks someday
3. Action rules
4. Action configs

When I did the fetcher configs, I just created a new table. However, that will get really annoying over time. Instead use one table with a simple name/type combo for everything.

This lets you do the following:

```
@dataclass
class MyCoolConfig(HMAConfig):
  enabled: bool
  privacy_groups: t.Set[int]
  action_rules: t.Dict[str, List[str]]

cfg = MyCoolConfig("Coolio", True, {123456}, ["and" => ["foo", "bar"])
update_config(cfg)  # Create
cfg = MyCoolConfig.get("Coolio")  # Fetch
cfg.enabled = False
update_config(cfg)  # Update
delete_config(cfg)  # Delete
print("Now only", len(MyCoolConfig.get_all()), "cool configs left")  # Get all
```

There are some tradeoffs being taken here:
Pro:
* Ultra simple
* Fast to use and integrate

Con:
* If you give access a role access to one config, you give access to all

To add a new config, what would be needed is:
1. Create a new class extending HMAConfig
2. Pass in the config table into the module, and add the connection to the config table.

Left as followups:
1. Replace syncer config with new model (heads up @schatten, @mengyangwang)
2. Add any new configs

Test Plan
---------

Added a unittest

Also wrote to a local copy of my database (update & delete) from an interactive python shell to verify all the things I was thinking about doing were possible:

image


hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,262 @@
...
+
+def _config_to_dynamodb_item(config) -> t.Dict[str, t.Any]:
+    """
+    Convert a config object into what is what goes into the put_item Item arg
+    """
+    item = {
+        field.name: _py_to_aws_field(field.type, getattr(config, field.name))
+        for field in fields(config)
+    }
+    del item["name"]
```
eriksendc, line 159: Up above I was going to ask about why "ConfigName" instead of just "name". What about changing the instance variable `name` to `config_name` and that way `name` (something fairly common someone might think to use for something) wouldn't be a reserved word, so to speak? I'm thinking you might have a config as follows:

ConfigType: EmergencyContact
ConfigName: ContactID1
Name: Brian <-- there are perhaps lots of Brians

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,262 @@
...
+    if in_type is Decimal:  # N
+        return py_field  # type: ignore # mypy/issues/10003
+    if in_type is str:  # S
+        return py_field  # type: ignore # mypy/issues/10003
+    if in_type is bool:  # BOOL
+        return py_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[str]:  # SS
+        return py_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[int]:  # SN
+        return {Decimal(s) for s in py_field}  # type: ignore # mypy/issues/10003
```
eriksendc, line 249: Probably a dumb question but why `Decimal(s)` here instead of `int(s)`?

eriksendc (approved):
I'm approving with a "hey think about the name --> config_name" change thing. I'd like to know more about thoughts on function name overloading, too. And I just don't understand line 245 of config.py, so I don't know if that's something to change or not.

eriksendc:
And the test that's failing should be fixed too, but assuming you'll do that. (Should we bother calling that kind of thing out?)

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,262 @@
...
+    # TODO - we should probably sanity check here to make sure all the fields
+    #        are the expected types, because lolpython. Otherwise, it will
+    #        fail to deserialize later
+    dynamodb.meta.client.put_item(
+        TableName=_TABLE_NAME,
+        Item=_config_to_dynamodb_item(config),
+    )
+
+
+def delete_config_by_name(config_type: str, name: str) -> None:
```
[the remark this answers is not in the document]
Dcallies: This is a copy paste error -_-

hasher-matcher-actioner/hmalib/common/config.py
Dcallies: `name` is a reserved variable in DynamoDB, which requires extra boilerplate to handle. Thus the column name is ConfigName, and it's special cased during translation.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,262 @@
...
+    """Delete a config by name only"""
```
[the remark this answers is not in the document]
Dcallies: Yes, those are the arguments to the function, but you can't do anything with the configs without type and so just dropped it.

hasher-matcher-actioner/hmalib/common/config.py
eriksendc: Sounds good!

hasher-matcher-actioner/hmalib/common/config.py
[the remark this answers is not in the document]
eriksendc: @Dcallies then just name it `delete_config_by_name_and_type()`?

hasher-matcher-actioner/hmalib/common/config.py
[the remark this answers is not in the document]
eriksendc: This comment should say `Delete a config by name and type`

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,266 @@
...
+from dataclasses import dataclass, fields, is_dataclass
+import typing as t
+
+import boto3
+from boto3.dynamodb.conditions import Attr
+
+T = t.TypeVar("T")
+TConfig = t.TypeVar("TConfig", bound="HMAConfig")
+
+dynamodb = boto3.resource("dynamodb")
```
Dcallies, line 20: This line is failing during test because there's no region.

One possible fix is to turn this is a standard "how to hide module-level side-effects" and convert it to

def get_dynamodb():
  return boto3.resource("dynamodb")

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -23,6 +23,31 @@ locals {
   te_data_folder     = module.hashing_data.threat_exchange_data_folder_info.key
 }
 
+### Config storage ###
```
BarrettOlson, line 26: Add ToDo to someday move to sub modules/consolidate?

hasher-matcher-actioner/terraform/fetcher/variables.tf
```diff
@@ -61,3 +61,8 @@ variable "collab_file" {
   description = "An optional file name of ThreatExchange Collaborations objects to prepopulate. See collabs_example.json for the correct formatting"
   type = string
 }
+
+variable "config_arn" {
```
schatten, line 65: Within a module, can we use the ARN alone to setup the hmalib.config.common module?

Can you give a pseudo-code example of how a lambda would use the ARN / additional environment variables to initialize?

Dcallies: Yea, check out how the fetcher does it, I was planning on basically duplicating the same thing. I might be able to just demonstrate in a followup diff to do the fetcher.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,282 @@
...
+    return boto3.resource("dynamodb")
+
+
+class HMAConfigSerializationError(ValueError):
+    pass
+
+
+@dataclass
+class HMAConfig:
+    """
```
schatten, line 46: Can you add a doc comment or a post_init validation to ensure no one is using ConfigName or ConfigType field names. Extremely unlikely, because python attributes should not be CamelCase, but just in case?

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,282 @@
...
+        return aws_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[str]:  # SS
+        return aws_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[int]:  # SN
+        return {int(s) for s in aws_field}  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[float]:  # SN
+        return {float(s) for s in aws_field}  # type: ignore # mypy/issues/10003
+
+    if origin is list:  # L
+        assert len(args) == 1
```
schatten, line 218: Can you explain why multiple args could be passed to the list type?

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,282 @@
...
+    if in_type is bool:  # BOOL
+        return aws_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[str]:  # SS
+        return aws_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[int]:  # SN
+        return {int(s) for s in aws_field}  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[float]:  # SN
+        return {float(s) for s in aws_field}  # type: ignore # mypy/issues/10003
+
+    if origin is list:  # L
```
schatten, line 217: Seems like we can do `origin is list || origin is set`. What do you think?

One more primitive supported and we can add better typing to 'ActionConfig'. I expect it to have `Set[Label]`, not `List[Label]`.
cc: @eriksendc

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -0,0 +1,282 @@
...
+        return py_field  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[int]:  # SN
+        return {i for i in py_field}  # type: ignore # mypy/issues/10003
+    if in_type is t.Set[float]:  # SN
+        # WARNING WARNING
+        # floating point is not truly supported in dynamodb
+        # See note above
+        return {Decimal(str(s)) for s in py_field}  # type: ignore # mypy/issues/10003
+
+    if origin is list:  # L
```
schatten, line 272: Same comment as above for sets.

hasher-matcher-actioner/hmalib/common/config.py
Dcallies: I need to return list or set (L215 and L218 are different). 

I looked into supporting dataclasses as an option, and saw it was possible, but don't want to implement it in first pass.

hasher-matcher-actioner/hmalib/common/config.py
Dcallies: Adding doc

hasher-matcher-actioner/hmalib/common/config.py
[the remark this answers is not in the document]
Dcallies: Discussed offline, would prefer to keep function name as is, but can update docbloc

hasher-matcher-actioner/hmalib/common/config.py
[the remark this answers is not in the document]
Dcallies: Discussed offline, would prefer to keep as is

eriksendc (approved):

hasher-matcher-actioner/hmalib/common/config.py
Dcallies: I thought that it would implicitly convert to tuple if you did List[str, str], but it appears to error, so I can safely remove these
```
 t.List[int, str]
  File "<stdin>", line 1, in <module>
  File "C:\Python39\lib\typing.py", line 262, in inner
    return func(*args, **kwds)
  File "C:\Python39\lib\typing.py", line 815, in __getitem__
    _check_generic(self, params, self._nparams)
  File "C:\Python39\lib\typing.py", line 199, in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} parameters for {cls};"
TypeError: Too many parameters for typing.List; actual 2, expected 1
```

hasher-matcher-actioner/hmalib/common/config.py
Dcallies: L274 and L278 are different

hasher-matcher-actioner/hmalib/common/tests/config_test.py
```diff
@@ -0,0 +1,152 @@
...
+
+    TABLE_NAME = "test-HMAConfig"
+
+    @staticmethod
+    def mock_aws_credentials():
+        """Mocked AWS Credentials for moto"""
+        os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+        os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+        os.environ["AWS_SECURITY_TOKEN"] = "testing"
+        os.environ["AWS_SESSION_TOKEN"] = "testing"
```
schatten, line 26: Would it make sense to pass along the region here as AWS_DEFAULT_REGION? Rather than adding it to all resource getters?

schatten (approved):
As discussed offline, punting on 'Set[SomeType]' serialization/deserialization.

Looks good to ship once you fix the test cases.

hasher-matcher-actioner/hmalib/common/tests/config_test.py
Dcallies: Yea, this is what I ended up doing to fix

PR #492, opened 2021-04-14
Modularize matches api

schatten:
Summary
---------
Move `matches` functionality into a module of its own. Add a middleware for easy JSON wrangling. 

Also adds some git-blame-friendly and comment-rewrap-friendly vscode extensions to the devcontainer.

Test Plan
---------
Post refactoring, ran the API locally using `python -m ...`,

Screen Shot 2021-04-14 at 15 09 19

 took screenshots of both APIs being served on localhost.

Screen Shot 2021-04-14 at 15 10 44

Screen Shot 2021-04-14 at 15 10 52

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/devcontainer.json
+++ b/hasher-matcher-actioner/.devcontainer/devcontainer.json
@@ -19,6 +19,8 @@
 		"rvest.vs-code-prettier-eslint",
 		"ms-azuretools.vscode-docker",
 		"hashicorp.terraform",
+		"eamodio.gitlens",
+		"stkb.rewrap"
 	],
 
 	"mounts": [
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -12,7 +12,9 @@ from bottle import response, error
 from hmalib import metrics
 from hmalib.common.logging import get_logger
 from hmalib.common.s3_adapters import ThreatExchangeS3PDQAdapter, S3ThreatDataConfig
-from hmalib.models import PDQMatchRecord, PipelinePDQHashRecord
+from hmalib.models import PipelinePDQHashRecord
+
+from .matches import get_matches_api
 
 # Set to 10MB for /upload
 bottle.BaseRequest.MEMFILE_MAX = 10 * 1024 * 1024
@@ -104,28 +106,6 @@ def image(key=None):
     return bytes_
 
 
-@app.get("/matches")
-def matches():
-    """
-    matches API endpoint:
-    returns style { matches: [MatchesResult] }
-    """
-    results = get_matches()
-    logger.debug(results)
-    return {"matches": results}
-
-
-@app.get("/match/<key>")
-def match_details(key=None):
-    """
-    matche details API endpoint:
-    return format: match_details : [MatchDetailsResult]
-    """
-    results = get_match_details(key)
-    logger.debug(results)
-    return {"match_details": results}
-
-
 @app.get("/hash/<key>")
 def hashes(key=None):
     """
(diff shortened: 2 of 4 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -0,0 +1,149 @@
...
+from mypy_boto3_dynamodb.service_resource import Table
+import typing as t
+
+from hmalib.models import PDQMatchRecord
+from .middleware import jsoninator, JSONifiable
+
+
+# TODO: Remove dependency on image folder key, use a common interface for
+# s3 and non-s3 HMA requests.
+IMAGE_FOLDER_KEY = os.environ["IMAGE_FOLDER_KEY"]
```
BarrettOlson, line 15: I think for now at least we want to keep all uses of os.environ are in the same module as `lambda_handler`

schatten: Agree, that's what the TODO is for. 

I plan to remove this soon. We'll need to do this for v0 when API uploads might make it necessary to **not** think about IMAGE_FOLDERs.

BarrettOlson (commented):
Will this mean we need to update: https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/webapp/src/Api.jsx#L35-L41 ?

schatten:
> Will this mean we need to update: https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/webapp/src/Api.jsx#L35-L41 ?

Yes. Absolutely! I'm working on that now. Should I put that into the same PR? 

Could do. Give me a few more minutes. :)

schatten:
> > Will this mean we need to update: https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/webapp/src/Api.jsx#L35-L41 ?
> 
> Yes. Absolutely! I'm working on that now. Should I put that into the same PR?
> 
> Could do. Give me a few more minutes. :)

I didn't realize how deep I was into extracting pieces from the UI. 😅

Here's what I propose:
- **Unless you are RC, approve this diff. I'll not merge until the UI diff is out and approved.**

hasher-matcher-actioner/.devcontainer/devcontainer.json
```diff
@@ -19,6 +19,8 @@
 		"rvest.vs-code-prettier-eslint",
 		"ms-azuretools.vscode-docker",
 		"hashicorp.terraform",
+		"eamodio.gitlens",
+		"stkb.rewrap"
```
BarrettOlson, line 23: Will this make our formatting non-standard for those using/not using the plugin?

schatten: No. This is only for comment wrapping. I could not find an equivalent black command for docstrings.

This is also not enabled by default. You have to press keys [opt + q] to get this to format the comment..

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
BarrettOlson: It's really easy to break the whole API silently atm like had to be fixed in #482 I feel strongly about being consistent with this. Can you add this (or a better CI test on os.environ) to the set of work before merge like the UI diff?

schatten: Sure. Let me cook something up.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -12,7 +12,9 @@
 from hmalib import metrics
 from hmalib.common.logging import get_logger
 from hmalib.common.s3_adapters import ThreatExchangeS3PDQAdapter, S3ThreatDataConfig
-from hmalib.models import PDQMatchRecord, PipelinePDQHashRecord
```
BarrettOlson, line 15: Heads up: be careful with your merge, I think this import is needed for recent already merged changes to `get_dashboard_matches`

hasher-matcher-actioner/.devcontainer/devcontainer.json
BarrettOlson: Okay, so an optionally layer for comments on top of prettier/black? works for me.

BarrettOlson (approved):
Approve with merge contingent on UI diff etc.
Thanks!

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -405,3 +311,12 @@ def get_hash_count() -> t.Dict[str, int]:
     pdq_data_files = pdq_storage.load_data()
 
     return {file_name: len(rows) for file_name, rows in pdq_data_files.items()}
+
+
+app.mount(
+    "/matches/v1/", get_matches_api(dynamodb_table=dynamodb.Table(DYNAMODB_TABLE))
```
eriksendc, line 317: Good with this PR with `/matches/` here. Discussed in room.

eriksendc (approved):

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -314,7 +314,10 @@ def get_hash_count() -> t.Dict[str, int]:
 
 
 app.mount(
-    "/matches/v1/", get_matches_api(dynamodb_table=dynamodb.Table(DYNAMODB_TABLE))
+    "/matches/",
+    get_matches_api(
+        dynamodb_table=dynamodb.Table(DYNAMODB_TABLE), image_folder_key=IMAGE_FOLDER_KEY
```
BarrettOlson, line 319: awesome, thanks! (even if we have to refactor it again later)

eriksendc (approved):

schatten:
Included in the rebased #500

PR #496, opened 2021-04-15
[hma] Create signal metadata db obj and surface on UI

BarrettOlson:
Summary
---------

The goal of this PR is getting us closer to closing the loop on reactions/opinions by building off of the existing structure in [descriptor.py](https://github.com/facebook/ThreatExchange/blob/master/python-threatexchange/threatexchange/descriptor.py). This takes the metadata from the index for matches and creates an object that can be updated later by a syncer/fetcher (or actioner) that can also be surfaced on the UI.

Feel free to review/provide feedback at the commit level as the 3 commits of this PR could probably be there own diffs but testing them really only works all together. 

Commits:
1) creates the new metadata object class + unit test
2) update pdq_matcher to create these object
3) updates the match details API and page to surface this metadata

What's next:
a) have the syncer/fetcher also update this metadata object
b) add the ability to send a post request from the details page to propagate an update the 'opinion' column

Test Plan
---------
By commits:
1) Added to test_models.py
`pytest hmalib`
2) After deployments, run `make dev_upload_test_data` and confirm the new object is in the dynamodb 
3) Same as two but update the dynamodb to have more elements to make sure nested cases and all special case tags work as expected. (see screenshot attached)

Screen Shot 2021-04-15 at 5 06 04 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -9,6 +9,7 @@ import datetime
 import typing as t
 from apig_wsgi import make_lambda_handler
 from bottle import response, error
+from enum import Enum
 
 from hmalib import metrics
 from hmalib.common.logging import get_logger
@@ -243,7 +244,7 @@ def get_matches() -> t.List[MatchesResult]:
 
 
 class MatchDetailsMetadata(t.TypedDict):
-    pg: str
+    dataset: str
     tags: t.List[str]
     opinion: str
 
@@ -289,7 +290,7 @@ def get_signal_details(
     table = dynamodb.Table(DYNAMODB_TABLE)
     return [
         MatchDetailsMetadata(
-            pg=metadata.pg_id,
+            dataset=metadata.ds_id,
             tags=[
                 tag
                 for tag in metadata.tags
@@ -300,7 +301,7 @@ def get_signal_details(
                     ThreatDescriptor.DISPUTED,
                 ]
             ],
-            opinion=get_opinion_from_tags(metadata.tags),
+            opinion=get_opinion_from_tags(metadata.tags).value,
         )
         for metadata in PDQSignalMetadata.get_from_signal(
             table, signal_id, signal_source
@@ -308,15 +309,22 @@ def get_signal_details(
     ]
 
 
-def get_opinion_from_tags(tags: t.List[str]):
+class OpinionString(Enum):
+    TP = "True Positive"
+    FP = "False Positive"
+    DISPUTED = "Unknown (Disputed)"
+    UNKNOWN = "Unknown"
+
+
+def get_opinion_from_tags(tags: t.List[str]) -> OpinionString:
     # see python-threatexchange descriptor.py for origins
     if ThreatDescriptor.TRUE_POSITIVE in tags:
-        return "True Positive"
+        return OpinionString.TP
     if ThreatDescriptor.FALSE_POSITIVE in tags:
-        return "False Positive"
+        return OpinionString.FP
     if ThreatDescriptor.DISPUTED in tags:
-        return "Unknown (Disputed)"
-    return "Unknown"
+        return OpinionString.DISPUTED
+    return OpinionString.UNKNOWN
 
 
 class HashResult(t.TypedDict):
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -13,7 +13,13 @@ from bottle import response, error
 from hmalib import metrics
 from hmalib.common.logging import get_logger
 from hmalib.common.s3_adapters import ThreatExchangeS3PDQAdapter, S3ThreatDataConfig
-from hmalib.models import PDQMatchRecord, PipelinePDQHashRecord, PDQRecordBase
+from hmalib.models import (
+    PDQMatchRecord,
+    PipelinePDQHashRecord,
+    PDQRecordBase,
+    PDQSignalMetadata,
+)
+from threatexchange.descriptor import ThreatDescriptor
 
 # Set to 10MB for /upload
 bottle.BaseRequest.MEMFILE_MAX = 10 * 1024 * 1024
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/hmalib/models.py
```diff
@@ -48,6 +56,90 @@ def from_sns_message(cls, message: str) -> "SNSMessage":
         raise NotImplementedError
 
 
+@dataclass
+class SignalMetadataBase(DynamoDBItem):
+    """
+    Base for signal metadata.
```
schatten, line 62: blocking: Would be super-hard to change prefixes once a database is built. Can we use source agnostic nomenclature NOW?

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -48,6 +56,90 @@ def from_sns_message(cls, message: str) -> "SNSMessage":
...
+    'PG' refers to privacy group which for the time being is
+    quivalent to collab (and in the long term could map to bank)
+    """
+
+    PG_PREFIX = "pg#"
+
+    signal_id: t.Union[str, int]
+    pg_id: str
+    updated_at: datetime.datetime  # ISO-8601 formatted
```
schatten, line 71: nit: it's either datetime or a formatted string. Can't be both.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -48,6 +56,90 @@ def from_sns_message(cls, message: str) -> "SNSMessage":
...
+    signal_source: str
+    signal_hash: str  # duplicated field with PDQMatchRecord having both for now to help with debuging/testing
+    tags: t.List[t.Any] = field(default_factory=list)
```
schatten, line 74: why Any?

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -48,6 +56,90 @@ def from_sns_message(cls, message: str) -> "SNSMessage":
...
+    """
+
+    SIGNAL_TYPE = "pdq"
+
+    def to_dynamodb_item(self) -> dict:
+        return {
+            "PK": self.get_dynamodb_signal_key(self.signal_source, self.signal_id),
+            "SK": self.get_dynamodb_pg_key(self.pg_id),
+            "SignalHash": self.signal_hash,
+            "SignalSource": self.signal_source,
```
schatten, line 100: blocking: Can you give examples of what SignalSource would be?

I'm concerned about the pg_key not being required to be globally unique.

What's preventing any other Source from having the same pg_id as another Source?

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -48,6 +56,90 @@ def from_sns_message(cls, message: str) -> "SNSMessage":
...
+            KeyConditionExpression=Key("PK").eq(
+                cls.get_dynamodb_signal_key(signal_source, signal_id)
+            )
+            & Key("SK").begins_with(cls.PG_PREFIX),
+            ProjectionExpression="PK, ContentHash, UpdatedAt, SK, SignalSource, SignalHash, Tags",
+            FilterExpression=Attr("HashType").eq(cls.SIGNAL_TYPE),
+        ).get("Items", [])
+        return cls._result_items_to_metadata(items)
+
+    @staticmethod
```
schatten, line 124: This gets confusing. If a class's method returns an object of the same type, perhaps we could standardize using @classmethod?

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -48,6 +56,90 @@ def from_sns_message(cls, message: str) -> "SNSMessage":
...
+    def _result_items_to_metadata(
+        items: t.List[t.Dict],
+    ) -> t.List["PDQSignalMetadata"]:
+        return [
+            PDQSignalMetadata(
```
schatten, line 129: nit: especially because we are inheriting fields from another dataclass, would prefer keyword arguments here.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -261,31 +266,59 @@ def get_match_details(content_id: str) -> t.List[MatchDetailsResult]:
...
+            ],
+            opinion=get_opinion_from_tags(metadata.tags),
+        )
+        for metadata in PDQSignalMetadata.get_from_signal(
+            table, signal_id, signal_source
+        )
+    ]
+
+
+def get_opinion_from_tags(tags: t.List[str]):
```
schatten, line 311: Can this return an enum instead? Or would that be too threat-exchange-y?

Also, are we sure all opinions are mutually exclusive? Can something be FALSE_POSITIVE and DISPUTED?

schatten (commented):
Overall, happy with the approach. I have a few questions. You can answer them inline. If you were to make changes, would be happy to review the delta commit.

Have two blocking questions. Would stamp once you answer those.

hasher-matcher-actioner/hmalib/models.py
BarrettOlson: They could, but I think that is okay because PK for this object incorporates the SignalSource as well. 'te' is what we have for now (i'll add a comment). It functions as a secondary prefix for PK atm (and SK for MatchRecord).

hasher-matcher-actioner/hmalib/models.py
BarrettOlson: sure are you okay with 'ds#' for dataset or have a specific one in mind?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
BarrettOlson: Sure, you want an enum, I'll get you an enum :) 
(long term I want to refactor pytx but didn't want to be blocked by a pypi push/release right now)

Assuming mutually exclusive based on all the `elif` of  https://github.com/facebook/ThreatExchange/blob/master/python-threatexchange/threatexchange/descriptor.py#L78-L92
happy to program it more defensively though (the js just becomes a little rougher to handle)

schatten (approved):
This looks great! Go ahead!
--
Cheers!

BarrettOlson:
Thanks @schatten! Going to merge but @eriksendc feel free to rip apart the jsx and I'll fix in a future PR :)

PR #497, opened 2021-04-16
Make actioner config

jeberl:
Summary
---------

Use @Dcallies 's config framework to make actioner configs. One additional level of complexity is that we have actioner types that have different parameter types but all must be unique by label (one label leads to a single config. See doc blocks for how to use the Actioner Config

Test Plan
---------

Added new test cases to config tests since I couldn't figure out how to get mocking to work in `actioner_tests.py`

mypy and pytest


hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -43,49 +45,115 @@ class ActionRule:
...
-        kwargs = {"url": self.url, "data": match_message.to_sns_message()}
-        self.request_type.call(**kwargs)
+class ActionPerformerConfig(config.HMAConfig):
+    """
+    An adapter to use the HMAConfig Class to store differnet types of
+    ActionPerformer configs. Since ActionPerformer configs are all unique by label but may
+    be of different types, we store the name of the concrete type and the
+    attributes of that type in AWS.
+
+    DO NOT CALL config.HMAConfig FUNCTIONS DIRECTLY
```
Dcallies, line 118: What does this mean?

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -43,49 +45,115 @@ class ActionRule:
...
-
-    request_type: t.Type[HTTPRequest]
-    url: TUrl
-
-    def perform_action(self, match_message: MatchMessage) -> None:
-        kwargs = {"url": self.url, "data": match_message.to_sns_message()}
-        self.request_type.call(**kwargs)
+class ActionPerformerConfig(config.HMAConfig):
+    """
+    An adapter to use the HMAConfig Class to store differnet types of
```
Dcallies, line 113: speeling: different

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -43,49 +45,115 @@ class ActionRule:
...
+    concrete_type: str
+    concrete_type_attrs: t.Dict[str, str]
+
+    @classmethod
+    def get_performer(cls, action_label: ActionLabel) -> t.Optional["ActionPerformer"]:
+        """
+        Load from AWS the correct actioner config as the correct concrete class if it exists
+        """
+        name = action_label.value
+        performer_config: t.Optional[ActionPerformerConfig] = super().get(name)
```
Dcallies, line 134: super.get() should be the same as self.get()

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -43,49 +45,115 @@ class ActionRule:
...
+            }
+            return subclasses_classes[performer_config.concrete_type](
+                action_label=action_label, **performer_config.concrete_type_attrs
+            )
+        return None
+
+    @classmethod
+    def update_performer(cls, action_performer: ActionPerformer) -> None:
+        """
+        Use this function to update the config in AWS instead of config.update_config directly
```
Dcallies, line 147: Hmm, this somewhat defeats the purpose of having a standard way of updating the configs if the first ever instance doesn't use it.

Dcallies (commented):
I think there's probably a solution that doesn't require the overrides to the config stuff, but I need to sync with you more.

My initial question was why there wasn't different config classes for each performer, since they each have different types. get_config_class, get, get_all are on the class so that you can fiddle with them if it makes sense.

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -43,49 +45,115 @@ class ActionRule:
...
+    @classmethod
+    def get_performer(cls, action_label: ActionLabel) -> t.Optional["ActionPerformer"]:
+        """
+        Load from AWS the correct actioner config as the correct concrete class if it exists
+        """
+        name = action_label.value
+        performer_config: t.Optional[ActionPerformerConfig] = super().get(name)
+        if performer_config:
+            subclasses_classes = {
+                cls.__name__: cls for cls in get_all_subclasses_rec({ActionPerformer})
```
eriksendc, line 137: I see how this works (I think... it's non-trivial to follow the reflection, but I think I got it). The pattern of projecting all subclasses into the dictionary to then dereference to get one seems a bit on the greedy side. In practice I'm not sure the number of actions will ever be so large as to cause any kind of perf / memory issue, but the pattern seems like one that we could find a way around.

eriksendc (approved):

jeberl:
@Dcallies Happy to sync offline. To answer your question about why can't we just make different config classes for each action performer, we want to performers to be unique per label but of different types. So we get an ActionLabel like `EnqueForReview` and we need to find which performer it's associated with. If each performer type is a different config we'd have to try each config type separately.

hasher-matcher-actioner/hmalib/common/actioner_models.py
jeberl: Saying not to call the underlying class directly. I think should be
```
DO NOT CALL config FUNCTIONS DIRECTLY
```

Dcallies (changes requested):
Blocking: Want to make sure I understand before we proceed forward, my current impression is that different types = different configs, and that would be beneficial, and I have some solutions that can solve ease of use as well.

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -43,49 +45,115 @@ class ActionRule:
...
+            }
+            return subclasses_classes[performer_config.concrete_type](
+                action_label=action_label, **performer_config.concrete_type_attrs
```
Dcallies, line 140: Re: offline conversation, what converts concrete_type_attrs to the appropriate type rather than str=> str?

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -233,7 +233,10 @@ def _aws_field_to_py(in_type: t.Type[T], aws_field: t.Any) -> T:
     # Another option would be adding a new class that adds methods to convert
     # to an AWS-friendly struct and back
     if origin is dict and args[0] is str:  # M
-        return {k: _aws_field_to_py(args[1], v) for k, v in aws_field.items()}  # type: ignore # mypy/issues/10003
+        if args[1] is not t.Any:
```
Dcallies, line 236: Hmm, This unfortunately breaks typing guarantees that the things you are putting in these maps are serializable. (The purpose of the exception on 241). 

I am aware of a solution that I think solves both of our problems, let me sync with you on next steps.

Dcallies (approved):
David and Jesse discussed offline, David will refactor to make it fit framework afterwards.

PR #498, opened 2021-04-16
Search matches api

schatten:
WARNING: Stacked diff: only review [67f4a413](https://github.com/facebook/ThreatExchange/pull/498/commits/67f4a4138dc3cfb791843d25aaa555762f7c8b7e)
---

Summary
---------
Allow query parameter based filtering for matches API. 

Either 'content_q=<content_id>' or ('signal_q=<signal_id>&signal_source=<source>')

Test Plan
---------
Ran locally and hit the APIs on chrome.

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
@@ -118,7 +118,19 @@ def get_matches_api(dynamodb_table: Table, image_folder_key: str) -> bottle.Bott
         """
         Returns all, or a filtered list of matches.
         """
-        records = PDQMatchRecord.get_from_time_range(dynamodb_table)
+        signal_q = bottle.request.query.signal_q or None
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -118,7 +118,19 @@ def matches() -> MatchSummariesResponse:
         """
         Returns all, or a filtered list of matches.
         """
-        records = PDQMatchRecord.get_from_time_range(dynamodb_table)
+        signal_q = bottle.request.query.signal_q or None
```
BarrettOlson, line 121: Looks good, my only concern is documentation, the endpoint is now complex enough the doc string should be more detailed and ideally include example usages (if not an API README)

BarrettOlson (approved):
Stamped with the same caveats as #492 :)

eriksendc (approved):

schatten:
Included in the rebased #500

PR #500, opened 2021-04-16
Support search by content / signal in "matches" page

schatten:
Rebases and abandons #492 #498. Includes those changes in a single commit. Only review [9ed70c32](https://github.com/facebook/ThreatExchange/pull/500/commits/9ed70c32efb4276042ad531db52c3db39091fa18) and [4ba9edc8](https://github.com/facebook/ThreatExchange/pull/500/commits/4ba9edc8e926154a67bbfd7620d4e275a3377167)
---

Summary
---------
In python land:
* Moves matches API and representation into their own module, uses `bottle` [mounts](https://bottlepy.org/docs/dev/tutorial.html#plugins-and-sub-applications) to have a 'sub' API.
* supports filtering matches by content id and signal source, signal id.

In Javascript land:
* Moves Matches -> pages/MatchIndex
* Creates a flex box based layout for showing all match results, the list is independently scrollable and on selecting any row, shows the relevant content match pane to its side.
* Abandons full dates in favour of relative times. eg. 2m ago instead of "Tue, 20 April, 2020 21:33:00"
* Have a box to select a signal or content id to show results from
* Link to contentId or signalId filtered pages

Test Plan
---------
The python code was tested locally using `$ python -m hmalib.lambdas.api.api_root` with the relevant env vars.

Screen Shot 2021-04-14 at 15 09 19

The following APIs were exercised and found to be working.
* `/matches/`
* `/matches/?signal_q=<foo>&signal_source=<bar>`
* `/matches/match/<contentId>/`

The code was deployed to AWS lamdbas using `$ make upload_docker`

A local copy of the UI was tested using npm start. TestData was added using P408767257.

The pages were found to be working and interactive.

The following actions were performed on the UI:
* Click a match to see it in the right pane.
  * the 'View matches for signal' link worked. It showed a reduced set of matches.
  * the 'View matches of this content` link worked. It showed a reduced set of matches.
  * The image is blurred by default. Only visible when hovered.
* The top-right signal / content filter worked.
  * For signals, I had to provide '<signal_source>|<signal_id>' to get it to work. This might change in the future.


hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -0,0 +1,81 @@
...
+  }, [contentId]);
+
+  useEffect(() => {
+    fetchImage(contentId)
+      .then(result => {
+        setImage(URL.createObjectURL(result));
+      })
+      .catch(() => {
+        setImage(
+          'https://image.shutterstock.com/image-vector/mission-failed-text-on-red-260nw-1633188751.jpg',
```
schatten, line 33: This will need to be fixed. Failed image placeholder can be better.

hasher-matcher-actioner/webapp/src/pages/MatchIndex.jsx
```diff
@@ -0,0 +1,251 @@
...
+    // .catch(err => console.log(err));
+  }, [searchInAttribute, searchQuery]);
+
+  return (
+    <>
+      <Spinner hidden={matchesData !== null} animation="border" role="status">
+        <span className="sr-only">Loading...</span>
+      </Spinner>
+      <Collapse in={matchesData !== null}>
+        <div className="row mt-3">
```
eriksendc, line 185: use `<Container>`, `<Row>` and `<Column>` here (and anywhere else in new code).

hasher-matcher-actioner/webapp/src/pages/MatchIndex.jsx
```diff
@@ -0,0 +1,251 @@
...
+import {
+  fetchAllMatches,
+  fetchMatchesFromContent,
+  fetchMatchesFromSignal,
+} from '../Api';
+import {timeAgo} from '../utils/DateTimeUtils';
+import ContentMatchPane from '../components/ContentMatchPane';
+import {useQuery} from '../utils/QueryParams';
+
+export default function MatchIndex() {
```
eriksendc, line 30: I think I get the divination of `MatchIndex`... it's like index.html, right? Take with a grain of salt given my time with React is not huge, but in WWW I think you often see something that matches the route or the main "entity" of the "page" (e.g. `<Matches>`). I don't ever remember seeing the word Index involved with a "root" component for a page.

eriksendc (changes requested):

hasher-matcher-actioner/webapp/src/App.jsx
```diff
@@ -3,16 +3,15 @@
  */
 
 import React from 'react';
-import {BrowserRouter as Router, Route} from 'react-router-dom';
-import {AnimatedSwitch} from 'react-router-transition';
```
BarrettOlson, line 7: Definitely fine to remove but it looks these change this broke/removed the padding/grid on some of the subpages

hasher-matcher-actioner/webapp/src/pages/MatchIndex.jsx
```diff
@@ -0,0 +1,251 @@
...
+  ] = useState({
+    contentId: undefined,
+    signalId: undefined,
+    signalSource: undefined,
+  });
+
+  return (
+    <div className="d-flex flex-column justify-content-start align-items-stretch h-100 w-100">
+      <div className="flex-grow-0">
+        <Container className="bg-dark text-light" fluid>
```
BarrettOlson, line 53: re: comment in App.jsx, I think this Container is what now gives the page the padding and why it is missing from the others. I think we should try and have consistent header.
(also super nit: but the dark gray or bg-dark clashes with the gray of selected tab in the sidebar)

hasher-matcher-actioner/webapp/src/pages/MatchIndex.jsx
```diff
@@ -0,0 +1,251 @@
...
+  }, [searchInAttribute, searchQuery]);
+
+  const history = useHistory();
+  const loadSearchPage = e => {
+    history.push(`?${localSearchInAttribute}=${localSearchQuery}`);
+    e.preventDefault(); // Prevent form submit
+  };
+
+  return (
+    <Form>
```
BarrettOlson, line 119: nit: I think it would be good to have a more intuitive 'clear' filter method then having to delete the text.

hasher-matcher-actioner/webapp/src/pages/MatchIndex.jsx
```diff
@@ -0,0 +1,251 @@
...
+                        onSelect({
+                          contentId: match.content_id,
+                          signalId: match.signal_id,
+                          signalSource: match.signal_source,
+                          [e.target.name]: e.target.value,
+                        })
+                      }
+                      key={`${match.signal_source}_${match.signal_id}_${match.content_id}`}>
+                      <td className="align-middle">{match.content_id}</td>
+                      <td className="align-middle">{match.signal_source}</td>
```
BarrettOlson, line 214: I think we've talked about this and know we are currently refining vernacular but signal_source != dataset

Can you add a note to change later if you don't want to scope that change of querying the other db object to get datasets in this diff? Thanks!

hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -0,0 +1,81 @@
...
+              View Matches for Signal
+            </Link>
+          </Col>
+        </Row>
+        <Row>
+          <Col md={3} className="p-4">
+            Content
+          </Col>
+          <Col className="p-4">
+            <Link to={`?contentId=images/${contentId}`}>
```
BarrettOlson, line 61: As it is currently structured there is no way to get to the "MatchDetails" (ContentDetails) subpage, can you add a 2nd link/button here for this? 
Otherwise the only way to get there is manually setting the URL or right after an upload (which is worse than not at all).

BarrettOlson (changes requested):
Overall really like the changes and the ability to filter the match results is definitely nice.
However the changes introduced here contrast a little with the other existing pages and don't flow as naturally as preferred. 

I'm okay with most of them being marked as bold TODOs for the purposes of getting merged but at a minimum I'd like for all pages to stay navigable. Thanks!

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -0,0 +1,192 @@
...
+            opinion=get_opinion_from_tags(metadata.tags).value,
+        )
+        for metadata in PDQSignalMetadata.get_from_signal(
+            table, signal_id, signal_source
+        )
+    ]
+
+
+class OpinionString(Enum):
+    TP = "True Positive"
```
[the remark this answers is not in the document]
schatten: This is actually a merge resolution from @BarrettOlson's changes. @BarrettOlson thoughts? Can you drive this input in a separate PR, perhaps? 

@eriksendc this is already on master. See [here](https://github.com/facebook/ThreatExchange/blob/0d07a800bbf25d8541f40b828e2dfd377395af9b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py#L312-L316).

hasher-matcher-actioner/webapp/src/App.jsx
schatten: That's intentional. I want sub-pages to have the capability to use real estate as they see fit. We can provide common layouts as compose helpers.

eg. Matches page needs to use _all_ the available space. Settings probably does not.

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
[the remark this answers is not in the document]
BarrettOlson: Sure, honestly this should be codified in python-threatexchange imo anyway, could you add a `# TODO naming` and I'll fix it in my next PR? (fine with `TRUE_POSITIVE` etc.)

hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -0,0 +1,81 @@
...
+          'https://image.shutterstock.com/image-vector/mission-failed-text-on-red-260nw-1633188751.jpg',
+        );
+      });
+  }, [contentId]);
+
+  return (
+    (hashDetails === null && <p>Loading...</p>) || (
+      <Container>
+        <Row>
+          <Col className="p-4">
```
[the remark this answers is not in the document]
schatten: You are correct. This portion, the ContentMatchPane is still a WIP and I'll make it a point to fix this when doing the final cut.

hasher-matcher-actioner/webapp/src/pages/MatchIndex.jsx
schatten: Yeah. I explained the reasoning for the padding as a reply to your comment above.

re: the dark gray. Totally up for grabs. We can try a few classes and see what works best. Leaving as-is for now because it is a non-functional aspect.

eriksendc (approved):

hasher-matcher-actioner/webapp/src/App.jsx
BarrettOlson: Okay, someone should own updating this soon for all subpages but not blocking.

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
```diff
@@ -0,0 +1,278 @@
...
+  fetchMatchesFromSignal,
+} from '../Api';
+import {timeAgo} from '../utils/DateTimeUtils';
+import ContentMatchPane from '../components/ContentMatchPane';
+import {useQuery} from '../utils/QueryParams';
+
+export default function Matches() {
+  const query = useQuery();
+  const searchQuery = query.get('contentId') || query.get('signalId');
+  let searchInAttribute;
```
BarrettOlson, line 33: not blocking: the state management of this variable confused me a little because of the ways it can be updated and assigned to local but can't think of a cleaner way at first glance.

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
```diff
@@ -0,0 +1,278 @@
...
+      </Spinner>
+      <Collapse in={matchesData !== null}>
+        <Row className="mt-3">
+          <Col>
+            <table className="table table-hover small">
+              <thead>
+                {/* TODO: Undecided nomenclature */}
+                <tr>
+                  <th>Content Id</th>
+                  <th>Matched in Dataset</th>
```
BarrettOlson, line 218: Can we change this to Source for now so it matches the other pages (even if we want to change it later when we're more decided on nomenclature etc)?

BarrettOlson (approved):
Accepted; rather get the API changes in sooner rather than later. We should try and get the other pages updated sooner rather than later imo (padding changes etc).

Made a few comments inline nothing blocking. Thanks!

PR #501, opened 2021-04-16
Actioner evaluate one rule

eriksendc:
Summary
---------

This PR introduces converting the TE concept of tags into "classifications", and wires them up to be sent through from matching to evaluation, and on to action execution.

Test Plan
---------

1. `make upload_docker`
2. `make dev_upload_test_data`
3. wait for hashing and matching to occur
4. validate CloudWatch logs for matcher shows classifications in MatchMessage
5. validate CloudWatch logs for action evaluator shows classifications in MatchMessage
6. validate webhook call contains classifications from MatchMessage

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
@@ -151,12 +151,19 @@ def get_threat_exchange_reaction_labels(
     """
     return [ThreatExchangeReactionLabel("SAW_THIS_TOO")]
 
-def react_to_threat_exchange(match_message: MatchMessage, threat_exchange_reaction_label: ThreatExchangeReactionLabel) -> None:
+
+def react_to_threat_exchange(
+    match_message: MatchMessage,
+    threat_exchange_reaction_label: ThreatExchangeReactionLabel,
+) -> None:
     """
     TODO implement
     Puts a ThreatExchangeReactionMessage on the queue to be processed asynchronously
     """
-    logger.info(f"The contents of a ThreatExchangeReactionMessage will contain match_message = {match_message} and threat_exchange_reaction_label = {threat_exchange_reaction_label}")
+    logger.info(
+        f"The contents of a ThreatExchangeReactionMessage will contain match_message = {match_message} and threat_exchange_reaction_label = {threat_exchange_reaction_label}"
+    )
+
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -149,6 +152,19 @@ def get_threat_exchange_reaction_labels(
...
+def react_to_threat_exchange(
+    match_message: MatchMessage,
+    threat_exchange_reaction_label: ThreatExchangeReactionLabel,
+) -> None:
+    """
+    TODO implement
+    Puts a ThreatExchangeReactionMessage on the queue to be processed asynchronously
+    """
+    logger.info(
+        f"The contents of a ThreatExchangeReactionMessage will contain match_message = {match_message} and threat_exchange_reaction_label = {threat_exchange_reaction_label}"
```
Dcallies, line 164: Nit: Split this line so it fits in ~800 characters, in py the easiest way is that adjacent strings are concatenated
```
  logger.info(
    "some line or another"
    "another line"
    f"an fstring {asdf}",  # comma goes here
  )

  # or
  logger.info(
    (
      "Another approach"
      "is to use wrapping parenthesis"
      "however, it's easy to accidentally add commas"
      "and turn it into a tuple"
    )
  )
```

Dcallies (commented):
Partial readthrough

schatten (approved):
LGTM. This is in line with what we have discussed. 👍

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -161,8 +161,10 @@ def react_to_threat_exchange(
     Puts a ThreatExchangeReactionMessage on the queue to be processed asynchronously
     """
     logger.info(
-        f"The contents of a ThreatExchangeReactionMessage will contain match_message = {match_message} and threat_exchange_reaction_label = {threat_exchange_reaction_label}"
+        "The contents of a ThreatExchangeReactionMessage will contain the following:"
     )
+    logger.ingo("match_message = %s", match_message)
+    logger.info("threat_exchange_reaction_label = %s", threat_exchange_reaction_label)
```
Dcallies: A potential downside of this approach is that if your log format adds a prefix (such as timestamp), it may make it hard to see these are all part of one message, since it will be interleveled with logs from other threads.

PR #503, opened 2021-04-16
[hma] Add post_apply method to update signal data of matches from fetcher

BarrettOlson:
Summary
---------

Draft for how a ~syncer/updater for the dynamodb items could work in the fetcher.

Test Plan
---------

So far tested manually, more thorough test plan pending feedback on the overall approach

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
@@ -26,6 +26,7 @@ import boto3
 from botocore.errorfactory import ClientError
 from hmalib.aws_secrets import AWSSecrets
 from hmalib.common.logging import get_logger
+from hmalib.common.s3_adapters import S3ThreatDataConfig
 from hmalib.models import PDQSignalMetadata
 from threatexchange import threat_updates as tu
 from threatexchange.api import ThreatExchangeAPI
@@ -52,21 +53,14 @@ class FetcherConfig:
     data_store_table: str
 
     @classmethod
-    @lru_cache(maxsize=1)  # probably overkill, but at least it's consistent
+    @lru_cache(maxsize=None)  # probably overkill, but at least it's consistent
     def get(cls):
         # These defaults are naive but can be updated for testing purposes.
         return cls(
-            s3_bucket=os.environ.get(
-                "THREAT_EXCHANGE_DATA_BUCKET_NAME",
-                "DEFAULT_BUCKET",
-            ),
-            s3_te_data_folder=os.environ.get(
-                "THREAT_EXCHANGE_DATA_FOLDER", "threat_exchange_data/"
-            ),
-            collab_config_table=os.environ.get(
-                "THREAT_EXCHANGE_CONFIG_DYNAMODB", "default-ThreatExchangeConfig"
-            ),
-            data_store_table=os.environ.get("DYNAMODB_TABLE", "default-HMADataStore"),
+            s3_bucket=os.environ["THREAT_EXCHANGE_DATA_BUCKET_NAME"],
+            s3_te_data_folder=os.environ["THREAT_EXCHANGE_DATA_FOLDER"],
+            collab_config_table=os.environ["THREAT_EXCHANGE_CONFIG_DYNAMODB"],
+            data_store_table=os.environ["DYNAMODB_DATASTORE_TABLE"],
         )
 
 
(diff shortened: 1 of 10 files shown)


hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -48,14 +49,24 @@ class FetcherConfig:
     s3_bucket: str
     s3_te_data_folder: str
     collab_config_table: str
+    data_store_table: str
 
     @classmethod
     @lru_cache(maxsize=1)  # probably overkill, but at least it's consistent
```
schatten, line 55: For Simple memoizations, prefer maxsize=None, it provides for a much simpler wrapper. 

https://github.com/python/cpython/blob/master/Lib/functools.py#L549-L564

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -48,14 +49,24 @@ class FetcherConfig:
...
     def get(cls):
+        # These defaults are naive but can be updated for testing purposes.
         return cls(
-            s3_bucket=os.environ["THREAT_EXCHANGE_DATA_BUCKET_NAME"],
-            s3_te_data_folder=os.environ["THREAT_EXCHANGE_DATA_FOLDER"],
-            collab_config_table=os.environ["THREAT_EXCHANGE_CONFIG_DYNAMODB"],
+            s3_bucket=os.environ.get(
+                "THREAT_EXCHANGE_DATA_BUCKET_NAME",
+                "DEFAULT_BUCKET",
```
schatten, line 61: Is this a magic value?

Shouldn't we stall execution if this value is not provided?

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -269,10 +290,35 @@ def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
...
+
+        for update in updated.values():
+            row = update.as_csv_row()
+            # example row format: ('<signal>', '<id>', '<time added>', '<tag1 tags2>')
+            # e.g ('096a6f9...064f', 1234567891234567, '2020-07-31T18:47:45+0000', 'true_positive hma_test')
+            PDQSignalMetadata(
+                signal_id=int(row[1]),
+                ds_id=privacy_group,
+                updated_at=datetime.now(),
+                signal_source="te",
```
schatten, line 317: Can we externalize this as a constant? It's getting reused in places and could become a source of errors.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -269,10 +290,35 @@ def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
...
+                signal_hash=row[0],
+                tags=row[3].split(" ") if row[3] else [],
+            ).write_to_table_if_already_exists(table)
```
schatten, line 320: This **feels** like a "patch" update and yet we are doing a PUT.

If there are other systems (there aren't any now), that update the PDQSignalMetadata object in any way, their updates will be lost if the signal were to be refetched.

Can we do this using [UpdateItem](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html) instead? And only update the bare minimum? (updated_at, tags)

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -23,6 +24,22 @@ class DynamoDBItem:
     def write_to_table(self, table: Table):
         table.put_item(Item=self.to_dynamodb_item())
 
+    def write_to_table_if_already_exists(self, table: Table) -> bool:
+        """
+        Only write to table if the objects with PK and SK already exist.
+        returns true if object existed and therefore put was successful
+        """
+        try:
+            table.put_item(
```
schatten, line 33: See note about using UpdateItem instead.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -269,10 +290,35 @@ def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
...
+        table = dynamodb.Table(self.data_store_table)
+
+        for update in updated.values():
+            row = update.as_csv_row()
+            # example row format: ('<signal>', '<id>', '<time added>', '<tag1 tags2>')
+            # e.g ('096a6f9...064f', 1234567891234567, '2020-07-31T18:47:45+0000', 'true_positive hma_test')
+            PDQSignalMetadata(
+                signal_id=int(row[1]),
+                ds_id=privacy_group,
+                updated_at=datetime.now(),
```
schatten, line 316: Why not use time_added?

schatten (changes requested):
Have questions about why PutItem is being used in what's clearly and UpdateItem path. 

Put does a replace and that will be destructive.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: Not my decorator but happy to take your word for it and make the change.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: because that means something completely different and is likely very source dependent. If you think it is worth propagating I can add it as its own field.

schatten: I thought that's the same. My bad. Thanks for clarifying.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: the fetcher can be used locally if the values are overridden. I'd prefer having a value here for ease of debugging and quickly changing/testing. Are you alright with `None` or want to me to get rid of the `.get` calls all together?

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: I wrote it this way to avoid the "What is the bare minimum?" question so this could be a DynamoDBItem level method but your point is a good one even if we want to treat the .pdq.te rows to be the ground truth, wiping out the old state when we get an update probably isn't ideal. Will change.

BarrettOlson:
@schatten this is going to have to be two different PRs because it needs the changes to python-threatexchange to be pushed for it to run in prod. (it also needs a significant rebase). I didn't want to you lose your review history so I published my changes to your feedback anyway here. If you want to review this as is, I will rebase and split it then put up the two PRs.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
schatten: None is fine. None is magic, but it is sensible magic.

schatten (approved):

PR #505, opened 2021-04-19
Give actioner secrets permssions

jeberl:
Summary
---------

In the vein of smaller PRs

This is part of implementing TE write backs

Test Plan
---------

`terraform -chdir=terraform validate && terraform -chdir=terraform apply` 
Test locally that my fetcher script can read as necessary

### DIFF
--- a/hasher-matcher-actioner/terraform/main.tf
+++ b/hasher-matcher-actioner/terraform/main.tf
@@ -19,8 +19,9 @@ locals {
   common_tags = {
     "HMAPrefix" = var.prefix
   }
-  pdq_file_extension = ".pdq.te"
-  te_data_folder     = module.hashing_data.threat_exchange_data_folder_info.key
+  pdq_file_extension   = ".pdq.te"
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/terraform/main.tf
```diff
@@ -265,4 +267,17 @@ module "actions" {
   log_retention_in_days = var.log_retention_in_days
   additional_tags       = merge(var.additional_tags, local.common_tags)
   measure_performance   = var.measure_performance
+  api_token_secret      = aws_secretsmanager_secret.api_token
+}
+
+### ThreatExchange API Token Secret ###
+
+resource "aws_secretsmanager_secret" "api_token" {
```
schatten, line 275: If this is specifically "threatexchange" api_token, can we scope the name as well? "threatexchange_api_token"?

schatten (approved):
Approve to unblock. It appears the API token is specific to threatexchange. But the name does not reflect that. 

Would love to see the name changed to something more explicit.

PR #506, opened 2021-04-19
Actions queues

eriksendc:
Summary
---------

Separate actions and reactions into separate queues.

Test Plan
---------

1. `terraform destroy`
2. `make upload_docker`
3. `terraform apply`
6. validate CloudWatch logs: see that action_performer got a MatchMessage
7. validate CloudWatch logs: see that reactioner got a MatchMessage

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
@@ -51,9 +51,14 @@ def lambda_handler(event, context):
 
     for sqs_record in event["Records"]:
         # TODO research max # sqs records / lambda_handler invocation
-        sns_notification = json.loads(sqs_record["body"])
+        sqs_record_body = json.loads(sqs_record["body"])
+
+        if sqs_record_body.get("Event") == "TestEvent":
+            logger.info("Disregarding test: %s", sqs_record_body)
+            continue
+
         match_message: MatchMessage = MatchMessage.from_sns_message(
-            sns_notification["Message"]
+            sqs_record_body["Message"]
         )
 
         logger.info("Evaluating match_message: %s", match_message)
--- a/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
@@ -20,6 +20,7 @@ logger = get_logger(__name__)
 sqs_client = boto3.client("sqs")
 
 ACTIONS_QUEUE_URL = os.environ["ACTIONS_QUEUE_URL"]
+REACTIONS_QUEUE_URL = os.environ["REACTIONS_QUEUE_URL"]
 
 
 def lambda_handler(event, context):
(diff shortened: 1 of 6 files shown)


hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -15,6 +17,10 @@
 from hmalib.lambdas.actions.action_performer import perform_label_action
 
 logger = get_logger(__name__)
+sqs_client = boto3.client("sqs")
+
+ACTIONS_QUEUE_URL = os.environ["ACTIONS_QUEUE_URL"]
+REACTIONS_QUEUE_URL = os.environ["REACTIONS_QUEUE_URL"]
```
jeberl, line 23: nit: David has advised me before that if we move these to a class it can be easier to mock in tests. See `FetcherConfig` class in `fetcher.py`

hasher-matcher-actioner/hmalib/lambdas/actions/reactioner.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    This is the main entry point for reacting to ThreatExchange. The action evaluator
+    sends a reaction message by way of the reactions queue and here's where they're
+    popped off and dealt with.
+    """
+    for sqs_record in event["Records"]:
+        # TODO research max # sqs records / lambda_handler invocation
+        sqs_record_body = json.loads(sqs_record["body"])
+        logger.info(f"sqs_record_body = {sqs_record_body}")
+        # TODO next PR will include implementation of the ReactionMessage class
```
jeberl, line 20: My fear is that the logic here is going to look almost identical to the logic in the action_performer. Unclear to me if that will become an issue

hasher-matcher-actioner/hmalib/lambdas/actions/reactioner.py
```diff
@@ -0,0 +1,22 @@
...
+logger = get_logger(__name__)
+
+
+def lambda_handler(event, context):
+    """
+    This is the main entry point for reacting to ThreatExchange. The action evaluator
+    sends a reaction message by way of the reactions queue and here's where they're
+    popped off and dealt with.
+    """
+    for sqs_record in event["Records"]:
```
jeberl, line 16: I've seen else where us check for test events. Do we want to do that here as well?

jeberl (approved):
I'm still not seeing the benefit of splitting these into multiple queues instead of keeping them together but we said disagree and commit so let's ship it!

hasher-matcher-actioner/hmalib/lambdas/actions/reactioner.py
eriksendc: Sounds good! Done!

hasher-matcher-actioner/hmalib/lambdas/actions/reactioner.py
eriksendc: @jeberl  Yup we'll have to see. One big difference is what gets persisted at the end of the day. In the case of actions, you'll be writing a row somewhere that's keyed off of the ID of the matched content. For reactions, it'll be something persisted by "banked content id" (indicator id).

PR #507, opened 2021-04-19
implement reactions in actioner

jeberl:
Summary
---------

Set up functions to react to descriptors in threat exchange as a result of a `MatchMessage`. This might later change to be a `ReactionMessage` @eriksendc  is working on this.

Specifically, this PR will react to all descriptors on all indicators that were involved in a match

Added new function to `python-threatexchange` will have to bump the version after this merges

also addressed the following deprecation warning in python_threatexchange:
```
ThreatExchange/python-threatexchange/threatexchange/api.py:145: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
```

Test Plan
---------

1. Added new test case to configs_test.py (@Dcallies is planning to move this to a new test file)
2. ran locally with real indicator ids and saw the reactions go through


hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -99,6 +105,44 @@ def call(self, **kwargs) -> Response:
...
+
+        descriptor_ids = {
+            descriptor_id["id"]
+            for indicator_id in indicator_ids
+            for descriptor_id in api.get_threat_descriptors_from_indicator(indicator_id)
+        }
+
+        for id in descriptor_ids:
+            api.react_to_threat_descriptor(id, self.reaction)
+            logger.warning("reacted %s to descriptor %s", self.reaction, id)
```
eriksendc, line 132: Why is this a `warning` and not `info`?

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -99,6 +105,44 @@ def call(self, **kwargs) -> Response:
         return delete(self.url)
 
 
+@dataclass
+class ReactActionPerformer(ActionPerformer):
```
eriksendc, line 109: In my mind, the concept of models is supposed to be along the lines of entities, or the definition of the types / shapes of objects. But this is processing (and the calling of webhooks) are not really entity / type -related at all. What about having this and the subclasses live in a separate module?

eriksendc (changes requested):

hasher-matcher-actioner/hmalib/common/actioner_models.py
jeberl: I'm good moving this to a separate module

eriksendc (approved):

hasher-matcher-actioner/hmalib/common/reactioner_models.py
```diff
@@ -0,0 +1,73 @@
...
+        raise NotImplementedError
+
+    def perform_action(self, match_message: MatchMessage) -> None:
+        api_key = AWSSecrets.te_api_key()
+        api = ThreatExchangeAPI(api_key)
+
+        indicator_ids = {
+            dataset_match_details.banked_content_id
+            for dataset_match_details in match_message.matching_banked_signals
+            if dataset_match_details.bank_source == "te"
```
schatten, line 28: "te" is being extracted to a constant in [another PR](https://github.com/facebook/ThreatExchange/pull/503/files#diff-5e4a23ec0c264ec4ce29b65fe691d4db71393e2972d2803b2444d1fd020020eaR23-R24). This should use the same constant. Avoid magic values as much as possible.

hasher-matcher-actioner/hmalib/common/reactioner_models.py
```diff
@@ -0,0 +1,73 @@
...
+            for descriptor_id in api.get_threat_descriptors_from_indicator(indicator_id)
+        }
+
+        for id in descriptor_ids:
+            api.react_to_threat_descriptor(id, self.reaction)
+            logger.info("reacted %s to descriptor %s", self.reaction, id)
+
+
+class ReactInReviewActionPerformer(ReactActionPerformer):
+    reaction = "IN_REVIEW"
```
schatten, line 43: Are these constants available in the python-threatexchange library? These seem very 'threatexchangey'. Should probably not be here.

python-threatexchange/threatexchange/api.py
```diff
@@ -226,7 +226,7 @@ def get_threat_descriptors(self, ids, **kwargs):
 
         descriptors = []
         for id, descriptor in response.items():
-            if includeIndicatorInOutput == False:
+            if not includeIndicatorInOutput:
```
schatten, line 229: nit: python variables should use snake_case

python-threatexchange/threatexchange/api.py
```diff
@@ -511,3 +510,21 @@ def _postThreatDescriptor(self, url, postParams, showURLs, dryRun):
         except urllib.error.HTTPError as e:
             responseBody = json.loads(e.read().decode("utf-8"))
             return [None, e, responseBody]
+
+    def get_threat_descriptors_from_indicator(
+        self, indicator_id: int, showURLs: bool = False
+    ) -> t.List[t.Dict[str, t.Any]]:
+        url = (
```
schatten, line 517: Would prefer [_get_graph_api_url](https://github.com/facebook/ThreatExchange/blob/d935e872ff89fd5dccc62f32951b17a0ffac3305/python-threatexchange/threatexchange/api.py#L335) used instead of handrolling URIs.

schatten (changes requested):
I'm not following the module well. And apologies if previous diffs passed me by.

**Are these really models?**
- When you say "models", the first thing to come to mind are entities stored in a database. As in [flask](https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/), [django](https://docs.djangoproject.com/en/3.2/topics/db/models/) and [tornado](https://pypi.org/project/tornado-models/)
- The doc string for ActionPerformer says that it is a configuration, when it is actually not. There is ActionPerformerConfig for that.

**Keeping business logic and configuration separate**
Users will interface with the configuration, not the business objects. ActionPerformerConfig is a configuration object. ActionPerformer and subclasses are business objects.

- The ActionPerformerConfig.update_performer method is actually ActionPerformerConfig.update_performer_config. Imagine a user updating the URL / methods on the UI. They will be dealing with the Config object, not the ActionPerformer object. An 
ActionPerformer instance need not be created just to update its config, right?


**Dynamic class loading instead of subclass search**
- Instead of using subclass finding, if you were to use a qualified class path like the [django project's middleware loader does here](https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/utils/module_loading.py#L7-L24), we could allow users to plug in their custom actions directly.
- That allows your ActionPerformerConfig.concrete_type to have values like 'hmalib.lambdas.actions.WebhookPostActionPerformer' or 'partnerorg.integrity.DoSomethingSeriousActionPerformer'.

hasher-matcher-actioner/hmalib/common/reactioner_models.py
jeberl: Will change once it lands

hasher-matcher-actioner/hmalib/common/reactioner_models.py
jeberl: Not stored anywhere in python-threatexchange currently. I think these ActionPerformer classes are ThreatExchange specific so not entirely sure the issue

python-threatexchange/threatexchange/api.py
jeberl: For this and the 2 comments above I agree with you. Do you think its worth another PR to clean this up and bump the version again or save it for another time?

jeberl:
@schatten replying to your top level comment:
1. the ActionPerformer and ActionPerformerConfig will change after @Dcallies 's diff. Will make sure comments are correct
2. I think, going from our earlier discussion theres a lot of confusion around the term "models" happy to change it but I think should be a change across the platform

PR #509, opened 2021-04-20
[hma] Have fetcher update dynamo for signal objects 

BarrettOlson:
Summary
---------

Second (and actually significant)  half of the split of #503

This modifies the fetcher to update `PDQSignalMetadata` this has already been mostly reviewed by @schatten as a part of #503. 

Test Plan
---------

I was able to confirm that the fetcher was able to update the db objects in a deployed lambda via cloudwatch log:

After the first fetch and a match was created I went to s3 and deleted the .pdq.te and .checkpoint files. 
I then triggered the fetcher to make sure it wold complete without error and checked the logs to make sure the db object was updated, (paste of one of the logs below.)

```
[INFO]	2021-04-20T17:55:37.849Z	730a56a1-06a8-4204-b448-e03e9425f554	Updated Signal Tags in DB for signal id: <REDACTED> source: te for privacy group: 303636684709969
```

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/s3_adapters.py
+++ b/hasher-matcher-actioner/hmalib/common/s3_adapters.py
@@ -20,6 +20,8 @@ HashRowT = t.Tuple[str, t.Dict[str, t.Any]]
 @dataclass
 class S3ThreatDataConfig:
 
+    SOURCE_STR = "te"
+
     threat_exchange_data_bucket_name: str
     threat_exchange_data_folder: str
     threat_exchange_pdq_file_extension: str
@@ -121,7 +123,7 @@ class ThreatExchangeS3Adapter:
                 {
                     "id": int(row["id"]),
                     "hash": row["hash"],
-                    "source": "te",  # default for now to make downstream easier to generalize
+                    "source": self.config.SOURCE_STR,  # default for now to make downstream easier to generalize
                     "privacy_groups": set(
                         [privacy_group]
                     ),  # read privacy group from key
--- a/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
@@ -28,6 +28,8 @@ from hmalib.aws_secrets import AWSSecrets
 from hmalib.common.config import HMAConfig
 from hmalib.common import config as hmaconfig
 from hmalib.common.logging import get_logger
+from hmalib.common.s3_adapters import S3ThreatDataConfig
+from hmalib.models import PDQSignalMetadata
 from threatexchange import threat_updates as tu
 from threatexchange.api import ThreatExchangeAPI
 from threatexchange.cli.dataset.simple_serialization import CliIndicatorSerialization
@@ -68,14 +70,17 @@ class FetcherConfig:
     s3_bucket: str
     s3_te_data_folder: str
     collab_config_table: str
+    data_store_table: str
 
     @classmethod
-    @lru_cache(maxsize=1)  # probably overkill, but at least it's consistent
+    @lru_cache(maxsize=None)  # probably overkill, but at least it's consistent
     def get(cls):
+        # These defaults are naive but can be updated for testing purposes.
         return cls(
             s3_bucket=os.environ["THREAT_EXCHANGE_DATA_BUCKET_NAME"],
             s3_te_data_folder=os.environ["THREAT_EXCHANGE_DATA_FOLDER"],
             collab_config_table=os.environ["THREAT_EXCHANGE_CONFIG_DYNAMODB"],
+            data_store_table=os.environ["DYNAMODB_DATASTORE_TABLE"],
         )
 
 
@@ -137,6 +142,7 @@ def lambda_handler(event, context):
             api.app_id,
             te_data_bucket,
             config.s3_te_data_folder,
+            config.data_store_table,
         )
         stores.append(indicator_store)
         indicator_store.load_checkpoint()
@@ -162,7 +168,9 @@ def lambda_handler(event, context):
         finally:
             if delta:
                 logging.info("Fetch complete, applying %d updates", len(delta.updates))
-                indicator_store.apply_updates(delta)
+                indicator_store.apply_updates(
+                    delta, post_apply_fn=indicator_store.post_apply
+                )
             else:
                 logging.error("Failed before fetching any records")
 
@@ -211,12 +219,14 @@ class ThreatUpdateS3PDQStore(tu.ThreatUpdatesStore):
         app_id: int,
         s3_bucket: t.Any,  # Not typable?
         s3_te_data_folder: str,
+        data_store_table: str,
     ) -> None:
         super().__init__(privacy_group)
         self.app_id = app_id
         self._cached_state: t.Optional[t.Dict] = None
         self.s3_bucket = s3_bucket
         self.s3_te_data_folder = s3_te_data_folder
+        self.data_store_table = data_store_table
 
     @property
     def checkpoint_s3_key(self) -> str:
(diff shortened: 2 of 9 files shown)


hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -121,7 +123,7 @@ def _parse_file(self, file_name: str, data_file: S3FileT) -> t.List[HashRowT]:
                 {
                     "id": int(row["id"]),
                     "hash": row["hash"],
-                    "source": "te",  # default for now to make downstream easier to generalize
+                    "source": self.config.SOURCE_STR,  # default for now to make downstream easier to generalize
```
jeberl, line 126: 👍 
will use this in the future

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -68,14 +70,17 @@ class FetcherConfig:
     s3_bucket: str
     s3_te_data_folder: str
     collab_config_table: str
+    data_store_table: str
 
     @classmethod
-    @lru_cache(maxsize=1)  # probably overkill, but at least it's consistent
```
jeberl, line 73: What's this for?

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -321,8 +331,13 @@ def _store_state(self, contents: t.Iterable["CliIndicatorSerialization"]):
             write_s3_text(txt_content, self.s3_bucket, self.data_s3_key)
             logger.info("%d rows stored for %d", len(items), self.privacy_group)
 
-    def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
+    def _apply_updates_impl(
+        self,
+        delta: tu.ThreatUpdatesDelta,
+        post_apply_fn=lambda x: None,
```
jeberl, line 337: Can we type the lambda inputs more? I would think that the post_apply function takes a single update but it looks like it takes a dict

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -333,10 +348,40 @@ def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
...
             else:
                 state[item.key] = item
+                updated[item.key] = item
 
         self._store_state(state.values())
         self._cached_state = state
 
+        post_apply_fn(updated)
+
+    def post_apply(self, updated: t.Dict = {}):
```
jeberl, line 358: same typing thing. Can we describe what the Dict is of

jeberl, line 358: This will also be useful for the eventual Retroactioner!

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -333,10 +348,40 @@ def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
...
+        """
+        After the fetcher applies an update, check to see if
+        matches for any of the updated signals and if so update
```
jeberl, line 361: nit
```suggestion
        matches for any of the updated signals exist and if so update
```

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -333,10 +348,40 @@ def _apply_updates_impl(self, delta: tu.ThreatUpdatesDelta) -> None:
...
+            # example row format: ('<signal>', '<id>', '<time added>', '<tag1 tags2>')
+            # e.g ('096a6f9...064f', 1234567891234567, '2020-07-31T18:47:45+0000', 'true_positive hma_test')
+            if PDQSignalMetadata(
+                signal_id=int(row[1]),
+                ds_id=str(self.privacy_group),
+                updated_at=datetime.now(),
+                signal_source=S3ThreatDataConfig.SOURCE_STR,
+                signal_hash=row[0],  # note: not used by update_tags_in_table_if_exists
+                tags=row[3].split(" ") if row[3] else [],
+            ).update_tags_in_table_if_exists(table):
```
jeberl, line 377: Why do we only update tags? is there an easy way to all the fields (except signal id and hash)

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -97,12 +98,44 @@ def to_dynamodb_item(self) -> dict:
...
             "SK": self.get_dynamodb_ds_key(self.ds_id),
             "SignalHash": self.signal_hash,
-            "SignalSource": self.signal_source,  # defaults to 'te' in the current pipeline
+            "SignalSource": self.signal_source,
             "UpdatedAt": self.updated_at.isoformat(),
             "HashType": self.SIGNAL_TYPE,
             "Tags": self.tags,
         }
 
+    def update_tags_in_table_if_exists(self, table: Table) -> bool:
```
jeberl, line 107: I'm going to flag that this seems like a good opportunity to use HMAConfig (though admittedly not a config, maybe we could change the name to HMAS3Adapter)

We could get rid of most of this boiler plate. It would change this fucntion to:
```
signal = PDQSignalMetadata.get(self.signal_source, self.signal_id)
signal.tags = new_tags
PDQSignalMetadata.update(signal)
```

hasher-matcher-actioner/setup.py
```diff
@@ -10,7 +10,7 @@
     install_requires=[
         "boto3",
         "boto3-stubs[essential,sns]==1.17.14.0",
-        "threatexchange[faiss,pdq_hasher]>=0.0.15",
+        "threatexchange[faiss,pdq_hasher]>=0.0.18",
```
jeberl, line 13: 👍

jeberl (changes requested):

eriksendc (approved):

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: https://github.com/facebook/ThreatExchange/pull/503#discussion_r615995597

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: The only meaningful field that isn't a part of lookup keys are tags and hash.

hasher-matcher-actioner/hmalib/models.py
BarrettOlson: For objects that use the single table model of the HMA-Datastore table I am more opposed to having it be dynamically generated like HMAConfig, indexes, typing, etc. become more complex so it is hard for me to see right now if tradeoff would be worth it.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: This is a method override so type granularity is complicated/could go wrong, where and how `_apply_updates_impl` uses the `post_apply_fn` method is dependent on the store class implementing it.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: This is the level of type granularity that is defined for `ThreatUpdateS3PDQStore._cached_state` 
If we want to refactor this further to for better type support I'm not opposed but this functions shouldn't (and I'm not sure mypy will let us) be more type restrictive than the class using it imo

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
jeberl: Do/should we change the updated_at field also? If so, maybe we should change the name to update_pdq_metadata.

non blocking

hasher-matcher-actioner/hmalib/models.py
jeberl: Indexes seems like a valid reason not to pursue right now. just planting the seed

jeberl (approved):
One comment I'd like to see but approving. No need to circle back

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: we already do, because 'update_at' is (or at least imo should be) updated every time an object in the db is updated (this field also exists for Match and Hash Records) and is not about the signal itself.

BarrettOlson:
> One comment I'd like to see but approving. No need to circle back

Thanks! update_at is updated there, it's when the object was last touched and is part of an index.  If you think that is overly confusing lmk and I'll see what I can do to make that convention clearer in another PR

PR #510, opened 2021-04-20
[hma][config] Add subtype helper

Dcallies:
Summary
---------

Well... here's a thing that does a thing. I went in circles on this for a while on decision points that probably didn't matter.

The key thing I hung on is where to do this:
```
     HMAConfig
          |
HMAConfigWithSubtype
          |
   HMAConfigSubtype
```

Or this

```
HMAConfig         HMAConfigWithSubtype
     |               - Returns instances of HMAConfigSubtype
HMAConfigSubtype
```

There are tradeoffs in either direction, but decided to go with former. 

I also made the choice(/mistake) to use metaclass magic to provide a default behavior on class name.

Test Plan
---------

Updated unittest, I believe its comprehensive.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/config.py
+++ b/hasher-matcher-actioner/hmalib/common/config.py
@@ -142,29 +142,39 @@ class _HMAConfigWithSubtypeMeta(type):
     """
 
     def __new__(metacls, cls_name: str, bases, cls_dict):
-        if cls_name != "HMAConfigWithSubtypes":
+        # Is this one of the bases?
+        if cls_name in ("HMAConfigWithSubtypes", "_HMAConfigSubtype"):
+            return super().__new__(metacls, cls_name, bases, cls_dict)
+        # Has a Subtype already been applied?
+        for base in bases:
+            if hasattr(base, "Subtype"):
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -126,13 +136,182 @@ def initialize(config_table_name: str) -> None:
...
+        cls_dict["Subtype"] = None  # Recursion guard, overwrite below
+
+        config_type = ""
+        # Provide magic default - last in MRO order or class name
+        for base in bases:
+            if hasattr(base, "CONFIG_TYPE"):
+                config_type = base.CONFIG_TYPE
+        if not config_type:
+            config_type = cls_name
+            cls_dict["CONFIG_TYPE"] = cls_name
```
Dcallies, line 162: These lines should not be needed, because SubType and CONFIG_TYPE have the same lifetimes, and we exit early above, this will just duplicate CONFIG_TYPE on subclasses

jeberl (approved):
Understand at about 60%. Does seem to solve my problem though so will stamp

PR #512, opened 2021-04-20
Messages for actions and reactions queues

eriksendc:
Summary
---------

This PR introduces two messages based off of MatchMessage: ActionMessage and ReactionMessage. They are instantiated by the action evaluator and consumed by the action performer and reactioner, respectively. Other code changes include moving Labels to actioner_models.py, removing Labels from the concept of matches and match records (we now have metadata and, later, classifications), and other minor cleanup like renaming things from SNS/sns-something-or-other to AWS/aws-something-or-other. 

Test Plan
---------

1. `terraform destroy`
2. `make upload_docker`
3. `terraform apply`
6. validate CloudWatch logs: see that action_performer got an ActionMessage
7. validate CloudWatch logs: see that reactioner got a ReactionMessage

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/actioner_models.py
+++ b/hasher-matcher-actioner/hmalib/common/actioner_models.py
@@ -69,7 +69,7 @@ TUrl = t.Union[t.Text, bytes]
 class ActionMessage(MatchMessage):
     """
     The action performer needs the match message plus which action to perform
-    TODO Create a reflection / introspection-based helper that implements 
+    TODO Create a reflection / introspection-based helper that implements
     to_ / from_aws_message code (and maybe from_match_message_and_label(), too)
     for ActionMessage and MatchMessage.
     """
--- a/hasher-matcher-actioner/hmalib/common/actioner_models.py
+++ b/hasher-matcher-actioner/hmalib/common/actioner_models.py
@@ -78,7 +78,9 @@ class ActionMessage(MatchMessage):
             {
                 "ContentKey": self.content_key,
                 "ContentHash": self.content_hash,
-                "MatchingBankedSignals": [x.to_dict() for x in self.matching_banked_signals],
+                "MatchingBankedSignals": [
+                    x.to_dict() for x in self.matching_banked_signals
+                ],
                 "ActionLabelValue": self.action_label.value,
             }
         )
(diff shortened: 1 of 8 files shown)


hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -52,23 +55,18 @@ def lambda_handler(event, context):
     for sqs_record in event["Records"]:
         # TODO research max # sqs records / lambda_handler invocation
         sqs_record_body = json.loads(sqs_record["body"])
-
```
eriksendc, line 55: After conversations with Barrett and Dipanjan, it was determined checking for test messages is only applicable to queue reads where the queue message originated as an S3 notification.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -47,12 +47,12 @@ def remove_content_key_prefix(key: str) -> str:
         return key[len(DynamoDBItem.CONTENT_KEY_PREFIX) :]
```
eriksendc, line 49: I renamed *SNS* / *sns* to *AWS* / *aws* because we're using the same functionality in both SNS and SQS contexts.

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -48,6 +65,95 @@ class ActionRule:
...
+        return cls(
+            match_message.content_key,
+            match_message.content_hash,
+            match_message.matching_banked_signals,
+            action_label,
+        )
+
+
+@dataclass
+class ReactionMessage(MatchMessage):
```
jeberl, line 111: Seems like a lot of this could be combined with the ActionMessage

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -48,6 +65,95 @@ class ActionRule:
 TUrl = t.Union[t.Text, bytes]
 
 
+@dataclass
+class ActionMessage(MatchMessage):
```
jeberl, line 69: Seems like a number of classes in this diff have the `to_aws_message` and `from_aws_message` functions. I think we should do a refactor and have these classes implement HMAConfig. That class handles the writing to and from dynamo and would help us remove a TON of boiler plate. Maybe a team discussion

jeberl (changes requested):
I'm still struggling to see the benefit of two queues here. It looks to me like 95% of the code is duplicated. If we do want to keep them as separate queues and separate classes can we at least try to combine as much of the code as possible into a super class?

hasher-matcher-actioner/hmalib/common/actioner_models.py
eriksendc: Seems like we can find a way to borrow some of the introspection / reflection code to where we have a helper to do this for any class. Definitely worth looking into at some point. V0 or V1, I'm not sure... But I'll add a TODO.

jeberl (approved):
good to ship

PR #494, opened 2021-04-15
Get privacy groups

mengyangwang:
Summary
---------

add new sync_privacy_groups function in fecher.py(class ThreatExchangeConfig) to fetch privacy groups and insert them to dynamoDB

Test Plan
---------

1. test local ``` python3 -m hmalib.lambdas.fetcher ```, new privacy groups should be added in wangmengyang-HMAConfig table
2. ```black . ```
3. ```mypy hmalib``` make sure no errors

PR #516, opened 2021-04-21
Create datasets api

mengyangwang:
Summary
---------
previous PR : https://github.com/facebook/ThreatExchange/pull/494
Task : T88063259, this task is to create ThreatExchange setting page, This PR is to create apis that can support below functions
1. get all datasets
2. update dataset(can only update fetcher_active and write_back)
3. sync datasets(only insert, will not overwrite existing config)
4. Enhance config framework, changes to below functions
   a. create_config, this will create config, exception when item exists
   b. update_config, this will update config, exception when item not exists

Future Actions
---------
 Create UI page to use new api 

Test Plan
---------

1. black hmalib
2. mypy hmalib
3.  run ``` python -m hmalib.lambdas.api.api_root``` with required environment params
4. test 3 apis /datasets/, /datasets/update, /datasets/sync
5. make docker
6. make upload_docker 
7.  terraform apply
8. test 3 apis in local using postman

https://user-images.githubusercontent.com/81996660/116473757-99b64080-a845-11eb-9de4-d7cf16bd9713.mov


hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+
+@dataclass
+class SyncOrUpdateDatasetResponse(JSONifiable):
+    response: str
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+
+def get_datasets_api(hma_config: str) -> bottle.Bottle:
```
schatten, line 42: nit: the variable is actually hma_config_table

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+                for collab in collabs
+            ]
+        )
+
+    @datasets_api.put("/update", apply=[jsoninator])
+    def update_datasets():
+        """
+        Update dataset
+        """
+        data = bottle.request.json
```
schatten, line 76: Can you also do a Dataset.from_json() here? That will give you type safety and potentially validation.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+        if data and data["privacy_group_id"]:
+            config = ThreatExchangeConfig(
+                str(data["privacy_group_id"]),
+                fetcher_active=data["fetcher_active"],
+                privacy_group_name=data["privacy_group_name"],
+            )
+            # Warning! Will stomp on existing configs (including if you disable them)
+            hmaconfig.update_config(config)
+            return SyncOrUpdateDatasetResponse(
```
schatten, line 85: Can you return a Datasets response here instead? A failure will be captured in the status code itself. A Datasets response here will allow the react app to auto-update on success.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+    def update_datasets():
+        """
+        Update dataset
+        """
+        data = bottle.request.json
+        if data and data["privacy_group_id"]:
+            config = ThreatExchangeConfig(
+                str(data["privacy_group_id"]),
+                fetcher_active=data["fetcher_active"],
+                privacy_group_name=data["privacy_group_name"],
```
schatten, line 81: Do not see the point of updating privacy_group_name here. I don't think that should be editable from the user side at all.

schatten (changes requested):
This is a great first cut. I have some comments on the response shape and the ability to edit the privacy_group_name.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: sure will do

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: sure sounds good to me

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: the current config framework will overwrite the exciting config, if privacy_group_name is not added then it will be missing in the updated config item. I am planning to modify config framework in a separate PR to handle this issue

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
jeberl, line 1: I really like that you made this its own file. This a good convention we should start to follow universally

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+import typing as t
+from dataclasses import dataclass, asdict
+from .middleware import jsoninator, JSONifiable
+from hmalib.common.config import HMAConfig
+from hmalib.common import config as hmaconfig
+from hmalib.lambdas.fetcher import *
+
+
+@dataclass
+class Dataset(JSONifiable):
```
jeberl, line 13: Whats the difference between this class and ThreatExchangeConfig. Can we reuse that?

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+            # Warning! Will stomp on existing configs (including if you disable them)
+            hmaconfig.update_config(config)
+            return SyncOrUpdateDatasetResponse(
+                response=f'{data["privacy_group_id"]} is successfully updated'
+            )
+
+    @datasets_api.post("/sync", apply=[jsoninator])
+    def sync_datasets():
+        """
+        Sync dataset
```
jeberl, line 92: Can we describe what sync means?
"Fetch new collaborations from ThreatExchnage and potentially update the configs stored in AWS"

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        sync_privacy_groups()
```
jeberl, line 94: Where is this function defined? Is it in fetcher? Did it come in another diff I missed?

jeberl:
Good with me once @schatten approves

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: sure will do

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: yes this is from fetcher, it's imported in the line 9, I will make the import more clear

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: this is class is created to easily converted to or from json, looks we can't reuse ThreatExchangeConfig

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,99 @@
...
+                for collab in collabs
+            ]
+        )
+
+    @datasets_api.put("/update", apply=[jsoninator])
+    def update_datasets():
+        """
+        Update dataset
+        """
+        dataSet = Dataset.from_dict(bottle.request.json)
```
schatten, line 80: nit: In python, use snake_case, please.

mengyangwang: yeah this is fixed now

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
schatten: From the [definition of sync_privacy_groups](https://github.com/facebook/ThreatExchange/blob/fabe18312252f6ae497c0dc347b9480a7c1a8c8e/hasher-matcher-actioner/hmalib/lambdas/fetcher.py#L178-L208), it appears that it will overwrite all the configs, setting fetcher_active to True.

schatten (changes requested):
Want to understand the sync_privacy_groups() API better. What does it do? 

As per present impl, it will overwrite all configs after fetching from threatexchange.

mengyangwang:
> Want to understand the sync_privacy_groups() API better. What does it do?
> 
> As per present impl, it will overwrite all configs after fetching from threatexchange.

yes it fetches privacy groups and overwrite all configs, set active to true for testing purpose.
I will submit another PR to change this logic, it fetch configs from threatexchange and compare with all current configs in dynamoDB. Then add missing configs, delete configs that no longer available, so that will not overwrite all configs

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: this is fixed, now it will only takes fetcher_active and write_back as param to update

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: > From the [definition of sync_privacy_groups](https://github.com/facebook/ThreatExchange/blob/fabe18312252f6ae497c0dc347b9480a7c1a8c8e/hasher-matcher-actioner/hmalib/lambdas/fetcher.py#L178-L208), it appears that it will overwrite all the configs, setting fetcher_active to True.

now added conditionExpression to not overwrite existing configs

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -306,8 +307,13 @@ def _convert_item(cls, item: t.Dict[str, t.Any]):
 # to make them easier to spot in the wild
 
 
-def update_config(config: HMAConfig) -> None:
-    """Update or create a config. No locking or versioning!"""
+def update_config(
```
schatten, line 310: The insert_only vs upsert difference, IMO, is significant.

How do you feel about two separate methods for this? 
`config.upsert_config` and `config.insert_config`. Explicit, top-level and hard to get wrong.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -322,10 +328,42 @@ def update_config(config: HMAConfig) -> None:
...
+        try:
+            get_dynamodb().meta.client.put_item(
+                TableName=_TABLE_NAME,
+                Item=_config_to_dynamodb_item(config),
+                ConditionExpression=Attr(key).not_exists(),
+            )
+        except ClientError as e:
+            # Ignore the ConditionalCheckFailedException, bubble up
+            # other exceptions.
+            if e.response["Error"]["Code"] != "ConditionalCheckFailedException":
```
schatten, line 341: This appears to swallow exceptions. So, the callsite will never know that the config was not updated. Would urge you to rethink how this is getting handled.

create/update is a simpler model. 

create -> Fails if already exists.
update -> Fails if doesn't exist.

Currently, we have 
* `put` -> Passes always
* `put if not exists` -> Passes always. Swallows exceptions.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -322,10 +328,42 @@ def update_config(config: HMAConfig) -> None:
...
+        get_dynamodb().meta.client.put_item(
+            TableName=_TABLE_NAME,
+            Item=_config_to_dynamodb_item(config),
+        )
+
+
+def update_config_attributes_by_type_and_name(
+    config_type: str, name: str, updates: dict
+):
+    """Delete a config by name (and type)"""
```
schatten, line 353: Outdated doc string.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -366,3 +404,23 @@ def _config_to_dynamodb_item(config) -> t.Dict[str, t.Any]:
     item["ConfigType"] = config.get_config_type()
     item["ConfigName"] = config.name
     return item
+
+
+def _get_update_params(body: dict):
```
schatten, line 409: How does this handle types?

Given that we are accepting dict in the update_config_attributes_by_type_and_name, it is possible that type information could be lost. 

In [Amazon's example](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GettingStarted.Python.03.html#GettingStarted.Python.03.03), they use the Decimal type when adding a number. We'd need to do something similar, right?

schatten: Also, this method's returns are not typed. In any new code we are writing, we must have types.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+
+@dataclass
+class SyncDatasetResponse(JSONifiable):
+    response: str
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+
+def sync_privacy_groups():
```
schatten, line 59: nit: This seems to business logic specific to be here. If it is possible for this code to be invoked outside of a lambda, consider moving to its own module outside of the `hmalib.lambdas` space.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    Implementation of the "sync_privacy_groups" function of threatexchange config.
```
schatten, line 61: Superfluous documentation.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+
+    1. call ThreatExchange API get_threat_privacy_groups_member
+    and get_threat_privacy_groups_owner to get the list of privacy groups
+
+    2. If the threat_updates_enabled is true, save it using config framework if it doesn't
+    exist in dynamoDB.
+
+    3. If one privacy group exists in dynamoDB but not in the result of get_threat_privacy_groups_owner
```
schatten, line 69: Confusing typo: in_use, not in_user.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+            # HMA can only read from privacy groups that have threat_updates enabled.
+            # # See here for more details:
+            # https://developers.facebook.com/docs/threat-exchange/reference/apis/threat-updates/v9.0
+            logger.info("Adding collaboration name %s", privacy_group.name)
+            priavcy_group_id_in_use.add(privacy_group.id)
+            config = ThreatExchangeConfig(
+                privacy_group.id,
+                # TODO Currently default to True for testing purpose,
+                # need to switch it to False before v0 launch
+                fetcher_active=True,
```
schatten, line 90: This is an important configuration value. The team has not had a chance to reflect on whether this should be active by default or not. 

Can we extract this to a constant in the interim, please?

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+                privacy_group_name=privacy_group.name,
+                in_use=True,
+                write_back=False,
```
schatten, line 93: ditto:

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+            # Warning! Will stomp on existing configs (including if you disable them)
```
schatten, line 95: I guess with your insert only changes, these comments are outdated?

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def update_privacy_group_in_use(priavcy_group_id_in_use: set) -> None:
+    collabs = ThreatExchangeConfig.get_all()
+    for collab in collabs:
+        if collab.privacy_group_id not in priavcy_group_id_in_use:
+            hmaconfig.update_config_attributes_by_type_and_name(
+                config_type="ThreatExchangeConfig",
+                name=str(collab.privacy_group_id),
+                updates={"in_use": False},
```
schatten, line 108: Follow up on type-safety as discussed [here](https://github.com/facebook/ThreatExchange/pull/516/files#diff-8c7b738eacee248e926da9c6d90088bc733c421aaaed4ce075c770fca8a295c7R409), a typo here "in_us" instead of "in_user" would succeed and we would never know what happened.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+                    fetcher_active=collab.fetcher_active,
+                    write_back=collab.write_back,
+                    in_use=collab.in_use,
+                )
+                for collab in collabs
+            ]
+        )
+
+    @datasets_api.put("/update", apply=[jsoninator])
+    def update_datasets():
```
schatten, line 144: Please add typing information.

schatten: The method name says 'update_datasets'. The plural made be think that we're updating multiple datasets together.

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+        if dataset["fetcher_active"] is not None:
+            updates["fetcher_active"] = dataset["fetcher_active"]
+        if dataset["write_back"] is not None:
+            updates["write_back"] = dataset["write_back"]
+        # Will only insert the item if privacy_group_id didn't exist
+        updated_dataset = hmaconfig.update_config_attributes_by_type_and_name(
+            config_type="ThreatExchangeConfig",
+            name=str(dataset["privacy_group_id"]),
+            updates=updates,
+        )
```
schatten, line 159: Could the updates dictionary building be done with

```python
updates = { k : dataset[k] for k in ["fetcher_active", "write_back"] }
```

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,171 @@
...
+
+    @datasets_api.put("/update", apply=[jsoninator])
+    def update_datasets():
+        """
+        Update dataset
+        """
+        dataset = bottle.request.json
+        updates = {}
+        if dataset["fetcher_active"] is not None:
```
schatten, line 150: Untyped request json is also hard to debug / understand / maintain. Consider adding an extension to jsoninator which does typed json processing and adds that as an argument to the view method.

Something like 
```python
@dataclass
class RequestType():
  <define types here>

@datasets_api.put("/update", apply=[jsoninator(RequestType)])
def update_datasets(typed_payload: RequestType):
  do_something()
```

schatten (changes requested):
Would have preferred the config.py changes as a separate PR. It is sizeable and more-importantly, reviewable separately from the rest of the PR.

I have various inline recommendations. 

While right now, the code works, from a better-engineering perspective, there are quite a few misses which I would want to catch early. Let me know if I can help with implementing any of the recommendations.

I am also fine to punt on this for now and revisit when we have time, but I'd strongly urge fixing these issues now.

hasher-matcher-actioner/hmalib/common/config.py
mengyangwang: > Would have preferred the config.py changes as a separate PR. It is sizeable and more-importantly, reviewable separately from the rest of the PR.
> 
> I have various inline recommendations.
> 
> While right now, the code works, from a better-engineering perspective, there are quite a few misses which I would want to catch early. Let me know if I can help with implementing any of the recommendations.
> 
> I am also fine to punt on this for now and revisit when we have time, but I'd strongly urge fixing these issues now.

yeah thanks for that!
and I absolutely agree to make changes now to avoid future work.
I will make those changes and separate the PR of config.py changes

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -310,7 +314,25 @@ def update_config(config: HMAConfig) -> None:
     get_dynamodb().meta.client.put_item(
         TableName=_TABLE_NAME,
         Item=_config_to_dynamodb_item(config),
+        ConditionExpression=Attr("ConfigType").not_exists(),
```
schatten, line 317: Dumb question: Will this suffice? Shouldn't we also check whether the ConfigName `.not_exists()`?

schatten: I strongly feel this will prevent creation of multiple configs of the same type.

hasher-matcher-actioner/hmalib/common/threatexchange_config.py
```diff
@@ -0,0 +1,61 @@
...
+                # TODO Currently default to True for testing purpose,
+                # need to switch it to False before v0 launch
+                fetcher_active=FETCHER_ACTIVE,
+                privacy_group_name=privacy_group.name,
+                in_use=True,
+                write_back=WRITE_BACK,
+            )
+            try:
+                hmaconfig.create_config(config)
+            except ClientError as e:
```
schatten, line 45: nit: No need to address in this PR. The 'duplicate' check should happen in the hmaconfig module itself. 

For users of that module, the storage is an implementation detail. Why must threatexchange_config need to know that a "ConditionalExpression" is being used to to prevent duplication?

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -0,0 +1,117 @@
...
+        collabs = ThreatExchangeConfig.get_all()
+        return DatasetsResponse(
+            datasets_response=[
+                Dataset(
+                    privacy_group_id=collab.privacy_group_id,
+                    privacy_group_name=collab.privacy_group_name,
+                    fetcher_active=collab.fetcher_active,
+                    write_back=collab.write_back,
+                    in_use=collab.in_use,
+                )
```
schatten, line 92: nit: what I have previously found succinct is to co-locate translations like these. Perhaps a `Dataset.from_collab classmethod` that does these translations?

Then, if anyone is adding an attribute to Dataset, adding the attribute translation is the sensible thing to do.

schatten (approved):
Hey @mengyangwang. This PR is good enough to merge subject to one check. I have a few nits littered throughout, but those are good candidates for subsequent PRs.

The only blocking comment is to verify that we can actually create_config for the same type multiple times. If you can add or point me to a test that proves this, you may merge this in.

hasher-matcher-actioner/hmalib/common/config.py
mengyangwang: yeah this is a little tricky, there is one post to explain the way dynamoDB works
https://www.alexdebrie.com/posts/dynamodb-condition-expressions/
Screen Shot 2021-04-29 at 5 17 14 PM
Basically when want to check duplicate, we just need to send PK, dynamoDB will automatically check both PK and SK

hasher-matcher-actioner/hmalib/common/threatexchange_config.py
mengyangwang: sure I'll mark it in to do list

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: yeah good catch! I added from_collab method for dataset class

hasher-matcher-actioner/hmalib/common/config.py
schatten: Checked this out on my devcontainer. Seems to be working fine. The ConfigName check is implicit.

PR #517, opened 2021-04-21
[hma][action] Convert configs to HMAConfigWithSubtypes

Dcallies:
Summary
---------

Followup from @jeberl's diff. Use new subtype to cover action case, hopefully works exactly as expected.

Test Plan
---------

mypy and pytest are green, will work with others for more.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/actioner_models.py
+++ b/hasher-matcher-actioner/hmalib/common/actioner_models.py
@@ -158,118 +158,69 @@ class ReactionMessage(MatchMessage):
 
 
 @dataclass
-class ActionPerformer:
+class ActionPerformer(config.HMAConfigWithSubtypes):
     """
-    A configuration of what action should be performed when a specific label
-    is added to a match. Each label can lead to exactly one action being taken.
+    An ActionPerfomer is the configuration + the code to perform an action.
 
-    To create a new type of ActionPerformer simply extend this class
-    and add the necessary fields as attributes. All these attributes will be stored in AWS
-    after update_aws_config is called. To retrieve from AWS the correct
-    action performer for a label use the function ActionPerformerConfig.get_performer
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -158,118 +158,69 @@ def from_match_message_and_label(
...
     """
 
-    action_label: ActionLabel
+    @staticmethod
+    def get_subtype_classes():
+        return [
+            WebhookPostActionPerformer,
+            WebhookGetActionPerformer,
+            WebhookPutActionPerformer,
+            WebhookDeleteActionPerformer,
```
jeberl, line 177: Decided to keep these separate?

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -158,118 +158,69 @@ def from_match_message_and_label(
...
-        }
-        del attrs["action_label"]
-        action_performer_config = ActionPerformerConfig(
-            name=action_performer.action_label.value,
-            concrete_type=action_performer.__class__.__name__,
-            concrete_type_attrs=attrs,
-        )
-        config.update_config(action_performer_config)
+    def call(self, _data: str) -> Response:
+        return delete(self.url)
```
jeberl, line 223: For this and GET we will need to find another way of including the match message in the request

jeberl (approved):
Looks good. Some comments of things I want to add later but the config structure seems correct

hasher-matcher-actioner/hmalib/common/actioner_models.py
Dcallies: Yea, asked you about it in chat, and wanted to avoid messing up any UI work you had in flight. Should be easy to compress later if you decide to.

PR #518, opened 2021-04-21
display actioner details

jeberl:
Summary
---------

Add more details to the actioner configs page

Next step, make it editable

Also some small changes that we requested from earlier PR

Test Plan
---------

Screen Shot 2021-04-21 at 3 10 13 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,3 @@ pdq/java/bazel-out/
 pdq/java/bazel-testlogs/
 .mypy_cache/
 .vscode/
-hasher-matcher-actioner/.devcontainer/devcontainer.json
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ pdq/java/bazel-out/
 pdq/java/bazel-testlogs/
 .mypy_cache/
 .vscode/
+hasher-matcher-actioner/.devcontainer/devcontainer.json
--- a/hasher-matcher-actioner/webapp/src/Settings.jsx
+++ b/hasher-matcher-actioner/webapp/src/Settings.jsx
@@ -2,15 +2,17 @@
  * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
  */
 
+/* eslint-disable react/prop-types */
+
 import React from 'react';
 import Card from 'react-bootstrap/Card';
 import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
-import Tabs from 'react-bootstrap/Tabs';
 import Tab from 'react-bootstrap/Tab';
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -17,3 +17,4 @@ pdq/java/bazel-out/
 pdq/java/bazel-testlogs/
 .mypy_cache/
 .vscode/
+hasher-matcher-actioner/.devcontainer/devcontainer.json
```
eriksendc, line 20: I see this is added here... Not sure if we want to do this... If we go this route we'll need a convention to call out when to go get a new copy, and we'll have to store the "official" team copy somewhere outside of source control. I'm pretty sure this should be left in as-is. @schatten what do you think?

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -44,44 +46,69 @@ function ActionLabelSettings() {
   return <>Action Label Settings coming here</>;
 }
 
-function WebhookActionerRender() {
-  return <Card>This is a card displaying a WebhookActionerRender</Card>;
+function WebhookActionerRender({url, headers, webhookType}) {
```
eriksendc, line 49: This naming is unconventional to me. We see this in the "framework" that was used for SignalOsmosis configs... How about something like `WebhookActioner` or `WebhookActionerDetails`? It's the `Renderer` bit that doesn't feel right.

hasher-matcher-actioner/webapp/src/Settings.jsx
```diff
@@ -44,44 +46,69 @@ function ActionLabelSettings() {
...
+  return (
+    <Card>
+      <h5>WebhookActionerRender</h5>
```
eriksendc, line 52: Use `<CardHeader>`, `<CardBody>` and `<CardFooter>`. And if you have a set of attributes, like URL, Webhook Type and Headers, you probably should put a table inside the `<CardBody>`.

eriksendc: https://react-bootstrap.github.io/components/cards

eriksendc (changes requested):
Please revert .gitignore, deal with renaming `WebhookActionerRenderer` and use the card-related React Bootstrap components as mentioned in inline comments.

.gitignore
jeberl: Didnt you need to change this file to set up the dev container? Change dipanjanm name to your unix name?

hasher-matcher-actioner/webapp/src/Settings.jsx
jeberl: Good with me

eriksendc (approved):

PR #519, opened 2021-04-21
Implemented action evaluation along with some tests.

eriksendc:
Summary
---------

This PR implements action evaluation (measuring up a match message against action rules and getting a resulting set of actions out the other side). Simple tests included! :)

Test Plan
---------

0. python -m py.test
1. `terraform destroy`
2. `make upload_docker`
3. `terraform apply`
6. validate CloudWatch logs: see that action_performer got an ActionMessage

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
+++ b/hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
@@ -1,0 +1,51 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import typing as t
+import unittest
+
+from hmalib.lambdas.actions.action_evaluator import get_action_labels
+from hmalib.models import MatchMessage, BankedSignal
+from hmalib.common.actioner_models import (
+    ActionLabel,
+    ActionRule,
+    ClassificationLabel,
+    Label,
+)
+
+
+class ActionRuleEvaluationTestCase(unittest.TestCase):
+    def test_get_action_labels(self):
+
+        enqueue_for_review_action_label = ActionLabel("EnqueueForReview")
+        bank_id = "12345"
+
+        banked_signal_without_foo = BankedSignal(
+            "67890", bank_id, "Test", ["Bar", "Xyz"]
+        )
+        banked_signal_with_foo = BankedSignal(
+            "67890", bank_id, "Test", ["Foo", "Bar", "Xyz"]
+        )
+
+        match_message_without_foo = MatchMessage(
+            "key", "hash", [banked_signal_without_foo]
+        )
+        match_message_with_foo = MatchMessage("key", "hash", [banked_signal_with_foo])
+
+        action_rules = [
+            ActionRule(
+                enqueue_for_review_action_label,
+                [Label("BankId", bank_id)],
+                [ClassificationLabel("Foo")],
+            )
+        ]
+
+        action_labels: t.Set[ActionLabel] = get_action_labels(
+            match_message_without_foo, action_rules
+        )
+
+        assert len(action_labels) == 1
+        assert action_labels.pop() == enqueue_for_review_action_label
+
+        action_labels = get_action_labels(match_message_with_foo, action_rules)
+
+        assert len(action_labels) == 0
--- a/hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
+++ b/hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
@@ -8,6 +8,7 @@ from hmalib.models import MatchMessage, BankedSignal
 from hmalib.common.actioner_models import (
     ActionLabel,
     ActionRule,
+    BankIDClassificationLabel,
     ClassificationLabel,
     Label,
 )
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -34,7 +35,7 @@ class ActionEvaluatorConfig:
     reactions_queue_url: str
 
     @classmethod
-    @lru_cache(maxsize=1)
+    @lru_cache(maxsize=None)
```
eriksendc, line 38: I saw a change from `1` to `None` in another PR. Asked Barrett about it and confirmed `None`'s the way to go.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -90,46 +95,69 @@ def lambda_handler(event, context):
...
-    action_rule: ActionRule, match_message: MatchMessage
+def action_rule_applies_to_classifications(
+    action_rule: ActionRule, classifications: t.Set[Label]
 ) -> bool:
     """
-    Evaluate if the action rule applies to the match message. Return True if the action rule's "must have"
-    labels are all present in the match message, and that none of the "must not have" labels are present
-    in the match message, otherwise return False.
+    Evaluate if the action rule applies to the classifications. Return True if the action rule's "must have"
+    labels are all present and none of the "must not have" labels are present in the classifications, otherwise return False.
```
eriksendc, line 153: Next PR is to implement action rule config, at which time I'll make `must_have_labels` and `must_not_have_labels` sets so they don't have to be converted from `List` to `Set` here.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -90,46 +97,69 @@ def lambda_handler(event, context):
...
+    classifications_by_match: t.List[t.Set[Label]] = list()
+
+    for banked_signal in match_message.matching_banked_signals:
+        classifications: t.Set[Label] = set()
+        classifications.add(Label("BankSource", banked_signal.bank_source))
+        classifications.add(Label("BankId", banked_signal.bank_id))
+        classifications.add(Label("BankedContentId", banked_signal.banked_content_id))
+        for classification in banked_signal.classifications:
+            classifications.add(ClassificationLabel(classification))
+        classifications_by_match.append(classifications)
```
jeberl, line 130: blocking: What's the difference between the classification labels stored in `banked_signal.classifications` and the ones you're adding here. Why aren't the BankSource BankId, and BankedContentID labels considered classification labels.

Conceptually, I think it makes more sense to me for all the labels to be added in the Matcher before it gets to the action evaluator.

Additionally, if we do think more labeling has to be done in the actioner I think we should add it to the MatchMessage rather than returning a separate set of ClassificationLabels

jeberl (changes requested):
This looks good and love the tests. I will keep building off that

One blocking comment is about where classification happens/ has it happened already. I think conceptually it makes more sense for the classification labels to be added in the Matcher rather than the Action Evaluator. Happy to discuss more

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
eriksendc: @jeberl I want to say I took this as guidance from @Dcallies. The concept of "what a match is" is determined by the matcher. Representing that in the consistent way that the action evaluator will evaluate is a concern of the evaluator. We'll have to check in with David and see where to go from here.

hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
```diff
@@ -0,0 +1,51 @@
...
+        match_message_without_foo = MatchMessage(
+            "key", "hash", [banked_signal_without_foo]
+        )
+        match_message_with_foo = MatchMessage("key", "hash", [banked_signal_with_foo])
+
+        action_rules = [
+            ActionRule(
+                enqueue_for_review_action_label,
+                [Label("BankId", bank_id)],
+                [ClassificationLabel("Foo")],
```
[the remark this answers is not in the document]
eriksendc: @jeberl ActionRule takes three parameters: an action label, a list of "must have" labels, and a list of "must not have" labels. So, positionally, `[Label("BankId", bank_id)]` is the list of "must have" labels, and `[ClassificationLabel("Foo")]` is the list of "must not have" labels.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
eriksendc: There are three particular attributes, `bank_source`, `bank_id`, and `banked_signal_id` that are IDs. It seemed like a good idea to leverage `Label` to create something where the "name" of the thing was the `key` and the ID value was the `value` of a `Label`. I'm happy to create `BankSourceClassificationLabel`, `BankIDClassificationLabel` and `BankedSignalIDClassificationLabel` if that makes more sense.

jeberl (approved):
Talked offline with Brian about where classification should happen. Both agreed that current state is not ideal but in't sure how to incorporate Davids comments. Accepting to unblock and Brian David and I will meet to reconcile our differences

PR #520, opened 2021-04-21
aws_dataclass

Dcallies:
Summary
---------

Breaks out the magic of converting dataclasses to aws-friendly dictionaries to another class. I have no idea if this works for more than just dynamodb items, and so this might be misnamed.

Bonus: 
* Supported for nested datastructures
* And arbitrary sets

Test Plan
---------

added unittest

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/aws_dataclass.py
+++ b/hasher-matcher-actioner/hmalib/common/aws_dataclass.py
@@ -7,7 +7,22 @@ There's likely already an existing library that exists somewhere that does
 this much better, but after spending 5 minutes looking, just try and write
 something on our own.
 
+How to use:
 
+    import typing as t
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -0,0 +1,154 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Helpers for converting python dataclasses to/from aws-friendly formats.
+
+There's likely already an existing library that exists somewhere that does 
+this much better, but after spending 5 minutes looking, just try and write
+something on our own.
+
+
```
Dcallies, line 10: maybe remove

eriksendc (approved):
LGTM

PR #524, opened 2021-04-22
Implement action rules as config-backed objects.

eriksendc:
Summary
---------

Implement action rules as config-backed HMAConfig objects.

Test Plan
---------

0. python -m py.test
1. `terraform destroy`
2. `make upload_docker`
3. `terraform apply`
4. `python -m hmalib.scripts.populate_config_db load_example_configs`
5. `make dev_upload_test_data`
6. wait for hashing and matching to occur
7. validate Cloudwatch logs: see that action_evaluator got an ActionRule back from the config table
8. validate CloudWatch logs: see that action_performer got an ActionMessage

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/actioner_models.py
+++ b/hasher-matcher-actioner/hmalib/common/actioner_models.py
@@ -72,7 +72,15 @@ class Action:
 
 
 @dataclass
-class ActionRule:
+class ActionRule(config.HMAConfig):
+    """
+    Action rules are config-backed objects that have a set of labels (both
+    "must have" and "must not have") which, when evaluated against the
+    classifications of a matching banked piece of content, lead to an action
+    to take (specified by the rule's action label). By convention each action
+    rule's name field is also the value field of the rule's action label.
+    """
+
     action_label: ActionLabel
     must_have_labels: t.List[Label]
     must_not_have_labels: t.List[Label]
--- a/hasher-matcher-actioner/hmalib/common/actioner_models.py
+++ b/hasher-matcher-actioner/hmalib/common/actioner_models.py
@@ -4,7 +4,7 @@ import hmalib.common.config as config
 import json
 import typing as t
 
-from dataclasses import dataclass, fields
+from dataclasses import dataclass, field, fields
 from hmalib.common.logging import get_logger
 from hmalib.models import BankedSignal, MatchMessage
 from requests import get, post, put, delete, Response
@@ -12,7 +12,7 @@ from requests import get, post, put, delete, Response
 logger = get_logger(__name__)
 
 
-@dataclass
+@dataclass(unsafe_hash=True)
 class Label:
     key: str
     value: str
@@ -29,39 +29,35 @@ class Label:
             return NotImplemented
         return self.key == another_label.key and self.value == another_label.value
 
-    def __hash__(self) -> int:
-        return self.value.__hash__()
+
+@dataclass(unsafe_hash=True)
+class ClassificationLabel(Label):
+    key = field(default="Classification", init=False)
 
 
-class LabelWithConstraints(Label):
-    _KEY_CONSTRAINT = "KeyConstraint"
-
-    def __init__(self, value: str):
-        super(LabelWithConstraints, self).__init__(self._KEY_CONSTRAINT, value)
+@dataclass(unsafe_hash=True)
+class BankSourceClassificationLabel(Label):
+    key = field(default="BankSourceClassification", init=False)
 
 
-class ClassificationLabel(LabelWithConstraints):
-    _KEY_CONSTRAINT = "Classification"
+@dataclass(unsafe_hash=True)
+class BankIDClassificationLabel(Label):
+    key = field(default="BankIDClassification", init=False)
 
 
-class BankSourceClassificationLabel(LabelWithConstraints):
-    _KEY_CONSTRAINT = "BankSource"
+@dataclass(unsafe_hash=True)
+class BankedContentIDClassificationLabel(Label):
+    key = field(default="BankedContentIDClassification", init=False)
 
 
-class BankIDClassificationLabel(LabelWithConstraints):
-    _KEY_CONSTRAINT = "BankID"
+@dataclass(unsafe_hash=True)
+class ActionLabel(Label):
+    key = field(default="Action", init=False)
 
 
-class BankedContentIDClassificationLabel(LabelWithConstraints):
-    _KEY_CONSTRAINT = "BankedContentID"
-
-
-class ActionLabel(LabelWithConstraints):
-    _KEY_CONSTRAINT = "Action"
-
-
-class ThreatExchangeReactionLabel(LabelWithConstraints):
-    _KEY_CONSTRAINT = "ThreatExchangeReaction"
+@dataclass(unsafe_hash=True)
+class ThreatExchangeReactionLabel(Label):
+    key = field(default="ThreatExchangeReaction", init=False)
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -29,39 +29,35 @@ def __eq__(self, another_label: object) -> bool:
...
-        return self.value.__hash__()
-
-
-class LabelWithConstraints(Label):
-    _KEY_CONSTRAINT = "KeyConstraint"
-
-    def __init__(self, value: str):
-        super(LabelWithConstraints, self).__init__(self._KEY_CONSTRAINT, value)
 
+@dataclass(unsafe_hash=True)
```
Dcallies, line 33:
```suggestion
@dataclass
```

Ignorable/YMMV: You can safely drop the hash generation here, because the number of fields hasn't changed, and so the same hash method will be generated

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -29,39 +29,35 @@ def __eq__(self, another_label: object) -> bool:
...
+class ClassificationLabel(Label):
+    key = field(default="Classification", init=False)
```
Dcallies, line 35: Whoops, I forgot to include `:str` in the thing I pasted you

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -72,7 +68,15 @@ class Action:
 
 
 @dataclass
-class ActionRule:
+class ActionRule(config.HMAConfig):
+    """
+    Action rules are config-backed objects that have a set of labels (both
```
Dcallies, line 73: Thanks for adding comments!

hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
```diff
@@ -34,6 +34,7 @@ def test_get_action_labels(self):
 
         action_rules = [
             ActionRule(
+                enqueue_for_review_action_label.value,
```
Dcallies, line 37: (Ignore if this is just temporary): Means you can only have one rule per action, is that correct? Two actions can't lead to a delete?

hasher-matcher-actioner/hmalib/common/actioner_models.py
eriksendc: tried this and it didn't work. Discussed further with David and leaving as is.

hasher-matcher-actioner/hmalib/common/actioner_models.py
eriksendc: np! :)

hasher-matcher-actioner/hmalib/common/actioner_models.py
Dcallies: It turns out this advice was wrong! python hash() is smart enough to notice you extended the class without redefining hash, and so hash might not be safe any more.

hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
eriksendc: @Dcallies This is just a simple test... you can definitely have another ActionRule in action_rules that has the same action label. The second rule just has to have a different name.

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -4,15 +4,15 @@
...
+from dataclasses import dataclass, field, fields
 from hmalib.common.logging import get_logger
 from hmalib.models import BankedSignal, MatchMessage
 from requests import get, post, put, delete, Response
 
 logger = get_logger(__name__)
 
 
-@dataclass
+@dataclass(unsafe_hash=True)
```
Dcallies, line 15: (ignorable): The alternative we discussed offline is making Label's to be frozen=True, but if there's anywhere that is modifying labels after creation it will start erroring

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -139,20 +144,22 @@ def get_classifications_by_match(match_message: MatchMessage) -> t.List[t.Set[La
...
+    #     ActionRule(
+    #         action_label.value,
+    #         action_label,
+    #         [
+    #             BankIDClassificationLabel("303636684709969"),
+    #             ClassificationLabel("true_positive"),
+    #         ],
+    #         [BankedContentIDClassificationLabel("3364504410306721")],
+    #     )
+    # ]
```
Dcallies, line 162: Should this be removed?

hasher-matcher-actioner/hmalib/common/actioner_models.py
eriksendc: fixed this.

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -285,7 +285,11 @@ module "actions" {
   log_retention_in_days = var.log_retention_in_days
   additional_tags       = merge(var.additional_tags, local.common_tags)
   measure_performance   = var.measure_performance
-  te_api_token_secret      = aws_secretsmanager_secret.te_api_token
+  te_api_token_secret   = aws_secretsmanager_secret.te_api_token
+  config_table = {
+    name = aws_dynamodb_table.hma_config.name
+    arn  = aws_dynamodb_table.hma_config.arn
+  }
```
Dcallies, line 292: nit: config_table = local.hma_config should work here, or you should be to make changes to the variable to make it work like fetcher does

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -132,8 +132,8 @@ module "fetcher" {
   additional_tags       = merge(var.additional_tags, local.common_tags)
   fetch_frequency       = var.fetch_frequency
 
-  te_api_token_secret  = aws_secretsmanager_secret.te_api_token
-  hma_config = local.hma_config
+  te_api_token_secret = aws_secretsmanager_secret.te_api_token
+  hma_config          = local.hma_config
```
Dcallies, line 136: Compare and contrast

Dcallies (approved):
Shipit. Heads up @jeberl

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
eriksendc: yep. Removed.

hasher-matcher-actioner/terraform/main.tf
eriksendc: refactored and made consistent.

hasher-matcher-actioner/terraform/main.tf
eriksendc: refactored and made consistent.

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -72,7 +68,15 @@ class Action:
 
 
 @dataclass
-class ActionRule:
+class ActionRule(config.HMAConfig):
+    """
+    Action rules are config-backed objects that have a set of labels (both
+    "must have" and "must not have") which, when evaluated against the
```
jeberl, line 74: nit 
```suggestion
    Action rules are config-backed objects that have two sets of labels (
    "must have" and "must not have") which, when evaluated against the
```

hasher-matcher-actioner/hmalib/scripts/populate_config_db.py
```diff
@@ -74,9 +84,18 @@ def load_defaults(_args):
             privacy_group_name="Test Config 2",
         ),
         WebhookPostActionPerformer(
-            name="EnqueForReviewAction",
+            name="EnqueueForReview",
```
jeberl, line 87: Should we make this action_label.value also?

jeberl (commented):
Looks like you merged already a few nits that might be nice in a follow up PR. Would also love your thought on making Labels Immutable

PR #529, opened 2021-04-23
Make Actioner Configs 

jeberl:
Summary
---------

Actioner configs are specifications the user will define for what should happen as a result of a match. For now we only support 1 actioner type - Send a Webhook - but in the future this could be expanded to other actioner types.

This diff only uses mocked data. In future PRs this will be read directly from AWS though the api layer

Also dis some small refactors of the settings page into multiple files per tab. Only need to review file ActionSettingsTab.jsx 

Test Plan
---------

https://user-images.githubusercontent.com/24800630/115926358-b8d15e80-a450-11eb-8a9d-f41f5395965a.mov


hasher-matcher-actioner/webapp/src/settings/ActionSettingsTab.jsx
```diff
@@ -0,0 +1,281 @@
...
+        description: 'Optional json object of headers to include in webhook',
+        default: '{}',
+      },
+    },
+    description:
+      'When a match occurs, a webhook will be sent to the specified url with data describing the match',
+  },
+};
+
+function WebhookActionerRender({url, webhookType, headers}) {
```
eriksendc, line 44: Please nix `Renderer` from the name

hasher-matcher-actioner/webapp/src/settings/ActionSettingsTab.jsx
```diff
@@ -0,0 +1,281 @@
...
+      <div hidden={editing}>
+        <Card.Header>
+          WebhookActioner
+          <br />
+          <Form.Text className="text-muted">
+            {ActionerTypes.WebhookActioner.description}
+          </Form.Text>
+        </Card.Header>
+        <Card.Body>
+          URL : {savedURL}
```
eriksendc, line 67: Still looking for formatting some other way (use a table, for example). Not required for this PR.

hasher-matcher-actioner/webapp/src/settings/ActionSettingsTab.jsx
```diff
@@ -0,0 +1,281 @@
...
+              setEditing(true);
+            }}>
+            Edit
+          </Button>
+        </Card.Footer>
+      </div>
+
+      <div hidden={!editing}>
+        <Card.Header>
+          <Form>
```
eriksendc, line 87: I'm not 100% sure, but when you get to wiring this up, you do have two forms here, one for the action "type" and one for all of the details... you may or may not find it necessary to have one outer <Form>... just something to be on the lookout for. It might be find to put <Form> just outside the card header, body and footer.

hasher-matcher-actioner/webapp/src/settings/ActionSettingsTab.jsx
```diff
@@ -0,0 +1,281 @@
...
+const ActionRenderers = {
+  WebhookActioner: WebhookActionerRender,
+};
+
+function ActionPerformerRow({name, type, params}) {
+  const [editing, setEditing] = useState(false);
+  const [savedName, setName] = useState(name);
+  const [tempName, setTempName] = useState(name);
+  return (
+    <tr>
```
eriksendc, line 175: I'm guessing if you run with developer tools open, you should be seeing an issue regarding "key" not being defined. You need that on this <tr>, I think. It needs to be set to something unique. See https://reactjs.org/docs/lists-and-keys.html for background / context.

eriksendc (changes requested):
`Renderer` crept back in. Please nix that. And I think you're probably seeing an issue in the browser console about `key` being undefined. I've made an inline comment about addressing that.

hasher-matcher-actioner/webapp/src/settings/ActionSettingsTab.jsx
jeberl: Might have to reorganize some of this when we get to multiple actioners anyway

hasher-matcher-actioner/webapp/src/settings/ActionSettingsTab.jsx
jeberl: Hmm. I'm not seeing that error. Might be hidden? or might be because its rows instead of list items

eriksendc (approved):

PR #531, opened 2021-04-27
Support typing request bodies using the jsoninator middleware

schatten:
Summary
---
Ammends jsoninator, enabling it to provide an optional type class for the request payload.

Test Plan
---
```
$ python -m py.test
```

Adds a dependency and tests. The test adds an in-process HTTP server. Both old (without typed req bodies) and new (with typed req bodies) APIs are tested.

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
@@ -83,10 +83,30 @@ def jsoninator(
     # I feel like I'm doing a lot of work to support a common API for typed
     # request payloads and untyped request payloads.  All this could be avoided
     # by having `jsoninator` and `jsoninator_typed_request(RequestType)`, but
-    # meh! This looks shinier.
+    # meh! This looks shinier. All the complexity is hidden behind a single
+    # method call.
 
-    # Can't use isinstance because Class types are t.Callable as well.
+    # Super verbose description of decorator BEGINS
+
+    # A decorator is just a function that returns another function where the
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
```diff
@@ -24,27 +28,95 @@
...
-        body = view_fn(*args, **kwargs)
-        response.content_type = "application/json"
-        return json.dumps(body.to_json())
+    # I feel like I'm doing a lot of work to support a common API for typed
+    # request payloads and untyped request payloads.  All this could be avoided
+    # by having `jsoninator` and `jsoninator_typed_request(RequestType)`, but
+    # meh! This looks shinier.
+
+    # Can't use isinstance because Class types are t.Callable as well.
+    if hasattr(view_fn_or_request_type, "from_dict"):
```
BarrettOlson, line 89: I think it might be good use a `etmli5` comment or two here. decorators are hard enough to understand and this is a level beyond even if you understand python closures. Maybe little more details on `view_fn_or_request_type` and the check? My understanding of this is that an objectively worse but more actuate name would be `view_fn_OR_request_type_and_view_fn`
Something like if the request_type "from_dict"

BarrettOlson (approved):
https://xkcd.com/353/ lied to me :)

mengyangwang (approved):

BarrettOlson:
@schatten FYI: this broke black, happy to stamp a quick PR to fix

PR #528, opened 2021-04-23
ActionRule: list to set refactor for must and must not have labels

eriksendc:
Summary
---------

The `ActionRule` dataclass had two lists of labels. It's more efficient if they're sets, as we'd been converting the lists to sets on each rule evaluation. Storing as sets straight away prevents the need for that conversion.

Test Plan
---------

0. python -m py.test
1. `terraform destroy`
2. `make upload_docker`
3. `terraform apply`
4. `python -m hmalib.scripts.populate_config_db load_example_configs`
5. upload "mini-castle.jpg"
6. wait for hashing, matching and actioning to occur
7. validate that the EnqueueMiniCastleForReview WebhookPostActionPerformer is exercised (https://webhook.site/#!/01cef721-bdcc-4681-8430-679c75659867) 
8. upload "sailboat-mast-and-sun.jpg"
9. wait for hashing, matching and actioning to occur 
10. validate that the EnqueueSailboatForReview WebhookPostActionPerformer is exercised (https://webhook.site/#!/fa5c5ad5-f5cc-4692-bf03-a03a4ae3f714)

PR #532, opened 2021-04-27
Actions add action rule to action message

eriksendc:
Summary
---------

This PR introduces a list of action rules into the action message class so that the rule(s) that is the source of the action being taken is eventually able to be persisted once the action is taken by the action performer. This work can be reviewed in the second commit.

While adding the action rules field to the action message, I found myself writing more .to_aws and .from_aws code. Instead of continuing down that road, I pivoted and tried out HasAWSSerialization. It worked with a couple of minor additions. This work can be reviewed in the first commit.

Other PRs for context:
* [PR 528 - ActionRule: list to set refactor for must and must not have labels](https://github.com/facebook/ThreatExchange/pull/528)
* [PR 519 - Implemented action evaluation along with some tests](https://github.com/facebook/ThreatExchange/pull/519)
* [PR 512 - Messages for actions and reactions queues](https://github.com/facebook/ThreatExchange/pull/512)

Test Plan
---------

0. python -m py.test
1. `terraform destroy`
2. `make upload_docker`
3. `terraform apply`
4. `python -m hmalib.scripts.populate_config_db load_example_configs`
5. upload "mini-castle.jpg"
6. wait for hashing, matching and actioning to occur
7. validate that the EnqueueMiniCastleForReview WebhookPostActionPerformer is exercised (https://webhook.site/#!/01cef721-bdcc-4681-8430-679c75659867) and that the payload delivered has the action rules list.


hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -49,6 +51,13 @@ def py_to_aws(py_field: t.Any, in_type: t.Optional[t.Type[T]] = None) -> T:
 
     check_type = origin or in_type
 
+    if isinstance(check_type, t.ForwardRef):
```
eriksendc, line 54: @Dcallies and I collaborated on getting the refactor in this PR to work. One thing I'd run into was that one of the fields of the base dataclass was being typed using a forward reference, which the code here doesn't support. So we decided to include a helpful error message in case someone else runs into this in the future.

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -511,16 +474,20 @@ class BankedSignal:
...
-            "Classifications": self.classifications,
-        }
 
-    @classmethod
-    def from_dict(cls, d: dict) -> "BankedSignal":
-        return cls(
-            d["BankedContentId"], d["BankId"], d["BankSource"], d["Classifications"]
-        )
+@dataclass
+class MatchMessage(HasAWSSerialization):
```
eriksendc, line 479: I had to move the match message dataclass to be after banked signal so that the typing of the matching banked signals field could be a direct reference.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -97,26 +102,35 @@ def lambda_handler(event, context):
...
+                    action_label_to_action_rules[action_rule.action_label].append(
+                        action_rule
+                    )
+                else:
+                    action_label_to_action_rules[action_rule.action_label] = [
+                        action_rule
+                    ]
+    action_label_to_action_rules = remove_superseded_actions(
+        action_label_to_action_rules
+    )
```
jeberl, line 132: Where did we land on adding classifications onto the match itself? What was davids perspective?

hasher-matcher-actioner/hmalib/models.py
jeberl, line 479: Really love this new from/aws functionality

jeberl (approved):
This looks good to me. The one thing I'm still stuck on is why classifications are separate from the match messgae. Why don't those come from the matcher

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
eriksendc: Had a talk with @Dcallies and we'll move applying classification labels to be a concern of the matcher in a future PR.

PR #534, opened 2021-04-28
Writebacker framework

jeberl:
Summary
---------

A framework for how writebacks should happen (formerly reactions). See tests for how to use

Test Plan
---------
1. Find an image in the copydays dataset from this website https://lear.inrialpes.fr/~jegou/data.php
2. Go to hma locally and upload file

Added tests
end to end test:
1. upload docker and update all lambdas
2. go to hma in UI and upload a copy days image
3. get hash from match details page and use `bunnylol threatdata <hash>` to get the indicator
4. Look for the indicator and then see that reactions happened


hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -56,8 +56,36 @@ class ActionLabel(Label):
...
+    value: str = field(default="FalsePositive", init=False)
+
+
+@dataclass(unsafe_hash=True)
+class TruePositiveReactionLabel(ReactionLabel):
+    value: str = field(default="TruePositive", init=False)
+
+
+@dataclass(unsafe_hash=True)
+class InReviewReactionLabel(ReactionLabel):
```
eriksendc, line 82: I'd say nix `InReviewReactionLabel` as we don't intend to expose it in v0

hasher-matcher-actioner/hmalib/common/actioner_models.py
```diff
@@ -160,20 +186,20 @@ def from_aws_message(cls, message: str) -> "ReactionMessage":
             parsed["ContentKey"],
             parsed["ContentHash"],
             [BankedSignal.from_dict(d) for d in parsed["MatchingBankedSignals"]],
-            ThreatExchangeReactionLabel(parsed["ReactionLabelValue"]),
+            ReactionLabel(parsed["ReactionLabelValue"]),
```
eriksendc, line 189: Be careful when merging... to_ and from_aws_message are no longer present in this class in master. And, by the way, it's moved to message_models.py. Note that from_match_message_and_label() _is_ still exists.

hasher-matcher-actioner/hmalib/common/tests/test_writebacker.py
```diff
@@ -0,0 +1,150 @@
...
+            self.match_message, reaction
+        )
+        event = {"Records": [{"body": reaction_message.to_aws_message()}]}
+
+        result = lambda_handler(event, None)
+        assert result == {"reactions_performed": {"te": "Wrote Back true positive"}}
+
+        os.environ["MOCK_TE_API"] = "False"
+
+    # def test_second_writebacker(self):
```
eriksendc, line 113: assuming you'll remove this commented stuff or resurrect it.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -84,21 +85,20 @@ def lambda_handler(event, context):
...
-                for threat_exchange_reaction_label in threat_exchange_reaction_labels:
-                    threat_exchange_reaction_message = (
-                        ReactionMessage.from_match_message_and_label(
-                            match_message, threat_exchange_reaction_label
-                        )
-                    )
-                    config.sqs_client.send_message(
-                        QueueUrl=config.reactions_queue_url,
-                        MessageBody=threat_exchange_reaction_message.to_aws_message(),
+        threat_exchange_reaction_labels = get_reaction_messages(
```
eriksendc, line 88: threat_exchange_reaction_labels is now a list of reaction messages, so a) this should be renamed and b) I don't understand why you're making another reaction message object at 94. Can't you just call send_message with each reaction message that came back from get_reaction_messages()?

hasher-matcher-actioner/hmalib/common/reactioner_models.py
```diff
@@ -1,71 +1,244 @@
...
+        raise NotImplementedError
+
+    def _writeback_impl(self, writeback_message: ReactionMessage) -> str:
         raise NotImplementedError
 
-    def perform_action(self, match_message: MatchMessage) -> None:
+    def perform_writeback(self, writeback_message: ReactionMessage) -> str:
+        writeback_options = self.writeback_options()
+        for writeback_option_cls in writeback_options:
+            writebacker = writeback_option_cls()
```
jeberl, line 99: Will move to dict look up

hasher-matcher-actioner/hmalib/common/reactioner_models.py
```diff
@@ -1,71 +1,244 @@
...
+@dataclass
+class ThreatExchangeWritebacker(Writebacker):
+    """
+    Writebacker parent object for all writebacks to ThreatExchange
+    """
+
+    source = "te"
+
+    @staticmethod
+    def writeback_options() -> t.List[t.Type["Writebacker"]]:
```
eriksendc, line 121: maybe use @lrucache here?

Dcallies (commented):
This was old, but might as well hit submit.

hasher-matcher-actioner/hmalib/common/label_models.py
```diff
@@ -0,0 +1,82 @@
...
+class ActionLabel(Label):
+    key: str = field(default="Action", init=False)
+
+
+@dataclass(unsafe_hash=True)
+class WritebackLabel(Label):
+    key: str = field(default="Writeback", init=False)
+
+    def __eq__(self, other) -> bool:
+        return self.value == other.value
```
Dcallies, line 62: (blocking): This is probably incorrect for the reason of the previous comment

hasher-matcher-actioner/hmalib/common/label_models.py
```diff
@@ -0,0 +1,82 @@
...
+        return {"K": self.key, "V": self.value}
+
+    @classmethod
+    def from_dynamodb_dict(cls, d: dict):
+        return cls(d["K"], d["V"])
+
+    def __eq__(self, another_label: object) -> bool:
+        if not isinstance(another_label, Label):
+            return NotImplemented
+        return self.key == another_label.key and self.value == another_label.value
```
Dcallies, line 29: This should be the default implementation, what happened?

(You may need to do `dataclass(eq=False)` on children to prevent it from being overwritten)

hasher-matcher-actioner/hmalib/common/label_models.py
```diff
@@ -0,0 +1,82 @@
...
+class WritebackLabel(Label):
+    key: str = field(default="Writeback", init=False)
+
+    def __eq__(self, other) -> bool:
+        return self.value == other.value
+
+
+@dataclass(unsafe_hash=True)
+class SawThisTooWritebackLabel(WritebackLabel):
+    value: str = field(default="SawThisToo", init=False)
```
Dcallies, line 67: nit: These classes are basically singletons, since there's only one possible value every instance can have. Here's an alternative:

```
class WritebackLabels:
  SAW_THIS_TOO = Label("Writeback", "SawThisToo")
  FALSE_POSITIVE = Label("Writeback", "FalsePositive")
```

hasher-matcher-actioner/hmalib/common/label_models.py
```diff
@@ -0,0 +1,82 @@
...
+    key: str = field(default="BankedContentIDClassification", init=False)
+
+
+@dataclass(unsafe_hash=True)
+class ActionLabel(Label):
+    key: str = field(default="Action", init=False)
+
+
+@dataclass(unsafe_hash=True)
+class WritebackLabel(Label):
```
Dcallies, line 58: Hmm, @eriksendc, this seems to be doubling down on action labels, which I wish were not the case. I've gotten to the point where I regret the creation of the Label class :( I don't want to live in a world where everything is a label.

Dcallies (commented):
@schatten if you have cycles, can you double check the stuff with `__eq__` and dataclass?

hasher-matcher-actioner/hmalib/common/label_models.py
Dcallies: Talked with @eriksendc offline, and the plan is to eventually refactor off label (either just rename or structure heirarchy change), my concerns are addressed!

hasher-matcher-actioner/hmalib/common/label_models.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
eriksendc, line 1: Please change the name of this module back to classification_models.py. We'll be making different classes for the things that aren't classifications (like action label and writeback label)

hasher-matcher-actioner/hmalib/common/evaluator_models.py
```diff
@@ -35,5 +25,5 @@ class ActionRule(HMAConfig):
     """
 
     action_label: ActionLabel
-    must_have_labels: t.Set[Label]
-    must_not_have_labels: t.Set[Label]
+    must_have_labels: t.Set[ClassificationLabel]
+    must_not_have_labels: t.Set[ClassificationLabel]
```
eriksendc, line 29: Was there a reason you changed the typing of these two to ClassificationLabel? It's not that it's a bad thing, but you haven't then made other compensating changes elsewhere. For example, line 142 in action_evaluator.py is a function that takes in classifications as just labels to be compared with rules' must have and must not have labels. Just wondering what problem you were trying to solve, and if it is necessary, maybe do another check for instances like line 142 in action_evaluator.py.

hasher-matcher-actioner/hmalib/common/label_models.py
```diff
@@ -0,0 +1,82 @@
...
+
+logger = get_logger(__name__)
+
+
+@dataclass(unsafe_hash=True)
+class Label:
+    key: str
+    value: str
+
+    def to_dynamodb_dict(self) -> dict:
```
eriksendc, line 19: Please check master (view this file in github) and I think you'll see the to_ and from_dynamodb_dict functions were removed in my version of the file.

hasher-matcher-actioner/hmalib/common/label_models.py
eriksendc: Blocking: I don't understand the need for this definition of __eq__. The version in Label should work, right?

hasher-matcher-actioner/hmalib/common/label_models.py
eriksendc: blocking: this breaks our pattern, and like David says, they're singletons / constants. What about adding some constants to WritebackLabel instead?

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+
+    @classmethod
+    @lru_cache(maxsize=None)
+    def get_writebacker_for_source(cls, source: str) -> t.Optional["Writebacker"]:
+        if cls.__name__ != "Writebacker":
+            raise ValueError(
+                "get_writebacker_for_source can only be called from the Writebacker class directly. eg Writebacker().get_writebacker_for_source"
+            )
+
+        print("subclasses", cls.__subclasses__())
```
eriksendc, line 75: assuming print() --> logger.info() or delete.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        if source not in sources_to_writebacker_cls.keys():
+            return None
+        return sources_to_writebacker_cls[source]()
+
+    def writeback_is_enabled(self) -> bool:
+        """
+        Users can switch on/off writebacks either globally or for individual sources
+        """
```
eriksendc, line 89: Blocking: I thought that we enable writebacks per "dataset" ("collaboration")? It's not at the source level. Mengyang is adding a field in the dataset config.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+    def perform_writeback(self, writeback_message: WritebackMessage) -> str:
+        writeback_options = self.writeback_options()
+        writeback_to_perform = writeback_message.writeback_label.value
+        if writeback_to_perform not in writeback_options:
+            return (
+                "Could not find writebacker for source "
+                + self.source
+                + " that can perform writeback "
+                + writeback_to_perform
+            )
```
eriksendc, line 111: Will this return raise an error or log something somewhere? I think if this kind of situation exists we'd want to surface an error in the logs, right?

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+
+        writebacker = writeback_options[writeback_to_perform]()
+        if writebacker.writeback_is_enabled:
+            return writebacker._writeback_impl(writeback_message)
+        return "Writeback {writebacker.__name__} not performed becuase it switched off"
```
eriksendc, line 116: it's

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+    Writebacker parent object for all writebacks to ThreatExchange
+    """
+
+    source = "te"
+
+    @staticmethod
+    @lru_cache(maxsize=None)
+    def writeback_options() -> t.Dict[str, t.Type["Writebacker"]]:
+        return {
+            "FalsePositive": ThreatExchangeFalsePositiveWritebacker,
```
eriksendc, line 131: The constants you define in WritebackLabel could be used here.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        TODO implement
+        Looks up from a config whether ThreatExchange writing back is enabled. Initially this will be a global
+        config, and this method will return True if writing back is enabled, False otherwise. At some point the
+        config for writing back to ThreatExchange may be on a per collaboration basis. In that case, the config
+        will be referenced for each collaboration involved (implied by the match message). If writing back
+        is enabled for a given collaboration, a label will be added to the match message
+        (e.g. "ThreatExchangeWritingBackEnabled:<collaboration-id>").
+        """
+        return True
```
eriksendc, line 147: I'm remembering that the team already decided that the granularity of enabling reactions to TE is at the collaboration level, hence Mengyang's work to add the toggle on a per collaboration basis. Please update the TODO comment to reflect this. When we implement, we just need to get all the configs and use that. But this will make things more complicated... You'll need to pass in the message and output a dictionary of "data set id" to bool to use later in what gets written back. I think you should do some deeper analysis of how your dynamic code works in this case. Will the necessary output of this method flow through?

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+        if writeback_to_perform not in writeback_options:
+            return (
+                "Could not find writebacker for source "
+                + self.source
+                + " that can perform writeback "
+                + writeback_to_perform
+            )
+
+        writebacker = writeback_options[writeback_to_perform]()
+        if writebacker.writeback_is_enabled:
```
eriksendc, line 114: Gating here won't work. You heed to deal with writebacks happening at a per-dataset level. I think this changes your pattern enough that you may want to be sure you can easily refactor.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -0,0 +1,240 @@
...
+            self.te_api.react_to_threat_descriptor(id, self.reaction)
+            logger.info("reacted %s to descriptor %s", self.reaction, id)
+        return (
+            "reacted "
+            + self.reaction
+            + " to "
+            + str(len(descriptor_ids))
+            + " descriptors"
+        )
+
```
eriksendc, line 229: Is _writeback_impl going to persist anything to be displayed in the UI? Seems like it should, like we still need in actioning.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -241,6 +222,4 @@ def get_threat_exchange_reaction_labels(
 
     match_message = MatchMessage("key", "hash", [banked_signal])
 
-    action_label_to_action_rules = get_actions_to_take(match_message, action_rules)
-
-    print(action_label_to_action_rules)
+    action_label_to_action_rules = get_actions_to_take(match_message, action_rules)
```
eriksendc, line 225: Looks like this isn't formatted with black. There should be an ending newline. Also, why nix the print statement? This was my debugging area.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -53,7 +54,7 @@ def get(cls):
         HMAConfig.initialize(os.environ["CONFIG_TABLE_NAME"])
         return cls(
             actions_queue_url=os.environ["ACTIONS_QUEUE_URL"],
-            reactions_queue_url=os.environ["REACTIONS_QUEUE_URL"],
+            writebacks_queue_url=os.environ["WRITEBACKS_QUEUE_URL"],
```
eriksendc, line 57: Not seeing the compensating change in terraform code for the change in this environment variable.

eriksendc (changes requested):
Once comments are addressed and end-to-end testing is done let me know and I'll re-review.

hasher-matcher-actioner/hmalib/common/label_models.py
jeberl: Fine with which ever but if the plan is to get rid of these labels anyway I dont think its worth doing in this PR

hasher-matcher-actioner/hmalib/common/writebacker_models.py
jeberl: Adding more logging through out this file

hasher-matcher-actioner/hmalib/common/writebacker_models.py
jeberl: I've tested overriding this property and it seems to work. I'm planning on implementing in a future PR. Can we see then?

hasher-matcher-actioner/hmalib/common/writebacker_models.py
jeberl: Updated the comment

I see your point about the refactor however I dont think it will be a heavy lift and I'd like to get this PR our soon

hasher-matcher-actioner/hmalib/common/writebacker_models.py
jeberl: Could do. Dont think thats on the v0 mural though

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
jeberl: Added and finished full e2e test

hasher-matcher-actioner/hmalib/common/writebacker_models.py
eriksendc: But as an error or just info?

hasher-matcher-actioner/hmalib/common/writebacker_models.py
eriksendc: yep

hasher-matcher-actioner/hmalib/common/writebacker_models.py
eriksendc: k

eriksendc (approved):
Approving to not block. Seems like the system error in `perform_writeback()` (if the writeback to perform isn't one of the expected ones) then that should log an error, not simply return. And it should probably raise an error, not return.

eriksendc:
Also, seeing a couple of broken tests, so please resolve that before merging.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
jeberl: Oops. Thought I added the print. Will add back

Black isn't complaining for me

PR #535, opened 2021-04-28
[hma] Submit Content API

BarrettOlson:
Summary
---------

Creates a new API by which a user can upload content to the system. Functions very similarly to  `/upload` 

this new endpoint will be `/submit/` and will take the following fields:
- submission_type: (always UPLOAD for now)
- content_id: (formerly   only used filename: this will correspond to s3 key (+ image prefix)
- content_type: str (always PHOTO for now)
- content_ref: encoded bytes of the image
- metadata: optional additional fields

Test Plan
---------
Used Postman and built out a UI that will be visible in my next PR (#536) to fully test upload functionality. Works just like the existing upload page except you can pick (must specify) an id for the content, instead of just using the filename.

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
@@ -3,6 +3,7 @@
 import bottle
 import boto3
 import base64
+import json
 
 from dataclasses import dataclass, asdict
 from mypy_boto3_dynamodb.service_resource import Table
@@ -38,8 +39,24 @@ class SubmitContentRequestBody(DictParseable):
 
 @dataclass
 class SubmitContentResponse(JSONifiable):
-    content_id_submitted: str
+    content_id: str
     submit_successful: bool
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+
+@dataclass
+class SubmitContentError(JSONifiable):
+    """
+    Warning: by defualt this will still return 200
+    you need to update bottle.response.status
+    if you want a specific return code.
+    ToDo update middleware.py to handle this.
+    """
+
+    content_id: str
+    message: str
 
     def to_json(self) -> t.Dict:
         return asdict(self)
@@ -59,7 +76,9 @@ def get_submit_api(
     submit_api = bottle.Bottle()
 
     @submit_api.post("/", apply=[jsoninator(SubmitContentRequestBody)])
-    def submit(request: SubmitContentRequestBody) -> SubmitContentResponse:
+    def submit(
+        request: SubmitContentRequestBody,
+    ) -> t.Union[SubmitContentResponse, SubmitContentError]:
         """
         Endpoint to allow for the general submission of content to the system
         """
@@ -78,12 +97,14 @@ def get_submit_api(
             )
 
             return SubmitContentResponse(
-                content_id_submitted=request.content_id, submit_successful=True
+                content_id=request.content_id, submit_successful=True
             )
 
         # Other possible submission types are not supported so just echo content_id for testing
-        return SubmitContentResponse(
-            content_id_submitted=request.content_id, submit_successful=False
+        bottle.response.status = 422
+        return SubmitContentError(
+            content_id=request.content_id,
+            message="submission_type not yet supported",
         )
 
     return submit_api
--- a/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
@@ -1,0 +1,89 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import bottle
+import boto3
+import base64
+
+from dataclasses import dataclass, asdict
+from mypy_boto3_dynamodb.service_resource import Table
+import typing as t
+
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -0,0 +1,89 @@
...
+s3_client = boto3.client("s3")
+dynamodb = boto3.resource("dynamodb")
+
+
+@dataclass
+class SubmitContentRequestBody(DictParseable):
+    submission_type: str  # TODO Enum
+    content_id: str
+    content_type: str  # TODO Enum
+    content_ref: t.Union[str, bytes]
```
eriksendc, line 24: I think another kind of submission might take a URL to the content. If we did that, maybe we could use content_ref for the URL and a submission type of, say, URL? More of a conversation / idea sharing comment.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -0,0 +1,89 @@
...
+            d["content_id"],
+            d["content_type"],
+            d["content_ref"],
+            d["metadata"],
+        )
+
+
+@dataclass
+class SubmitContentResponse(JSONifiable):
+    content_id_submitted: str
```
eriksendc, line 41: can we change this to content_id? I think in most APIs I've seen you keep a name consistent if it represents the same thing.

eriksendc (changes requested):

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -0,0 +1,89 @@
...
+    A Closure that includes all dependencies that MUST be provided by the root
+    API that this API plugs into. Declare dependencies here, but initialize in
+    the root API alone.
+    """
+
+    # A prefix to all routes must be provided by the api_root app
+    # The documentation below expects prefix to be '/submit/'
+    submit_api = bottle.Bottle()
+
+    @submit_api.post("/", apply=[jsoninator(SubmitContentRequestBody)])
```
schatten, line 61: Woohoo! `jsoninator` gets its first RequestType usage!

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -0,0 +1,89 @@
...
+                Body=fileContents,
+                Bucket=image_bucket_key,
+                Key=f"{image_folder_key}{fileName}",
+            )
+
+            return SubmitContentResponse(
+                content_id_submitted=request.content_id, submit_successful=True
+            )
+
+        # Other possible submission types are not supported so just echo content_id for testing
```
schatten, line 84: Throw a 400 instead? Or a [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) if you're feeling like it.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -0,0 +1,89 @@
...
+    @submit_api.post("/", apply=[jsoninator(SubmitContentRequestBody)])
+    def submit(request: SubmitContentRequestBody) -> SubmitContentResponse:
+        """
+        Endpoint to allow for the general submission of content to the system
+        """
+
+        assert isinstance(request, SubmitContentRequestBody)
+        logger.debug(f"Content Submit Request Received {request.content_id}")
+
+        if request.submission_type == "UPLOAD":
```
schatten, line 70: Food for thought. If instead of upload being the sole submission type, we supported 'URL', we could hit two birds with one stone.

S3 uploads can be painlessly achieved on the client side using amplify [as described here](https://docs.amplify.aws/lib/storage/upload/q/platform/js#browser-uploads).

--- 

Also, b64 encoding adds processing overhead on client and on server. Perhaps multi-part forms would be more up your alley? You can't use JSON for serializing metadata, but you can upload file + metadata. Here's a [bottle example](https://stackoverflow.com/questions/15050064/how-to-upload-and-save-a-file-using-bottle-framework).

schatten (changes requested):
From a code organisation and structure perspective, this is great.

I have a few suggestions about the request payload. b64 encoding is resource consuming esp. for larger files. It will cause the object to be loaded into the javascript runtime on the clientside and prevent streaming on the server.

Would advise we use multi-part streaming or two phase submission. I have linked to examples for both.

schatten (approved):
Based on offline discussion, and [Amplify's weirdness](https://stackoverflow.com/questions/57855644/aws-amplify-breaks-binary-posts-to-aws-api-gateway), approving the status quo for using b64 to encode and upload file content!

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
BarrettOlson: Yes, that is the plan!

BarrettOlson:
Thanks @eriksendc and thanks @schatten for accepting given the context, URL support sounds like something to prioritize so I'll work on that next (after the UI+storage) :)  

Going to merge, delta since last review: 
- Changed response to have `content_id`
- Also made it return 422 for not yet supported submission_types because I was feeling it (found small bug and fixed it in middleware.py because of it)

PR #536, opened 2021-04-28
[hma][ui] Connect Submit Content API to new UI Page

BarrettOlson:
Summary
---------

Builds off of the API Created in #535

URL Support and recording the metadata are both critical features that I will work on adding. 

For the purpose of this PR this page is a slightly nicer upload page with room to grow :)  

Thanks! 

Test Plan
---------

Demo video: (bridge was a repeat upload else the video would have to wait 2-3 minutes for the hash and matches to show up)

https://user-images.githubusercontent.com/7664526/116619587-a3f04180-a90e-11eb-898f-16ad0c9ff283.mov

### DIFF
--- a/hasher-matcher-actioner/webapp/src/components/SubmitContentFields.jsx
+++ b/hasher-matcher-actioner/webapp/src/components/SubmitContentFields.jsx
@@ -1,0 +1,243 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+import React from 'react';
+import {
+  Col,
+  Button,
+  Form,
+  InputGroup,
+  Image,
+  Accordion,
+  Card,
+} from 'react-bootstrap';
+import PropTypes from 'prop-types';
+
+import {SUBMISSION_TYPE} from '../utils/constants';
+
+const inputsShape = {
+  submissionType: PropTypes.oneOf(Object.keys(SUBMISSION_TYPE)),
+  contentId: PropTypes.string,
+  contentType: PropTypes.string,
+  contentRef: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.shape({raw: PropTypes.file, preview: PropTypes.string}),
+  ]),
+};
+
+export function ContentIdAndTypeField({inputs, handleInputChange}) {
+  return (
+    <Form.Group>
+      <Form.Row>
+        <Form.Label>Content ID and Type</Form.Label>
+        <InputGroup>
+          <Form.Control
+            onChange={handleInputChange}
+            type="text"
+            name="contentId"
+            placeholder="Enter a unique identifier for content"
+            required
+            value={inputs.contentId}
+          />
+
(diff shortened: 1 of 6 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -9,7 +9,6 @@
 import typing as t
 from apig_wsgi import make_lambda_handler
 from bottle import response, error
-from enum import Enum
```
BarrettOlson, line 12: moved this here
 https://github.com/facebook/ThreatExchange/pull/536/commits/1241af8380e82e09a1c39fd4fb118c3fabdba056#diff-ec198432ec143dc2d5426b6e26a45bc5c364ee67f28e659f7bc44f6eba3f1546L133

hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
```diff
@@ -0,0 +1,346 @@
...
+  Card,
+  Spinner,
+} from 'react-bootstrap';
+import {submitContent, submitContentUpload} from '../Api';
+
+const SUBMISSION_TYPE = Object.freeze({
+  UPLOAD: 'Direct Upload',
+  RAW: 'Raw Value (example only)',
+  S3_OBJECT: 'S3 Object (example only)',
+  PRESIGNED_URL: 'Presigned URL (example only)',
```
eriksendc, line 25: I think just URL might be an option, too. If a partner is using a VPC, they might not bother with needing to go through the trouble of pre-signing a URL. Also, signed URLs may expire. It might work best for a partner for the URL to be available longer / indefinitely... Anyways, I suggest adding URL as another submission type.

eriksendc (changes requested):

hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
BarrettOlson: thanks for the feedback, this PR was still in draft mode when you reviewed so the history inline won't be tracked because I am force pushing.

BarrettOlson: changed PRESIGNED_URL to just URL for now (doesn't matter because neither work yet haha)

schatten (approved):
LGTM.

hasher-matcher-actioner/webapp/src/components/SubmitContentFields.jsx
```diff
@@ -0,0 +1,243 @@
...
+      <Form.Text className="text-muted mt-0">
+        Warning currently behavior will overwrite content with the same id
+      </Form.Text>
+    </Form.Group>
+  );
+}
+
+ContentIdAndTypeField.propTypes = {
+  inputs: PropTypes.shape(inputsShape),
+  handleInputChange: PropTypes.func,
```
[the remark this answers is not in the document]
BarrettOlson: thanks!

PR #539, opened 2021-04-29
devcontainers: infer unixname from env; add github cli

schatten:
Summary
---------
Finally makes it possible to not have changes to the devcontainer file which reduces the possibility of merge conflicts. 

Test Plan
---------
Rebuilt container and verified that the default user when I open terminal is my unixname.


BarrettOlson (approved):
Nice!

PR #540, opened 2021-04-30
Storm submissions API with the same image repeatedly

schatten:
Summary
---------
* Do not be alarmed by the access token. It is an expired JWT and non-reverse-engineer-able.
* Also slight adjustment to the README to allow for redeploying of webapp

Test Plan
---------
Copy pasted script on an ec2 instance. Ran it with a violating image. Verified that matches appeared on the matches page ~15 seconds later.

### DIFF
--- a/hasher-matcher-actioner/scripts/storm_submissions_api
+++ b/hasher-matcher-actioner/scripts/storm_submissions_api
@@ -7,6 +7,11 @@ Storms the submissions API for a running HMA instance with images.
 Try to resist the urge to add any hmalib dependencies here. This script can be
 run on a server which does not have hmalib.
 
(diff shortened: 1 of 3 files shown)


BarrettOlson (approved):
Approving to unblock.

One thing I'd really like to see either in the PR summary or even the docstring is example usage. (with dummy values for API urls)  e.g. what does the help text look like in the terminal and what should be copied if I wanted to run it lazily using copy pasta+replace. 

Regression tests would also be nice as this will likely need to be changed soon because I think submission_type "UPLOAD" is going to become two phase. However we can probably go without for now because I can be mindful of it and handle the adjustments with the other changes post merge. Thanks!

PR #542, opened 2021-05-03
Add zsh to devcontainer, track history for it

schatten:
Summary
---------
* also create bashrc and zshrc files that are copied at build
  * Allows further customizing of prompts and general shell goodness

Test Plan
---------
- Rebuilt container fired it up.
- verified that bash history is retained by tailing bash_history on host
- verified that zsh history is retained by tailing zsh_history on host after switching to zsh on guest


BarrettOlson (approved):
LGTM. 

We will probably want a '.tmux.config' as well if that ends up working (didn't see it in the test plan), unless you like the defaults :)

schatten:
> We will probably want a '.tmux.config' as well if that ends up working (didn't see it in the test plan), unless you like the defaults :)

Haven't used tmux at all. I'll leave that as an exercise to you. :)

PR #544, opened 2021-05-03
Create datasets api

mengyangwang:
Summary
---------
previous PR : https://github.com/facebook/ThreatExchange/pull/516
Task : T88063259, this task is to create ThreatExchange setting page, This PR is to create delete api to be used in ThreatExchange setting page

Future Actions
---------
 Create UI page to use the delete api 

Test Plan
---------

1. black hmalib
2. mypy hmalib
4. make docker
5. make upload_docker 
6.  terraform apply
7. test delete api in local and verify the result in DynamoDB

https://user-images.githubusercontent.com/81996660/116912102-c2f41980-ac15-11eb-8c99-89f0afd4b5c6.mov


hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
```diff
@@ -58,7 +58,7 @@ def to_json(self) -> t.Dict:
 
 
 @dataclass
-class SyncDatasetResponse(JSONifiable):
+class SyncDeleteDatasetResponse(JSONifiable):
```
schatten, line 61: nit: this seems like it is for an API that syncs-deletes  which is probably not what you were going for? Perhaps some other name would do better?

hasher-matcher-actioner/terraform/api/main.tf
```diff
@@ -83,7 +84,7 @@ data "aws_iam_policy_document" "api_root" {
   }
   statement {
     effect    = "Allow"
-    actions   = ["dynamodb:GetItem", "dynamodb:Query", "dynamodb:Scan", "dynamodb:PutItem", "dynamodb:UpdateItem"]
+    actions   = ["dynamodb:GetItem", "dynamodb:Query", "dynamodb:Scan", "dynamodb:PutItem", "dynamodb:DeleteItem"]
```
schatten, line 87: You need update AND delete, right?

schatten (approved):
Approve to unblock. If you have a strong reason to remove updateItem privs, go ahead.

There is a nit that you may want to address, but totally up to you.

hasher-matcher-actioner/terraform/api/main.tf
mengyangwang: actually for now we just need put_item and delete_item, update_item is not needed for now

hasher-matcher-actioner/hmalib/lambdas/api/datasets_api.py
mengyangwang: yeah I agree! I will change this before I merge it to master

PR #545, opened 2021-05-03
[hma] enable naive url upload to submit api

BarrettOlson:
Summary
---------

#536 Added upload file. This PR follow the same path but has the API pull filecontents from a url and store it in s3.

Test Plan
---------

https://user-images.githubusercontent.com/7664526/116917105-39941580-ac1c-11eb-8fe4-2fce2c42b269.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
@@ -142,3 +142,5 @@ def get_submit_api(
                 content_id=request.content_id,
                 message="submission_type not yet supported",
             )
+
+    return submit_api
--- a/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
@@ -96,7 +96,7 @@ def get_submit_api(
         assert isinstance(request, SubmitContentRequestBody)
         logger.debug(f"Content Submit Request Received {request.content_id}")
 
-        if request.submission_type == "UPLOAD":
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -99,12 +109,38 @@ def submit(
...
-        )
+        elif request.submission_type == "URL":
+            fileName = request.content_id
+            url = request.content_ref
+            response = requests.get(url)
+            # TODO better checks that the URL actually worked...
+            if response and response.content:
+                # TODO a whole bunch more validation and error checking...
+
+                # Right now this makes a local copy in s3 but future changes to
```
schatten, line 120: This is great. You have identified the path of least resistance. I would say how the submit api and the hasher lambdas works together is an internal impl detail. It is perfectly okay to NOT give the api/user an option.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -99,12 +109,38 @@ def submit(
...
                 content_id=request.content_id, submit_successful=True
             )
-
-        # Other possible submission types are not supported so just echo content_id for testing
-        bottle.response.status = 422
-        return SubmitContentError(
-            content_id=request.content_id,
-            message="submission_type not yet supported",
-        )
+        elif request.submission_type == "URL":
```
schatten, line 112: Use `SubmissionType.URL.name` instead?

schatten (approved):
minor nits aside, this is a great way to enable features without redesign or substantial changes. MOVE FAST!

PR #543, opened 2021-05-03
Actions rule config UI

eriksendc:
Summary
---------

This PR introduces a first draft of the Action Rules (in Settings). 

This PR is organized as a "stacked diff". It's easiest to review per commit:

1. Added the concept of a route at the tab level so that page refreshes don't go back to the default tab.
2. Added the Action Rules tab without any real functionality (just getting the main component in place).
3. Added the main functionality: a table with the rules configured so far; the ability to add a rule; editing and deleting a rule is mocked (incomplete functionality).
4. Added validation: Name required and must be unique; Labeled As required; Action required.
5. (skip - just fixing formatting)
6. Added delete functionality.
7. Added edit functionality.

Test Plan
---------

Run the web app, navigate to Settings --> Action Rules then exercise the UI. You should see what's captured in the attached video.

https://user-images.githubusercontent.com/8974354/116909103-b79eef00-ac11-11eb-933a-28e7bab014f0.mp4

PR #546, opened 2021-05-03
Create threat exchange tab

mengyangwang:
Summary
---------
previous PR : https://github.com/facebook/ThreatExchange/pull/516 , https://github.com/facebook/ThreatExchange/pull/544
Task : T88063259
     this task is to create ThreatExchange setting page.
     this PR is to create the UI page of ThreatExchange 

Future Actions
---------
 1. Change buttons to icons and adding more tooltips
 2. Add copy text feature for privacy group name(currently has style conflict, might need to rewrite "CopyableTextField" module later)
 3. Add matcher active feature to disable matching from privacy group level 

Test Plan
---------
test localhost UI changes
1. test show all privacy groups with loading spin
2. test sync function, it should automatically add privacy groups
3. test update function for each active privacy group
4. test delete function for in_active privacy group

https://user-images.githubusercontent.com/81996660/116940407-edf16400-ac3b-11eb-9c60-4e1eb65ca4e2.mov

### DIFF
--- a/hasher-matcher-actioner/webapp/src/settings/ThreatExchangeSettingsTab.jsx
+++ b/hasher-matcher-actioner/webapp/src/settings/ThreatExchangeSettingsTab.jsx
@@ -1,0 +1,244 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+import React, {useEffect, useState} from 'react';
+import {
+  Col,
+  Row,
+  Button,
+  Toast,
+  Form,
+  Spinner,
+  Tooltip,
+  OverlayTrigger,
+  Card,
+} from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import {CopyableTextField} from '../utils/TextFieldsUtils';
+import {
+  fetchAllDatasets,
+  syncAllDatasets,
+  udpateDataset,
+  deleteDataset,
+} from '../Api';
+
+export default function ThreatExchangeSettingsTab() {
+  const [datasets, setDatasets] = useState([]);
+  const [loading, setLoading] = useState(true);
+  const [syncing, setSyncing] = useState(true);
+  const [showToast, setShowToast] = useState(false);
+  const [toastBody, setToastBody] = useState(null);
+  const onPrivacyGroupSave = privacyGroup => {
+    udpateDataset(
+      privacyGroup.privacyGroupId,
+      privacyGroup.localFetcherActive,
+      privacyGroup.localWriteBack,
+    ).then(response => {
+      datasets[
+        datasets.findIndex(
+          item => item.privacy_group_id === response.privacy_group_id,
+        )
+      ] = response;
+      setDatasets(datasets);
+    });
+  };
+  const onPrivacyGroupDelete = privacyGroupId => {
+    deleteDataset(privacyGroupId).then(() => {
+      const filteredDatasets = datasets.filter(
+        item => item.privacy_group_id !== privacyGroupId,
+      );
+      setDatasets(filteredDatasets);
+    });
+  };
+  const onSync = () => {
+    syncAllDatasets().then(syncResponse => {
+      setSyncing(true);
+      if (syncResponse.response === 'Dataset is update-to-date') {
+        fetchAllDatasets().then(response => {
+          setDatasets(response.datasets_response);
+        });
+      } else {
+        alert('Errors when syncing privacy groups. Please try again later');
+      }
+    });
+  };
+  useEffect(() => {
+    fetchAllDatasets(setLoading(false)).then(response => {
+      setLoading(true);
+      setDatasets(response.datasets_response);
+    });
+  }, []);
+  return (
+    <>
+      <Card.Header>
+        <Row className="mt-3">
+          <h2 className="mt-2">ThreatExchange Privacy Groups </h2>
+          <OverlayTrigger
+            key="syncButton"
+            placement="right"
+            overlay={
+              <Tooltip id="tooltip-right">
+                Fetch privacy groups from ThreatExchange
+              </Tooltip>
+            }>
+            <Button
+              variant="primary"
+              onClick={() => {
+                setSyncing(false);
+                setToastBody('Privacy groups are up to date!');
+                setShowToast(true);
+                onSync();
+              }}
+              style={{marginLeft: 10}}>
+              <Spinner
+                hidden={syncing}
+                as="span"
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/webapp/src/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -0,0 +1,244 @@
...
+} from '../Api';
+
+export default function ThreatExchangeSettingsTab() {
+  const [datasets, setDatasets] = useState([]);
+  const [loading, setLoading] = useState(true);
+  const [syncing, setSyncing] = useState(true);
+  const [showToast, setShowToast] = useState(false);
+  const [toastBody, setToastBody] = useState(null);
+  const onPrivacyGroupSave = privacyGroup => {
+    udpateDataset(
```
schatten, line 33: nit: update spelling is wrong.

schatten: nonblocking:
As @eriksendc's comment yesterday. You could change the state here so that user's can't click the button while the API requests is processing.

Or, you could just update the state assuming the API call would succeed. This is called optimistic updates [1]. In essence, update the state 'before' the API call succeeds. In the then() block, you 'reset' the state and do a failed toast.

1: https://medium.com/@_erikaybar/optimistic-ui-updates-in-react-9e139ffa2e45

hasher-matcher-actioner/webapp/src/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -0,0 +1,244 @@
...
+      const filteredDatasets = datasets.filter(
+        item => item.privacy_group_id !== privacyGroupId,
+      );
+      setDatasets(filteredDatasets);
+    });
+  };
+  const onSync = () => {
+    syncAllDatasets().then(syncResponse => {
+      setSyncing(true);
+      if (syncResponse.response === 'Dataset is update-to-date') {
```
schatten, line 57: Just use the status code? The response string is rarely part of the contract and introduces an implicit dependency.

schatten: IMO, if you get a 'then' call, you can assume the API call succeeded. If it were to fail, the `catch` method would be called.

The pattern I have seen most often: (I am biased. Not an expert)
```javascript
const onSaveHandler = () => {
  makeAPICall()
    .then(() => {
       ...updateState()
       ...successMessage()
    })
    .catch(() => {
       ...resetState()
       ...failedMessage()
    })

}

```

hasher-matcher-actioner/webapp/src/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -0,0 +1,244 @@
...
+                Fetch privacy groups from ThreatExchange
+              </Tooltip>
+            }>
+            <Button
+              variant="primary"
+              onClick={() => {
+                setSyncing(false);
+                setToastBody('Privacy groups are up to date!');
+                setShowToast(true);
+                onSync();
```
schatten, line 91: nit: All of this should be in the onSync() method.. 

No matter which component calls onSync(), these things would happen. Doing it in a function also allows you to later centralize common behaviors.

hasher-matcher-actioner/webapp/src/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -0,0 +1,244 @@
...
+              </div>
+            )}
+            {inUse ? null : (
+              <div>
+                <Button
+                  variant="secondary"
+                  onClick={() => {
+                    setToastBody('The privacy group is deleted!');
+                    setShowToast(true);
+                    onDelete(privacyGroupId);
```
schatten, line 221: nit: Same comment as before. Try to centralize behaviors in the handler functions.

I would even suggest allowing the top-level component `ThreatExchangeSettingsTab` to determine whether toasts are to be shown at all or not. 

Low level component like a card should not be able to affect state on their parents. If you find a component altering state of its siblings, reconsider. Since `ThreatExchangeSettingsTab` renders the `Toast` and the `ThreatExchangePrivacyGroupCard`, they are siblings.

Some of this is just experience with react. You may not find this in the standard guides.

schatten (approved):
Requesting changes, but these are not binding, feel free to revisit in a later PR.

I am yet to articulate this in the README, but here's a stab:
- within webapp/src, there are two directories, `components/` and `pages/`.
- All top-level pages should go into `pages/`, this includes settings.
  - A top-level page is one that calls APIs eg. settings, matches, dashboard
- Components required by those pages that are either large-enough or re-useable can be moved to `components/`

# Other nits
* As noted inline, I am concerned the cards are able to create toasts and set the body on their own.
* `ThreatExchangeSettingsTab` owns the cards and the toasts. It should determine when and what toasts to show.

BarrettOlson:
> Requesting changes, but these are not binding, feel free to revisit in a later PR.
> 
> I am yet to articulate this in the README, but here's a stab:
> 
> * within webapp/src, there are two directories, `components/` and `pages/`.
> * All top-level pages should go into `pages/`, this includes settings.
>   
>   * A top-level page is one that calls APIs eg. settings, matches, dashboard
> * Components required by those pages that are either large-enough or re-useable can be moved to `components/`
> 
> # Other nits
> * As noted inline, I am concerned the cards are able to create toasts and set the body on their own.
> * `ThreatExchangeSettingsTab` owns the cards and the toasts. It should determine when and what toasts to show.

@schatten I just raised a similar point about file structure in #543

mengyangwang:
> > Requesting changes, but these are not binding, feel free to revisit in a later PR.
> > I am yet to articulate this in the README, but here's a stab:
> > 
> > * within webapp/src, there are two directories, `components/` and `pages/`.
> > * All top-level pages should go into `pages/`, this includes settings.
> >   
> >   * A top-level page is one that calls APIs eg. settings, matches, dashboard
> > * Components required by those pages that are either large-enough or re-useable can be moved to `components/`
> > 
> > # Other nits
> > 
> > * As noted inline, I am concerned the cards are able to create toasts and set the body on their own.
> > * `ThreatExchangeSettingsTab` owns the cards and the toasts. It should determine when and what toasts to show.
> 
> @schatten I just raised a similar point about file structure in #543

I think we can move src/settings to src/pages/settings, I can do this in a separate diff

PR #547, opened 2021-05-04
Implement disabling/enabling writebacks

jeberl:
Summary
---------

Writebacks can be enabled disabled at the dataset (privacy group) level. 
Changes
1. Small refactor of writebacks to enable this. Now when evaluating a match for writebacks, each piece of matched content can be written back separately so some matched content might writeback and others might now
2. Implemented mocking functionality for HMAConfigs. You can now use the function `mock_create_config` which will simulate a dynamodb write
3. Implemented a function `cached_get` for HMAConfigs which is identical to `get` but with a cache

Test Plan
---------

Added test files
`pytest`

Also e2e tested


hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -29,6 +29,8 @@
 from hmalib.common.logging import get_logger
 from hmalib.common.s3_adapters import S3ThreatDataConfig
 from hmalib.common.signal_models import PDQSignalMetadata
+from hmalib.common.fetcher_modules import ThreatExchangeConfig
```
mengyangwang, line 32: I think there are some other places that we need to import ThreatExchangeConfig. For example, in populate_config_db.py, line 19 "from hmalib.lambdas.fetcher import ThreatExchangeConfig". Do we need to change all of them?

jeberl: Good call out. I wonder why mypy didn't realize that

mengyangwang (approved):

hasher-matcher-actioner/hmalib/common/fetcher_modules.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
eriksendc, line 1: Was the file name meant to be _modules.py? (as opposed to _models.py)

hasher-matcher-actioner/hmalib/common/classification_models.py
```diff
@@ -16,10 +16,21 @@ class Label:
...
+            isinstance(other, Label)
+            and self.key == other.key
+            and self.value == other.value
+        )
+
 
 @dataclass(unsafe_hash=True)
 class ClassificationLabel(Label):
-    key: str = field(default="Classification", init=False)
+    key: str
```
eriksendc, line 33: As discussed, let's put things back the way they were a few PRs ago:

1. Put this back to having the defaulted field
2. Put the type back to Label for the sets in ActionRule
3. Change initializations of Classification objects to have a single parameter.
4. Anything else to get back to the previous state w.r.t. this topic.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -137,13 +149,13 @@ def writeback_options() -> t.Dict[
...
-        Looks up from a config whether ThreatExchange writing back is enabled. Initially this will be a per
-        collaboration lookup
-        """
-        return True
+    def writeback_is_enabled(self, writeback_signal: BankedSignal) -> bool:
+        privacy_group_id = writeback_signal.bank_id
+        privacy_group_config = ThreatExchangeConfig.cached_get(privacy_group_id)
+        if isinstance(privacy_group_config, ThreatExchangeConfig):
+            return privacy_group_config.write_back
+        # If no config, dont write back
```
eriksendc, line 157: Not finding a config warrants a logger.warn() here.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -118,6 +120,7 @@ def get_actions_to_take(
             if action_rule_applies_to_classifications(
                 action_rule, banked_signal.classifications
             ):
+                print("rule applies to classifications")
```
eriksendc, line 123: Nix print()

eriksendc (changes requested):
Great stuff! Please address or rebut comments and then I can take another look.

eriksendc (approved):
LGTM!

schatten:
@jeberl I have the same [concerns as David](https://github.com/facebook/ThreatExchange/pull/534/files/b0cffa5e93528276aa0ea14be02f0466bad7474f#r622456728) had in #534 with [__eq__](https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/hmalib/common/classification_models.py#L19-L28) in `Label`.

Can we address this some other way instead? Perhaps a utility method that checks attributes?

This change is hard to go back on because we might introduce implicit, untested dependencies as we write more code. So if we discover something later, it will be painful to fix.

jeberl:
@schatten Would like to discuss this more because I think, as we're currently thinking about Labels, this is what we think of for equality. I would argue there is no conceptual difference between `Label("BankIDClassification" "bank 4")` and `BankIDClassificationLabel("bank 4")`

 The reason I chose to override eq was the exact reason you don't want to override it - bugs that are really difficult to detect. Specifically, the issue was when we serialize labels they aren't always deserialized as the same class they we're before

I see that our concept of Labels could change but i'm not seeing how it might change our concept of equality. Let me know and lets talk through it more

PR #549, opened 2021-05-04
Store actions

jeberl:
Summary
---------
 # How to review this PR
Only review the last commit: https://github.com/facebook/ThreatExchange/pull/549/commits/ab0d2b0952fa3af0c76f4d83b43ae40330880a95

Other commits are from #547

# Summary 
Store the result of actions successfully performed in dynamo db. Specifically we create a PDQActionRecord that stores the hash, the updated time, the signal type and the label of the action taken

Test Plan
---------

e2e test:
1. `make upload docker`
2. update lambdas: `sha=$(aws ecr describe-images --repository-name hma-lambda-dev --image-ids imageTag=$(whoami) | jq -r '.imageDetails[0].imageDigest') && terraform -chdir=terraform apply -var "hma_lambda_docker_uri=521978645842.dkr.ecr.us-east-1.amazonaws.com/hma-lambda-dev@$sha"`
3. from localhost upload image
4. after ~7 minutes watch the action record appear in the DynamoDB

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/classification_models.py
+++ b/hasher-matcher-actioner/hmalib/common/classification_models.py
@@ -30,7 +30,7 @@ class Label:
 
 @dataclass(unsafe_hash=True)
 class ClassificationLabel(Label):
-    key: str
+    key: str = field(default="Classification", init=False)
 
 
 @dataclass(unsafe_hash=True)
--- a/hasher-matcher-actioner/hmalib/common/classification_models.py
+++ b/hasher-matcher-actioner/hmalib/common/classification_models.py
@@ -16,10 +16,21 @@ class Label:
     key: str
     value: str
 
+    def __eq__(self, other) -> bool:
+        # Concrete type doesnt matter for equality, only key, value
+        # eg the following are equal:
+        #   ClassificationLabel("Classification", "true_positive")
+        #   Label("Classification", "true_positive")
+        return (
+            isinstance(other, Label)
+            and self.key == other.key
+            and self.value == other.value
+        )
+
 
 @dataclass(unsafe_hash=True)
 class ClassificationLabel(Label):
-    key: str = field(default="Classification", init=False)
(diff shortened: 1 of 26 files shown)


hasher-matcher-actioner/hmalib/common/classification_models.py
```diff
@@ -16,10 +16,21 @@ class Label:
...
+        return (
+            isinstance(other, Label)
+            and self.key == other.key
+            and self.value == other.value
+        )
+
 
 @dataclass(unsafe_hash=True)
 class ClassificationLabel(Label):
-    key: str = field(default="Classification", init=False)
```
jeberl, line 22: Need this because it was overriding child classes when deserialized 

eg.
ClassificationLabel("BankIDClassification", "bank 4") -> Label("Classification", "bank 4")

eriksendc (approved):
I approve the changes to the very last commit

BarrettOlson (changes requested):
Happy to prioritize a review of this but first can you please update the summary and test plan with a bit more context? 
I take it only the top commit is to be reviewed because the rest are from another PR. can you link it?

Particularly it would be helpful do document or show in the test plan what fields are stored and possibly add a unit test similar to those that store MatchRecords/SignalMetadata. Thanks!

BarrettOlson (changes requested):
I don't think this accomplishes what we need for storing a record of actions taken. Happy to discuss more as I could be mistaken.

It seems that this just stores the most recent action label for a content id and the only context of the action taken that is being recorded in the hash and type both of which could be looked up elsewhere and might not be relevant. (See nit inline about not using PDQRecord)

I think we want to store all action taken for a content_id and the context as to why this action was taken (e.g. what rules triggered this and when). @eriksendc does this sound correct?

BarrettOlson:
Closing as most of the functionality of this PR will hopefully be handled in #583

PR #552, opened 2021-05-06
Convert Descriptor ID to Indicator ID 

jeberl:
Summary
---------

We were in correctly storing the indicator ID as the banked content ID when fetching from ThreatUpdates.

1. Change simple serialization in `python-threatexchange` to store both indicator id and descriptor roll up in CSV output. (will need to bump pip version in later PR) This effects how csv's are stored in s3
2. Change the post apply function in the fetcher which reads csv rows and creates PDQMetadata objects to now store the indicator ID rather thann the descriptor ID
3. Change the s3_adapters to read the new column names from the CSV
4. Change the ThreatExchange writebacker to recognize that `id` now refers to indicator id not descriptor ID 

NOTE: After this lands terraform destroy is probably needed

Test Plan
---------
Retested after rebase

1. Make upload_docker and update lambdas
2. delete all data in s3
3. refetch ThreatExchange data in s3 and confirm the csv columns are correct
4. Watch as index is rebuilt
5. Upload image using HMA UI
6. Watch match, copy ID, and using ID tool confirm that the ID is the correct indicator ID:
Screen Shot 2021-05-06 at 9 27 41 AM


BarrettOlson (commented):
This introduces a breaking change to python-threatexchange, @jeberl and I discussed possible alternatives and tradeoffs involved to this. The outcome of that discussion was that he is going to explore implementing a non-breaking serialization to work in HMA. 

Putting the PR back in draft mode for now.

python-threatexchange/threatexchange/descriptor.py
```diff
@@ -177,7 +177,6 @@ def merge(self, descriptor: ThreatDescriptor) -> None:
     def as_row(self) -> t.Tuple[int, str, str]:
         """Simple conversion to CSV row"""
         return self.first_descriptor_id, self.added_on, " ".join(self.labels)
-
```
jeberl, line 180: removing

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -354,4 +358,18 @@ def write_s3_text(txt_content: io.StringIO, bucket, key: str) -> None:
     logging.basicConfig(level=logging.INFO)
     # This will only kinda work for so long - eventually will
     # need to use a proper harness
-    lambda_handler(None, None)
+    # lambda_handler(None, None)
+
+    csv.field_size_limit(65535)  # dodge field size problems
```
BarrettOlson, line 363: rm before merge.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -255,18 +255,19 @@ def load_state(self, allow_cached=True):
                 csv.field_size_limit(65535)  # dodge field size problems
                 for row in csv.reader(txt_content):
                     items.append(
-                        CliIndicatorSerialization(
-                            "HASH_PDQ",
+                        HMASerialization(
                             row[0],
```
BarrettOlson, line 259: I missed this in #553 but having the constructor args for `indicator_type` and `indicator` be swapped between `HMASerialization` and `CliIndicatorSerialization` messed with my head. Maybe worth a comment.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -307,24 +308,27 @@ def post_apply(self, updated: t.Dict = {}):
...
                 updated_at=datetime.now(),
                 signal_source=S3ThreatDataConfig.SOURCE_STR,
                 signal_hash=row[0],  # note: not used by update_tags_in_table_if_exists
-                tags=row[3].split(" ") if row[3] else [],
+                tags=row[4].split(" ") if row[4] else [],
             ).update_tags_in_table_if_exists(table):
                 logger.info(
                     "Updated Signal Tags in DB for signal id: %s source: %s for privacy group: %d",
-                    row[1],
+                    row[0],
```
BarrettOlson, line 331: I think we still want this to be 1...

BarrettOlson (approved):
Overall, look good. approving to unblock. A few nit in line to please address before merge as well as two others: 
- This will break `make dev_upload_test_signal` can you just comment it out in the makefile if you don't want to go through the trouble of changing it? 
- can you delete?https://github.com/facebook/ThreatExchange/blob/9224eb7a00c61ffcc489bcccd8f0a3a3d11caf9d/hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_indexer.py#L25 (as it is unused and this change make it inaccurate)  

Thanks!

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
jeberl: Ah. Didnt think of that either. A comment would be useful but don't think this is the right place. Probably in the `HMASerialization` class

PR #553, opened 2021-05-06
Create CLI Serializer with Indicator ID

jeberl:
Summary
---------

HMA needs the indicator IDs. Instead of editing the existing CLI serializer we chose to create a new class so that the change would not be breaking

Test Plan
---------

Added test at end of file: ran with `python3 -m threatexchange.cli.dataset.simple_serialization`
e2e tests coming in later PR that will use this serializer


python-threatexchange/threatexchange/cli/dataset/simple_serialization.py
```diff
@@ -87,3 +90,60 @@ def load(
...
+        rollup: SimpleDescriptorRollup,
+    ):
+        self.indicator_id = indicator_id
+        self.indicator_type = indicator_type
+        self.indicator = indicator
+        self.rollup = rollup
+
+    def as_csv_row(self) -> t.Tuple:
+        """As a simple record type for the threatexchange CLI cache"""
+        return (self.indicator_id, self.indicator) + self.rollup.as_row()
```
BarrettOlson, line 114: can you remove the descriptor id from this version of the serialization?

python-threatexchange/threatexchange/cli/dataset/simple_serialization.py
```diff
@@ -50,7 +53,7 @@ def te_threat_updates_fields(cls):
     def store(
         cls, state_dir: pathlib.Path, contents: t.Iterable["CliIndicatorSerialization"]
     ) -> t.List[pathlib.Path]:
-        # Stores in multiple files
+        # Stores in multiple files split by data type
```
BarrettOlson, line 56: nit/ignorable: indicator type

python-threatexchange/threatexchange/cli/dataset/simple_serialization.py
```diff
@@ -87,3 +90,61 @@ def load(
...
+        rollup: SimpleDescriptorRollup,
+    ):
+        self.indicator_id = indicator_id
+        self.indicator_type = indicator_type
+        self.indicator = indicator
+        self.rollup = rollup
+
+    def as_csv_row(self) -> t.Tuple:
+        """indicator details and descriptor rollup without descriptor ID"""
+        return (self.indicator_id, self.indicator) + self.rollup.as_row_without_id()
```
BarrettOlson, line 115: nit: to keep it even more consistent could be swap this order of indicator and id? I like the 'signal' itself being the first column personally (not sure if our system actually still depends on it anywhere)

python-threatexchange/threatexchange/cli/dataset/simple_serialization.py
```diff
@@ -87,3 +90,61 @@ def load(
...
+            # Violate your warranty with class state! Not threadsafe!
+            csv.field_size_limit(path.stat().st_size)  # dodge field size problems
+            with path.open("r", newline="") as f:
+                for row in csv.reader(f):
+                    ret.append(
+                        cls(
+                            row[0],
+                            indicator_type,
+                            row[1],
+                            SimpleDescriptorRollup.from_row(row[2:]),
```
BarrettOlson, line 147: Does this still work or does it expect an id?

jeberl: You're right. This is expecting a descriptor idea. Didn't realize this earlier because I wasn't actually applying updates yet so had nothing to load. I think this means we either need to store the descriptor ID or stop using SimpleDescriptorRollup. I think the former is better

python-threatexchange/threatexchange/cli/dataset/simple_serialization.py
```diff
@@ -87,3 +90,87 @@ def load(
                         )
                     )
         return ret
+
+
+class HMASerialization(CliIndicatorSerialization):
+    """
+    A Serialization for HMA Similar to CliIndicatorSerialization but with
```
BarrettOlson, line 97: nit: add note to doc string that this serialization currently stores the "first descriptor" id as well as indicator id, future refactors should consider either removing this id or adding all visible descriptor ids.

BarrettOlson (approved):
LGTM

PR #554, opened 2021-05-07
[Multipart] Graphs and cloudwatch based counts for dashboard

schatten:
How to Review
---
* [Diff 1/4](https://github.com/facebook/ThreatExchange/pull/554/commits/b033d86c155c5b51ef811cc3e5c2b613d0c42b8a)  Query count sum and graphs from cloudwatch 
* [Diff 2/4](https://github.com/facebook/ThreatExchange/pull/554/commits/4473803a3095f60eb9b353ff351b713671d036f9)  Stats API using cloudwatch metrics backend; actions not integrated yet 
* [Diff 3/4](https://github.com/facebook/ThreatExchange/pull/554/commits/7b636788f7a1ab824c15a2ef5049e37167454174)  Pass along measure_performance envvar to api_root lambda 
* [Diff 4/4](https://github.com/facebook/ThreatExchange/pull/554/commits/6eb8abd6436e9c8ba3729c7d010f98c7bcd3dead)  Rewrite Dash.jsx using the graph based stats APIs 

Summary
---------
Uses cloudwatch instead of dynamodb to show counts of hashes and matches. Have not removed old files or APIs yet. Will do as subsequent pull request.

Supports 3 choices of graph time spans. 1h, 24h, 7d.

**Future Plans**
* Extract the graphing widget into its own component. Prettify.
* Add cloudwatch metrics (3-4 lines) to actions, fetcher, syncer
* Add graphing widgets to their respective pages

Test Plan
---------
new-dashboard-gyazo

* Published new docker image from local
* Updated tfvar to use new docker SHA
* Ran webapp locally and recorded the above GIF


hasher-matcher-actioner/webapp/package.json
```diff
@@ -19,7 +19,9 @@
     "react-bootstrap": "^1.5.2",
     "react-dom": "^17.0.1",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "4.0.3"
+    "react-scripts": "4.0.3",
+    "uplot": "^1.6.9",
+    "uplot-react": "^1.0.2"
```
schatten, line 24: Installing this requires you to add `--legacy-peer-deps` argument to react because the authors have not explicitly allowed use with reacted 0.17. 

React + npm versions are a mess now. Signs of improvement, but all of this will take a while.

hasher-matcher-actioner/hmalib/metrics/query.py
```diff
@@ -0,0 +1,87 @@
...
+    }[period] or timedelta(days=1)
+
+    return datetime.now() - delta
+
+
+def _period(period: MetricTimePeriod):
+    """
+    Granularity of AWS statistics.
+    1 minute for HOURS_1; returns 60 data points
+    10 minutes for HOURS_24; returns 144 data points
```
BarrettOlson, line 45: this is doc string correct?

hasher-matcher-actioner/hmalib/metrics/query.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import functools
+from dataclasses import dataclass
+from enum import Enum
+import typing as t
+import boto3
+from datetime import datetime, timedelta
+
+from . import measure_performance, METRICS_NAMESPACE
```
BarrettOlson, line 10: nit: should these be changed to cached methods not raw variables set in __init__.py?

hasher-matcher-actioner/hmalib/metrics/query.py
```diff
@@ -64,23 +88,32 @@ def get_count_with_graph(
...
+            StartTime=start_time,
+            EndTime=end_time,
             Period=_period(time_period),
         )["Datapoints"]
 
         total = int(functools.reduce(lambda acc, s: acc + s["Sum"], stats, 0))
 
-        graph_data = [(s["Timestamp"], int(s["Sum"])) for s in stats]
+        graph_data = [
+            (s["Timestamp"].replace(tzinfo=None), int(s["Sum"])) for s in stats
```
BarrettOlson, line 113: Having the browser handle timezone? nit: maybe add a comment explaining that/why the removal

hasher-matcher-actioner/webapp/src/Api.jsx
```diff
@@ -67,6 +67,16 @@ export function fetchDashboardCardSummary(path) {
   return apiGet(`/${path}`);
 }
 
+export const StatsTimeSpans = Object.freeze({
```
BarrettOlson, line 70: nit: comment to where this maps to a python enum? 

Also I created `utils/constants.jsx` a few PRs ago, if you think we should instead put enums/constants inline I'm open to it, I mostly just think we should try to be consistent.

hasher-matcher-actioner/webapp/src/components/GraphWithNumberWidget.jsx
```diff
@@ -0,0 +1,63 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// import uPlot from 'uplot';
+import UplotReact from 'uplot-react';
+import '../../node_modules/uplot/dist/uPlot.min.css';
+
+/**
+ * Reduces the width required for displaying a number. eg. 4,000 -> 4k.
+ * 2,000,000 -> 2M etc.
```
BarrettOlson, line 14: Should the m be `M` or `m` 
Personally don't have a strong opinion but let's have the comment match the returned value?

Rabbit Hole warning on stack exchange meta: https://meta.stackexchange.com/questions/87251/use-m-for-million-instead-of-m

hasher-matcher-actioner/webapp/src/components/GraphWithNumberWidget.jsx
```diff
@@ -0,0 +1,63 @@
...
+  series: [
+    {},
+    {
+      stroke: 'rgba(5, 141, 199, 1)',
+      fill: 'rgba(255, 0, 0, 0.3)',
+    },
+  ],
+};
+
+// const data = [
```
BarrettOlson, line 46: rm?

hasher-matcher-actioner/webapp/src/components/GraphWithNumberWidget.jsx
```diff
@@ -0,0 +1,63 @@
...
+//   [1546300800, 1546387200], // x-values (timestamps)
+//   [35, 71], // y-values (series 1)
+// ];
+
+export default function GraphWithNumberWidget({number, graphData}) {
+  console.log(number);
```
BarrettOlson, line 52: rm.

hasher-matcher-actioner/webapp/src/pages/Dash.jsx
```diff
@@ -0,0 +1,132 @@
...
+          <div>Show statistics for the last</div>
+          <DropdownButton
+            as={ButtonGroup}
+            id="dropdown-time-span-picker"
+            variant="secondary"
+            title={getDisplayTimeSpan(timeSpan)}>
+            {Object.entries(StatsTimeSpans)
+              .map(entry => entry[1]) // Get values only, no keys.
+              .map((timeSpanChoice, index) => (
+                <Dropdown.Item
```
BarrettOlson, line 95: nit/aside/ignorable: is dev tools/lint okay with this? I had problems with trying to use index as a part of keys in map, looks like eventKey is serving a different purpose than key would but either way I'd expect a warning.

hasher-matcher-actioner/webapp/src/pages/Dash.jsx
```diff
@@ -0,0 +1,132 @@
...
+    fetchStats(timeSpan).then(response => {
+      setStatCards(response.cards);
+    });
+  }, [timeSpan]);
+
+  return (
+    <Container className="m-4">
+      <Row>
+        <Col xs={6} className="mb-4 d-flex justify-content-end">
+          <div>Show statistics for the last</div>
```
BarrettOlson, line 86: super nit: add some padding to this or the button as they are touching in the gif

BarrettOlson (approved):
Overall, love it! Didn't see anything blocking but added a few comments inline for you to consider in this PR or future ones. 

I do think the signals this provides are valuable but somewhat distinct from those (not mocked) in the existing dash. For example things like "total signal hashes" or "total unique content hashes" Don't have arguments either way but if we do just mostly replace the existing dash lets do so with a user story in mind :)

schatten (commented):
> Overall, love it! Didn't see anything blocking but added a few comments inline for you to consider in this PR or future ones.
> 
> I do think the signals this provides are valuable but somewhat distinct from those (not mocked) in the existing dash. For example things like "total signal hashes" or "total unique content hashes" Don't have arguments either way but if we do just mostly replace the existing dash lets do so with a user story in mind :)

Thanks for the super-detailed review, @BarrettOlson. I'll address them in a commit on this PR.

About 'total' counts, I think that is less useful at a dashboard level. There should be some way to arrive at those numbers, but the way I see it, the intent of the current story is to demonstrate that the system is working as expected. Graphs in that way were the most significant way to address it.

If we don't see eye to eye here, we can create another card/issue to track the "totals" story.

hasher-matcher-actioner/webapp/src/Api.jsx
schatten: Addressed in https://github.com/facebook/ThreatExchange/pull/554/commits/dd4fddd81acb1f8ee34b3f985abbb05652c8cd1c

hasher-matcher-actioner/hmalib/metrics/query.py
schatten: Do not see either changing during the lifetime of a lambda. So, could go either way. 

Leaving as is until ☝️ changes.

hasher-matcher-actioner/hmalib/metrics/query.py
schatten: Yes. Browser / graphing library handles timezones pretty well.

There is an additional reason, will add as comment.

hasher-matcher-actioner/webapp/src/pages/Dash.jsx
schatten: Irked me too. :) Decided to drop unless someone pointed out. Thanks for the nudge.

hasher-matcher-actioner/webapp/src/pages/Dash.jsx
schatten: Yeah. Event key was different. Key threw the warning. Fixed with 
```javascript
                  key={`time-span-picker-${timeSpanChoice}`}
```

Issue #555, opened 2021-05-07
[HMA] Bootstrap with copydays data and link from settings page


schatten:
User Story
---
A partner who does not yet have access to ThreatExchange deploys an instance of the HMA system to their cloud. They should be able to test things out.

We have used the [copy-days data set](https://lear.inrialpes.fr/~jegou/data.php) within ThreatExchange to provide test data. We should have a route for this without needing access to ThreatExchange.

**Acceptance Criteria**
When this Issue is closed, 
* a fresh installation of HMA should include PDQ hashes for *all* images from one of the copy days datasets. All is important. Imagine uploading an image that's not in our index and waiting for hours for an action. Would be disappointing.
* one of the settings page should link to the copy-days page and call out the link for the dataset we have indexed.

Technicals
---
We presently have a subset of the [sample data](https://github.com/facebook/ThreatExchange/blob/fb407da7ea907bda1d68802719cdef8d325e0347/hasher-matcher-actioner/sample_data/pdq.csv). Note that it is only 137 images hashes. That needs to be taken to 812. The actual number of images in the dataset.

Once done, verify that as part of setup, we are uploading the csv to the index location on the `hashing-data` s3 bucket.

Then make the UI changes to address the linking bits from the acceptance Criteria.

schatten:
Using 
```shell
$  ls | xargs threatexchange --config ~/.txconfig hash --signal-type pdq photo > ~/copydays-pdq-hashes.txt
```

to generate a list of PDQ hashes. Verified that some of the hashes existed in the [sample file](https://github.com/facebook/ThreatExchange/blob/fb407da7ea907bda1d68802719cdef8d325e0347/hasher-matcher-actioner/sample_data/pdq.csv).

PR #560, opened 2021-05-07
Copydays sample data uploaded to threat_exchange_data/; refresh sample data to include PDQ for all 812 images

schatten:
Summary
---------
As part of #555, need to provide full copydays data when HMA is setup. This adds the sample data computed using [command noted here](https://github.com/facebook/ThreatExchange/issues/555#issuecomment-834542181).

Then, it uses terraform null resources to update this within `threat_exchange_data/` s3 key.

Reviewer, please note that this would mean that the it would appear that the data came from ThreatExchange. The indicator ids in the new sample file are completely random and may or may not exist in FB Ent Universe. I don't know what the ramifications of that are. Please advise if this is not the correct direction.

Test Plan
---------
Ran `terraform apply` and verified that the index got updated. Logs from the indexer are shown below.

Screen Shot 2021-05-07 at 12 42 45


BarrettOlson (commented):
LGTM! : when #552 is merged `copydays-jpg1-pdq-hashes.csv` will need to be reformatted.  
(will leave official stamp to @jeberl)

jeberl (approved):
Lgtm

PR #562, opened 2021-05-08
Address miscellaneous comments left over from PR 543

eriksendc:
Summary
---------

PR 543 was a big "stacked diff" that I merged after it was stable, but before some final pr comments were addressed. This PR addresses the following:
1. Use deconstructing parameter lists instead of a deconstructing assignment of props.
2. Moved modules into appropriate directories.
3. Broke components out into separate modules.
4. Moved tab event keys into an array to streamline validation in Settings.jsx.
5. Defined defaultActionRule object (had been using two objects being defined inline with the same properties).
6. Simplified syntax of a few onClick()'s.
7. In preparation for driving all API calls from ActionRuleSettingsTab, passed actions down as props to child components.
8. Moved css class action-rule-classification-column to _settings.scss.
9. Added props typing.

Test Plan
---------

Run the web app, navigate to Settings. Ensure the ThreatExchange and Action Rules tabs function without errors when they're functionalities are exercised. You should see what's captured in the attached video.

https://user-images.githubusercontent.com/8974354/117700422-ff82c080-b193-11eb-9415-9b8bfa846c2f.mp4


hasher-matcher-actioner/webapp/src/pages/Settings.jsx
```diff
@@ -10,17 +10,14 @@ import ActionRuleSettingsTab from './settings/ActionRuleSettingsTab';
...
   const history = useHistory();
-  if (
-    tab === undefined ||
-    !tab ||
-    (tab !== 'threatexchange' &&
-      tab !== 'pipeline' &&
-      tab !== 'actions' &&
-      tab !== 'action-rules')
-  ) {
+  if (tab === undefined || !tab || !tabEventKeys.includes(tab)) {
```
schatten, line 20: Aah! Thanks for this!

schatten (approved):
Love it!

PR #565, opened 2021-05-10
Remove dashboard APIs, Components and make new Dashboard the default route

### LINKED ISSUE #567: [HMA] Support "counting" objects in DynamoDB
A bunch of user-stories all have the same underlying technology dependence.

Sample user-stories:
1. "See total number of hashes and matches"
2. "See total number of content hashes and matches split by privacy group"
3. "Dashboard page has counter of hashes/matches that links to AWS cloudwatch page"

Counting using the Scan functionality is expensive and not recommended for production web applications.

I'm planning to store the counters as separate entries on the same table that contain the data. The operations on the counters (inc, dec, get) will be done using [DynamoDB Transactions](https://advancedweb.hu/how-to-keep-accurate-counts-in-dynamodb/) as documented in the link. 

The way I'm thinking right now, base classes to do counted CRUD operations might be tricky. I'm instead thinking of decorators. Pseudo code below.

Eg:
```python
class SomeBrandNewRecord(DynamoDBItem):
    @get_counter('some_brand_new_record')
    def count(self, table: Table):
      # The work is done by the decorator, okay for this method to return None.
      pass

    @increment_counter('some_brand_new_record')
    def write_to_table(self, table: Table):
      return super(<>).write_to_table(self, table)
```

If transactions can't be done using the decorators like I suggest above, I might have to relook at how best to do this.

schatten:
Summary
---------
- Some dashboard routes were housed in api_root. They were moved into their own module as part of #554 The older ones are being cleaned up here.
- The Dashboard page was left as is in #554 to get feedback and for comparisons. That is being removed and replaced by the new page here.
- As @BarrettOlson had flagged, the total counts are required in the finished dashboard. They are now being served via the API based on functionality added in #571. Once merged, this PR fixes #567 

Test Plan
---------
* Ran black and mypy
* With the new python code deployed to lambdas, each page in the HMA ui was opened locally by running `npm start` and its results were eyeballed.

Screen Shot 2021-05-10 at 14 41 51

Notice how the dashboard link is active now because while the route is `/`.

### Update
New stat card shows total and time span numbers below the graph.

Screen Shot 2021-05-12 at 18 03 31


hasher-matcher-actioner/webapp/src/pages/Dashboard.jsx
```diff
@@ -126,13 +127,23 @@ function StatCard({statName, timeSpan}) {
...
+          </Col>
+          <Col xs={4}>
+            <h1>{getDisplayNumber(card.time_span_count)}</h1>
+            <small className="text-muted">
+              in the last {getDisplayTimeSpan(card.time_span)}.
+            </small>
+          </Col>
+          <Col xs={4}>
+            <h1>{getDisplayNumber(card.total_count)}</h1>
+            <small className="text-muted">since HMA is online.</small>
```
BarrettOlson, line 144: nit/not blocking: awkward wording (and this value maybe doesn't need to be a large as the count for the time range.) 

something like: "of <total_count> processed in total." somewhere in the card is better imo

hasher-matcher-actioner/hmalib/lambdas/api/stats.py
```diff
@@ -98,11 +105,14 @@ def default_stats() -> StatResponse:
             metric_time_period,
         )
 
+        total_count = stat_name_to_count[bottle.request.query.stat_name]()
```
BarrettOlson, line 108: nit/unsure:`.get(..., <noop>)` instead of `[...]` to avoid key error if we want to send `bottle.request.query.stat_name = "actions"` before it is ready on the backend and not have the API error out?

BarrettOlson (approved):
LGTM!  Added some nit/questions inline.

hasher-matcher-actioner/hmalib/lambdas/api/stats.py
schatten: https://github.com/facebook/ThreatExchange/pull/565/files#diff-719551ffa714ce37c8b382114df348cb3c82750380efc186fbe2292c8875b51aR85-R88 would catch this I believe.

Also the error would nudge the dev to implement functionality. Leaving this as is.

hasher-matcher-actioner/webapp/src/pages/Dashboard.jsx
schatten: I see what you mean. I'm not sure either. Would defer to more experienced hands on copy writing. Happy to revisit when we are doing a design/copy thingy.

BarrettOlson:
@schatten I think we should consider the possible confusion that could result from "Photo Matched" count being much larger than the number of rows on the matches page due to repeated submissions. While not as likely in a day-to-day prod usage, I think repeat submissions will be very common in evaluating the systems functionality. So we might what to over communicate how the calcs/count work at least at first.

hasher-matcher-actioner/hmalib/lambdas/api/stats.py
BarrettOlson: Ahh, of course github hid that check from me, thanks! 
(links broken for me but figured it was the same file and just went to the line numbers in the url haha)

PR #566, opened 2021-05-10
Copydays information block in ThreatExchange settings page

### LINKED ISSUE #555: [HMA] Bootstrap with copydays data and link from settings page

schatten:
Summary
--------
Add an information block giving instructions on how to download images from copydays and use them to match.

This is 1 of 2 diffs that fixes #555 The other diff is #560 

Test Plan
---------
No python files were touched. A screenshot from the ThreatExchange settings page is attached.

Screen Shot 2021-05-10 at 19 58 35

Updated the blurb to also reflect the test privacy group id.

Screen Shot 2021-05-11 at 13 54 04

### DIFF
--- a/hasher-matcher-actioner/webapp/src/components/CopydaysInformationBlock.jsx
+++ b/hasher-matcher-actioner/webapp/src/components/CopydaysInformationBlock.jsx
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+import React from 'react';
+import {Card} from 'react-bootstrap';
+
+/**
+ * Super simple informational component. Drop in anywhere.
+ */
+export default function CopydaysInformationBlock() {
+  return (
+    <Card bg="light" body style={{maxWidth: '720px'}}>
+      <h2>Test Photos</h2>
+      <p>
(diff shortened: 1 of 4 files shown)


jeberl (approved):
One other change but approving to unblock

PR #568, opened 2021-05-11
add headers to WebhookActionPerformer

mengyangwang:
Summary
---------
This PR is to add "headers" to WebhookActionPerformer in DynamoDB

Future Actions
---------
 Create apis to create/update/read/delete data in DynamoDB 

Test Plan
---------

1. black hmalib
2. mypy hmalib
4. delete all WebhookActionPerformer in DynamoDB, then run ```python -m hmalib.scripts.populate_config_db load_example_configs```
5. test changes in DynamoDB
Screen Shot 2021-05-11 at 3 00 58 PM

https://user-images.githubusercontent.com/81996660/117886542-5dd79e00-b27d-11eb-9396-59c1dd7ee692.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/actioner_models.py
+++ b/hasher-matcher-actioner/hmalib/common/actioner_models.py
@@ -45,6 +45,7 @@ class WebhookActionPerformer(ActionPerformer):
     """Superclass for webhooks"""
 
     url: str
+    headers: str
 
     def perform_action(self, match_message: MatchMessage) -> None:
         self.call(data=json.dumps(match_message.to_aws()))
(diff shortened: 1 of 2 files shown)


jeberl (changes requested):
Looks great but can we do an end-2-end test that these headers actually get propagated in the HTTP request?

Trigger an action which should happen when you upload an image that isn't the sailboat. Then confirm that the webhook is called with the right headers by going to the webhook address stored in the action. Can show you how to do this if it's unclear

jeberl (approved):
Thanks for adding the e2e test! Ship it!

PR #569, opened 2021-05-11
Connect Api to writebacker

jeberl:
Summary
---------

Connect the api to the Writeback Framework so that true and false positives can be sent.

Future PRs will implement and e2e test False Positive and True positive writebacks

Test Plan
---------

mypy && make_uplaod docker

From UI press true-positive. Look at logs to see that writeback message is enqueued

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/message_models.py
+++ b/hasher-matcher-actioner/hmalib/common/message_models.py
@@ -106,7 +106,6 @@ class WritebackMessageConfig:
     @classmethod
     @lru_cache(maxsize=None)
     def get(cls):
-        HMAConfig.initialize(os.environ["CONFIG_TABLE_NAME"])
         return cls(
             writebacks_queue_url=os.environ["WRITEBACKS_QUEUE_URL"],
             sqs_client=boto3.client("sqs"),
@@ -147,7 +146,7 @@ class WritebackMessage(HasAWSSerialization):
         opinion_change_to_writeback_type = {
             PendingOpinionChange.MARK_TRUE_POSITIVE: WritebackTypes.TruePositive,
             PendingOpinionChange.MARK_FALSE_POSITIVE: WritebackTypes.FalsePositive,
-            PendingOpinionChange.REMOVE_OPINION: WritebackTypes.RemoveTrueFalsePositive,
+            PendingOpinionChange.REMOVE_OPINION: WritebackTypes.RemoveOpinion,
         }
 
         writeback_type = opinion_change_to_writeback_type.get(
--- a/hasher-matcher-actioner/hmalib/common/message_models.py
+++ b/hasher-matcher-actioner/hmalib/common/message_models.py
@@ -1,5 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
+import os
+import boto3
 import typing as t
 from dataclasses import dataclass, field
 from hmalib.common.classification_models import (
@@ -9,9 +11,13 @@ from hmalib.common.classification_models import (
     ClassificationLabel,
     Label,
     WritebackTypes,
+    PendingOpinionChange,
 )
 from hmalib.common.evaluator_models import ActionLabel, ActionRule
 from hmalib.common.aws_dataclass import HasAWSSerialization
+from hmalib.common.config import HMAConfig
+from mypy_boto3_sqs import SQSClient
+from functools import lru_cache
 
 
 @dataclass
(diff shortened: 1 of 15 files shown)


hasher-matcher-actioner/hmalib/common/message_models.py
```diff
@@ -89,25 +95,76 @@ def from_match_message_action_label_and_action_rules(
...
         )
+
+    @classmethod
+    def from_banked_signal_and_opinion_change(
+        cls, banked_signal: BankedSignal, opinion_change: PendingOpinionChange
+    ) -> "WritebackMessage":
+        opinion_change_to_writeback_type = {
+            PendingOpinionChange.MARK_TRUE_POSITIVE: WritebackTypes.TruePositive,
+            PendingOpinionChange.MARK_FALSE_POSITIVE: WritebackTypes.FalsePositive,
+            PendingOpinionChange.REMOVE_OPINION: WritebackTypes.RemoveTrueFalsePositive,
```
BarrettOlson, line 150: Why not `RemoveOpinion` to match the enum? I'm not tied to that name but `RemoveTrueFalsePositive` is a little too confusing imo.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -54,7 +53,6 @@ def get(cls):
         HMAConfig.initialize(os.environ["CONFIG_TABLE_NAME"])
```
BarrettOlson, line 53: env values used in the API lambda but not declared in the top level module are hard to debug and easy to miss. We've gotten around this before by having this init occur within api_root.py. Thoughts?

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -239,6 +239,8 @@ module "api" {
   config_table          = local.config_table
   measure_performance   = var.measure_performance
   te_api_token_secret   = aws_secretsmanager_secret.te_api_token
+
```
BarrettOlson, line 242: nit: random new line?

hasher-matcher-actioner/hmalib/common/message_models.py
```diff
@@ -89,25 +95,76 @@ def from_match_message_action_label_and_action_rules(
...
+    Simple holder for getting typed environment variables
+    """
+
+    writebacks_queue_url: str
+    sqs_client: SQSClient
+
+    @classmethod
+    @lru_cache(maxsize=None)
+    def get(cls):
+        HMAConfig.initialize(os.environ["CONFIG_TABLE_NAME"])
```
BarrettOlson, line 109: (repeated note): env values used in the API lambda but not declared in the top level module are hard to debug and easy to miss. We've gotten around this before by having this init occur within api_root.py. Thoughts?

BarrettOlson (approved):
Stamping to unblock. Changes asked about and inline nits can be addressed as part of future PRs in this area since this is just connecting the API to the queue.

Please double check this doesn't break anything e2e with os.env stuff (terraform looked correct so I think it is unlikely but it has happened before haha)

hasher-matcher-actioner/hmalib/common/message_models.py
jeberl: I was thinking that "Opinion" is too general and "Ingested" or "Seen" wiritebacks might be considered and opinion. However I think you're right that Opinion is clearer

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
jeberl: Bit confused. Don't think this is called directly from an api call. This is called after a match occurs?

hasher-matcher-actioner/terraform/main.tf
jeberl: Added that to separate from the ones above

hasher-matcher-actioner/hmalib/common/message_models.py
jeberl: I follow your comment here. Will move to `matches.get_matches_api` to mirror `datasets_api`

hasher-matcher-actioner/terraform/main.tf
BarrettOlson: again complete nit but what separates this variables from the ones above, the previous 5 lines/vars seem like a random collection?

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
BarrettOlson: Ah my mistake I missed this files path on first review. 
Edit: env values used in the ~~API~~ *any* lambda but not declared in the top level module are hard to debug and easy to miss....

Ideally imo this would be a constant declaration at the top of the file like in `api_root.py` (just for ease of refactor later and quick checks between the file and tf) but because this file as a `lambda_handler` method and something like this already being done in `fetcher.py` it is probably fine.

PR #571, opened 2021-05-12
DynamoDB Counts

schatten:
Summary
---------
Uses `transact_write_items`, a low level API to write records and update a counter item in the same transaction.

Partially fulfills requirements for #567 

Design
---
Creates sparse dynamodb items in the same table that contains the items being written. Uses `transact_write_items` to update a counter for every write. Updates by key are super-fast in ddb, so performance impact is negligible.

Rather than inferring the class name, requires that the class explicitly provide a name so that future refactoring does not reset the counters. 

Test Plan
---------
```
$ python -m py.test # Have added tests specific to this functionality
$ python -m black hmalib
$ python -m mypy hmalib
```

Also used the following script in the REPL to update counters in my prefixed dynamodb instance.

```python
from hmalib.models import PipelinePDQHashRecord
import boto3
tbl = boto3.resource('dynamodb').Table('dipanjanm-HMADataStore')
p = PipelinePDQHashRecord.get_from_content_id(tbl, 'images/storm-2021-05-03-a2f956db-4eb0-4405-affc-3fe54a262db8-200100.jpg')
p.write_to_table(tbl)
```

Sample AWS UI output shown below.
Screen Shot 2021-05-12 at 09 33 24

### DIFF
--- a/hasher-matcher-actioner/hmalib/models.py
+++ b/hasher-matcher-actioner/hmalib/models.py
@@ -6,6 +6,7 @@ import typing as t
 import json
 from dataclasses import dataclass, field
 from mypy_boto3_dynamodb.service_resource import Table
+from mypy_boto3_dynamodb import Client
 from boto3.dynamodb.conditions import Attr, Key, And
 from botocore.exceptions import ClientError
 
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/models.py
```diff
@@ -108,6 +193,14 @@ def to_sqs_message(self) -> dict:
             "key": self.content_id,
         }
 
+    @classmethod
+    def is_total_countable(cls):
+        return True
+
+    @classmethod
+    def get_total_count_keyname(cls):
+        return "PipelinePDQHashRecord"
```
BarrettOlson, line 202: >Rather than inferring the class name, requires that the class explicitly provide a name so that future refactoring does not reset the counters.

I have mixed feelings about this. It seems super easy to miss and I don't love the name of some of these classes so I think we will want to change it eventually. Specific counts are also something I think would be easier to migrate if needed because it is only one command per record type.  
Thoughts on adding an assert or check to log a warning if the inferred class no longer matches this hard coded value as a middle ground for now? (that way the counts wont reset but the dev will see an error/warning?)

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -16,14 +19,96 @@
...
+        transact_items: t.List[TransactWriteItemTypeDef] = [
+            {
+                "Put": {
+                    "Item": self.to_dynamodb_item(),
+                    "TableName": table.table_name,
+                },
+            },
+        ]
+
+        if self.__class__.is_total_countable():
```
BarrettOlson, line 95: question: is `__class__` needed here for just defensive programing? (want to check my understanding of inheritance and classmethods)

hasher-matcher-actioner/hmalib/models.py
```diff
@@ -16,14 +19,96 @@
...
+                        },
+                        # if_not_exists takes care of initializing the count if it hasn't been done yet.
+                        "UpdateExpression": "SET WriteCount = if_not_exists(WriteCount, :zero) + :one",
+                        "TableName": table.table_name,
+                        "ExpressionAttributeValues": {":one": 1, ":zero": 0},
+                    },
+                }
+            )
+
+        client.transact_write_items(TransactItems=transact_items)
```
BarrettOlson, line 111: Any way to sanity check the perf cost then add a comment of its impact or that it is negligible? (if it is non-negligible we may want to refactor in a enable/disable flag in a future PR)

BarrettOlson (approved):
Love it! Added a few (non-blocking) discussion topics /questions inline.

hasher-matcher-actioner/hmalib/models.py
schatten: Just being explicit. You could just as easily do `self.is_total_countable()`. Nothing in python prevents this.

hasher-matcher-actioner/hmalib/models.py
schatten: Used the script below to do a REPL based perf test.
```python
from hmalib.models import PipelinePDQHashRecord
import boto3
tbl = boto3.resource('dynamodb').Table('dipanjanm-HMADataStore')
p = PipelinePDQHashRecord.get_from_content_id(tbl, 'images/storm-2021-05-03-a2f956db-4eb0-4405-affc-3fe54a262db8-200100.jpg')
p.write_to_table(tbl)

# perf tests
import timeit
timeit.timeit(stmt="p.write_to_table(tbl)", number=200, globals={'p': p, 'tbl': tbl})
```

For 200 writes
| Configuration| Time in Seconds|
|--------------|------------------|
|`[put_item + update_item]` | 14 
|`[put_item]`|13|
|`put_item` using High Level API| 10|

Deferring making this a repeatable test, but it appears the handwavy cost is around 40%. 

Most of it appears to be because of a transaction lock. I'm inferring this from the second configuration taking nearly as much time as the first one.

hasher-matcher-actioner/hmalib/models.py
schatten: If it were to be missed completely, the count operations would throw a `NotImplementedError`.

If the class name is changed, the counts continue to work. Unless the dev changes the return value of this method. They should only change that if they have handwritten records to the dynamodb. I think this is where I would want to draw the line.

BarrettOlson: Okay, I would say let's not have it match the class name from the start at all then if that is the desired outcome (to avoid confusion/a naive find+replace refactor) but this has definitely become more a nit imo.

PR #572, opened 2021-05-12
devcontainers: update dev container to handle custom dotfiles

BarrettOlson:
Summary
---------

Updates `devcontainer.json` to have support of both zsh and bash using the new prefered settings. (Users can set default for new terminal in their global settings) 

Test Plan
---------
Rebuild container and made sure apply and other dev commands still worked both for `bash` and `zsh`


schatten (approved):
Approving to unblock! Thanks for taking the jump into devcontainers.

Would only writing to `.profile` suffice? Do we need to write to `.zshrc` too?

BarrettOlson:
> Would only writing to `.profile` suffice? Do we need to write to `.zshrc` too?
If we sourced `.profile` in the default `.zshrc` that should be fine. However, as far as I can tell there is no harm in appending to both for now and that way entry level dotfiles users/improvements only need to care about one file. 
(Also, not sure how standard ~.profiles is for this images/distro so if something does go wrong with it in an update non custom dotfile folks won't have to worry about losing history)

PR #573, opened 2021-05-12
Action performer api

mengyangwang:
Summary
---------
This PR is to create create/read/update/delete apis to for ActionPerformer in DynamoDB
 

Test Plan
---------

1. black hmalib
2. mypy hmalib
4. python -m hmalib.lambdas.api.api_root
5. test all apis in postman

https://user-images.githubusercontent.com/81996660/118007684-5adebb00-b31a-11eb-8869-5c7c68f14d53.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
@@ -1,0 +1,115 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import bottle
+import typing as t
+from dataclasses import dataclass, asdict
+from .middleware import jsoninator, JSONifiable, DictParseable
+from hmalib.common.config import HMAConfig, update_config
+from hmalib.common import config as hmaconfig
+from hmalib.common.actioner_models import ActionPerformer
+
+
+@dataclass
+class Action(JSONifiable, DictParseable):
+    name: str
+    config_subtype: str
+    url: str
+    headers: str
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+    @classmethod
+    def from_dict(cls, d: dict) -> "Action":
+        return cls(
+            d["name"],
+            d["config_subtype"],
+            d["url"] if d["url"] else "",
+            d["headers"] if d["headers"] else "",
+        )
+
+
+@dataclass
+class ActionsResponse(JSONifiable):
+    actions_response: t.List[Action]
+
+    def to_json(self) -> t.Dict:
+        return {
+            "actions_response": [action.to_json() for action in self.actions_response]
+        }
+
+
+@dataclass
+class DeleteActionResponse(JSONifiable):
+    response: str
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+
+@dataclass
+class CreateActionResponse(JSONifiable):
+    response: str
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+
+def get_actions_api(hma_config_table: str) -> bottle.Bottle:
+    # The documentation below expects prefix to be '/actions/'
+    actions_api = bottle.Bottle()
+    HMAConfig.initialize(hma_config_table)
+
+    @actions_api.get("/", apply=[jsoninator])
+    def get_actions() -> ActionsResponse:
+        """
+        Returns all action configs.
+        """
+        action_configs = ActionPerformer.get_all()
+        return ActionsResponse(
+            actions_response=[
+                Action.from_dict(config.__dict__) for config in action_configs
+            ]
+        )
+
+    @actions_api.post("/update", apply=[jsoninator(Action)])
+    def update_action(request: Action) -> t.Optional[Action]:
+        """
+        Update an action url and headers
+        """
+        sub_types = ActionPerformer.get_subtype_classes()
+        for sub_type in sub_types:
+            if sub_type.get_config_subtype() == request.config_subtype:
+                config = sub_type.getx(request.name)
+                config.url = request.url
+                config.headers = request.headers
+                updated_config = hmaconfig.update_config(config)
+                return Action.from_dict(updated_config.__dict__)
+
+        raise ValueError("The request is invalid!")
+
+    @actions_api.post("/create", apply=[jsoninator(Action)])
+    def create_action(request: Action) -> CreateActionResponse:
+        """
+        create an action
+        """
+        sub_types = ActionPerformer.get_subtype_classes()
+        for sub_type in sub_types:
+            if sub_type.get_config_subtype() == request.config_subtype:
+                config = sub_type(
+                    name=request.name, url=request.url, headers=request.headers
+                )
+                hmaconfig.create_config(config)
+                return CreateActionResponse(response="The action config is created.")
+        raise ValueError("The request is invalid!")
+
+    @actions_api.post("/delete/<key>", apply=[jsoninator])
+    def delete_action(key=None) -> DeleteActionResponse:
+        """
+        Delete an action
+        """
+        config = ActionPerformer.getx(str(key))
+        hmaconfig.delete_config(config)
+        return DeleteActionResponse(response="The action config is deleted.")
+
+    return actions_api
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -46,6 +47,8 @@
 IMAGE_FOLDER_KEY_LEN = len(IMAGE_FOLDER_KEY)
 
 # Override common errors codes to return json instead of bottle's default html
+
+
```
jeberl, line 51: Nit: delete this

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,115 @@
...
+from hmalib.common import config as hmaconfig
+from hmalib.common.actioner_models import ActionPerformer
+
+
+@dataclass
+class Action(JSONifiable, DictParseable):
+    name: str
+    config_subtype: str
+    url: str
+    headers: str
```
jeberl, line 17: Can we make these `parameters` a dict so that they can have different fields for differnt action performer subtypes?

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,115 @@
...
+import typing as t
+from dataclasses import dataclass, asdict
+from .middleware import jsoninator, JSONifiable, DictParseable
+from hmalib.common.config import HMAConfig, update_config
+from hmalib.common import config as hmaconfig
+from hmalib.common.actioner_models import ActionPerformer
+
+
+@dataclass
+class Action(JSONifiable, DictParseable):
```
jeberl, line 13: `Action` is a little vague here. If we keep this class as distinct from `ActionPerformer` can we name it `UIAction`

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,115 @@
...
+    @actions_api.get("/", apply=[jsoninator])
+    def get_actions() -> ActionsResponse:
+        """
+        Returns all action configs.
+        """
+        action_configs = ActionPerformer.get_all()
+        return ActionsResponse(
+            actions_response=[
+                Action.from_dict(config.__dict__) for config in action_configs
+            ]
```
jeberl, line 72: Maybe it makes sense to combine `ActionPerformer` and `Action` classes?

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+
+    @actions_api.post("/update", apply=[jsoninator(Action)])
+    def update_action(request: Action) -> t.Optional[Action]:
+        """
+        Update an action url and headers
+        """
+        sub_types = ActionPerformer.get_subtype_classes()
+        for sub_type in sub_types:
+            if sub_type.get_config_subtype() == request.config_subtype:
```
jeberl, line 82: nit: Maybe it makes more sense to refactor the `get_subtype_classes` function as a dict mapping from class name to class. eg.

```
def get_subtype_classes():
        return {
            "WebhookPostActionPerformer" : WebhookPostActionPerformer,
            "WebhookGetActionPerformer" : WebhookGetActionPerformer,
            "WebhookPutActionPerformer" : WebhookPutActionPerformer,
            "WebhookDeleteActionPerformer" : WebhookDeleteActionPerformer,
        }
```

You could then rewrite these lines as:
```
config = ActionPerformer.get_subtype_classes()[request.config_subtype].getx(request.name)

File: hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py [hunk 6]
```diff
@@ -0,0 +1,115 @@
...
+    def create_action(request: Action) -> CreateActionResponse:
+        """
+        create an action
+        """
+        sub_types = ActionPerformer.get_subtype_classes()
+        for sub_type in sub_types:
+            if sub_type.get_config_subtype() == request.config_subtype:
+                config = sub_type(
+                    name=request.name, url=request.url, headers=request.headers
+                )
```
jeberl, line 101: Same as above

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,115 @@
...
+    def to_json(self) -> t.Dict:
+        return asdict(self)
+
+
+@dataclass
+class CreateActionResponse(JSONifiable):
+    response: str
+
+    def to_json(self) -> t.Dict:
+        return asdict(self)
```
jeberl, line 55: The naming of these classes makes me think they are related but the code makes it seem like they aren't. Can we rename to make this a bit clearer what they each are for?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -309,5 +312,10 @@ def get_signal_hash_count() -> t.Dict[str, int]:
 
 app.mount("/stats/", get_stats_api(dynamodb_table=dynamodb.Table(DYNAMODB_TABLE)))
 
+app.mount(
+    "/actions/",
+    get_actions_api(hma_config_table=HMA_CONFIG_TABLE),
+)
+
```
jeberl, line 319: Love this modularization

jeberl (changes requested):
This looks great and thanks for the through end to end testing. Most of my comments are just renaming and simplifying

jeberl (approved):
Approving to unblock but please makes changes

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
mengyangwang: yes, when I think more about this, looks ```Action``` is not needed

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
mengyangwang: yes there is actually already a method ```_get_subtypes_by_name```, I'll use this method instead of looping through the array

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,107 @@
...
+    def fetch_all_actions() -> FetchAllActionsResponse:
+        """
+        Returns all action configs.
+        """
+        action_configs = ActionPerformer.get_all()
+        return FetchAllActionsResponse(
+            actions_response=[config.__dict__ for config in action_configs]
+        )
+
+    @actions_api.post("/update", apply=[jsoninator(CreateUpdateActionRequest)])
```
eriksendc, line 74: This should be `@actions_api.put("/", ...`

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        config = ActionPerformer._get_subtypes_by_name()[request.config_subtype].getx(
+            request.name
+        )
+        for key, value in request.fields.items():
+            setattr(config, key, value)
+        hmaconfig.update_config(config)
+        return UpdateActionResponse(response="The action config is updated.")
+
+    @actions_api.post("/create", apply=[jsoninator(CreateUpdateActionRequest)])
```
eriksendc, line 87: This should be `@actions_api.post("/", ...`

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        create an action
+        """
+        config = ActionPerformer._get_subtypes_by_name()[request.config_subtype](
+            **{"name": request.name, **request.fields}
+        )
+        hmaconfig.create_config(config)
+        return CreateActionResponse(response="The action config is created.")
+
+    @actions_api.post("/delete/<key>", apply=[jsoninator])
```
eriksendc, line 98: This should be `@actions_api.delete("/<key>", ...`

eriksendc (commented):
In the future we should follow the standard web api conventions of using get, post, put and delete, and then you don't need extra words in the endpoint url.

mengyangwang:
> In the future we should follow the standard web api conventions of using get, post, put and delete, and then you don't need extra words in the endpoint url.

Thanks Brian! I agree with you that we should use the standard way for new apis. I will submit another PR to refactor it!

PR #574, opened 2021-05-12
devcontainers: add less

BarrettOlson:
Summary
---------

Add `less` to the docker image

Test Plan
---------

Rebuilt the container.


schatten:
You won't believe I was just about to make the exact same change!

PR #576, opened 2021-05-12
Implement False Positive Writebacks

jeberl:
Summary
---------

This PR...
1. Implements False positive writebacks as a DISAGREE_WITH_TAGS reaction
2. Creates stub class for RemoveOpinion writebacks
3. Changes the descriptions of writebacks to match our v0 implementation plan
4. Removed the INGESTED writebacker which we determined was not worthy of implementing in v0

Test Plan
---------

`mypy hmalib && pytest` 
NOTE: No e2e test since the only code written (step 1 above) is using an already tested class. More e2e tests coming in later PRs


BarrettOlson (approved):
Stamping to unblock. Looks like it is mostly comment/mock return updates anyway. Let me know if you want me to dig into the existing write backer code to give a more thorough review (probably can wait for next PR imo) Thanks!

PR #577, opened 2021-05-12
Add FixedWidthCenterAlignedLayout layout and migrates some pages

### LINKED ISSUE #558: Use Bootstrap Grid "The Right Way"
Some pages (ThreatExchange Settings, Actions UI) are not using the Bootstrap grid the right way. You can see this because those pages' content is flush up against the left-hand nav. Check out action rule settings as a model for the pattern to use.

schatten:
Summary
---------
@eriksendc had called out in #558 that some pages are not using bootstrap the right way. This PR adds a new Layout component which makes it easy for folks to get on with the program and use one common way for laying out a page.

Other than the matches page which has a multi-pane design, all other pages can potentially be migrated. As part of this, few have been and more will be added if this is accepted. For practical purposes, this fixes #558.

I feel the settings pages could do with their own layout because of their tabbed nature. If this appeals to folks, I'll take a stab at that.

Test Plan
---------
Screenshots for the updated pages are shown below.

## Dashboard
Screen Shot 2021-05-12 at 19 54 43

## Content Details / Summary
Screen Shot 2021-05-12 at 19 54 58

## Signals
Screen Shot 2021-05-12 at 19 55 05

## Upload
Screen Shot 2021-05-12 at 19 55 12

## Submit Content
Screen Shot 2021-05-12 at 19 55 21

### DIFF
--- a/hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
+++ b/hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
@@ -8,9 +8,9 @@ import {
   Button,
   Row,
   Form,
-  Container,
   Collapse,
   Spinner,
+  Alert,
 } from 'react-bootstrap';
 import {Link} from 'react-router-dom';
 
@@ -23,6 +23,7 @@ import {
   OptionalMetadataField,
   NotYetSupportedField,
 } from '../components/SubmitContentFields';
+import FixedWidthCenterAlignedLayout from './layouts/FixedWidthCenterAlignedLayout';
 
 const FORM_DEFAULTS = {
   submissionType: undefined,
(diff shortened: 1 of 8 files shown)


eriksendc (approved):
LURV IT!

BarrettOlson (approved):
nit/nonblocking: This might come down to personal preference but are we sure we want a fixed width? The whitespace on both sides could be utilized for the more detail heavy pages. 

For example the Detail/Summary page has a lot of fields (more pending) and the extra space on large screens from the container being 'fluid' might help.

Even so I'd prefer we merge this as a standard layout we can adjust is better than the current ad hoc imo. Thanks!

hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
```diff
@@ -100,134 +101,128 @@ export default function SubmitContent() {
...
-              <p>
-                Currently only <b>images</b> using the submisson type{' '}
-                <b>Direct Upload</b> or <b>URL</b> are supported.
-              </p>
-            </div>
-            <div className="px-4 py-2" />
-          </Row>
-        </Container>
-      </div>
-      <Container className="mt-4" style={{overflowY: 'auto'}}>
```
BarrettOlson, line 120: Could you double check that adding a bunch of additional fields is still visible/scrollable? this override was originally to allow for that but it might not be needed anymore

schatten: I'll check.

schatten:
> nit/nonblocking: This might come down to personal preference but are we sure we want a fixed width? The whitespace on both sides could be utilized for the more detail heavy pages.
> 
> For example the Detail/Summary page has a lot of fields (more pending) and the extra space on large screens from the container being 'fluid' might help.
> 
> Even so I'd prefer we merge this as a standard layout we can adjust is better than the current ad hoc imo. Thanks!

Exactly! The matches page has still not been updated because it will remain full-width. I say we create more such layouts. eg. `FullWidthScrollableLayout`,  `FullWidthNonScrollableLayout` and then allow pages to choose.

If you think the detail / summary page needs more space, we can definitely go for full width.

hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
BarrettOlson: Feel free to just create an issue if so and I can fix it in my next PR (I want to update "metadata" to be 
"additional fields" anyway.

PR #583, opened 2021-05-17
[HMA][Multipart] Content Object and ActionEvent Storage + API 

BarrettOlson:
How to Review
---------

- [Diff 1/5](https://github.com/facebook/ThreatExchange/pull/583/commits/65d7ae4d3d96fd9003545ce2536ba3589831f407) Creates models for content and action record objects (and unit test)
- [Diff 2/5](https://github.com/facebook/ThreatExchange/pull/583/commits/706fc17de2aa513225186999e2eb4ae58a85f6a7) Begins recording ActionEventRecords in action_performer lambda
- [Diff 3/5](https://github.com/facebook/ThreatExchange/pull/583/commits/e284c33657a9c822cd976eed5df99ebc005da697) Moves existing related endpoints to sub-api `/content/<x>` and expose new models on same sub-api
- [Diff 4/5](https://github.com/facebook/ThreatExchange/pull/583/commits/47cada24ca7bb2fe8c6853a73fb63e075e800ec2) Begins recording content objects in submit API
- [Diff 5/5](https://github.com/facebook/ThreatExchange/pull/583/commits/cc723f21f05a34efbea1fc3c95265132f8c39492) WIP: Surfaces new action records and content details on summary (`/matches/<content_id>`) page

Summary
---------

This PR is my first pass at creating an interface/model for storing (and actually storing):

1) Content Objects (and additional submissions) including metadata/additional fields (Diff 1 & 4)
2) A record of actions taken in action_evaluator (Diff 1 & 2)

It also:

3) Refractors some legacy endpoints outside of `api_root.py` (Diff 3)
4) Exposed this new content and action record endpoints  (Diff 3)

And finally: (slight WIP but happy to merge as is)

5) Adds these new values to the content summary pages (Diff 5)

Test Plan
---------
See test of the e2e in the video below 
Not yet fully tested:
- ActionRecords end to end. (ui works with mocked values however) -> (need ability to config rules)
- Throughput of system with additional storage step -> (need to update store scripts)

https://user-images.githubusercontent.com/7664526/118521121-b50bc180-b708-11eb-98f9-4d74f1ce259b.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/content_models.py
+++ b/hasher-matcher-actioner/hmalib/common/content_models.py
@@ -27,7 +27,7 @@ class ContentObjectBase(DynamoDBItem):
 
 
 @dataclass
-class ActionEventRecord(ContentObjectBase, JSONifiable):
+class ActionEvent(ContentObjectBase, JSONifiable):
     """
     Record of action events in the action performer
     keyed based on action_label and time of event.
@@ -38,18 +38,18 @@ class ActionEventRecord(ContentObjectBase, JSONifiable):
     ACTION_TIME_PREFIX = "action_time#"
     DEFAULT_PROJ_EXP = "PK, SK, ActionLabel, ActionRules"
 
-    time_performed: datetime.datetime
+    performed_at: datetime.datetime
     action_label: str
-    action_rules: t.List
+    action_rules: t.List[str]
 
     @staticmethod
-    def get_dynamodb_action_time_key(time_performed: datetime.datetime) -> str:
-        return f"{ActionEventRecord.ACTION_TIME_PREFIX}{time_performed.isoformat()}"
+    def get_dynamodb_action_time_key(performed_at: datetime.datetime) -> str:
+        return f"{ActionEvent.ACTION_TIME_PREFIX}{performed_at.isoformat()}"
 
     def to_dynamodb_item(self) -> dict:
         return {
             "PK": self.get_dynamodb_content_key(self.content_id),
-            "SK": self.get_dynamodb_action_time_key(self.time_performed),
+            "SK": self.get_dynamodb_action_time_key(self.performed_at),
             "ActionLabel": self.action_label,
             "ActionRules": self.action_rules,
         }
@@ -59,7 +59,7 @@ class ActionEventRecord(ContentObjectBase, JSONifiable):
         Used by '/content/action-history/<content-id>' in lambdas/api/content
         """
         result = asdict(self)
-        result.update(time_performed=self.time_performed.isoformat())
+        result.update(performed_at=self.performed_at.isoformat())
         return result
 
     @classmethod
@@ -67,7 +67,7 @@ class ActionEventRecord(ContentObjectBase, JSONifiable):
         cls,
         table: Table,
         content_id: str,
-    ) -> t.List["ActionEventRecord"]:
+    ) -> t.List["ActionEvent"]:
 
         items = table.query(
             KeyConditionExpression=Key("PK").eq(
@@ -77,19 +77,19 @@ class ActionEventRecord(ContentObjectBase, JSONifiable):
             ProjectionExpression=cls.DEFAULT_PROJ_EXP,
         ).get("Items", [])
 
-        return cls._result_item_to_action_record(items)
+        return cls._result_item_to_action_event(items)
 
     @classmethod
-    def _result_item_to_action_record(
+    def _result_item_to_action_event(
         cls,
         items: t.List[t.Dict],
-    ) -> t.List["ActionEventRecord"]:
+    ) -> t.List["ActionEvent"]:
         return [
-            ActionEventRecord(
+            ActionEvent(
                 content_id=cls.remove_content_key_prefix(
                     item["PK"],
                 ),
-                time_performed=datetime.datetime.fromisoformat(
+                performed_at=datetime.datetime.fromisoformat(
                     item["SK"][len(cls.ACTION_TIME_PREFIX) :]
                 ),
                 action_label=item["ActionLabel"],
@@ -116,10 +116,10 @@ class ContentObject(ContentObjectBase, JSONifiable):
     content_type: str
     content_ref: str
     content_ref_type: str
-    additional_fields: t.Optional[t.Set]
-    submissions: t.List
-    created_on: datetime.datetime
+    submission_times: t.List[datetime.datetime]
+    created_at: datetime.datetime
     updated_at: datetime.datetime
+    additional_fields: t.Set[str] = field(default_factory=set)
 
     @staticmethod
     def get_dynamodb_content_type_key(content_type: str) -> str:
@@ -127,15 +127,15 @@ class ContentObject(ContentObjectBase, JSONifiable):
 
     def to_json(self) -> t.Dict:
         """
-        TODO docstring
+        Used by '/content/<content-id>' in lambdas/api/content
         """
         result = asdict(self)
         result.update(
             additional_fields=list(
                 self.additional_fields if self.additional_fields else set()
             ),
-            submissions=[s.isoformat() for s in self.submissions],
-            created_on=self.created_on.isoformat(),
+            submission_times=[s.isoformat() for s in self.submission_times],
+            created_on=self.created_at.isoformat(),
             updated_at=self.updated_at.isoformat(),
         )
         return result
@@ -153,8 +153,8 @@ class ContentObject(ContentObjectBase, JSONifiable):
                 "ContentRef": self.content_ref,
                 "ContentRefType": self.content_ref_type,
                 "AdditionalFields": self.additional_fields,
-                "Submissions": [s.isoformat() for s in self.submissions],
-                "CreatedOn": self.created_on.isoformat(),
+                "SubmissionTimes": [s.isoformat() for s in self.submission_times],
+                "CreatedOn": self.created_at.isoformat(),
                 "UpdatedAt": self.updated_at.isoformat(),
             }
         """
@@ -166,15 +166,15 @@ class ContentObject(ContentObjectBase, JSONifiable):
             # If ContentRef exists it needs to match or BAD THING(tm) can happen...
             ConditionExpression=Or(Attr("ContentRef").not_exists(), Attr("ContentRef").eq(self.content_ref)),  # type: ignore
             # Unfortunately while prod is happy with this on multiple lines pytest breaks...
-            UpdateExpression="""SET ContentRef = :cr, ContentRefType = :crt, Submissions = list_append(if_not_exists(Submissions, :empty_list), :s), CreatedOn = if_not_exists(CreatedOn, :co), UpdatedAt = :u ADD AdditionalFields :af""",
+            UpdateExpression="""SET ContentRef = :cr, ContentRefType = :crt, SubmissionTimes = list_append(if_not_exists(SubmissionTimes, :empty_list), :s), CreatedAt = if_not_exists(CreatedAt, :c), UpdatedAt = :u ADD AdditionalFields :af""",
             ExpressionAttributeValues={
                 ":cr": self.content_ref,
                 ":crt": self.content_ref_type,
                 ":af": self.additional_fields
                 if self.additional_fields
                 else {"Placeholder"},
-                ":s": [s.isoformat() for s in self.submissions],
-                ":co": self.created_on.isoformat(),
+                ":s": [s.isoformat() for s in self.submission_times],
+                ":c": self.created_at.isoformat(),
                 ":u": self.updated_at.isoformat(),
                 ":empty_list": [],
             },
@@ -214,9 +214,9 @@ class ContentObject(ContentObjectBase, JSONifiable):
             additional_fields=item["AdditionalFields"],
             # Notes careful not using this version to write back to the table...
             # Will dup previous submissions...
-            submissions=[
-                datetime.datetime.fromisoformat(s) for s in item["Submissions"]
+            submission_times=[
+                datetime.datetime.fromisoformat(s) for s in item["SubmissionTimes"]
             ],
-            created_on=datetime.datetime.fromisoformat(item["CreatedOn"]),
+            created_at=datetime.datetime.fromisoformat(item["CreatedAt"]),
             updated_at=datetime.datetime.fromisoformat(item["UpdatedAt"]),
         )
--- a/hasher-matcher-actioner/hmalib/common/content_models.py
+++ b/hasher-matcher-actioner/hmalib/common/content_models.py
@@ -1,0 +1,222 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import datetime
+from enum import Enum
+import typing as t
+
+from dataclasses import dataclass, field, asdict
+from mypy_boto3_dynamodb.service_resource import Table
+from boto3.dynamodb.conditions import Attr, Key, Or
+from botocore.exceptions import ClientError
+
+from hmalib.lambdas.api.middleware import JSONifiable
+from hmalib.models import DynamoDBItem
+
+
+@dataclass
+class ContentObjectBase(DynamoDBItem):
+    """
+    Base content obeject for ActionRecords and ContentObjects
+    Both of which used content_id as there PK
+    (This is similar to Hash and Match Records
+    but can easily be moved to it's own db table as they do not
+    use the other indexes)
+    """
+
+    content_id: str
+
+
+@dataclass
+class ActionEventRecord(ContentObjectBase, JSONifiable):
+    """
+    Record of action events in the action performer
+    keyed based on action_label and time of event.
+    For the time being the action_rules in the action message
+    being preformed are also stored as a list serialized json blobs
+    """
+
+    ACTION_TIME_PREFIX = "action_time#"
+    DEFAULT_PROJ_EXP = "PK, SK, ActionLabel, ActionRules"
+
+    time_performed: datetime.datetime
+    action_label: str
+    action_rules: t.List
+
+    @staticmethod
+    def get_dynamodb_action_time_key(time_performed: datetime.datetime) -> str:
+        return f"{ActionEventRecord.ACTION_TIME_PREFIX}{time_performed.isoformat()}"
+
+    def to_dynamodb_item(self) -> dict:
+        return {
+            "PK": self.get_dynamodb_content_key(self.content_id),
+            "SK": self.get_dynamodb_action_time_key(self.time_performed),
+            "ActionLabel": self.action_label,
+            "ActionRules": self.action_rules,
+        }
+
+    def to_json(self) -> t.Dict:
+        """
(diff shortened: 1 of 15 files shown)


hasher-matcher-actioner/hmalib/common/tests/test_content_models.py
```diff
@@ -0,0 +1,177 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
schatten, line 1: Yay! Tests!

hasher-matcher-actioner/hmalib/common/tests/test_content_models.py
```diff
@@ -0,0 +1,177 @@
...
+        cls.mock_dynamodb2 = mock_dynamodb2()
+        cls.mock_dynamodb2.start()
+        cls.create_mocked_table()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.mock_dynamodb2.stop()
+
+    @classmethod
+    def create_mocked_table(cls):
```
schatten, line 46: At some point, we might want to abstract out common features for tests. 

I also like starting "afresh".  See: https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/hmalib/tests/test_pdq_models.py#L346-L353

This allows you to have a clean dynamodb using a context-manager. 

Note: Not saying that you need to do this here. This is good!

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+    Record of action events in the action performer
+    keyed based on action_label and time of event.
+    For the time being the action_rules in the action message
+    being preformed are also stored as a list serialized json blobs
+    """
+
+    ACTION_TIME_PREFIX = "action_time#"
+    DEFAULT_PROJ_EXP = "PK, SK, ActionLabel, ActionRules"
+
+    time_performed: datetime.datetime
```
schatten, line 41: nit: To be consistent with other 'time' related attributes, `performed_at` might be better suited.

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+    action_label: str
+    action_rules: t.List
```
schatten, line 43: nit: List of?

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+    def write_to_table(self, table: Table):
+        """
+        Write operations for this object need to be special cased (to avoid overwritting)
+        Therefore we do not implement `to_dynamodb_item`
+
+        If you're curious it would ~look like this:
+        def to_dynamodb_item(self) -> dict:
+            return {
+                "PK": self.get_dynamodb_content_key(self.content_id),
+                "SK": self.get_dynamodb_content_type_key(self.content_type),
```
schatten, line 152: Can you explain why content_type may be different for the same content_key?

I know we have discussed this in the past, but might help others who read this.

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+     "This content_id exists, [x] overwrite? (I know what you're doing)"
+    is likely better.
+    """
+
+    CONTENT_TYPE_PREFIX = "content_type#"
+
+    content_type: str
+    content_ref: str
+    content_ref_type: str
+    additional_fields: t.Optional[t.Set]
```
schatten, line 119: nit: prefer that set itself is not optional. Collections, if optional, can always have length 0.

```python
additional_fields: t.Set[<type>] = field(default_factory=set)
```

This prevents unnecessary things like:
```python
if content_obj.additional_fields:
  for field in content_obj.additional_fields:
    <do something>
```

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+    submissions: t.List
+    created_on: datetime.datetime
```
schatten, line 121: `s/on/at`?

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+    updated_at: datetime.datetime
+
+    @staticmethod
+    def get_dynamodb_content_type_key(content_type: str) -> str:
+        return f"{ContentObject.CONTENT_TYPE_PREFIX}{content_type}"
+
+    def to_json(self) -> t.Dict:
+        """
+        TODO docstring
```
schatten, line 130: LOL

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        result = asdict(self)
+        result.update(
+            additional_fields=list(
+                self.additional_fields if self.additional_fields else set()
+            ),
+            submissions=[s.isoformat() for s in self.submissions],
```
schatten, line 137: nit: Was surprised to gather that `self.submissions` are just times. Better names?

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+            created_on=self.created_on.isoformat(),
+            updated_at=self.updated_at.isoformat(),
+        )
+        return result
+
+    def write_to_table(self, table: Table):
+        """
+        Write operations for this object need to be special cased (to avoid overwritting)
+        Therefore we do not implement `to_dynamodb_item`
```
schatten, line 146: If you are overriding `write_to_table`, what is the extra safety offered by not implementing `to_dynamodb_item`?

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+                "SK": self.get_dynamodb_content_type_key(self.content_type),
+                "ContentRef": self.content_ref,
+                "ContentRefType": self.content_ref_type,
+                "AdditionalFields": self.additional_fields,
+                "Submissions": [s.isoformat() for s in self.submissions],
+                "CreatedOn": self.created_on.isoformat(),
+                "UpdatedAt": self.updated_at.isoformat(),
+            }
+        """
+        table.update_item(
```
schatten, line 161: `put_item` also takes ConditionExpressions AFAIK. Perhaps explain why you are using `update_item`?

schatten (commented):
Will have a multipart review for quicker feedback. This is feedback for 
* [Diff 1/5](https://github.com/facebook/ThreatExchange/pull/583/commits/65d7ae4d3d96fd9003545ce2536ba3589831f407#) Creates models for content and action record objects (and unit test)

Nothing blocking, but perhaps explain why you chose update_item instead of put_item. And explain the rationale behind content_type being SK.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -61,6 +69,15 @@ def lambda_handler(event, context):
 
         perform_label_action(action_message, action_message.action_label)
 
+        ActionEventRecord(
+            content_id=action_message.content_key,
+            time_performed=datetime.datetime.now(),
+            action_label=action_message.action_label.value,
+            # Hack: the label rules model is not something I don't fully understand yet...
+            # rn this just says f-it let's make a list of json blobs we can recover and store it.
+            action_rules=[rule.to_aws_json() for rule in action_message.action_rules],
```
schatten, line 78: Is the eventual goal to save a reference to the immutable and versioned action_rule object?

This tells the rules, but doesn't say why the rules existed. If there were an audit, this would be hard to explain without some normalizing.

schatten: Why the rule existed would be: __\<User x\> removed condition 'Y' on \<date\>.__

schatten (commented):
non-blocking: unclear why action rules are getting denormalized here.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -216,6 +160,15 @@ def get_signal_hash_count() -> t.Dict[str, t.Tuple[int, str]]:
     ),
 )
 
+app.mount(
```
schatten, line 163: <3 it!

hasher-matcher-actioner/hmalib/lambdas/api/content.py
```diff
@@ -0,0 +1,111 @@
...
+    API that this API plugs into. Declare dependencies here, but initialize in
+    the root API alone.
+    """
+
+    # A prefix to all routes must be provided by the api_root app
+    # The documentation below expects prefix to be '/content/'
+    content_api = bottle.Bottle()
+    image_folder_key_len = len(image_bucket_key)
+
+    @content_api.get("/<key>", apply=[jsoninator])
```
schatten, line 58: Warning label on bottle: this means that `\<key\>` can't contain slashes.

hasher-matcher-actioner/hmalib/lambdas/api/content.py
```diff
@@ -0,0 +1,111 @@
...
+    def content(key=None) -> t.Optional[ContentObject]:
+        """
+        Content object for a given ID
+        see hmalib/commom/content_models.ContentObject for specific fields
+        """
+        return ContentObject.get_from_content_id(dynamodb_table, key)
+
+    @content_api.get("/action-history/<key>", apply=[jsoninator])
+    def action_history(key=None) -> ActionHistoryResponse:
```
schatten, line 67: nit: I was wondering what action_history_key is. Perhaps use content_key or content_id as the variable name?

schatten (commented):
- [Review 3/5](https://github.com/facebook/ThreatExchange/pull/583/commits/e284c33657a9c822cd976eed5df99ebc005da697) Moves existing related endpoints to sub-api `/content/<x>` and expose new models on same sub-api

blocking: Should acknowledge that `/` is not allowed in content_id. Or fix it.

Other than that, looks good to me. General nitty comments littered throughout.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -115,6 +117,23 @@ def to_json(self) -> t.Dict:
...
+):
+    # TODO add a confirm overwrite path for this
+    submit_time = datetime.datetime.now()
+    ContentObject(
+        content_id=request.content_id,
+        content_type="PHOTO",
+        content_ref=f"{image_folder_key}{request.content_id}",  # raw bytes + tmp urls are a bad idea atm, assume s3 object for now
+        content_ref_type=request.submission_type,
+        additional_fields=set(request.metadata) if request.metadata else set(),
+        submissions=[submit_time],
```
schatten, line 131: Implies re-submissions overwrite. Please confirm.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -142,6 +161,9 @@ def submit(
         if request.submission_type == SubmissionType.UPLOAD.name:
             fileName = request.content_id
             fileContents = base64.b64decode(request.content_ref)
+
+            record_content_submission(dynamodb_table, image_folder_key, request)
```
schatten, line 165: I suspect this is being done before s3_client.put_object to prevent a case where hasher starts to hash but content object doesn't exist yet. Document?

schatten (commented):
- [Review 4/5](https://github.com/facebook/ThreatExchange/pull/583/commits/47cada24ca7bb2fe8c6853a73fb63e075e800ec2) Begins recording content objects in submit API

Confirm that submission does an overwrite as noted below. Otherwise, nothing blocking.

schatten (approved):
- [Review 5/5](https://github.com/facebook/ThreatExchange/pull/583/commits/cc723f21f05a34efbea1fc3c95265132f8c39492) WIP: Surfaces new action records and content details on summary (`/matches/<content_id>`) page

Approving to unblock. Have left various comments on individual reviews. This one is good to go. For the others, there are some requests for comments or additional clarification. Overall, this looks good and the only blocking thing is.

> blocking: Should acknowledge that / is not allowed in content_id. Or fix it. 

Review Experience: ⭐ ⭐ ⭐ ⭐ ⭐

hasher-matcher-actioner/hmalib/common/tests/test_content_models.py
BarrettOlson: Yes! thought the same thing when writing it: https://github.com/facebook/ThreatExchange/issues/594

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: I mean, I don't think it should (same id w/different type) be honestly but we don't outright enforce it not being the case currently. The same is true for signal id and signal type. 

Do you mean to call out how a piece of content can produce multiple signals and they would share a content id but be different types?

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: `s/submissions/submission_times

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
```diff
@@ -61,6 +69,15 @@ def lambda_handler(event, context):
 
         perform_label_action(action_message, action_message.action_label)
 
+        ActionEventRecord(
+            content_id=action_message.content_key,
+            time_performed=datetime.datetime.now(),
+            action_label=action_message.action_label.value,
```
eriksendc, line 75: Maybe blocking? Similar to action the hack for action rules, this shouldn't just be the action label (which is included in the action rule) but the entire Action config object. Similar to action rules, eventually we can have a pointer to an ID + version instead.

eriksendc (changes requested):

hasher-matcher-actioner/webapp/src/pages/ContentDetails.jsx
```diff
@@ -34,15 +40,25 @@ export default function ContentDetails() {
 
   // TODO fetch actions once endpoint exists
   useEffect(() => {
-    setActions([]);
+    fetchContentActionHistory(id).then(result => {
+      if (result && result.action_history) {
+        setActionHistory(result.action_history);
+      }
+    });
+  }, []);
```
eriksendc, line 48: Is there a reason you have `, []` on this line and at line 54?

schatten: Unless we specify what attribute changes to useEffect() on, react gets into an infinite loop. `fetchContentActionHistory` changes state, causing the effect to be triggered again which changes state which .....

The `[]` tells react that useEffect should only be called once. Normally, you'd specify which attributes (state, props) you want to react to, but in this case, we want to react to nothing.

hasher-matcher-actioner/hmalib/common/content_models.py
schatten: The SK being content type makes me feel that we are expecting multiple items in this partition.

However, I can now see that this is more due to Content Items cohabiting a table which has PK and SK.

So, tell me if this is accurate. For HMADataStore

| PartitionKey | SortKey | ItemType |
|-------------|-----------|-----------|
|`c#images/129901.jpg`| `type#pdq` | `PipelinePDQHashRecord`|
|`c#images/129901.jpg`| `s#te#<indicator_id>` | `PDQMatchRecord`|
|`c#images/129901.jpg`| `content_type#photo` | `ContentObject`|
|`c#images/129901.jpg`| `action_time#2021-05-01.....` | `ActionEventRecord`|

I think we are doing a good job of naming right now, but for PipelinePDQHashRecord and PDQMatchRecord, it is not immediately apparent (looking at PK and SK alone) that we are looking at items of that type.

For the 3rd and 4th row, it is abundantly clear that I will receive an object of type ContentObject and ActionEventRecord by looking at the PK and SK design. But for row 1 and 2, it is less clear.

We could have the SK for hash records be `hash_type#pdq`. Similarly, for match records, it could be `signal_match#te#<indicator>` or something similar.

Anyway, all said and done, I think we need to be documenting this somewhere.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
BarrettOlson: yeah... now that I understand what you are saying this is a larger refactor and probably a different field on this object entirely. For the purposes of this PR I am going to punt but lets create an issue for it and (based on what the team thinks) likely address it before v0 is 'complete' ...

eriksendc: Works for me. Issue created: https://github.com/facebook/ThreatExchange/issues/596

eriksendc (approved):

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
BarrettOlson: Does imply but I handled that with 
`SubmissionTimes = list_append(if_not_exists(SubmissionTimes, :empty_list), :s)` in `UpdateExpression`
meaning all new values are appended. 

This gets messy if we start allowing query then update (in which case we'd want to null out submission or do something else fancy.) 

Will add comment.

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: because it should not be used. If you did some fancy stuff with super() things could get weird. Until it has a use case I am breaking convention by not implementing it this comment is more about explaining why. Please feel free to suggest edits

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: I don't think it has `UpdateExpression` which is needed for the SET and ADD stuff.

hasher-matcher-actioner/hmalib/lambdas/actions/action_performer.py
BarrettOlson: yeah, this is a little bit of a mess... 
Storing the whole rule object as is for now to avoid falling down another rabbit hole in this PR. 

Larger issues around recording action for an audit exist beyond the scope of this PR and might be worth a design review. cc: @eriksendc  and #596

hasher-matcher-actioner/hmalib/lambdas/api/content.py
BarrettOlson: no more <key>! :)

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: Agree and think it is accurate. The need to document is potentially issue worthy at this point.

PR #584, opened 2021-05-17
Delete descriptor

jeberl:
Summary
---------

Add function to delete a ThreatDescriptor to pytx3. This is necessary for the RemoveOpinion writebacker in HMA as the ThreatDescriptor will be deleted if the user elects to do so.

The same functionality could also be done by setting the item to expired (expired descriptors are automatically deleted) however this seemed cleaner.

Test Plan
---------

Ran function locally and saw that descriptor was deleted.

I checked the code manually to confirm that threat_updates would register the delete and I belive it would in `EntThreatDescriptorCollaborationSyncTrigger::genOnDirectDeletionOnlyChangesets` Will check with end to end test of all writebacks in future diff

### DIFF
--- a/python-threatexchange/threatexchange/api.py
+++ b/python-threatexchange/threatexchange/api.py
@@ -544,7 +544,7 @@ class ThreatExchangeAPI:
             self._base_url
             + "/"
             + str(indicator_id)
-            + "?fields=descriptors&access_token="
+            + "?fields=descriptors{privacy_members,indicator,type,owner}&access_token="
             + self.api_token
         )
 
--- a/python-threatexchange/threatexchange/api.py
+++ b/python-threatexchange/threatexchange/api.py
@@ -483,13 +483,39 @@ class ThreatExchangeAPI:
 
         return self.upload_threat_descriptor(postParams, showURLs, dryRun)
 
+    def delete_threat_descriptor(
+        self, descriptor_id, showURLs, dryRun
+    ) -> t.List[t.Any]:
+        url = (
+            self._base_url
+            + "/"
+            + str(descriptor_id)
+            + "?access_token="
+            + self.api_token
+        )
+        if showURLs:
+            print()
+            print("(DELETE) URL:")
+            print(url)
+        if dryRun:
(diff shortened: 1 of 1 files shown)


BarrettOlson:
@jeberl Can you move this to its own branch?

jeberl:
Oops. Yes forgot to push that

python-threatexchange/threatexchange/api.py
```diff
@@ -483,13 +483,39 @@ def copy_threat_descriptor(self, postParams, showURLs, dryRun):
 
         return self.upload_threat_descriptor(postParams, showURLs, dryRun)
 
+    def delete_threat_descriptor(
```
BarrettOlson, line 486: nit: can we add a note about how this actually requests the expiration of a td that would result in it no longer being visible in `threat_updates` and its eventual deletion? (I think that is what this is actually doing on the backend right?). We wouldn't want folks to misconstrue the capabilities of the API.

BarrettOlson: I worry that DELETE is too ambiguous in the print statements etc.

BarrettOlson: maybe just call it `expire_threat_descriptor` and mention that it should be deleted within x-time in the doc string?

BarrettOlson (changes requested):
RC to make sure you see my comment but if you feel strongly about calling it delete let me know and I can unblock you

python-threatexchange/threatexchange/api.py
jeberl: It doesn't expire the descriptor, it deletes the descriptor. What expires is the edge from collaboration to indicator.

I also was confused at this and had to look through the threat updates code to get a refresher but the descriptor is still deleted

Quick refresher:
1. When a threat descriptor expires it is deleted async
2. On all descriptor mutations (including deletes) the Collaboration -> Indicator edge is refreshed with a new update time and an expire time if necessary

BarrettOlson (approved):
Stamped

PR #561, opened 2021-05-07
use matcher_active to pause on matching, use dompurify to sanitize innerHTML

mengyangwang:
Summary
---------
previous PR : https://github.com/facebook/ThreatExchange/pull/516 , https://github.com/facebook/ThreatExchange/pull/544
Task : T88063259
     this task is to create ThreatExchange setting page.
     this PR is to do following things:
1. use dompurify to sanitize privacy description which contains HTML tags
2. use matcher_active to pause on matching 
3. Add copy text feature for privacy group name

Future Actions 
---------
 1.Change buttons to icons and adding more tooltips
 2. add last_updated field to HMAConfig table

Test Plan
---------
1. black hmalib
2. mypy hmalib
3. python -m py.test to pass all tests
5. make docker
6. make upload_docker 
7.  terraform apply
Test changes in localhost:
1. test if copy feature works
2. test if description links are working
3. test turn on and off matcher_active

https://user-images.githubusercontent.com/81996660/117511154-489c0000-af5b-11eb-9d0f-ba9a40b48770.mp4

PR #587, opened 2021-05-17
fix matcher_active

mengyangwang:
Summary
---------
previous PR : https://github.com/facebook/ThreatExchange/pull/561
this PR is to fix the issue with matcher_active:
1. fix the issue when disable matcher_active, it will still write the matching record to DynamoDB and publish to SQS queue
2. add lru cache(5 mins) for getting the matcher_active config from dynamoDB

Test Plan
---------
1. black hmalib
2. mypy hmalib
5. make docker
6. make upload_docker 
7.  terraform apply
Test changes in localhost:
1. disable matcher_active, upload an image, no matching records should be written to dynamoDB

https://user-images.githubusercontent.com/81996660/118562991-c3bf9c00-b73b-11eb-81d2-0f973a30315a.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
@@ -5,6 +5,7 @@ import os
 import pickle
 import boto3
 import datetime
+import time
 import typing as t
 from mypy_boto3_sns import SNSClient
 
@@ -17,6 +18,7 @@ from hmalib.common.signal_models import PDQSignalMetadata
 from hmalib.common.logging import get_logger
 from hmalib.common.config import HMAConfig
 from hmalib.common.fetcher_models import ThreatExchangeConfig
+from functools import lru_cache
 
 logger = get_logger(__name__)
 s3_client = boto3.client("s3")
@@ -50,7 +52,8 @@ def get_index(bucket_name, key):
     return result
 
 
-def get_privacy_group_matcher_active(privacy_group_id: str) -> bool:
+@lru_cache(maxsize=1)
+def get_privacy_group_matcher_active(privacy_group_id: str, _) -> bool:
     config = ThreatExchangeConfig.get(privacy_group_id)
     if not config:
         logger.warning("Privacy group %s is not found!", privacy_group_id)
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -50,7 +53,8 @@ def get_index(bucket_name, key):
     return result
 
 
-def get_privacy_group_matcher_active(privacy_group_id: str) -> bool:
+@lru_cache(maxsize=1)
```
schatten, line 56: Note: This will only cache for a single privacy group.

Is this intentional?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -103,58 +107,62 @@ def lambda_handler(event, context):
                 )
                 privacy_group_list = metadata.get("privacy_groups", [])
                 metadata["privacy_groups"] = filter(
```
schatten, line 109: BTW, just found out. The filter() does not produce an actual list. You need to do `list(filter(x))` to get a list.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -103,58 +107,62 @@ def lambda_handler(event, context):
...
-                    lambda x: get_privacy_group_matcher_active(str(x)),
+                    lambda x: get_privacy_group_matcher_active(
+                        str(x), time.time() // CACHED_TIME
```
schatten, line 111: If you are using time.time(), wouldn't that keep changing the argument list? How is this caching? Seems like every call is cache-busted. I'm not following.

schatten (changes requested):
From our discussion, the functional requirement seems to have been satisfied. We are only pushing messages and storing records if there are privacy groups active.

For the non-functional requirement, I'm not sure how this is caching. Using `time.time()` will cache bust every request. Additionally, the maxsize=1 will ensure ddb calls need to be made *every* time.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
mengyangwang: yeah I was wrong, I changed it to 128, do you think it's good enough?

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
mengyangwang: I am actually using ```time.time() // CACHED_TIME```, ```CACHED_TIME``` is default to 300. So ```time.time() // CACHED_TIME``` will change every 5 mins(300 seconds), which means after 5 mins, the cached result will be expired

mengyangwang: Added a comment to explain why using ```time.time() // CACHED_TIME```, will consider to optimize in the v1 release

schatten (approved):

Issue #588, opened 2021-05-18
Add Unique ID to Config Framework


eriksendc:
Right now Name is the primary key. Change this so that there's also a unique key (allows for updates instead of deletes / creates).

PR #590, opened 2021-05-18
Action Rules API

eriksendc:
Summary
---------

This PR adds GET, POST, PUT and DELETE web api methods at the /action-rules endpoint:
1. GET gets all the action rules
2. POST creates a new action rule (errors out if one already exists with the same name)
3. PUT modifies an existing action rule, or, if the name changes, creates a new action rule and deletes the old one
4. DELETE deletes an existing action rue (errors out if a rule doesn't exist)

Test Plan
---------

https://user-images.githubusercontent.com/8974354/118708493-caa4e800-b7e9-11eb-9f34-c18eb4429a9a.mp4


hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -137,6 +137,14 @@ def get_all(cls: t.Type[TConfig]) -> t.List[TConfig]:
                     ret.append(obj)
         return ret
 
+    @classmethod
+    def exists(cls: t.Type[TConfig], name: str) -> bool:
+        _assert_initialized()
+        ret = cls.get(name)
+        if not ret:
+            return False
+        return True
```
eriksendc, line 146: For whatever reason I was afraid to try `return not not ret`. Is that something I should try?

BarrettOlson: worth double checking based on the impl of get but I think `return bool(cls.get(name))` would work

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+from hmalib.common import config as hmaconfig
+from hmalib.common.evaluator_models import ActionRule
+from hmalib.common.classification_models import Label, ActionLabel
+
+
+logger = get_logger(__name__)
+
+
+@dataclass
+class ActionRulesAPIRequest(DictParseable):
```
BarrettOlson, line 22: Naming nit, I saw this in dataset_api as well. Personally, I feel the fact that this file is in the `/api/` folder and the classes have names with `Request` and `Response` makes the use of API in both the base filename and classnames a little superfluous. Thoughts?

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+    action_rule: ActionRule
+
+    @classmethod
+    def from_dict(cls, d: dict) -> "ActionRulesAPIRequest":
+        ar = ActionRule.from_aws(d["action_rule"])
+        logger.info("Deserialized ActionRule: %s", ar)
```
BarrettOlson, line 28: nit/ignorable: probably more of a debug level log unless you think we care about this enough every time it happens

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+    action_rules: t.List[ActionRule] = field(default_factory=list)
+
+    def to_json(self) -> t.Dict:
+        return {
+            "error_message": self.error_message,
+            "action_rules": [action_rule.to_aws() for action_rule in self.action_rules],
+        }
+
+
+def handle_unexpected_error(e: Exception):
```
BarrettOlson, line 44: I like this, I wonder if we want it as a top level decorator for all endpoints...

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+    logger.error("Unexpected error: %s", sys.exc_info()[0])
+    logger.exception(e)
+    response.status = 500
+
+
+def get_action_rules_api(hma_config_table: str) -> bottle.Bottle:
+    # The documentation below expects prefix to be '/action-rules/'
```
BarrettOlson, line 51: nit: s/expects/requires OR endpoint below are prefixed with...

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Creates an action rule.
+        """
+        logger.info("request: %s", request)
+        error_message = ""
+
+        try:
+            hmaconfig.create_config(request.action_rule)
+        except ClientError as e:
+            # TODO this test for "already exists" should be moved to a common place
```
BarrettOlson, line 85: +1 just a helper with _if_exist in the name is something I added for privacy group 'configs'

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+                )
+                logger.exception(e)
+            response.status = 500
+        except Exception as e:
+            error_message = "Unexpected error."
+            handle_unexpected_error(e)
+
+        return ActionRulesAPIResponse(error_message)
+
+    @action_rules_api.put("/<old_name>", apply=[jsoninator(ActionRulesAPIRequest)])
```
BarrettOlson, line 105: Why not have them both past in as part of the request (or have `oldname` be a query param) and just have the endpoint called `/update/`?

it seems like `/<key>` type endpoints are more trouble than they are worth due to the possibility of '/' in the names and other tbd routing concerns.

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+            logger.warning(
+                "An attempt was made to update an action rule named either '%s' or '%s' but neither exist.",
+                request.action_rule.name,
+                old_name,
+            )
+            response.status = 500
+
+        return ActionRulesAPIResponse(error_message)
+
+    @action_rules_api.delete("/<name>", apply=[jsoninator])
```
BarrettOlson, line 141: same here as for  `/update/` have the endpoint more general and have a query param for `name` ?

hasher-matcher-actioner/terraform/api/main.tf
```diff
@@ -197,6 +197,22 @@ resource "aws_apigatewayv2_route" "hma_apigateway_post" {
...
+
+resource "aws_apigatewayv2_route" "hma_apigateway_delete" {
+  api_id             = aws_apigatewayv2_api.hma_apigateway.id
+  route_key          = "DELETE /{proxy+}"
+  authorization_type = "JWT"
+  authorizer_id      = aws_apigatewayv2_authorizer.hma_apigateway.id
+  target             = "integrations/${aws_apigatewayv2_integration.hma_apigateway.id}"
+}
+
 resource "aws_apigatewayv2_integration" "hma_apigateway" {
```
BarrettOlson, line 216: did you figure our why we have this with `POST` but not the other verbs? If so could you document?

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+        if ActionRule.exists(request.action_rule.name):
+            try:
+                hmaconfig.update_config(request.action_rule)
+            except Exception as e:
+                error_message = "Unexpected error."
+                handle_unexpected_error(e)
+        elif ActionRule.exists(old_name):
+            try:
+                hmaconfig.create_config(request.action_rule)
+                hmaconfig.delete_config_by_type_and_name("ActionRule", old_name)
```
BarrettOlson, line 126: I think you've already brought it up but this feels weird to me... 
The whole need to create delete thing just for a rename makes me feel like we should treat names as ~immutable in terms of the endpoint but add the ability to "copy and rename" on the UI. This would send a request for create and then could provide a modal that asks the user if they want to delete the old copy. That way changing the name is viewed (accurately as having a more significant overhead).

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: I'm pretty sure DELETE HTTP requests aren't supposed to have a body. A quick google turned up this: 

From the HTTP/1.1 spec: A payload within a DELETE request message has no defined semantics; sending a payload body on a DELETE request might cause some existing implementations to reject the request.

It's pretty standard to put the ID of the thing you're deleting into the pathing. Check out the first big code block here: https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: I'm pretty sure the standard form of a web api for PUTs is to have the ID of the thing you're updating be in the pathing. Check out the first code block of https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
BarrettOlson: right, I agree on not having a body for any GET and DELETE requests but what are your thoughts on having this be a query param instead of the endpoint path?

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
BarrettOlson: I agree that is it standard, I just know we've had issue with path due to `/` in names etc. and this was a quick way to try and handle that instead. Not worth blocking on, just means the sooner we can add field validations the better.

hasher-matcher-actioner/hmalib/common/config.py
eriksendc: Ok. Using `return bool(cls.get(name))`

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: Sounds good. Nix'd API in the names.

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: Sounds good.

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: I agree. I think when we're done with V0 a thing to come back and standardize is API response consistency, including error handling. Added a TODO and filed Issue 599: https://github.com/facebook/ThreatExchange/issues/599

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: Sounds good. Went with this:

`# The endpoints below imply a prefix of '/action-rules'`

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
```diff
@@ -0,0 +1,165 @@
...
+    logger.exception(e)
+    response.status = 500
+
+
+def get_action_rules_api(hma_config_table: str) -> bottle.Bottle:
+    # The documentation below expects prefix to be '/action-rules/'
+    action_rules_api = bottle.Bottle()
+    HMAConfig.initialize(hma_config_table)
+
+    @action_rules_api.get("/", apply=[jsoninator])
```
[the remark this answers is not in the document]
eriksendc: Easy peasy! Ignored! :)

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: Yeah I'd rather have the fact that names with slashes break the system and push us toward implementing unique IDs be the solution.

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: I agree the current state is suboptimal. Adding a unique id column to the config table seems to be the right way to go... I added Issue 588 yesterday to track that: https://github.com/facebook/ThreatExchange/issues/588

hasher-matcher-actioner/hmalib/lambdas/api/action_rules_api.py
eriksendc: A query param is still not the standard web api way to do things. We should fix the root cause of that (documented in https://github.com/facebook/ThreatExchange/issues/588 I submitted yesterday) as opposed to putting in a bunch of workarounds that we might forget to deal with.

hasher-matcher-actioner/terraform/api/main.tf
eriksendc: Trial and error. I added an aws_apigatewayv2_integration named hma_apigateway_put and hma_apigateway_delete and hma_apigateway_get and paired them with their corresponding aws_apigatewayv2_route resource, and got an error "integration_method must be POST when integration_type is AWS_PROXY".

PR #591, opened 2021-05-18
[HMA] Make sample data match-able by creating pg config

BarrettOlson:
Summary
---------

Changes look more significant than reality imo (but I can split it up if desired). 

Context: 
- #560 and #566 added sample signals.
- However due to the change of #552 the .cvs needed an extra column.
- And due to the changes of #561 a config object needs to be present for the privacy group in the db for matching metadata to be created/recorded.

This PR:
- Adds the newly missing column (all zeros for the now unused fake descriptor id) 
- Add a button on the TE setting page to create the config so matches will be recorded.
- Along the way I did some refactoring to avoid copy pasting code and updated privacy_group_id to allow for type string.
- This also resulted in the creation of a new endpoint for creating a PG Config to the `dataset_api` however its only use in the UI is currently prepopulated. 

Test Plan
---------
See video (I also submitted a photo in the sample dataset and made sure a match was found, created, and visible on the UI.)

https://user-images.githubusercontent.com/7664526/118692664-4a29bb80-b7d8-11eb-9308-11ffc7ae9217.mov


hasher-matcher-actioner/hmalib/common/threatexchange_config.py
```diff
@@ -13,11 +13,47 @@
...
 logger = get_logger(__name__)
+
+# TODO Currently default to True for testing purpose,
+# need to switch it to False before v0 launch
 FETCHER_ACTIVE_DEFAULT = True
 WRITE_BACK_DEFAULT = True
 MATCHER_ACTIVE_DEFAULT = True
 
 
+def create_privacy_group_if_not_exists(
```
mengyangwang, line 24: this is nice!

mengyangwang (approved):

mengyangwang:
LGTM!

PR #585, opened 2021-05-17
Version 20

jeberl:
Summary
---------

Bump version to 0.0.20

PR #592, opened 2021-05-18
Edits to pytx3 necessary for writebacker

jeberl:
Summary
---------

After #584 realized there were 2 more edits to pytx3 I needed for writebacks. Would like to add these before I bump the version to 0.0.20 in #585 

Specifically, two changes:
1. Add function to remove a Reaction from a ThreatDescriptor
2. Add a `fields` argument to get_threat_descriptors function

Some context on 2
In writing the Writebacker I need to load the current privacy_groups from our ThreatDescriptor (to see if one needs to be removed). However any app can only see privacy groups for descriptor it owns. Therefore, we cannot ask for this field when we are loading all descriptors for a given indicator (error was thrown) and must make a second call to the api to load these private fields. A cursory look at graph API documentation did not reveal an easy way to only load the privacy_members field for descriptors that did not throw an error.

Test Plan
---------

Implemented writebacker using these functions locally. Saw that it worked. video of e2e test coming in writebacker diff

### DIFF
--- a/python-threatexchange/threatexchange/api.py
+++ b/python-threatexchange/threatexchange/api.py
@@ -197,6 +197,25 @@ class ThreatExchangeAPI:
         showURLs = kwargs.get("showURLs", False)
         includeIndicatorInOutput = kwargs.get("includeIndicatorInOutput", True)
 
+        default_fields = [
+            "raw_indicator",
+            "type",
+            "added_on",
+            "last_updated",
+            "confidence",
+            "owner",
+            "privacy_type",
+            "review_status",
+            "status",
+            "severity",
+            "share_level",
+            "tags",
+            "description",
+            "reactions",
+            "my_reactions",
+        ]
+        fields = kwargs.get("fields", default_fields)
+
         # Check well-formattedness of descriptor IDs (which may have come from
         # arbitrary data on stdin).
         for id in ids:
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/api.py
```diff
@@ -544,7 +591,7 @@ def get_threat_descriptors_from_indicator(
             self._base_url
             + "/"
             + str(indicator_id)
-            + "?fields=descriptors{privacy_members,indicator,type,owner}&access_token="
```
jeberl, line 547: This fails if any descriptor returned is owned by a different app

BarrettOlson: blocking, not sure I understand this change don't we just want to remove owner?

BarrettOlson (changes requested):
looks good for the most part but I'm confused about the last line, can you elaborate?

python-threatexchange/threatexchange/api.py
jeberl: sorry, `privacy_members` leads to failure if the owner ID is not our app. Could just remove privacy members but I think it might be better to return all default fields not just indicator, type, and owner

BarrettOlson: oh, this is undoing your change in #584, missed that. good to go

python-threatexchange/threatexchange/api.py
```diff
@@ -231,17 +251,23 @@ def get_threat_descriptors(self, ids, **kwargs):
...
+                #         {
+                #            "id": "03465026013486502",
+                #            "text": "uploaded_by_hma"
+                #         }
+                #      ]
+                #   },
+                #
+                # Canonicalize the tag ordering and simplify the
+                # structure to simply an array of tag-texts
+                tags = descriptor.get("tags", {"data": []})["data"]
```
BarrettOlson, line 267: if you don't know python well this line is a bit rough haha

BarrettOlson (approved):
Stamp

Issue #594, opened 2021-05-18
[HMA] generalize tests of dynamo db models (create test baseclass?)


BarrettOlson:
`mock_dynamodb2` is used in a very similar way in multiple tests
including [hmalib/tests/test_pdq_models.py](https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/hmalib/tests/test_pdq_models.py) and hmalib/common/tests/test_content_models.py. 

We can definitely abstract out common features for reuse in these tests.

PR #595, opened 2021-05-18
[HMA][MultiPart] Count matches per privacy group

schatten:
### Multi-part diff. Review in this order
* [Diff 1/5] [terraform: Move datastore definition to its own module](https://github.com/facebook/ThreatExchange/pull/595/commits/f3718518ef84e5588164f0aa5b336c46a6677db7)
* [Diff 2/5] [ Add methods to increment count of matches by privacy group ](https://github.com/facebook/ThreatExchange/pull/595/commits/671a2e5b3115d8c82191f3b100e215f26c53cd28)
* ~[Diff 3/5] [ Add PrivacyGroups to PDQ Match Record ](https://github.com/facebook/ThreatExchange/pull/595/commits/f7232a18d25079d5e60e88019e2be85d0f446d65)~ Reverted!
* [Diff 4/5] [ Stream-processing lambda: listens to DDB INSERT and MODIFY and updates counts](https://github.com/facebook/ThreatExchange/pull/595/commits/7224d326e7c08f70986cc180300e0f92d17276f4)
* ~[Diff 5/5] [ terraform: connect stream processing lambda to ddb updates ](https://github.com/facebook/ThreatExchange/pull/595/commits/c9cca15d2eae4e0d115714356ff7a6ec7dccd4ae)~ Reverted!

* [Diff 9/9] [ Second queue on matches SNS topic to increment pg counters ](https://github.com/facebook/ThreatExchange/pull/595/commits/3faafb04c4b90a0c6d6b7b40f2695870b4eea401)

Summary
---------
Adds a stream processor. This processor listens to updates to dynamodb and where required updates the counters of matches by privacy group.

A subsequent PR will show this information in the UI.

Test Plan
---------
New tests were added, so 
```
$ python -m py.test 
$ python -m mypy hmalib
$ python -m black hmalib
```

A fresh dynamodb instance was created and the storm script used to push some images to it.

Because of how my HMA instance is setup, the image matched 3 separate privacy groups.

As can be seen in the screenshot, all of their counters were incremented.
Screen Shot 2021-05-18 at 16 03 21

# Update

After the new changes using an SQS queue were done, I ran the storm test script against a different image to get counts updated
Screen Shot 2021-05-21 at 12 05 27
.


hasher-matcher-actioner/terraform/main.tf
```diff
@@ -309,3 +307,5 @@ resource "aws_secretsmanager_secret_version" "te_api_token" {
   secret_id     = aws_secretsmanager_secret.te_api_token.id
   secret_string = var.te_api_token
 }
+
```
BarrettOlson, line 329: super nit/aside: do we want to have a standard auto formatter for tf files, I've noticed some inconsistencies (beside the whitespace in this PR) that could mess with the blame...

hasher-matcher-actioner/hmalib/tests/test_pdq_models.py
```diff
@@ -387,3 +388,44 @@ def test_writes_inc_counters_only_for_the_updated_class(self):
             record.write_to_table(self.table)
 
             assert 0 == models.PDQMatchRecord.get_total_count(self.table)
+
+
+class MatchByPrivacyGroupCounterTestCase(CountersTest):
```
BarrettOlson, line 396: [Diff 2/5] nit: this test doesn't strike me as pdq specific, add a TODO note to refactor?

hasher-matcher-actioner/hmalib/common/count_models.py
```diff
@@ -0,0 +1,55 @@
...
+    Updates and retrieves counts of matches by Privacy Group.
+
+    Everything is a class method so you can extend this class, override what you
+    see fit and get the behavior without writing more than you need to. This
+    class might in the future become a base class for all counters.
+    """
+
+    @classmethod
+    def get_all_counts(cls, table: Table) -> t.Dict[str, int]:
+        response = table.scan(FilterExpression=Attr("PK").eq(cls._get_pkey()))
```
BarrettOlson, line 21: [Diff 2/5] scan could be unnecessarily pricey if we store this in the primary datastore. If we know can look up the SKs we want before had we should probably use them even if we want all counts... 
At the very least we could add a filter based on SK's prefix...

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
```diff
@@ -116,6 +120,7 @@ def lambda_handler(event, context):
                     signal_id,
                     metadata["source"],
                     metadata["hash"],
+                    metadata["privacy_groups"],
```
BarrettOlson, line 123: [Diff 3/5] `PDQSignalMetadata` has this information and can be looked up (we can add a helper to the class to make it easier).

Is your concern about retaining this info at all or at the very moment of the match?
- If at all I believe it is covered. 
- If at the very moment of the match then sure (Though I don't love the likely storage overlap and the different sources of truth to pick between presenting on the detail page but we can handle it) 

The fact that a signal is in a PG is not context the match record itself needs to care about imo; my understanding is that this is something `PDQSignalMetadata` and the match message would handle.

[not sure] but even if it is a at the very moment concern the match message downstream might be a better place to record this.

hasher-matcher-actioner/hmalib/lambdas/stream/ddb_stream_processor.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Listens to events on the dynamodb table and increments various counters.
+
+    Presently supported:
+    ---
+    1. split of hash and matches by privacy group
+
+    Planned
+    ---
+    1. move total counters here
```
BarrettOlson, line 62: [Diff 4/5] Also add TODO note where total counters are currently implemented stating that they are going to be moved (to avoid someone copying that pattern in the meantime)

hasher-matcher-actioner/terraform/stream-processor/main.tf
```diff
@@ -0,0 +1,107 @@
...
+
+data "aws_iam_policy_document" "stream_processor_iam_policy_document" {
+  statement {
+    effect    = "Allow"
+    actions   = ["dynamodb:GetItem", "dynamodb:Query", "dynamodb:Scan", "dynamodb:UpdateItem"]
+    resources = ["${var.datastore.arn}*"]
+  }
+  statement {
+    effect    = "Allow"
+    actions   = ["dynamodb:GetItem", "dynamodb:Query", "dynamodb:Scan", "dynamodb:PutItem", "dynamodb:DeleteItem"]
```
BarrettOlson, line 70: [Diff 5/5] feels like an over Allow, are all these actions being used? if not can we wait to add them until they are?

BarrettOlson (changes requested):
Looks good! RC-ing to check understanding of the change and talk about where we want to retain PG info...

Possibly a dumb question but worth documenting the answer to even so:
- We are effectively creating a lambda that is triggered when a change to the datastore is made yes?
- This lambda (depending on the event) then updates the datastore
- Won't this cause the lambda to be triggered again? (if not what prevents that/can we add additional protections to do so)
  - It shouldn't infinite loop because the second event trigger will be ignored i.e. not update again but it still feels wasteful

hasher-matcher-actioner/hmalib/common/count_models.py
schatten: I do not think it is pricey if the PK is the same. It will go to the same partition so the DTUs utilized should be limited.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
schatten: You are right. This information is already present in PDQSignalMetadata. I'll rework to change how the information is inferred.

hasher-matcher-actioner/hmalib/tests/test_pdq_models.py
schatten: https://github.com/facebook/ThreatExchange/issues/594#issue-894714795

Does this work? Haha. Went to file an issue. Saw you'd already done that. Resolved.

hasher-matcher-actioner/terraform/stream-processor/main.tf
schatten: Evidently all of these are necessary to be able to read ddb streams.

schatten: Nope. Copy-pasted. I'll limit it to the read only ones.

hasher-matcher-actioner/hmalib/lambdas/match_counter.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Presently supported:
+    ---
+    1. split of hash and matches by privacy group
+    """
+    records_table = dynamodb.Table(DYNAMODB_TABLE)
+    counters = defaultdict(lambda: 0)
+
+    for sqs_record in event["Records"]:
+        # TODO research max # sqs records / lambda_handler invocation
```
BarrettOlson, line 32: ignore: is this comment just going to be in every lambda we write haha?

hasher-matcher-actioner/hmalib/lambdas/match_counter.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Listens to events on a queue attached to the match SNS topic and increments
+    various counters.
+
+    Presently supported:
+    ---
+    1. split of hash and matches by privacy group
+    """
+    records_table = dynamodb.Table(DYNAMODB_TABLE)
+    counters = defaultdict(lambda: 0)
```
BarrettOlson, line 29: ignore: yo dawg, I heard you liked lambda

hasher-matcher-actioner/terraform/counters/main.tf
```diff
@@ -0,0 +1,157 @@
...
+  assume_role_policy = data.aws_iam_policy_document.lambda_assume_role.json
+  tags = merge(
+    var.additional_tags,
+    {
+      Name = "MatchCounterLambdaRole"
+    }
+  )
+}
+
+data "aws_iam_policy_document" "match_counter_iam_policy_document" {
```
BarrettOlson, line 112: do we need more than "dynamodb:UpdateItem" for ["${var.datastore.arn}*"]? I am not seeing the lambda do anything else?

hasher-matcher-actioner/terraform/counters/variables.tf
```diff
@@ -0,0 +1,66 @@
...
+
+variable "datastore" {
+  description = "DynamoDB Table to store hash and match information into"
+  type = object({
+    name = string
+    arn  = string
+  })
+}
+
+variable "config_table" {
```
BarrettOlson, line 48: Could be missing something but it doesn't look like this or anything to do with the "config_table" is used but all are passed in. Could be from a copy paste and I think it is likely we will want to add it eventually to enable configurations after tf apply but could you note it as currently unused  in a comment if you want to keep it in the meantime? 

(my [not blocking] preference is to reduce the amount of things we give to what is being used for now)

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -110,6 +110,26 @@ module "pdq_signals" {
...
+  lambda_docker_info = {
+    uri = var.hma_lambda_docker_uri
+    commands = {
+      match_counter = "hmalib.lambdas.match_counter.lambda_handler"
+    }
+  }
+  log_retention_in_days = var.log_retention_in_days
+  measure_performance   = var.measure_performance
+  matches_sns_topic_arn = aws_sns_topic.matches.arn
+  config_table = {
```
BarrettOlson, line 127: .

BarrettOlson (approved):
Nice! approving to unblock. It feels like there is some leftover code from copying that ideally would get cleaned up for comment acknowledged before merge, however my/the teams other tf work is enough of a glass house that I don't think it is at all blocking haha. 

Thanks!

PR #597, opened 2021-05-18
Implement RemoveOpinion ThreatExchange Writebacker

jeberl:
Summary
---------

Last unimplemented writebacker option RemoveOpinion

Test Plan
---------

1. Added unit tests
2.Tested all writebackers together by trigging them locally (not through UI). 
    a) Find an indicator that exists in HMA but that our app doesnt have a descriptor for 
    b) Ensure that our app has not reacted to any of the other descriptors yet
    c) trigger a FalsePositive writeback and watch as other descriptor is reacted to
    d) trigger a RemoveOpinion writeback and watch as reaction is removed
    e) trigger a TruePositive writeback and watch as a descriptor is created with the right fields and privacy group
    f) trigger a RemoveOpinion writeback and watch as descriptor is deleted
3. Tested that Triggering from UI works as well

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/writebacker_models.py
+++ b/hasher-matcher-actioner/hmalib/common/writebacker_models.py
@@ -239,7 +239,7 @@ class ThreatExchangeRemoveOpinionWritebacker(ThreatExchangeWritebacker):
     """
 
     def _writeback_impl(self, writeback_signal: BankedSignal) -> str:
-
+        privacy_group_id = writeback_signal.bank_id
         indicator_id = writeback_signal.banked_content_id
         descriptors = self.te_api.get_threat_descriptors_from_indicator(indicator_id)
         my_descriptor = self.my_descriptor(descriptors)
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -234,10 +247,78 @@ class ThreatExchangeRemoveOpinionWritebacker(ThreatExchangeWritebacker):
...
+    ) -> str:
+        new_privacy_groups = [
+            pg
+            for pg in my_descriptor["privacy_members"]
+            if isinstance(pg, str) and pg != privacy_group_id
+        ]
+        if not new_privacy_groups:
+            response = self.te_api.delete_threat_descriptor(
+                my_descriptor["id"], True, False
+            )
```
jeberl, line 299: Add comment here: "If no other privacy groups we should delete the descriptor"

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -151,6 +151,23 @@ def te_api(self) -> ThreatExchangeAPI:
         api_key = AWSSecrets().te_api_key()
         return ThreatExchangeAPI(api_key)
 
+    def my_descriptor(
+        self, descriptors: t.List[t.Dict[str, t.Any]]
+    ) -> t.Optional[t.Dict[str, t.Any]]:
+        print(descriptors)
```
BarrettOlson, line 157: .

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -151,6 +151,23 @@ def te_api(self) -> ThreatExchangeAPI:
...
+        for descriptor in descriptors:
```
BarrettOlson, line 158: suggest adding a comment or docstring like:
`loop until we find the one descriptor we own for a given indicator (if it exists)`

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -180,8 +193,8 @@ def _writeback_impl(self, writeback_signal: BankedSignal) -> str:
             "review_status": "REVIEWED_MANUALLY",
             "status": "MALICIOUS",
         }
-        print(descriptors)
 
+        # If we already have a descriptor we can copy it and upload make sure to never expire it
         if my_descriptor:
```
BarrettOlson, line 198: nit: is this a function call or a variable/property? (probably alright to leave if you rename `self.my_descriptor(...)`)

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -180,8 +193,8 @@ def _writeback_impl(self, writeback_signal: BankedSignal) -> str:
             "review_status": "REVIEWED_MANUALLY",
             "status": "MALICIOUS",
         }
-        print(descriptors)
 
+        # If we already have a descriptor we can copy it and upload make sure to never expire it
```
BarrettOlson, line 197: same nits as before: `s\copy\update` and `s\make sure to never expire it\set it to not expire`

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -234,10 +247,78 @@ class ThreatExchangeRemoveOpinionWritebacker(ThreatExchangeWritebacker):
...
+                f"No descriptor to remove for indicator {writeback_signal.banked_content_id}"
+            )
+
+        logs.extend(self.remove_false_positive_from_descriptors(other_desriptors))
+
+        for log_message in logs:
+            if "Error" in log_message:
+                logger.error(log_message)
+            else:
+                logger.info(log_message)
```
BarrettOlson, line 285: based on `perform_writeback` this will double log...

BarrettOlson (approved):
Looks mostly good! added a few minor concerns inline. Please address before merging.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
jeberl: I like "copy" here because the function is `copy_threat_descriptor`. Reworded otherwise

BarrettOlson: again nit but because it has come up twice, 'copy' is a fault in the function name the comment is trying to correct not reinforce...
The call outs I think we want to address here or in a docstring because it is overly confusing are:
1) That this will just result in 1 descriptor even though the word copy is in the method name
2) This descriptor is set to not expire but could be later if something changes (that is why I was against the word 'never')
@jeberl

PR #601, opened 2021-05-19
For match metrics, add and use correct metric

schatten:
Summary
---------
So far, we were using the number of times the index was searched as a number of matches metric. This is patently false. The number of matches is actually the number of times match records were written.

This adds a new metric, measures it, and uses it to report back stats.

Test Plan
---------
Deployed to prod and saw the number of matches (as viewed in the dashboard graphs) drop.

Note: There is still some work to be done to ensure things are accurate. I'm not sure they are at this point.


BarrettOlson (approved):
LGTM

Issue #596, opened 2021-05-18
ActionEvent should include the full details of the action that was executed


eriksendc:
Right now we persist an ActionLabel and an ActionRule as a part of ActionEvent. ActionRule includes all the bits of the rule that lead to the action being taken. ActionLabel simply names that action that was taken, but doesn't include the URL, HTTP verb, headers, etc. that were used when that action execution was performed. We should have that persisted some way, like we do with the action rule, for audit purposes.

PR #602, opened 2021-05-19
[HMA] store action performer as well in action event

BarrettOlson:
Summary
---------

Addition to Diff 2/5 of #583 from @eriksendc suggestion

Test Plan
---------

Same as #583


schatten (approved):
LOKTM: Looks OK to me. 

I'm wondering whether this is simply kicking the versioned objects can down the road or taking on loads of tech debt on behalf of our partners.

This works, but not in a way I'm happy with. I see #596 has already been filed and am fine with merging this in as tech-debt to speed v0 up.

BarrettOlson:
I concur but as noted, I am prioritizing a more complete system over a tech-debt free one with this PR. The point you're making is a good one that I agree with though and is worth discussing further. Thanks

PR #603, opened 2021-05-19
Can't switch directly from TP <-> FP

jeberl:
Summary
---------

Writebacks can't currently handle switching directly from True Positive to False Positive. The quickest v0 solution is to disable this and require users to first remove a previous opinion before adding a new opinion. We could potentially change this in the future but for v0 this should work

Decided that a tool tip wasn't really necessary here as having only one enabled action "Remove Opinion" seems to make it clear how to change an opinion. if anyone feels particularly strongly I can add one but already a lot of text/explanations on this page

Test Plan
---------

Screen Shot 2021-05-19 at 5 30 06 PM

### DIFF
--- a/hasher-matcher-actioner/webapp/src/components/OpinionTableCell.jsx
+++ b/hasher-matcher-actioner/webapp/src/components/OpinionTableCell.jsx
@@ -45,7 +45,8 @@ export default function OpinionTableCell({
         <Dropdown.Item
           size="sm"
           disabled={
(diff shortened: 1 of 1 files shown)


BarrettOlson (approved):
Sure haha

BarrettOlson:
I do feel strongly that the options and behavior at a minimum need a 'nux' type intro if not a new tooltip but for now stamping to unblock.

jeberl:
What's a " 'nux' type intro "?

PR #604, opened 2021-05-19
Close the loop on Writebacks

jeberl:
Summary
---------

Ensure that writebacks flow all the way through the system. From changing opinion in UI -> API -> dynamoDB -> Writebacker -> ThreatExchange ->Fetcher -> dynamoDB -> API -> UI.

A few necessary/helpful changes:
1. Edit fetcher to update the PendingOpinionChange property on a fetch. @BarrettOlson determined that the necessary logic would be to see if the opinion has changed and if so remove any pending change.
1. typing in signal_models to ensure signal_id is always a str
2. typing in fetcher to convert privacy group ID to str

Test Plan
---------

From UI do the following and watch changes be propagated through the system. 
1. Upload an image which our test app currently doesnt have a descriptor for but a noter app has a descriptor in the privacy group
2. Watch match and see that opinion is "Unknown"
3. Change opinion to True Positive. Confirm that a new descriptor is created for our app and the opinion changes from Unknown -> Unknown + pending -> TruePositive
4. Remove opinion. Confirm that the descriptor is delete and the opinion changes from TruePositive -> True Positive +  pending -> Unknown
5. Change opinion to False Positive. Confirm that we react DiSAGREE_WITH_TAGS to the other apps descriptor and the opinion changes from Unknown -> Unknown + pending -> FalsePositive
6. Remove opinion. Confirm our reaction is removed and the opinion changes from FalsePositive -> True FalsePositive +  pending -> Unknown

NOTE: Testing this took about 10 minutes because of time taken to flow through the whole system the full system so I didn't make a video.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/signal_models.py
+++ b/hasher-matcher-actioner/hmalib/common/signal_models.py
@@ -55,6 +55,7 @@ class PDQSignalMetadata(SignalMetadataBase):
     """
 
     SIGNAL_TYPE = "pdq"
+    SIGNAL_TYPE_PROJECTION_EXPRESSION = "PK, ContentHash, UpdatedAt, SK, SignalSource, SignalHash, Tags, PendingOpinionChange"
 
     def to_dynamodb_item(self) -> dict:
         return {
@@ -122,7 +123,6 @@ class PDQSignalMetadata(SignalMetadataBase):
         table: Table,
         signal_id: t.Union[str, int],
         signal_source: str,
-        ds_id: str = "",
     ) -> t.List["PDQSignalMetadata"]:
         """
         Load the metadata objects relaed to this signal.
@@ -132,9 +132,9 @@ class PDQSignalMetadata(SignalMetadataBase):
             KeyConditionExpression=Key("PK").eq(
                 cls.get_dynamodb_signal_key(signal_source, signal_id)
             )
-            & Key("SK").begins_with(cls.DATASET_PREFIX + ds_id),
-            ProjectionExpression="PK, ContentHash, UpdatedAt, SK, SignalSource, SignalHash, Tags, PendingOpinionChange",
+            & Key("SK").begins_with(cls.DATASET_PREFIX),
             FilterExpression=Attr("HashType").eq(cls.SIGNAL_TYPE),
+            ProjectionExpression=cls.SIGNAL_TYPE_PROJECTION_EXPRESSION,
         ).get("Items", [])
         return cls._result_items_to_metadata(items)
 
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -315,23 +317,56 @@ def post_apply(self, updated: t.Dict = {}):
...
             # example row format: ('<raw_indicator>', '<indicator-id>', '<descriptor-id>', '<time added>', '<space-separated-tags>')
             # e.g (10736405276340','096a6f9...064f', '1234567890', '2020-07-31T18:47:45+0000', 'true_positive hma_test')
-            if PDQSignalMetadata(
-                signal_id=int(row[1]),
+            new_tags = row[4].split(" ") if row[4] else []
+
+            # Figure out if we have a new opinion about this indicator and clear out a pending change if so
+            # If this is a new indicator without metadata we also should have no pending opinion about it
+            new_pending_opinion_change = PendingOpinionChange.NONE
+            for signal in PDQSignalMetadata.get_from_signal(
```
BarrettOlson, line 328: Unfortunately the way `get_from_signal` works (`query`) is a fair amount slower than way we'd implement `get_from_signal_and_ds_id` (`get_item`). 
This is bad because for every updated item we run the query. If you want help with writing something `get_from_signal_and_ds_id` tomorrow morning or you'd just like me to write it (more than happy to because it's a gap I left in that model when I created it) let me know but I do think it is blocking

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -315,23 +317,56 @@ def post_apply(self, updated: t.Dict = {}):
...
+                if signal.ds_id == str(self.privacy_group):
+
+                    # Threat Updates guarentees there is either 0 or 1 special tags on a descriptor
+                    opinion_tags = ThreatDescriptor.SPECIAL_TAGS
+                    old_opinion = [tag for tag in signal.tags if tag in opinion_tags]
+                    new_opinion = [tag for tag in new_tags if tag in opinion_tags]
+
+                    # If our opinion changed, set the pending opinion change to None
+                    # otherwise keep it unchanged
+                    new_pending_opinion_change = (
```
BarrettOlson, line 341: nit: wondering if it is worth checking for the case where it hasn't changed but it is the pending state already (I think it is possible if we fetch right when a change is requested on the UI but that is pretty niche)

BarrettOlson (changes requested):
Nice! unfortunately I think `get_from_signal` is a little too inefficient to use here. Other than that, looks good!

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -303,6 +305,41 @@ def _apply_updates_impl(
 
         post_apply_fn(updated)
 
+    def get_new_pending_opinion_change(
+        self, metadata: PDQSignalMetadata, new_tags: t.List[str]
+    ):
+        # Figure out if we have a new opinion about this indicator and clear out a pending change if so
+
+        # Threat Updates guarentees there is either 0 or 1 opinion tags on a descriptor
```
BarrettOlson, line 313: nit: python-threatexchange/threatexchange/descriptor.ThreatDescriptor.from_te_json is what actually guarantees this

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -315,23 +352,49 @@ def post_apply(self, updated: t.Dict = {}):
...
+                S3ThreatDataConfig.SOURCE_STR,
+                str(self.privacy_group),
+            )
+
+            if metadata:
+                new_pending_opinion_change = self.get_new_pending_opinion_change(
+                    metadata, new_tags
+                )
+            else:
+                # If this is a new indicator without metadata we should have no pending opinion about it
```
BarrettOlson, line 372: if you get here you can just return...

hasher-matcher-actioner/hmalib/common/signal_models.py
```diff
@@ -123,35 +124,53 @@ def get_from_signal(
...
+    @classmethod
+    def get_from_signal_and_ds_id(
+        cls, table: Table, signal_id: t.Union[str, int], signal_source: str, ds_id: str
+    ) -> t.Optional["PDQSignalMetadata"]:
+        item = table.get_item(
+            Key={
+                "PK": cls.get_dynamodb_signal_key(signal_source, signal_id),
+                "SK": cls.DATASET_PREFIX + ds_id,
+            },
+            ProjectionExpression=cls.SIGNAL_TYPE_PROJECTION_EXPRESSION,
```
BarrettOlson, line 150: get_item doesn't need this if you want all the fields anyway.

hasher-matcher-actioner/hmalib/common/writebacker_models.py
```diff
@@ -322,9 +322,9 @@ def remove_descriptor_from_privacy_group(
 
     def remove_false_positive_from_descriptors(self, descriptors) -> t.List[str]:
         ret = []
+        reaction = ThreatExchangeFalsePositiveWritebacker.reaction
```
BarrettOlson, line 325: why this change?

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -315,23 +352,49 @@ def post_apply(self, updated: t.Dict = {}):
...
                 ds_id=str(self.privacy_group),
                 updated_at=datetime.now(),
                 signal_source=S3ThreatDataConfig.SOURCE_STR,
                 signal_hash=row[0],  # note: not used by update_tags_in_table_if_exists
-                tags=row[4].split(" ") if row[4] else [],
-            ).update_tags_in_table_if_exists(table):
+                tags=new_tags,
+                pending_opinion_change=new_pending_opinion_change,
+            )
+            if metadata.update_tags_in_table_if_exists(table):
```
BarrettOlson, line 384: nit: having two update calls is less than ideal but not going to block on it (can you add a #TODO comment)

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -360,17 +423,14 @@ def write_s3_text(txt_content: io.StringIO, bucket, key: str) -> None:
     logging.basicConfig(level=logging.INFO)
     # This will only kinda work for so long - eventually will
     # need to use a proper harness
```
BarrettOlson, line 425: can we just delete this? the `if __name__ == "__main__":` code should just be in a test harness if we want to commit it. It super easy to get out of date.

BarrettOlson (approved):
Cool cool. Added a few things to address inline but nothing fully blocking.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
jeberl: Don't we still need to create the metadata?

hasher-matcher-actioner/hmalib/common/signal_models.py
jeberl: Want to keep the fields returned same as `get_from_signal`

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: nope (and your code does not anyway)

hasher-matcher-actioner/hmalib/common/signal_models.py
BarrettOlson: For get_item though PROJECTION_EXPRESSION is just to filter out fields you don't want.

PR #608, opened 2021-05-20
[HMA] make pdq_matcher write of metadata be conditional 

BarrettOlson:
Summary
---------

This changes the matcher to only update/create a metadata object  if it is not found in the table. Once initially created it is the fetcher's job to keep the item up to date.

Test Plan
---------
Test of `write_to_table_if_not_found`
```
import boto3
from hmalib.common.signal_models import PDQSignalMetadata

dynamodb = boto3.resource("dynamodb")
table = dynamodb.Table("barrett-HMADataStore")

print(
    PDQSignalMetadata(
        "example_id_1",
        "example_pg_1",
        datetime.datetime.now(),
        "te",
        "hash_val",
        ["tag1"],
    ).write_to_table_if_not_found(table)
)

print(
    PDQSignalMetadata(
        "example_id_1",
        "example_pg_1",
        datetime.datetime.now(),
        "te",
        "hash_val",
        ["tag1"],
    ).write_to_table_if_not_found(table)
)


``` 
output:
```
True
False
```

Tested changes to matcher via the UI by uploading an image twice and checking that UpdatedAt no longer changes in the table (because the values are the same add after create updates should be coming from the UI or the fetcher not the matcher) .

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_matcher.py
@@ -133,8 +133,8 @@ def lambda_handler(event, context):
                         ).write_to_table(records_table)
 
                     for pg in metadata.get("privacy_groups", []):
-                        # TODO: we might be able to get away with some 'if exists/upsert' here
-                        # TODO: @BarrettOlson Write if don't exist or only update given fields is now needed!
+                        # Only update the metadata if it is not found in the table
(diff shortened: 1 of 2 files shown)


BarrettOlson:
> Love it. Dunno whether you noticed or not. This only works because we are using transactions. In transactions, if one of the condition expressions fail, the entire transaction is avoided.
> 
> If we had used batch_writes here, we would have counters incrementing for failed writes!

I did! I had to test it to make sure it was that convenient haha

PR #617, opened 2021-05-21
[HMA] Replace Upload page and refresh Submit Page

BarrettOlson:
Summary
---------

This PR does a number of small things to unfortunately connected files which made my usually strategy of breaking up at commit level rough/impossible. Still willing to try if this review becomes onerous. 

Largest Changes:
- swaps "Upload" side bar page with "Submit Content"
- Changes `SubmissionType` enum (python & JS) to only the currently supported methods
- Adds new supported method (Post URL Upload)
  - This is what was used and tested on the old upload pages (this allows it to be used on the submit page)
 
Renaming:
- `ContentIdAndTypeField` ->  `ContentUniqueIdField` 
  - done to avoid confusion as the only type we support atm is PHOTO and should be broken up once we support additional
- `metadata` -> `AdditionalFields` (and all associated fields/helpers) 
  - change was already the case at the storage layer this just has the submit layer match
- `content_ref`  -> `content_bytes_url_or_file_type` (in submit request body)
  - don't love the name but the field changes meaning based on  `SubmissionType`
  -  and it NOT match content_ref in the storage layer so it need to be renamed to avoid confusion

Minor:
- some text string on the UI and comments were updated 
- form alignment on submit page
- Add additional field button

Test Plan
---------
This should actually change very little functionality.

https://user-images.githubusercontent.com/7664526/119198938-2ecce380-ba58-11eb-93be-623f250dee84.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/submit.py
@@ -64,21 +64,19 @@ class InitUploadRequestBody(DictParseable):
         return cls(d["content_id"], d["file_type"])
 
 
-# TODO use enum in storage class
 class SubmissionType(Enum):
-    UPLOAD = "Direct Upload"
-    URL = "URL"
-    RAW = "Raw Value (example only)"
-    S3_OBJECT = "S3 Object (example only)"
+    POST_URL_UPLOAD = "Upload"
+    DIRECT_UPLOAD = "Direct Upload (~faster but only works for images < 3.5MB)"
+    FROM_URL = "From URL"
 
 
 @dataclass
 class SubmitContentRequestBody(DictParseable):
-    submission_type: str  # TODO Enum
+    submission_type: str  # Enum SubmissionType names
     content_id: str
-    content_type: str  # TODO Enum
-    content_ref: t.Union[str, bytes]
-    metadata: t.Optional[t.List]
+    content_type: str  # Only photo supported
+    content_bytes_url_or_file_type: t.Union[str, bytes]
+    additional_fields: t.Optional[t.List]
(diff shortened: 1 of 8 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -147,65 +147,112 @@ def get_submit_api(
...
+        return SubmitContentError(
+            content_id=request.content_id,
+            message="not yet supported",
+        )
 
+    def from_url(
+        request: SubmitContentRequestBody,
+    ) -> t.Union[SubmitContentResponse, SubmitContentError]:
+        """
+        TODO xyz doctring
```
eriksendc, line 207: ?

hasher-matcher-actioner/webapp/src/Api.jsx
```diff
@@ -139,24 +139,48 @@ export async function submitContent(
...
+export async function submitContentPostURLUpload(
   submissionType,
   contentId,
   contentType,
-  contentRef,
-  metadata,
+  content,
+  additionalFields,
+) {
+  const postURL = await apiPost('/submit/', {
```
eriksendc, line 161: This name threw me for a sec... it's the response object from the apiPost call... What about `response` or `submitContentResponse` (whatever matches the python dataclass that's returned)? Nit-ish. Non-blocking.

eriksendc (approved):

hasher-matcher-actioner/webapp/src/Api.jsx
BarrettOlson: Thanks! will update name and add comment.

PR #619, opened 2021-05-22
[HMA] Add count of matches / privacy group to API and to UI

schatten:
Summary
---------
Since we had already added hash counts to the ThreatExchange settings tab, it made sense to add the match counts there as well. I have not added Tate's feedback or played around with the UI. Just replaced last modified with match count.

Last modified is getting tricky because the source of information could be from many places. The config itself is updated. The hash and match counts are updated by separate processes. So, just dropped it from the UI.

Also, removed the /hash_counts/ API and added everything to the same /datasets/ call.

It seems more and more likely that the signals page could go away. There is no information on that page that is not anywhere in the ThreatExchange settings page. And showing the same information in two places becomes confusing.

Test Plan
---------
```
$ python -m mypy .
$ python -m black .
$ python -m py.test 
```

Screen Shot 2021-05-22 at 19 19 27

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -1,19 +1,14 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 import os
-import traceback
 import bottle
 import boto3
 import json
-import base64
-import datetime
 import typing as t
 from apig_wsgi import make_lambda_handler
 from bottle import response, error
 
-from hmalib import metrics
 from hmalib.common.logging import get_logger
-from hmalib.common.s3_adapters import ThreatExchangeS3PDQAdapter, S3ThreatDataConfig
 
 from .action_rules_api import get_action_rules_api
 from .actions_api import get_actions_api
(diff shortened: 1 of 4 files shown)


mengyangwang (approved):
this looks so nice!

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -114,46 +93,6 @@ class SignalSourceSummary(t.TypedDict):
     updated_at: str
 
 
-def get_signals() -> t.List[SignalSourceSummary]:
```
BarrettOlson, line 117: This breaks the signal page, if you want to go through with this change and not update `"/signals"` please also delete the endpoint, `webapp/src/pages/Signals.jsx`, and its associated sidebar item.

BarrettOlson (changes requested):
RC to flag minor breaking change. (okay if you just want to delete the page)

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
schatten: Thanks for the flag. Already prepping a PR for removing the signals sidebar.

I'll drop the new one on your plate so we can merge both in together and not break master.

BarrettOlson: stamped then.

BarrettOlson (approved):

eriksendc (approved):

PR #605, opened 2021-05-19
actions UI

mengyangwang:
Summary
---------
previous PR: https://github.com/facebook/ThreatExchange/pull/573, https://github.com/facebook/ThreatExchange/pull/581
this PR is to update actions tab to read/create/update/delete action performers from dynamoDB

Test Plan
---------

https://user-images.githubusercontent.com/81996660/118894954-331bc400-b8d3-11eb-9e24-26094c1b95ed.mov

PR #620, opened 2021-05-24
make action performer name and webhook type editable

mengyangwang:
Summary
---------

In previous PR [605](https://github.com/facebook/ThreatExchange/pull/605), I've made the field 'Action Name' and 'webhook type' non-editable. 
This PR is to make these two fields editable, the changes involved:
1. Made changes to ```/actions``` api, added two url params ```/actions/<old_name>/<old_config_sub_stype>```
2. Compare old_name with updated name and old_config_sub_stype with updated config sub type, if either one is different then delete the old config and create the new config. Otherwise will update old config
3. Made changes to UI related files to make those two fields editable

Test Plan
---------
1. ```mypy hmalib```
2. ```python -m pytest```
3. ```black hmalib```
4. ```make docker```
5. ```make upload_docker```
6. ```terraform apply```
7. test changes in localhost

https://user-images.githubusercontent.com/81996660/119396801-5c599d00-bca3-11eb-9184-b76e14a5a231.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
@@ -4,7 +4,7 @@ import bottle
 import typing as t
 from dataclasses import dataclass, asdict
 from .middleware import jsoninator, JSONifiable, DictParseable
-from hmalib.common.config import HMAConfig, update_config
+from hmalib.common.config import HMAConfig
 from hmalib.common import config as hmaconfig
 from hmalib.common.actioner_models import ActionPerformer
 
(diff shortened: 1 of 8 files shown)


schatten (approved):
Looks good. Since you are allowing edits to the name, do you think it makes sense to get the user to confirm name changes?

If action names are changed, any action rules using them will no longer work, right? That could be something that the user is not expecting.

Also, another question inline.

mengyangwang:
> Looks good. Since you are allowing edits to the name, do you think it makes sense to get the user to confirm name changes?
> 
> If action names are changed, any action rules using them will no longer work, right? That could be something that the user is not expecting.
> 
> Also, another question inline.

That's right, if name is changed then the action rules using this action performer will no longer work. 
I think the confirmation is a good idea! I'll add that

jeberl:
> Looks good. Since you are allowing edits to the name, do you think it makes sense to get the user to confirm name changes?

+1 . Maybe also don't allow a name change if its attached to a rule? This may be more complicated so I'm ok punting

hasher-matcher-actioner/hmalib/lambdas/api/actions_api.py
```diff
@@ -71,17 +71,26 @@ def fetch_all_actions() -> FetchAllActionsResponse:
...
         """
         Update an action url and headers
         """
-        config = ActionPerformer._get_subtypes_by_name()[request.config_subtype].getx(
-            request.name
-        )
-        for key, value in request.fields.items():
-            setattr(config, key, value)
-        hmaconfig.update_config(config)
+        if old_name != request.name or old_config_sub_stype != request.config_subtype:
```
jeberl, line 84: Maybe a comment about how these fields are immutable in our storage model so we must delete and create?

jeberl (approved):
Looks good. few optional comments

PR #622, opened 2021-05-24
Remove the SIgnals page from the App and the Sidebar

schatten:
Summary
---------
All of the functionality that the signals page was offering has been duplicated in the ThreatExchange settings page. The new functionality coming in (match counts) #619 also went in to that page.

Having this page could become confusing to users. Thus, removing it.

Test Plan
---------
Local `npm start` and viewed the app.

Screen Shot 2021-05-24 at 17 01 23


BarrettOlson (approved):
LGTM

PR #623, opened 2021-05-24
[hma] Add ActionHistoryTable to Details page

BarrettOlson:
Summary
---------

Makes ActionHistory its own component for the details page. The table shows all actions for a piece of content giving action label and the time it was performed  (most recent first) by default it only displays the two most recent actions but users can see the total count and hit "...see all" to show the rest.

PR also moves opinion change submitted toast to the top right.

Test Plan
---------

Screen Shot 2021-05-24 at 5 30 45 PM
Screen Shot 2021-05-24 at 5 31 01 PM
Screen Shot 2021-05-24 at 5 31 11 PM
Screen Shot 2021-05-24 at 5 31 28 PM
Screen Shot 2021-05-24 at 5 31 42 PM

### DIFF
--- a/hasher-matcher-actioner/webapp/src/components/ActionHistoryTable.jsx
+++ b/hasher-matcher-actioner/webapp/src/components/ActionHistoryTable.jsx
@@ -1,0 +1,105 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ */
+
+import React, {useState, useEffect} from 'react';
+import {Col, Collapse, Row, Table, Button} from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import Spinner from 'react-bootstrap/Spinner';
+
+import {fetchContentActionHistory} from '../Api';
+import {CopyableTextField} from '../utils/TextFieldsUtils';
+import {formatTimestamp} from '../utils/DateTimeUtils';
+
+const DEFAULT_NUM_ROWS = 2;
+
+export default function ActionHistoryTable({contentKey}) {
+  const [actionHistory, setActionHistory] = useState(null);
+  const [showAll, setShowAll] = useState(false);
+
+  useEffect(() => {
+    fetchContentActionHistory(contentKey).then(result => {
+      if (result && result.action_history) {
+        result.action_history.sort(
+          (event1, event2) =>
+            new Date(event2.performed_at).getTime() -
+            new Date(event1.performed_at).getTime(),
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/webapp/src/components/ActionHistoryTable.jsx
```diff
@@ -0,0 +1,105 @@
...
+import {CopyableTextField} from '../utils/TextFieldsUtils';
+import {formatTimestamp} from '../utils/DateTimeUtils';
+
+const DEFAULT_NUM_ROWS = 2;
+
+export default function ActionHistoryTable({contentKey}) {
+  const [actionHistory, setActionHistory] = useState(null);
+  const [showAll, setShowAll] = useState(false);
+
+  useEffect(() => {
```
schatten, line 20: Component doing API calls? Page should do the call and pass props to component.

schatten (approved):
nit about a Component making an API call. To me the hard line between Pages and Components is exactly that.

Also, would Love it if instead of just showing two rows, you added an extra row with ellipsis indicating that there are more rows. That's just an example. Could be some other visual feedback as long as it is _in_ the table.

In this case, we are relying on users to **notice** the 'show all' link _outside_ the table to know that there are more rows.

Approving to unblock. Both my comments are nits.

BarrettOlson:
> nit about a Component making an API call. To me the hard line between Pages and Components is exactly that.

Hmm I might be missing something but avoiding "Component making an API call." is semi at odds with the reasoning behind the creation and usage of react hooks to enables stateful components. This is likely a longer conversation but since the pattern is not new to this PR I am going to leave it for now as the same changes I'd make would also be needed for `/webapp/src/components/ContentMatchTable.jsx`.

> Also, would Love it if instead of just showing two rows, you added an extra row with ellipsis indicating that there are more rows. That's just an example. Could be some other visual feedback as long as it is _in_ the table.
> 
> In this case, we are relying on users to **notice** the 'show all' link _outside_ the table to know that there are more rows.

I originally tried something like this. I'm not sure the picture makes it clear but the "Total Action Events 4        -----------                        ...see all" is in fact that last row of the table. When I had just the `ellipsis` there it was actually less clear to me and stuff like making the last row gray ended up looking awful imo. I think the ideal implementation might not be achievable without a different table component as a base.  

> 
> Approving to unblock. Both my comments are nits.

Thanks, I appreciate the flexibility you're giving me to ignoring them for now

PR #625, opened 2021-05-25
[hma] clean up outdated scripts

BarrettOlson:
Summary
---------

The HMA system outgrew its initial test harness. In prep for writing a new one I want to do some clean up first.

We also don't need a top level 'tests/scripts' directory for hasher-matcher-actioner that is redundant with 'scripts/` when it only has one file.

Test Plan
---------

Commands were already outdated/redundant. (however: https://xkcd.com/1172/)


eriksendc (approved):

eriksendc (approved):
I'm assuming from what you wrote in the summary that `check_deployed_instance.py` will be replaced by something. Is that right?

PR #628, opened 2021-05-31
Add user-pool name to terraform outputs

schatten:
Summary
---------
Output user pool name. This is required to make the 'Grant users access to HMA' wiki entry smoother.

Test Plan
---------
Ran `terraform output`

Screen Shot 2021-05-31 at 19 22 00

### DIFF
--- a/hasher-matcher-actioner/terraform/authentication/outputs.tf
+++ b/hasher-matcher-actioner/terraform/authentication/outputs.tf
@@ -4,6 +4,10 @@ output "webapp_and_api_user_pool_id" {
   value = var.use_shared_user_pool ? var.webapp_and_api_shared_user_pool_id : aws_cognito_user_pool.webapp_and_api_user_pool[0].id
 }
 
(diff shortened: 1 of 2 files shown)


BarrettOlson (approved):
Approving to unblock but am a little confused about the value add, probably I just need to see the wiki page. 
Note: The Id is already present as an output from the submodule and this change to surface name wont work for in the case of "use_shared_user_pool" etc. being set in .tfvars because the name is not known by the local tf apply.

schatten:
> Approving to unblock but am a little confused about the value add, probably I just need to see the wiki page.
> Note: The Id is already present as an output from the submodule and this change to surface name wont work for in the case of "use_shared_user_pool" etc. being set in .tfvars because the name is not known by the local tf apply.

The ID is not visible on the AWS UI, so won't be useful. Will catch up with you or @eriksendc on the shared user pool stuff. I'm not sure I fully understand it.

PR #631, opened 2021-06-01
[hma][scripts] Move part of storm_submit_api to util file for reuse

BarrettOlson:
Summary
---------

- (title)Moves part of storm_submit_api to util file for reuse
- Adds option to get `RefreshToken` instead of `IdToken` to `get_auth_token` (going to be used in soak testing)

Long term this will make the scripts more maintainable but the short term is a little inconvenience when it comes to using ec2 instances. (requires the copying of two files)

Test Plan
---------

Made sure both storm scripts still worked with changes.


schatten (approved):
Looks great! Thanks for extracting this out into a module. This would help partners start integrating far more easily.

PR #634, opened 2021-06-01
Matches page design feedback

### LINKED ISSUE #630: Matches Page + Content Detail Design Feedback
* Header / Sidebar title should be consistent
* Preview pane could have more information from the Content Detail pages.
* Last Seen → Uploaded / Submitted At
* Links can be renamed to ‘filter matches for the same ....’
* Bring the filter option closer to the list it is filtering. Above the scrollable pane.
* Stronger division between the panes

schatten:
Fixes #630. Can review commit by commit or all at once.

Summary
---------
Changes are mostly visual.
* Added border to sidebar
* Wrote a new layout for full width pages, made matches page use that layout
* Moved the content / signal search bar to the table header
* Added more content to the content details page
* Fancier empty state for no match selected

Test Plan
---------
Clicked around on prod data.

Screenshots attached.

### Empty state for matches
localhost_3000_matches (1)

### When additional fields are not available
localhost_3000_matches

### When additional fields are available
localhost_3000_matches

### DIFF
--- a/hasher-matcher-actioner/webapp/src/Sidebar.jsx
+++ b/hasher-matcher-actioner/webapp/src/Sidebar.jsx
@@ -7,13 +7,10 @@ import PropTypes from 'prop-types';
 import {NavLink} from 'react-router-dom';
 import {Auth} from 'aws-amplify';
 import './styles/_sidebar.scss';
-import {SeparatorBorder} from './utils/constants';
 
 export default function Sidebar({className}) {
   return (
-    <div
-      className={`${className} d-flex flex-column`}
-      style={{borderRight: SeparatorBorder}}>
+    <div className={`${className} d-flex flex-column`}>
       <div className="px-2 pt-2 pb-4Right">
         <a className="navbar-brand alert-link" href="/">
           HMA
--- a/hasher-matcher-actioner/webapp/src/Sidebar.jsx
+++ b/hasher-matcher-actioner/webapp/src/Sidebar.jsx
@@ -10,8 +10,10 @@ import './styles/_sidebar.scss';
 
 export default function Sidebar({className}) {
   return (
-    <div className={`${className} d-flex flex-column`}>
-      <div className="px-2 pt-2 pb-4">
+    <div
+      className={`${className} d-flex flex-column`}
+      style={{borderRight: '1px solid #ececec'}}>
+      <div className="px-2 pt-2 pb-4Right">
         <a className="navbar-brand alert-link" href="/">
           HMA
         </a>
--- a/hasher-matcher-actioner/webapp/src/Sidebar.jsx
+++ b/hasher-matcher-actioner/webapp/src/Sidebar.jsx
@@ -7,12 +7,13 @@ import PropTypes from 'prop-types';
 import {NavLink} from 'react-router-dom';
 import {Auth} from 'aws-amplify';
 import './styles/_sidebar.scss';
+import {SeparatorBorder} from './utils/constants';
 
 export default function Sidebar({className}) {
   return (
     <div
       className={`${className} d-flex flex-column`}
-      style={{borderRight: '1px solid #ececec'}}>
+      style={{borderRight: SeparatorBorder}}>
       <div className="px-2 pt-2 pb-4Right">
         <a className="navbar-brand alert-link" href="/">
           HMA
--- a/hasher-matcher-actioner/webapp/src/Sidebar.jsx
+++ b/hasher-matcher-actioner/webapp/src/Sidebar.jsx
@@ -53,7 +53,6 @@ export default function Sidebar({className}) {
             activeClassName="text-white bg-secondary rounded"
             to="/settings"
             className="nav-link px-2">
-            <span className="glyphicon glyphicon-cog" />
             Settings
           </NavLink>
         </li>
--- a/hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
+++ b/hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
@@ -78,10 +78,12 @@ export default function ContentMatchPane({contentId, signalId, signalSource}) {
             )}
 
             <p className="mt-4">
-              <Link to={`/matches/${contentId}`}>See all details</Link>
+              <Link to={`/matches/${contentId}`}>
+                Open the details page for content
+              </Link>
             </p>
 
-            <h4 className="mt-4">Filter search results</h4>
+            <h4 className="mt-4">Filter results on this page</h4>
             <li>
               <Link to={`?signalId=${signalSource}|${signalId}`}>
                 Show matches for this signal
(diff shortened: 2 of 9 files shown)


hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -36,43 +38,61 @@ export default function ContentMatchPane({contentId, signalId, signalSource}) {
...
+
+            <h4 className="mt-4">Additional Fields</h4>
+            {contentDetails && contentDetails.additional_fields.length !== 0 ? (
+              contentDetails.additional_fields.map(field => <li>{field}</li>)
+            ) : (
+              <p>No additional fields provided</p>
+            )}
+
+            <p className="mt-4">
+              <Link to={`/matches/${contentId}`}>See all details</Link>
```
BarrettOlson, line 81: This should be worded to make it clear that it will take the user to a new page.

hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -36,43 +38,61 @@ export default function ContentMatchPane({contentId, signalId, signalSource}) {
...
+            </p>
+
+            <h4 className="mt-4">Filter search results</h4>
```
BarrettOlson, line 84: We should make this clear that the following is distinct from the other links in this pane. See below:

hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -36,43 +38,61 @@ export default function ContentMatchPane({contentId, signalId, signalSource}) {
...
+            <li>
+              <Link to={`?signalId=${signalSource}|${signalId}`}>
+                Show matches for this signal
```
BarrettOlson, line 87: What about something like making this or a part of it a button? 
e.g.
"Show only the matches for this signal: <Button>Apply Filter</Button>"

hasher-matcher-actioner/webapp/src/components/ContentMatchPane.jsx
```diff
@@ -36,43 +38,61 @@ export default function ContentMatchPane({contentId, signalId, signalSource}) {
...
+              </Link>
+            </li>
+
+            <li>
+              <Link to={`?contentId=images/${contentId}`}>
```
BarrettOlson, line 92: .

hasher-matcher-actioner/webapp/src/utils/constants.jsx
```diff
@@ -37,3 +37,5 @@ export const StatNames = Object.freeze({
   HASHES: 'hashes',
   MATCHES: 'matches',
 });
+
+export const SeparatorBorder = '1px solid #ececec';
```
BarrettOlson, line 41: Why have this in js vs a .css file?

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
```diff
@@ -118,54 +132,29 @@ function MatchListFilters({searchInAttribute, searchQuery}) {
...
+            </InputGroup.Prepend>
             <FormControl
               ref={inputRef}
               onChange={e => setLocalSearchQuery(e.target.value)}
               value={localSearchQuery || ''}
-              htmlSize="40"
             />
             <InputGroup.Append>
-              <Button onClick={loadSearchPage}>Search</Button>
+              <Button onClick={loadSearchPage}>Filter</Button>
```
BarrettOlson, line 154: The word 'Search' is all over this file...
I agree the Filter is better but I think we should also refactor `setLocalSearchInAttribute` etc., while we are here to things like `setLocalFilterInAttribute` (to avoid the code drifting from the user experience/docs)

hasher-matcher-actioner/webapp/src/Sidebar.jsx
```diff
@@ -7,11 +7,14 @@ import PropTypes from 'prop-types';
...
 import './styles/_sidebar.scss';
+import {SeparatorBorder} from './utils/constants';
 
 export default function Sidebar({className}) {
   return (
-    <div className={`${className} d-flex flex-column`}>
-      <div className="px-2 pt-2 pb-4">
+    <div
+      className={`${className} d-flex flex-column`}
+      style={{borderRight: SeparatorBorder}}>
```
BarrettOlson, line 16: why not update `webapp/src/styles/_sidebar.scss` ?

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
```diff
@@ -48,48 +50,60 @@ export default function Matches() {
...
-            <div className="px-4 py-2">
+    <FullWidthLocalScrollingLeftAlignedLayout title="Matches">
+      <Container className="h-100 v-100" fluid>
+        {/* ^ This container is everything below the header */}
+        <Row className="d-flex align-items-stretch h-100">
+          {/* Each child of this row spans half the available space dividing into left and right pane. */}
+          <Col
+            md={6}
+            className="d-flex flex-column justify-content-start h-100 px-0"
+            style={{borderRight: SeparatorBorder}}>
```
BarrettOlson, line 61: @eriksendc and I had a back and forth about in code style awhile ago, Brain did we determine a standard practice we wanted to follow or no?

BarrettOlson (changes requested):
Love the changes! I had enough nit/questions in line that I think it is worth RC-ing for discussion purposes (as a few were structural). However if you want to address this in a future PR let me know and I am willing to can stamp to unblock :)

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
schatten: Would be happy if you or @eriksendc pointed me to a place where we are doing the current standard way of doing things. 

Consider this answering all questions about `SeparatorBorder`: My understanding was that we want to stay as close to bootstrap defaults as possible so our changes would be 'local' only. Which is why inline-styles made most sense. But I'm open to other solves for this.

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
schatten: Makes sense. Will do.

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
eriksendc: You can still override without it being inline, right? Seems like if we're going to have a standard separator boarder then we'd want to centralize this... It's just not what I've seen, this constructing of an inline stile with a constant. What about having a class and then using it somewhere?

schatten:
@BarrettOlson as discussed offline, you are good with the change set as long as the inline-style confusion is cleared.

Based on @eriksendc's suggestion, I went ahead and moved the variable to our own variables.scss. So this is now entirely out of javascript. Since the border can be on any side, using one class would not have cut it. I'm merging this in current state. 

If we want to rethink how these styles are done, we can put up another PR.

PR #637, opened 2021-06-01
[hma][tf] hotfix for those using shared user pool

BarrettOlson:
Summary
---------

#628 broke tf for those with use_shared_user_pool="True"

Test Plan
---------

`$ terraform -chdir=terraform plan`


schatten (approved):
Never knew this. Things are clearer now. Thanks!

PR #638, opened 2021-06-01
add validation rules when for update and delete

mengyangwang:
Summary
---------

In previous PR https://github.com/facebook/ThreatExchange/pull/620, I made action's name and type editable. Since action rules are dependent on actions, delete an action or update an action's name may cause issues. In this PR, I made below changes:

- Fetch and save all action rules when refresh the action page
- If an action has one or more action rules
     1. disable the delete button, display a help icon to show the reason why the delete button is disabled
     2. In edit mode, make action name non-editable, display a hover over message to show the reason why it's non-editable

Test Plan
---------
Screen Shot 2021-06-01 at 5 39 08 PM
Screen Shot 2021-06-01 at 5 39 19 PM
Screen Shot 2021-06-01 at 5 39 42 PM
Screen Shot 2021-06-01 at 5 39 32 PM


jeberl:
Love this pR and thanks for taking the time to make this page more user friendly

jeberl (approved):

PR #641, opened 2021-06-03
[hma][scripts] Add API methods to script utils

BarrettOlson:
Summary
---------

Expands `HasherMatcherActionerAPI` to support a number of API endpoints useful for testing. 

Sneaky side fixes (can breakout to own PR if requested):
- Remove vestigial `MatchSummary` field
- Fixed an issue in `authentication-shared/main.tf` and adds to `authentication/main.tf` so `get_auth_token` will work for all deployments with the right local aws credentials.

Test Plan
---------

Manually confirmed all new methods work (see bottom of `scripts/script_utils.py`) EXCEPT for `create_action` and `create_action_rule` which I plan to test as part of 'test deployments script' in my next PR.

### DIFF
--- a/hasher-matcher-actioner/scripts/script_utils.py
+++ b/hasher-matcher-actioner/scripts/script_utils.py
@@ -5,6 +5,7 @@
 Utilities for hma scripts
 """
 
+import os
 import json
 import base64
 import requests
@@ -29,9 +30,15 @@ class HasherMatcherActionerAPI:
         refresh_token: str = None,
     ) -> None:
         self.api_url = api_url
-        self.api_token = api_token
         self.client_id = client_id
         self.refresh_token = refresh_token
+        self.session = requests.Session()
+        self.session.headers.update(
+            {
+                "content-type": "application/json",
+                "authorization": api_token,
+            }
+        )
 
     def _refresh_token(self):
         """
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/scripts/script_utils.py
```diff
@@ -30,14 +62,38 @@ def _get_header(self) -> t.Dict[str, str]:
     def _get_request_url(self, api_path: str) -> str:
         return urljoin(self.api_url, api_path)
 
+    def _get_with_auth(
```
schatten, line 65: A more flexible way to do this is to expose a `requests.Session` object [1]. That way users have access to the full gamut of requests capabilities while re-using the auth. A session object has almost all `request.*` methods available.

Imagine having to write `_post_multi_part_with_auth()` and `_put_with_auth()` and `_delete_with_auth()`

[1]: https://docs.python-requests.org/en/master/user/advanced/

schatten: Oh, and did I mention it respects Keep-Alive and re-uses the underlying TCP connection for higher performance?

hasher-matcher-actioner/scripts/script_utils.py
```diff
@@ -86,7 +134,137 @@ def send_single_submission_url(
             headers={"content-type": "image/jpeg"},
         )
 
+    def get_content_hash_details(
+        self,
+        content_id: str,
+        api_path: str = "/content/hash/",
+    ):
```
schatten, line 141: typing? Can we re-use the representation classes from hmalib here? I guess that is what you were planning to do. Totally okay to do that as separate PR.

schatten: all new API methods in this module seem to lack that. You can address my concerns as a reply to this message.

schatten (changes requested):
Would prefer using requests.Session instead of creating proxy methods. If you do it now, adding DELETE and PUT calls will be simpler and have the full array of requests' feature set.

Other than that, nit about typing. You mentioned that you'll use hmalib as a dependency. If we can return objects of the representation classes (return types in API methods), that would reduce the need to refer to attributes in the objects. eg. 
```python
        return response.json().get("action_history", [])
```

hasher-matcher-actioner/scripts/script_utils.py
BarrettOlson: I am intentionally avoiding having this class import hmalib so for things like `scripts/storm_submissions_api` we only need this file atm. 

For now scripts that are deeply tied to representation classes from hmalib can import it themselves if/when that becomes impossible I will move them into this class as well. For the time being this file is structured more like `webapp/src/Api.jsx`

hasher-matcher-actioner/scripts/script_utils.py
BarrettOlson: sure! I was avoiding cookies but since we control the API and don't set any the concern is likely unfounded.

hasher-matcher-actioner/scripts/script_utils.py
```diff
@@ -62,31 +100,169 @@ def send_single_submission_url(
...
         put_response = requests.put(
             response_json["presigned_url"],
             data=file,
             headers={"content-type": "image/jpeg"},
         )
 
+    def get_content_hash_details(
+        self,
+        content_id: str,
+        api_path: str = "/content/hash/",
```
schatten, line 124: Why is the `api_path` an argument? Not seeing the utility here? Do we expect this to change?

schatten (approved):
Approving to unblock. Do not see why the URL is not internal to the method but can be overridden at callsites. More a question than an RC.

Other than that, LGTM.

hasher-matcher-actioner/scripts/script_utils.py
BarrettOlson: I found overriding helpful for debugging and testing. If I create `/content/new-hash/` for example that I want to function the same but not replace the old impl just yet.
It's roughly the same number of lines but if these methods become too much like Swiss Army Knife Anti-patterns I'll refactor.

PR #644, opened 2021-06-03
change information icon background color if privacy group is not in use

mengyangwang:
Summary
---------

In ThreatExchange tab, if a privacy group is not in use, the header background will be gray, but the information icon stays blue. 

Test Plan
---------

test changes locally
  before:
Screen Shot 2021-06-03 at 5 49 02 PM
  after:
Screen Shot 2021-06-03 at 5 49 44 PM


BarrettOlson (approved):
LGTM.

PR #659, opened 2021-06-11
[hma][scripts] add start of submit_content_test and helpers

BarrettOlson:
Summary
---------

Add start of submit_content_test and helpers building off of #641 

`python3 scripts/submit_content_test.py` will create an pg_config for sample data, an action, and an action-rule and then submit an image. 

Test Plan
---------

I confirmed that all the expected config objects are created by the script given a refresh token. Note this is test script is a WIP and the checking object e2e is yet to be implemented. (next step)


hasher-matcher-actioner/scripts/submit_content_test.py
```diff
@@ -0,0 +1,165 @@
...
+    pass
+
+
+if __name__ == "__main__":
+    # If you want hard code tests for methods you can do so here:
+
+    # Add init values for the api:
+    # `os.environ.get(...)` will enable us pass values from `terraform output` in sh
+
+    HMA_TOKEN = os.environ.get("HMA_TOKEN", "")
```
schatten, line 141: nit: Why do these look like constants?

hasher-matcher-actioner/scripts/submit_content_test.py
```diff
@@ -0,0 +1,165 @@
...
+from script_utils import HasherMatcherActionerAPI
+
+from hmalib.common.evaluator_models import Action, ActionRule
+from hmalib.common.classification_models import Label, ActionLabel, ClassificationLabel
+from hmalib.common.actioner_models import ActionPerformer, WebhookGetActionPerformer
+
+
+class DeployedInstanceTestHelper:
+    """
+    Class around testing a deployed instance of HMA from Content Submission to Hash - Match - Action
```
schatten, line 20: I'm not able to understand the purpose of this class.
* The public methods are merely proxying their arguments to `self.api`. 
* What's the value that this class is adding? The constant for SAMPLE_PG_ID? Even that is used in a separate public function, not a method of this class.
* Given that `_refresh_token()` is also a private method of the `HasherMatcherActionerAPI`, it seems the staleness of the token is an internal matter of that class.

hasher-matcher-actioner/scripts/submit_content_test.py
```diff
@@ -0,0 +1,165 @@
...
+        api_url: str,
+        api_token: str = "",
+        client_id: str = None,
+        refresh_token: str = None,
+    ) -> None:
+        if not api_token and (not client_id or not refresh_token):
+            print(
+                "Test requires an api_token OR a client_id + refresh_token to function"
+            )
+            exit(1)
```
schatten, line 37: nit: Calling `exit()` from a class method can lead to hard-to-debug issues. Instead raise an exception.

I understand that this is not production code, so am fine without this being addressed.

hasher-matcher-actioner/scripts/submit_content_test.py
```diff
@@ -0,0 +1,165 @@
...
+                if key not in {"name", "config_subtype"}
+            },
+        )
+
+    def create_action_rule(
+        self,
+        action_rule: ActionRule,
+    ):
+        # Need to give the api a json like dict object (just like is used in aws)
+        self.api.create_action_rule(action_rule.to_aws())
```
schatten, line 83: nit: Shouldn't serialization be a functionality of the `HasherMatcherActionerAPI` class?

The API class should deal in domain object types like 'ActionRule'. The serialization to_aws() is incidental.

hasher-matcher-actioner/scripts/submit_content_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def set_up(helper: DeployedInstanceTestHelper):
```
schatten, line 86: Is there idempotency in this method? What if I call set_up twice?

schatten (changes requested):
I'm happy with the functionality that the submit_content_test script exposes, but am conflicted about the class design and responsibilities. Have a look at inline comments.

Given that this is not production code, my review is too strict. But, if you see this class being re-used in other tests, some more thought is merited.

hasher-matcher-actioner/scripts/submit_content_test.py
BarrettOlson: because they were, this is matching the style in our lambdas. Happy to change it if you'd prefer. 

https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py#L32-L35

hasher-matcher-actioner/scripts/submit_content_test.py
BarrettOlson: This class is allowing self.api to stay at a low enough level it can be used independently of `hmalib`
This class is handles the specifics of `hmalib`

hasher-matcher-actioner/scripts/submit_content_test.py
BarrettOlson: See above, API is independent of `hmalib` at the moment otherwise I'd agree.

hasher-matcher-actioner/scripts/submit_content_test.py
BarrettOlson: yes because the request fails if it already exists. I'll add a doc string

hasher-matcher-actioner/scripts/submit_content_test.py
BarrettOlson: changed them, thanks for flagging.

hasher-matcher-actioner/scripts/submit_content_test.py
BarrettOlson: However, for the 2nd two points I agree:
- removed the constant from the class 
- added docstring todo for eventual adjustment of refresh responsibility.

schatten: > This class is allowing self.api to stay at a low enough level it can be used independently of `hmalib`
> This class is handles the specifics of `hmalib`

This module (submit_content_test) **imports** hmalib. So you can't use it without having hmalib on `PYTHONPATH`. If you want to avoid imports at the callsite, I'm not seeing what problem that's trying to solve.

schatten (approved):
Approving to unblock. 

I do not see the value in staying independent of hmalib at this point. Setting up your environment for an integration test is totally acceptable in my book.

Do you foresee partners using this script? Or is it just going to be for regression testing?

BarrettOlson:
> Approving to unblock.
Thanks!

> I do not see the value in staying independent of hmalib at this point. Setting up your environment for an integration test is totally acceptable in my book.
> Do you foresee partners using this script? Or is it just going to be for regression testing?

(Talked offline) Summary response to both points: 
I have a gut feeling that keeping a utils lib independent of hmalib will be helpful for extending both testing and something partners could use like a CLI. 
For now the benefit is keeping the total number of files needed to run  `storm_submission_api` lower.

PR #665, opened 2021-06-14
fix sync undefined issue

mengyangwang:
Summary
---------

this PR is to fix the issue that ThreatExchange tab is broken when click sync button. The issue is that we have changed one param name in datasets api, but didn't make all corresponding changes in UI

Test Plan
---------

before fix:
199494540_602475757383245_8708732885548711817_n

after fix:
Screen Shot 2021-06-14 at 12 29 26 PM


hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -70,16 +70,22 @@ export default function ThreatExchangeSettingsTab() {
         setShowToast(true);
       });
   };
+
+  const refreshDatasets = () => {
+    fetchAllDatasets(setLoading(false)).then(response => {
```
BarrettOlson, line 75: I'm confused by this line, `setLoading(false)` as a parameter for `fetchAllDatasets.` 
When is it called?

mengyangwang: yeah I think that's wrong, it should not be a parameter, I should call ```setLoading(false)``` before calling ```fetchAllDatasets```, I made the change

hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -70,16 +70,31 @@ export default function ThreatExchangeSettingsTab() {
         setShowToast(true);
       });
   };
+
+  const refreshDatasets = () => {
+    setLoading(false);
```
BarrettOlson, line 75: hmm, don't we want the opposite? We are starting a request therefore loading is `true` or am I missing something?

mengyangwang: the name might be confusing, because it's equals to hidden, so when starting a request I set hidden as false

BarrettOlson: then can we call it hidden or switch the conditional? Otherwise the PR looks good to me :)

BarrettOlson (approved):
Stamping to unblock. Please swap `Loading`'s values or rename it before merge. Thanks!

hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
mengyangwang: Thanks for pointing that out! I am still using the name loading and I've switched the conditional to ```hidden=!loading```, so it looks more logical

hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -141,7 +149,7 @@ export default function ThreatExchangeSettingsTab() {
       </Card.Header>
       <Card.Body>
         <Row className="mt-3">
-          <Spinner hidden={loading} animation="border" role="status">
+          <Spinner hidden={!loading} animation="border" role="status">
```
eriksendc, line 152: I took the new version of this file and put it in my local environment and I'm still getting an error because there is a time when datasets is undefined, and so there won't be a .length property. It's the same error as this morning. I think you need `!datasets || datasets.length === 0` as the condition at line 155. But that then leads to another error at line 177, because again, there is a time when datasets is undefined, so `some` cannot be referenced.

I think you need to do a terraform desctroy then terraform apply to recreate the scenario I'm in.

BarrettOlson: `datasets` is initialized to `[]` so I didn't think that would be an issue. Does a bad API request make it undefined? It feels like this check might just hide an API level issues.

PR #666, opened 2021-06-14
fix undefined datasets issue

mengyangwang:
Summary
---------

this PR is to fix the issue that ThreatExchange tab is broken when click sync button. The issue is that we have changed one param name in datasets api, but didn't make all corresponding changes in UI

Test Plan
---------

before fix:

199494540_602475757383245_8708732885548711817_n

after fix:

Screen Shot 2021-06-14 at 12 29 26 PM

### DIFF
--- a/hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
+++ b/hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
@@ -77,7 +77,6 @@ export default function ThreatExchangeSettingsTab() {
       .then(response => {
         setLoading(false);
         setDatasets(response.threat_exchange_datasets);
-        setDatasets(undefined);
       })
       .catch(() => {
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.jsx
```diff
@@ -152,7 +152,7 @@ export default function ThreatExchangeSettingsTab() {
           <Spinner hidden={!loading} animation="border" role="status">
             <span className="sr-only">Loading...</span>
           </Spinner>
-          {datasets.length === 0
+          {!datasets || datasets.length === 0
```
BarrettOlson, line 155: (also commented this on #665) `datasets` is initialized to `[]` so I didn't think that would be an issue. Does a bad API request make it undefined? It feels like this check might just hide an API level issues.

eriksendc (approved):

PR #671, opened 2021-06-15
Add UI URL to terraform outputs

jeberl:
Summary
---------

To make the [Installation of HMA](https://github.com/facebook/ThreatExchange/wiki/Installation) easier, include in the terraform outputs the link to view the UI at

Test Plan
---------

```
$ terraform -chdir=terraform apply
...
Terraform will perform the following actions:

Plan: 0 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + ui_url = "jesses-org-jeberl-hma-webapp.s3-website-us-east-1.amazonaws.com"

...
```

Go to jesses-org-jeberl-hma-webapp.s3-website-us-east-1.amazonaws.com and confirm it leads to the HMA UI

### DIFF
--- a/hasher-matcher-actioner/terraform/webapp/outputs.tf
+++ b/hasher-matcher-actioner/terraform/webapp/outputs.tf
@@ -7,3 +7,7 @@ output "s3_bucket_name" {
 output "cloudfront_distribution_domain_name" {
   value = var.include_cloudfront_distribution ? aws_cloudfront_distribution.webapp[0].domain_name : "no-cloudfront-distribution"
 }
+
+output "ui_url" {
+  value = aws_s3_bucket.webapp.website_endpoint
+}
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/terraform/webapp/outputs.tf
```diff
@@ -7,3 +7,7 @@ output "s3_bucket_name" {
 output "cloudfront_distribution_domain_name" {
   value = var.include_cloudfront_distribution ? aws_cloudfront_distribution.webapp[0].domain_name : "no-cloudfront-distribution"
 }
+
+output "ui_url" {
+  value = aws_s3_bucket.webapp.website_endpoint
```
BarrettOlson, line 12: did you test this`include_cloudfront_distribution` set to false? I think this might error.

If you look at the check above you can have this return empty string if set to false.

BarrettOlson (approved):
Stamped. Please check for inline concern before merge though :)

hasher-matcher-actioner/terraform/webapp/outputs.tf
jeberl: Talked with Barrett offline. Confirmed this works with `include_cloudfront_distribution` set to true/false and if a shared user pool is used

PR #673, opened 2021-06-15
[hma][hotfix] fix broken override of call in WebhookActionPerformer subclass

BarrettOlson:
Summary
---------

When writing tests I discovered that  `WebhookGetActionPerformer` and `WebhookDeleteActionPerformer` don't override`WebhookActionPerformer.call()` correctly. This results in an key arg error for all calls to `WebhookActionPerformer.perform_action()`

Meaning Get and Delete Actions won't trigger correctly. 

Test Plan
---------

two line class override fix.


jeberl (approved):
Nice find. Thanks for the fix

PR #676, opened 2021-06-15
Improve Install flow

jeberl:
Summary
---------

Some improvement to make the [install flow](https://github.com/facebook/ThreatExchange/wiki/Installation) easier/work.

1. Make `org` a default variable for terraform
2. Fix bug where user name prefix had trailing whitespace which was causing issues with Docker urls

Test Plan
---------

Followed the steps in the install guide: https://github.com/facebook/ThreatExchange/wiki/Installation

### DIFF
--- a/hasher-matcher-actioner/Makefile
+++ b/hasher-matcher-actioner/Makefile
@@ -24,7 +24,6 @@ upload_docker:
 
 terraform/terraform.tfvars:
 	echo 'hma_lambda_docker_uri = "${DOCKER_URI}:${DOCKER_TAG}"' > terraform/terraform.tfvars
-	echo 'organization = "${USER}-org"' >> terraform/terraform.tfvars
 
 terraform/backend.tf:
 	./scripts/write_backend_config.sh > terraform/backend.tf
--- a/hasher-matcher-actioner/Makefile
+++ b/hasher-matcher-actioner/Makefile
@@ -8,7 +8,7 @@ shell-or-die = $\
 
 USER ?= $(call shell-or-die,whoami)
 REPOSITORY_NAME = hma-lambda-dev
-DOCKER_TAG ?= ${USER} # todo - change to ${USER}-latest?
+DOCKER_TAG ?= ${USER}# todo - change to ${USER}-latest?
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/Makefile
```diff
@@ -24,6 +24,7 @@ upload_docker:
 
 terraform/terraform.tfvars:
 	echo 'hma_lambda_docker_uri = "${DOCKER_URI}:${DOCKER_TAG}"' > terraform/terraform.tfvars
+	echo 'organization = "${USER}-org"' >> terraform/terraform.tfvars
```
BarrettOlson, line 27: not blocking: I don't like that this puts `user` in the name of things twice. Especially because that does not guarantees uniqueness really any better then `example-org` (Consider in the case of user='john'). 

Personally, if you don't want it to be required I would just give it a default value in terraform.tfvars.example

hasher-matcher-actioner/Makefile
```diff
@@ -8,7 +8,7 @@ shell-or-die = $\
 
 USER ?= $(call shell-or-die,whoami)
 REPOSITORY_NAME = hma-lambda-dev
-DOCKER_TAG ?= ${USER} # todo - change to ${USER}-latest?
+DOCKER_TAG ?= ${USER}# todo - change to ${USER}-latest?
```
BarrettOlson, line 11: the fact that could matter is a "someone new to Makefile"'s worst nightmare haha

BarrettOlson (approved):
Stamping to unblock.
(I don't love the default org based on user though)

hasher-matcher-actioner/Makefile
eriksendc: The issue is that the user really needs to be aware about the need to set `organization`. As @BarrettOlson says, common names will break this. Even with `example-org` it's no different. I strongly prefer users have to set `organization` manually and that's called out in the docs so that it's clear as to why. (`organization` was introduced because some "scopes" of naming uniqueness in AWS are to the region, not to the AWS account.)

Blocking (until further discussion) don't default `organization`

eriksendc (changes requested):

hasher-matcher-actioner/Makefile
jeberl: Removed default

eriksendc (approved):

PR #680, opened 2021-06-18
[hma][scripts] basic changes to storm script

BarrettOlson:
Summary
---------
Based on a patch from @schatten. minor but worth merging before continuing work on the scripts. 

- Slightly changes the `HasherMatcherActionerAPI` submit method interface.
- Make output in `storm_submissions_api` depend on a flag.
- Catch and print request errors

Test Plan
---------
Ran the script live and in parallel.

### DIFF
--- a/hasher-matcher-actioner/scripts/storm_submissions_api
+++ b/hasher-matcher-actioner/scripts/storm_submissions_api
@@ -16,7 +16,8 @@ TODO:
 [x] Make a single successful request with a stolen JWT
 [x] Make threadpooled requests
 [x] Simple Stats for the submission API calls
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/scripts/storm_submissions_api
```diff
@@ -40,6 +43,12 @@ TOKEN = ""
 # However it can be used here for easier clean up between storms that do not need the UI.
 ID_SEPARATOR = "-"
 
+# hostnames are especially useful when storming form more than one instance
```
schatten, line 46: Thanks for these comments!

schatten (approved):
LGTM.

PR #681, opened 2021-06-21
[hma] Add basic autogen-ed dashboard to terraform

BarrettOlson:
Summary
---------

As part of perf testing @schatten built and I expanded a dashboard to monitor lambda invocations, error, latency, etc. along with estimated queue delays. This adds a slightly simplified version to terraform that we'll be created when apply is run. 

Test Plan
---------

Screen Shot 2021-06-22 at 8 31 23 PM


hasher-matcher-actioner/terraform/dashboard/main.tf
```diff
@@ -0,0 +1,162 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+# Due to limitations in terraform and the aws_cloudwatch_dashboard resource
```
schatten, line 3: There is a beautiful [templatefile](https://www.terraform.io/docs/language/functions/templatefile.html) function which allows you to store the JSON in a separate file and yet do templating with tfvars / outputs.

schatten (changes requested):
Overall, thanks for taking the lead on this. I believe this is super-useful to all customers.

I also believe the dashboard is _most_ useful when we lay out things that are related along known axes. I had gotten used to the 

| functions | hashing | matching | action eval | action perform |
|----|----|----|----|----|
|Lambda Invocations / Errors| graph | graph | graph | graph |
|Lambda latencies| graph | graph | graph | graph |
|Lambda Triggering Queue| graph | graph | graph | graph |

display and I think that is easier to understand and reason about as you get used to the dashboard.

There is also value in splitting the fetcher completely. The fetcher operates in a different plane and it not having data will immediately raise questions for anyone seeing this dashboard. 

In short: If I were doing it, YMMV. Create as many specific-use dashboards as you want, store dashboard configs in dedicated files and bring in using `templatefile`.

nits: inline

hasher-matcher-actioner/terraform/dashboard/variables.tf
```diff
@@ -0,0 +1,37 @@
...
+
+variable "datastore" {
+  description = "DynamoDB Table to store hash and match information into"
+  type = object({
+    name = string
+    arn  = string
+  })
+}
+
+variable "lambdas_to_monitor" {
```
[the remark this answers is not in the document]
BarrettOlson: you lose order with maps (using list of objects to get around this)

hasher-matcher-actioner/terraform/dashboard/main.tf
BarrettOlson: beautiful is a word, while I love templates, unfortunately it doesn't play super well with `jsonencode` because it returns a string value (double encode nightmare) but I found a way to use it.

BarrettOlson:
> I also believe the dashboard is _most_ useful when we lay out things that are related along known axes. I had gotten used to the

Agree. Done :) 
 
> There is also value in splitting the fetcher completely. The fetcher operates in a different plane and it not having data will immediately raise questions for anyone seeing this dashboard.

Agree. Removed it for now (still visible in total concurrent with other lesser used lambda) 

> In short: If I were doing it, YMMV. Create as many specific-use dashboards as you want, store dashboard configs in dedicated files and bring in using `templatefile`.

In tooling with this my mileage seemed to vary :P 
Currently we still have modules variables/inputs as a bottleneck so I think what you are describing is a tad overkill. I might be misunderstanding but having dedicated set of dashboard configs won't allow for as much reuse as would be ideal without some gross nesting (unless dashboard configs = widget configs). 

Regardless, let me know what you think about these changes. Thanks!

schatten (approved):
Love it. Especially the screenshot! Anyone who uses HMA will definitely find this valuable.

Can you figure out a way to link to this cloudwatch dashboard from the HMA home page? That would be epic!

PR #682, opened 2021-06-21
[hma] add check to handle sample ds privacy_group_id

BarrettOlson:
Summary
---------

"inria-holidays-test" is not an `int` however our fetcher expects one. If someone creates a local pg config and mistakenly enables the fetcher, the fetcher will get stuck in a retry loop. This PR makes things marginally better by at least checking if the ID is an int (so at least we won't fail on our built-in sample hash set).

Test Plan
---------

Before
`[ERROR] ValueError: invalid literal for int() with base 10: 'inria-holidays-test' Traceback (most recent call last):`
...and retry again and again

After
`Fetch skipped because privacy_group_id(inria-holidays-test) is not an int`

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
@@ -85,7 +85,7 @@ class FetcherConfig:
         )
 
 
-def is_int(int_string):
+def is_int(int_string: str):
     """
     Checks if string is convertible to int.
     """
--- a/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
@@ -85,6 +85,17 @@ class FetcherConfig:
         )
 
 
+def is_int(int_string):
+    """
+    Checks if string is convertible to int.
+    """
+    try:
+        int(int_string)
+        return True
+    except ValueError:
+        return False
+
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -152,10 +169,6 @@ def lambda_handler(event, context):
             else:
                 logging.error("Failed before fetching any records")
 
-    # TODO add TE data to indexer
-
-    return {"statusCode": 200, "body": "Sure Yeah why not"}
```
jeberl, line 157: Lol.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -85,6 +85,17 @@ def get(cls):
         )
 
 
+def is_int(int_string):
```
jeberl, line 88: nit: maybe call this `is_intish`

jeberl (approved):
Take or leave the comment!

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
BarrettOlson: We cast it to an int directly in this function and a few lines after the call. What values are you thinking of as you suggested this change?

PR #684, opened 2021-06-21
[HMA][minor] batch hash record writes and add limit check on metrics

BarrettOlson:
Summary
---------

Attempts to address some of the smaller things @schatten and I saw as part of benchmarking. 

The metric reporter will check if its report will error and will now skip is so. 
- Errors results in a retry of the full set of queue events even if the only thing to fail was metric records

The pdq_hasher will now create hashes for all s3 event notifications before attempting to write to db and add to sqs.
- The main theorized advantage here is that an s3 error will not results in rewrite and re-enqueue.

(Draft until #681 and #683 and reviewed are merged)

Test Plan
---------

Ran with stress test with #681, the `PutMetricData` error went away, I did not hit the hasher/dynamodb/s3 hard enough yet to have a statistically significant impact for the hasher change.


hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -70,18 +72,19 @@ def lambda_handler(event, context):
...
 
-            hash_record.write_to_table(records_table)
-
-            # Publish to SQS queue
-            sqs_client.send_message(
-                QueueUrl=OUTPUT_QUEUE_URL,
-                MessageBody=json.dumps(hash_record.to_sqs_message()),
-            )
+    for hash_record in hash_records:
+        hash_record.write_to_table(records_table)
```
schatten, line 80: This is probably worse for the the load numbers.

Before, for one loop, we had
```
|---hash---ddb-write---hash---ddb-write---hash---ddb-write
```

Now, we do
```
|---hash---hash---hash---ddb-write-ddb-write-ddb-write-|
```

The second one will be spikier load, right? I'm forgetting the problem we're trying to solve here? What kind of errors would reduce if we did this?

schatten (commented):
The limit on metrics is good to go. I'm concerned about the objective and outcome of the batching. When we discussed batching, I was talking about [BatchWriteItem](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html).

If our goal was to reduce the write capacity units so that we are not throttled, I don't believe this will help either. From the docs:

> Parallel processing reduces latency, but each specified put and delete request consumes the same number of write capacity units whether it is processed in parallel or not.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
BarrettOlson: "The main theorized advantage here is that an s3 error will not results in rewrite and re-enqueue."

So no help to dynamodb or sqs yet but if s3 throttles this helps. The next step would be to batch write but I didn't want to change write interface as part of the first pass.

BarrettOlson: @schatten you're right, we don't have enough information to know if this will help so I will skip this change for now.

PR #685, opened 2021-06-22
Add and Remove Classifications from ActionRules

jeberl:
Summary
---------

Add the ability to add and remove classifications to ActionRules from the UI. This PR follows from #679. See comments there for context

Also changes default values for ClassificationType to be empty per discussion on #679. Decided that a default of "=" was justifiable

Also adds some validation and shows+describes errors
- Must have at least one Classification that is an "equals" rule
- Every Classification's value must be non-empty

Also shows that all the above also works when creating a new rule

Test Plan
---------

https://user-images.githubusercontent.com/24800630/123155687-656e9300-d436-11eb-960d-e3d4e0993b93.mov


jeberl:
One issue when a new ActionRule is created. Will address tomorrow

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.jsx
```diff
@@ -33,7 +42,12 @@ export default function ActionRuleFormColumns({
             const newClassification = classification;
             newClassification.classification_type = e.target.value;
             onClassificationChange(newClassification);
-          }}>
+          }}
+          isInvalid={
+            showErrors &&
+            classification.classification_type === classificationTypeTBD
+          }>
+          <option value={classificationTypeTBD}> - </option>
```
eriksendc, line 50: Let's change `-` to `Select...`

If the pick list was wider I'd say it should be `Select Classification Type...` following the pattern used for selecting action. But since the width is limited, `Select...` will do.

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.jsx
```diff
@@ -113,13 +135,32 @@ export default function ActionRuleFormColumns({
...
+              };
               return renderClassificationRow(
                 classification,
                 onClassificationChange,
+                onClassificationDelete,
                 index,
               );
             })
           : []}
+        <Button
```
eriksendc, line 151: This button "dances around", sometimes coming at the end of the label, sometimes not. Maybe put `<br />` just before it, or surround the `<Form.Label>` above at 120 with `<Row><Col>`.

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.jsx
eriksendc: Maybe change the default option for the action select (line 177) to just be `Select...`, too, for consistency (what you're selecting is implied, at least in the action column).

eriksendc (changes requested):
Totally awesome! Very minor changes requested. Let me know what you think.

eriksendc (approved):

PR #691, opened 2021-06-25
Typescript

jeberl:
Summary
---------

Adding TypeScript to our webapp. Benefit of typescript is that it doens't have to be used universally. A Typescript file (.tsx) can interact with .jsx files and visa versa. This means that we can transfer over to typescript piecemeal

Converted one file to TypeScript as a proof of concept. ActionRuleFormColumns.tsx 

In doing so, needed to upgrade to a newer version of prettier which caused a few changes elsewhere.

Also needed to switch to the typescript parser for ESLint which required some other changes elsewhere.

Test Plan
---------

- Show that typescript will work natively in devcontainer and will show typing errors
- Show that typescript files will build
- Show that prettier will auto-format on save for .jsx files
- Show that prettier will auto-format on save for .tsx files also
- Check all pages changed in UI and ensure still work the same


hasher-matcher-actioner/webapp/.eslintrc.json
```diff
@@ -1,24 +1,35 @@
...
+      "jsx": true,
+      "tsx": true
     },
     "ecmaVersion": 12,
     "sourceType": "module"
   },
-  "plugins": ["react", "prettier"],
+  "plugins": ["react", "prettier", "@typescript-eslint"],
   "rules": {
+    "no-use-before-define": "off",
```
jeberl, line 27: The eslint error `no-use-before-define` breaks when using the `typescript-eslint` parser so we override it with `typescript-eslint`'s own version of the error. This is direct from the `typescript-eslint` documentation

hasher-matcher-actioner/webapp/package.json
```diff
@@ -57,6 +61,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "prettier": "^2.2.1"
+    "prettier": "2.3.1",
```
jeberl, line 64: typescript requires a higher version of prettier

hasher-matcher-actioner/webapp/src/components/HolidaysDatasetInformationBlock.jsx
```diff
@@ -21,7 +21,7 @@ export function HolidaysDatasetInformationBlock({samplePGExists, refresh}) {
       `This group will become disabled on "Sync" as it does not have TE
       counterpart. Simply delete and create again if you wish to keep match
       sample data after syncing`,
-    ).then(() => {});
```
jeberl, line 24: A eslint requested change. Tested this by deleting and recreating the holiday dataset in the UI

hasher-matcher-actioner/webapp/src/components/HolidaysDatasetInformationBlock.jsx
```diff
@@ -72,5 +74,5 @@ HolidaysDatasetInformationBlock.propTypes = {
 
 HolidaysDatasetInformationBlock.defaultProps = {
   samplePGExists: false,
-  refresh: () => {},
```
jeberl, line 75: eslint doenst like do-nothing functions so made nullable

hasher-matcher-actioner/webapp/src/components/OpinionTableCell.jsx
```diff
@@ -55,16 +55,14 @@ export default function OpinionTableCell({
...
-              ? () => {}
-              : () => {
-                  setNewPendingOpinionChange(
-                    PENDING_OPINION_CHANGE.MARK_FALSE_POSITIVE,
-                  );
-                  setShowModal(true);
-                }
-          }>
+          onClick={() => {
+            if (!updatePending) {
```
jeberl, line 59: Dito on do nothing functions

hasher-matcher-actioner/webapp/src/components/OpinionTableCell.jsx
```diff
@@ -80,16 +78,14 @@ export default function OpinionTableCell({
             updatePending &&
             pendingOpinionChange === PENDING_OPINION_CHANGE.MARK_TRUE_POSITIVE
           }
-          onClick={
-            updatePending
-              ? () => {}
```
jeberl, line 85: .

hasher-matcher-actioner/webapp/src/components/settings/ActionPerformer/ActionPerformerRows.jsx
```diff
@@ -20,14 +20,10 @@ export default function ActionPerformerRows({
   canNotDeleteOrUpdateName,
 }) {
   const [editing, setEditing] = useState(edit);
-  const [
```
jeberl, line 23: formatting changes

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -3,14 +3,36 @@
  */
```
jeberl, line 4: This is the one file that is now TypeScript

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -101,7 +123,7 @@ export default function ActionRuleFormColumns({
       <td>
         <Form.Label>
           Name
-          <span hidden={!showErrors || name}> (required)</span>
+          <span hidden={!showErrors || !!name}> (required)</span>
```
jeberl, line 126: `!showErrors || name` returns a string if `showErrors` is true

`!showErrors || !!name` always returns a bool

hasher-matcher-actioner/webapp/src/pages/Dashboard.jsx
```diff
@@ -69,42 +69,31 @@ function toUFlotFormat(graphData) {
   return [timestamps, values];
```
jeberl, line 69: These changes were just ordering. Dont use a function before you define it. Tested by looking at Dashboard

hasher-matcher-actioner/webapp/src/pages/Matches.jsx
```diff
@@ -29,71 +29,6 @@ import {useQuery} from '../utils/QueryParams';
 import FullWidthLocalScrollingLeftAlignedLayout from './layouts/FullWidthLocalScrollingLeftAlignedLayout';
```
jeberl, line 29: Same. Just moving functions around

hasher-matcher-actioner/webapp/src/pages/settings/SignalSettingsTab.jsx
```diff
@@ -10,11 +10,48 @@ import ListGroup from 'react-bootstrap/ListGroup';
 import Button from 'react-bootstrap/Button';
```
jeberl, line 10: More moving around

eriksendc (approved):
Looks good to me and I approve. But check with @BarrettOlson to see if he wants to review before you consider this done.

hasher-matcher-actioner/webapp/.eslintrc.json
BarrettOlson: nit: link to documentation/if comments allowed provide the context here for future readers.

hasher-matcher-actioner/webapp/package.json
BarrettOlson: nit: might was well leave the `^`

hasher-matcher-actioner/webapp/src/components/HolidaysDatasetInformationBlock.jsx
BarrettOlson: hmm, I think not waiting on the future at all is a slight change in logic but the only case I could maybe see that mattering is if they click off the page immediately. Probably fine

hasher-matcher-actioner/webapp/.eslintrc.json
```diff
@@ -1,24 +1,35 @@
...
+    "@typescript-eslint/no-use-before-define": ["error"],
     "react/jsx-filename-extension": [
       "warn",
       {
-        "extensions": [".js", ".jsx"]
+        "extensions": [".js", ".jsx", ".tsx"]
```
BarrettOlson, line 32: to keep the pattern consistent/incase we end up having .ts files can you add them as well?

hasher-matcher-actioner/webapp/.prettierrc.json
```diff
@@ -7,7 +7,7 @@
   "endOfLine": "auto",
   "overrides": [
     {
-      "files": ["*.js", "*.jsx"]
+      "files": ["*.js", "*.jsx", "*.tsx"]
```
BarrettOlson, line 10: .

hasher-matcher-actioner/webapp/README.md
```diff
@@ -57,7 +57,7 @@ Please run the following checks/formating before submitting a PR.
 
 ```
 # linter
-npx eslint src --ext .js,.jsx --fix
+npx eslint src --ext .js,.jsx,.tsx --fix
```
BarrettOlson, line 60: .

hasher-matcher-actioner/webapp/package.json
```diff
@@ -14,13 +14,16 @@
     "clipboard-copy": "^4.0.1",
     "date-fns": "^2.21.0",
     "dompurify": "^2.2.8",
+    "eslint": "^7.29.0",
     "node-sass": "^5.0.0",
+    "prettier": "^2.3.1",
```
BarrettOlson, line 19: nit: I don't know if these needed to added beyond devDependencies do you have errors without them?

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -3,14 +3,36 @@
...
-import {PropTypes} from 'prop-types';
+import PropTypes from 'prop-types';
 import Form from 'react-bootstrap/Form';
 import Row from 'react-bootstrap/Row';
 import Col from 'react-bootstrap/Col';
 import Button from 'react-bootstrap/Button';
 
 export const classificationTypeTBD = 'TBD';
 
+type Action = {
```
BarrettOlson, line 14: This is a larger concern probably beyond the scope of this PR but I would strongly prefer things like `Action` and `Classification` actually correspond to something on the backend in python. 

Right now these models makes sense for the UI but don't map back as well. The more the UI interface and the actually backing structure diverge the more difficult to develop on and imo less intuitive the system will become.

BarrettOlson: At a minimum can you add some docstrings to these types?

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -3,14 +3,36 @@
...
+  id: string;
+  name: string;
+};
+
+type Classification = {
+  classificationType: string;
+  equals: boolean;
```
BarrettOlson, line 21: is it confusing what this fields is saying, equal_to vs not_equal_to check when a match is found?

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -60,7 +82,7 @@ export default function ActionRuleFormColumns({
         <Form.Control
           as="select"
           required
-          value={classification.equals}
+          value={classification.equals ? 'true' : 'false'}
```
BarrettOlson, line 85: I think 'Equal' vs 'Not Equal' here might help

BarrettOlson (approved):
Nice! nits and such inline but nothing really about the typescript change itself. Bit of a shame this messes with the blame as much as it does but given we are adding typing needing `no-use-before-define` makes sense.

hasher-matcher-actioner/webapp/package.json
jeberl: They were automatically generated. Can try without

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
jeberl: This doesnt map 1:1 with anything in python. Its effectively a pointer an Action. The Action Settings tab would have the 1:1 mapping with `WebhookActionPerformer` in python

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
jeberl: Gonna rename the class and the field

hasher-matcher-actioner/webapp/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
BarrettOlson, line 1: what does this do?

PR #693, opened 2021-06-27
PDQ Hasher now supports directly downloading images, bypassing S3 completely.

schatten:
Summary
---------

# Diffs to review in order:

### [33b670db](https://github.com/facebook/ThreatExchange/pull/693/commits/33b670db85796419aa5ce207064e5c99945c7aec) Add message models to cover explicit URL message and automatic S3 Upload events
Introduce a new model for messaging between API submission lambda and pdq(other?) hashing lambda `URLImageSubmissionMessage`. Also formalizes the S3 event message used so far into `S3ImageSubmissionBatchMessage`.

### [d1cacafd3](https://github.com/facebook/ThreatExchange/pull/693/commits/d1cacafd38adcf4c7edaf6dd51d6b08574ee3c45) Add image_sources: download bytes for any ImageSubmissionMessage
Introduces image_sources. These are generic pluggable classes that take submission messages defined in the previous diff and get bytes from s3 or the internet. Includes test for the new URL image source.

### [371c5cb07](https://github.com/facebook/ThreatExchange/pull/693/commits/371c5cb0784293f2c27ab8bf734f7702f63415e0) Use image_sources in pdq hashing; backwards compatible with s3 upload auto-generate events
pdq_hasher now supports image_sources. Both the old style s3 uploads and the newly introduced url image sources.

### [ade507117](https://github.com/facebook/ThreatExchange/pull/693/commits/ade507117ac2218626433e271b7be528f569291b) Publish SNS message for URL image submissions rather than copying to s3
submit API now uses the `URLImageSubmissionMessage` to pass a message to the hashing lambda rather than downloading and writing to s3 on its own.

### [b841270](https://github.com/facebook/ThreatExchange/pull/693/commits/b84127057d44d8ca863478dd343bbf5eb8f7d68c) Use == for type checks rather than is
More details in [Issue #697](#697)

### [d905dba4](https://github.com/facebook/ThreatExchange/pull/693/commits/d905dba465779f181059d13455b170ff380889cc) devcontainer: use pytest test framework in vscode's python extension
You probably want to ignore this one. 

### [5631ebe44c](https://github.com/facebook/ThreatExchange/pull/693/commits/5631ebe44c62107fd39adceedd4b4b473cb7f430)  Add enums for ContentType and ContentRefType 
Leaving these to be open-ended strings was making it confusing for me. Hopefully others too.

### [c06fa75](https://github.com/facebook/ThreatExchange/pull/693/commits/c06fa7598f07ab85cbb5ac59bb9b8bbd78d4a24e) Formalize convention for naming s3 keys in the default image bucket 
Prior to this, the convention on how to create the s3 keys was split across various lambdas. This brings everything home. If you are 'using' image_prefix, somewhere, that's wrong, it should be moved to image_sources.py

### [27cf88ca](https://github.com/facebook/ThreatExchange/pull/693/commits/27cf88ca79ee9965f98b82dd703f3fefd12c2dad) Add image prefix envvar to pdq-hasher
This allows it to use S3BucketImageSource.

# Future work
* Verify that matched previews can be rendered for non-expiring URLs.
* Interesting images (matches) will be saved to S3 separately.

Test Plan
---------
```
$ make upload_docker
$ <swap out docker image hash>
$ terraform -chdir terraform apply
```

Upload image from s3 and upload image via the UI, using a URL.

Screen Shot 2021-06-27 at 17 19 38

**Logs from a URL submission**

Screen Shot 2021-06-27 at 17 22 03

**Logs from an S3 upload**

I have repeated the test plan after the fixes. Uploaded one image each of DIRECT_UPLOAD, UPLOAD and URL. Verified that all caused hashes to be published and their content details pages loaded with the hashes.

Note: The image on the content details does not load when we submit via URL. Have verified for UPLOAD and DIRECT_UPLOAD.

This is because of sort of a CORS issue. I plan to fix this in another pull request.


hasher-matcher-actioner/hmalib/common/image_sources.py
```diff
@@ -0,0 +1,76 @@
...
+      defaults)
+    """
+
+    def get_image_bytes(self, identifier: str) -> bytes:
+        r = requests.get(identifier)
+        r.raise_for_status()
+        return r.content
+
+
+def get_image_bytes(
```
BarrettOlson, line 68: I don't see this method being used anywhere, was it for debug/future utility? 
can you delete or add a docstring? Since it shares the name of the class method a rename or suffix might be helpful.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -29,49 +33,57 @@ def lambda_handler(event, context):
...
+        if URLImageSubmissionMessage.could_be(message):
+            images_to_process.append(
+                URLImageSubmissionMessage.from_sqs_message(message)
+            )
+        elif S3ImageSubmissionBatchMessage.could_be(message):
+            images_to_process.extend(
+                S3ImageSubmissionBatchMessage.from_sqs_message(
+                    message
+                ).image_submissions
+            )
```
BarrettOlson, line 73: nit: log warning for message that isn't a test event and doesn't fit the 2 'could_be'

hasher-matcher-actioner/hmalib/common/tests/test_image_sources.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import unittest
+import io
+
+from PIL import Image
```
BarrettOlson, line 6: do we need to add to `requirements-dev.txt` ?

hasher-matcher-actioner/hmalib/common/message_models.py
```diff
@@ -167,3 +172,85 @@ def send_to_queue(self) -> None:
             QueueUrl=config.writebacks_queue_url,
             MessageBody=self.to_aws_json(),
         )
+
+
+@dataclass
+class URLImageSubmissionMessage:
```
BarrettOlson, line 178: nit: the fact that the base version of one submission is `...Message` and the other isn't is confusing to me.

Having a `URLImageSubmissionMessage` contain a DTO like  `URLImageSubmission` class
OR
having `S3ImageSubmission`-> `S3ImageSubmissionMessage`
+  `S3ImageSubmissionBatchMessage` -> `S3ImageSubmissionBatchMessages` 
would help imo.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -29,49 +33,57 @@ def lambda_handler(event, context):
...
-        for s3_record in message["Records"]:
-            bucket_name = s3_record["s3"]["bucket"]["name"]
-            key = unquote_plus(s3_record["s3"]["object"]["key"])
-
-            # Ignore Folders and Empty Files
-            if s3_record["s3"]["object"]["size"] == 0:
-                logger.info("Disregarding empty file or directory: %s", key)
-                continue
+        images_to_process: t.List[
+            t.Union[URLImageSubmissionMessage, S3ImageSubmission]
```
BarrettOlson, line 61: see note on `hasher-matcher-actioner/hmalib/common/message_models.py` changes around class naming.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -206,35 +216,22 @@ def from_url(
...
-                content_id=request.content_id, submit_successful=True
-            )
-        else:
-            bottle.response.status = 400
-            return SubmitContentError(
-                content_id=request.content_id,
-                message="url submitted could not be read from",
-            )
+        return SubmitContentResponse(
+            content_id=request.content_id, submit_successful=True
```
BarrettOlson, line 233: The hasher doesn't know about `IMAGE_FOLDER_KEY` therefore the way that class handles writing `PipelinePDQHashRecord` for s3 vs URL submissions ~breaks. 

Where this was 'fixed' (or goes wrong depending on perspectives) can be seen here: https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/hmalib/lambdas/api/submit.py#L124

Right now the system (and the UI) assumes the content_id stored in the dyanmodb is prefixed with `IMAGE_FOLDER_KEY`

I think we do want to change this but the move fast for now change is to add the prefix here.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -29,49 +33,57 @@ def lambda_handler(event, context):
...
+                bytes_: bytes = image_sources.get_image_bytes(image)
+
+            logger.info("Generating PDQ hash for submission: %s", repr(image))
 
             with metrics.timer(metrics.names.pdq_hasher_lambda.hash):
                 pdq_hash, quality = pdq_hasher.pdq_from_bytes(bytes_)
 
             hash_record = PipelinePDQHashRecord(
-                key, pdq_hash, datetime.datetime.now(), quality
+                image.content_id, pdq_hash, datetime.datetime.now(), quality
```
BarrettOlson, line 86: This is where the missing `IMAGE_FOLDER_KEY` prefix goes wrong for URLImageSubmissionMessage

BarrettOlson (changes requested):
Looks good to me! 

Only blocking change has to do with the hacky content_id stuff 
(makes me think we might want to create a submssion_id for s3 keys instead)

hasher-matcher-actioner/hmalib/common/tests/test_failing_dataclass_serialization.py
```diff
@@ -0,0 +1,29 @@
...
+@dataclass
+class NotSoSimple(aws_dataclass.HasAWSSerialization):
+    e: t.Set[float] = field(default_factory=lambda: {1.3, 4.2})
+
+
+"""
+I have no idea what's going on. If you comment out the class defined above and
+run tests, they fail with a deserialization error.
+
+My numerous attempts at figuring out exactly what is wrong have failed. The
```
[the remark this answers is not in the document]
schatten: I think the fault is in python typing. I dived into the HasAWSSerialization, and there does not seem to be any reflection that would cause this.

We'll fight this another day.

hasher-matcher-actioner/hmalib/common/tests/test_image_sources.py
schatten: It is already there as part of threatexchange. No code here that is not already in pdq_hasher.py

BarrettOlson: To check my understanding, if we remove it from python-threatexchange e.g. change the image processing lib this would then break here? (If so my gut is telling me it is worth having as a requirement then but I'll leave it up to you)

hasher-matcher-actioner/hmalib/common/image_sources.py
schatten: It is actually what the pdq_hasher lambda uses to obtain the image. Introduced in a later commit. I see your confusion. I'll add a docstring though.

https://github.com/facebook/ThreatExchange/pull/693/files#diff-dca029e7f27b7a31a5d85e9275abf7d49bc1535026d25e6173e4d4dcbe76cca3R78

hasher-matcher-actioner/hmalib/common/tests/test_image_sources.py
schatten: Well, your gut makes sense to me. :P I'll add it.

hasher-matcher-actioner/hmalib/common/image_sources.py
BarrettOlson: I missed this because I mistook `image_sources` of `image_sources.get_image_bytes(image)` as an object not a module. Given that we called the same methods on objects in the test. Can we please change the name of one or both?

hasher-matcher-actioner/hmalib/common/message_models.py
schatten: I see the confusion. I'll clarify in docstrings.

The issue is that S3ImageSubmission is not a message. It is part of the S3ImageSubmissionBatchMessage. IMO, a message is not a collection of other messages.

BarrettOlson: (again nit:) Right, I don't think you should call `S3ImageSubmission` a message. At the same time URLImageSubmissionMessage is a submission but S3ImageSubmissionBatchMessage is not. What I dislike is half way making a distinction between a message and submission depending on the source with the resulting access pattern:
https://github.com/facebook/ThreatExchange/pull/693/files/76e74290de70346957561f6a210a44c7d75f9c77#diff-dca029e7f27b7a31a5d85e9275abf7d49bc1535026d25e6173e4d4dcbe76cca3R64-R72

hasher-matcher-actioner/hmalib/common/tests/test_failing_dataclass_serialization.py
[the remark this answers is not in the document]
Dcallies: The best part of python framework hacking is trying to debug it when it fails! This is generally why I suggest avoiding python magic if possible.

hasher-matcher-actioner/hmalib/common/tests/test_failing_dataclass_serialization.py
```diff
@@ -0,0 +1,29 @@
...
+message_models.py. If you swap out the type of image_submissions:
+t.List[S3ImageSubmission]
+
+to any a primitive list like `image_submissions: t.List[int]`, there are no
+failures. Spooky!
+
+The failure exception looks like:
+
+E           hmalib.common.aws_dataclass.AWSSerializationFailure: Deserialization
+error: Expected typing.Set[float] got <class 'list'> ([Decimal('1.3'),
```
Dcallies, line 27: I have an inkling of what happened - decimal must be used instead of float for some amazon services, so it attempts to serialize as that. There's likely a missing special case in the expectations.

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -64,27 +64,27 @@ def py_to_aws(py_field: t.Any, in_type: t.Optional[t.Type[T]] = None) -> T:
...
+    if in_type == bool:  # BOOL
         return py_field  # type: ignore # mypy/issues/10003
-    if in_type is t.Set[str]:  # SS
+    if in_type == t.Set[str]:  # SS
         return py_field  # type: ignore # mypy/issues/10003
-    if in_type is t.Set[int]:  # SN
+    if in_type == t.Set[int]:  # SN
         return {i for i in py_field}  # type: ignore # mypy/issues/10003
-    if in_type is t.Set[float]:  # SN
+    if in_type == t.Set[float]:  # SN
```
eriksendc, line 87: @schatten  Just wondering about this change and learning about Python. Why was this necessary? What's the difference here?

eriksendc: @schatten thanks for explaining this in the issue you filed! :)

Dcallies (approved):
Nice work!

EDIT: Whoops, this is in the middle of your diff stack - can you please pull out these changes to their own PR and submit them?

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -64,27 +64,27 @@ def py_to_aws(py_field: t.Any, in_type: t.Optional[t.Type[T]] = None) -> T:
             f"Expected {check_type} got {type(py_field)} ({py_field!r})"
         )
 
-    if in_type is int:  # N
+    if in_type == int:  # N
```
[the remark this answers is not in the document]
schatten: Because of how github's comments are tied to the SHA, doing this would change the history for this PR. 

If I am not able to merge this in by Wednesday, I'll move to a dedicated PR and rebase.

hasher-matcher-actioner/.devcontainer/zshrc
```diff
@@ -12,4 +12,8 @@ setopt INC_APPEND_HISTORY
 # Make path respect local binaries from pip
 export PATH=$PATH:~/.local/bin
 
+# Add directory containing hmalib to PYTHONPATH. This allows vscode
+# testdiscovery to work
+export PYTHONPATH=$PYTHONPATH:/workspaces/ThreatExchange/hasher-matcher-actioner
```
BarrettOlson, line 17: Ignorable (for now) could you either add this to the Dockerfile or create a `.profile` for this. I don't use the default zshrc.
Ref: https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/.devcontainer/Dockerfile#L130-L132

BarrettOlson (changes requested):
Really love this! I think it is a great way to handle the `IMAGE_FOLDER_KEY` issue.

However requesting changes because this breaks `matches.py` and therefore the matches and details page. Should be as simple as making the same changes you made to `content.py` to `matches.py`

After that, happy to stamp as the rest looks great!

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -35,7 +35,10 @@
 HMA_CONFIG_TABLE = os.environ["HMA_CONFIG_TABLE"]
 DYNAMODB_TABLE = os.environ["DYNAMODB_TABLE"]
 IMAGE_BUCKET_NAME = os.environ["IMAGE_BUCKET_NAME"]
-IMAGE_FOLDER_KEY = os.environ["IMAGE_FOLDER_KEY"]
+IMAGE_FOLDER_KEY = os.environ[
+    "IMAGE_FOLDER_KEY"
+]  # Misnamed, this is a prefix, not a key, if renaming, use IMAGE_PREFIX
```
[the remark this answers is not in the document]
schatten: This is a TODO for when we rename. Haven't touched it yet.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -2,76 +2,114 @@
...
     records_table = dynamodb.Table(DYNAMODB_TABLE)
 
     for sqs_record in event["Records"]:
-        sns_notification = json.loads(sqs_record["body"])
-        message = json.loads(sns_notification["Message"])
+        message_body = json.loads(sqs_record["body"])
+        message = json.loads(message_body["Message"])
 
-        if message.get("Event") == "s3:TestEvent":
+        if message.get("Event", None) == "s3:TestEvent":
```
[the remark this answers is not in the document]
schatten: Did not know that it never raises KeyError. Fixing!

schatten:
@BarrettOlson Have addressed both your latest review comments. Thanks for the matches flag. I had not tested for it.

Here's a screenshot for a recent upload. It works!

Screen Shot 2021-06-29 at 22 56 51

BarrettOlson (approved):
LGreatTM

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -94,7 +97,6 @@ class SignalSourceSummary(t.TypedDict):
     "/matches/",
     get_matches_api(
         dynamodb_table=dynamodb.Table(DYNAMODB_TABLE),
-        image_folder_key=IMAGE_FOLDER_KEY,
```
BarrettOlson, line 97: itsdone

jeberl (commented):
Couple questions. Not blocking

1. When someone submits content via a URL, how do they determine whether the image is saved in s3 or not? Is it never saved?
2. For connecting to partners own s3 banks, having a function that can be directly called to upload content without saving to s3 would be useful. Would that function currently be `hmalib.lambdas.api.submit.get_submit_api().post_url_upload` ?

schatten:
>     1. When someone submits content via a URL, how do they determine whether the image is saved in s3 or not? Is it never saved?

At present, no, images submitted via URL are **never** uploaded to s3. In an [internal wp post](https://fb.workplace.com/groups/1936087780027428/posts/2595869387382594/?comment_id=2595870550715811) @BarrettOlson mentioned and @Dcallies concurred that we could be saving matched images to s3. That will happen in a second PR.

schatten:
>  2. For connecting to ~internal banks~ existing s3 buckets, having a function that can be directly called to upload content without saving to s3 would be useful. Would that function currently be `hmalib.lambdas.api.submit.get_submit_api().post_url_upload` ?

Even for existing buckets, I would recommend **not** making a copy of the image. Submit the URL, because it is signed, the hasher will download and process it.

If there is a match, it will be copied over. If not, the content API will still return 'no matches found'. Is there a benefit to having the image copied over to HMA's private bucket?

jeberl:
@schatten 
Yep. Don't intend to copy the image. When I have a url to the image and i'd like to send it to the hasher, the function I should call is `hmalib.lambdas.api.submit.get_submit_api().post_url_upload` correct?

schatten:
> @schatten
> Yep. Don't intend to copy the image. When I have a url to the image and i'd like to send it to the hasher, the function I should call is `hmalib.lambdas.api.submit.get_submit_api().post_url_upload` correct?

I would prefer you call the API via the API Gateway. It is not just the code that is run, there are permissions that the submission lambda has, those might evolve over time. Keeping the integrations lambda's permissions up-to-date with the submission lambdas will be a pain. Call [the API as shown here](https://github.com/facebook/ThreatExchange/blob/master/hasher-matcher-actioner/scripts/script_utils.py#L130), less maintenance going forward.

PR #694, opened 2021-06-28
git ignore react file

jeberl:
Summary
---------

This file seems to be necessary for Typescript compilation but not needed. Standard practice is to add to .gitignore.

More context: https://github.com/facebook/create-react-app/issues/6560

Test Plan
---------

npm still compiles


BarrettOlson (approved):

BarrettOlson (changes requested):
extra q?

BarrettOlson (approved):

PR #695, opened 2021-06-28
[hma][tf] Move sample data upload to after indexer creation

BarrettOlson:
Summary
---------

`resource "null_resource" "provide_sample_pdq_data_holidays"` is created in terraform/hashing-data.
However this means the file with the sample signals is added to s3 before the indexer is created and listening for `aws_sns_topic.threat_exchange_data` notifs.

This means that until more datasets are created and files added to this s3 prefix (likely by syncing the fetcher) OR the indexer is triggered manually no index exists and the matcher fails.

This moves the upload of the sample set to s3 to be after the indexer creation.

Test Plan
---------

`terraform -chdir=terraform apply` and before making any changes to PGs etc make sure an index is actually created.

### DIFF
--- a/hasher-matcher-actioner/terraform/hashing-data/main.tf
+++ b/hasher-matcher-actioner/terraform/hashing-data/main.tf
@@ -77,26 +77,6 @@ resource "aws_s3_bucket_object" "threat_exchange_data" {
     }
   )
 }
-resource "null_resource" "provide_sample_pdq_data_holidays" {
-  # To force-update on existing deployment, taint and apply terraform again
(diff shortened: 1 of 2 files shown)


jeberl (approved):
This looks good to me. This was likely going to be necessary for local data sets also right?

BarrettOlson:
> This looks good to me. This was likely going to be necessary for local data sets also right?

What do you mean by local datasets? I think in most cases extra data sets are separate from deployment a part from this built in one) but depending on how local banks are implemented we do want to make sure the indexer is notified to the changes.

PR #699, opened 2021-06-29
[hma] Update package.json dep versions

BarrettOlson:
Summary
---------

npm has advised us on package vulnerabilities. Most of them are server side and therefore not actually relevant to us. Regardless some can be addressed by updating the desired versions in package.json. 

However, due to dependency issues not all can be resolved but can likely wait for the associated package to update. This change fixes 3 moderate and 6 high vulnerabilities as well as added the recommended packages for typescript based on https://create-react-app.dev/docs/adding-typescript 

Test Plan
---------
```
$ npm install 

$ ESLINT_NO_DEV_ERRORS=true npm start

# check that the site still functioned as expected. 
```


jeberl (approved):
Yeah Sure. Why not.

PR #703, opened 2021-06-30
[hma][tf] fix sqs queue_batch_size inconsistency (null wasn't good enough)

BarrettOlson:
Summary
---------

the behavior of `null` in a part of a conditional expression is recommended here:  
https://www.terraform.io/docs/language/expressions/types.html 

However it was not behaving consistently so just added the defaults that should have been used automatically instead of `null`

Test Plan
---------

tf apply and checked that the queue values actually changed to the correct values based on `set_sqs_windows_to_min`


jeberl (approved):
!! Exciting

PR #704, opened 2021-06-30
Submit s3 bucket script

schatten:
Summary
---------
List objects in an s3 bucket [with prefix] and submit all as signed urls to an HMA instance

Test Plan
---------
Screen Shot 2021-06-30 at 18 00 42

Verified that we got these images were processed on the cloudwatch dashboard.


hasher-matcher-actioner/scripts/script_utils.py
```diff
@@ -127,6 +127,30 @@ def send_single_submission_url(
         )
         put_response.raise_for_status()
 
+    def submit_single_url(
```
BarrettOlson, line 130: `submit_from_url` instead?

until we batch the `single` is redundant anyway imo

hasher-matcher-actioner/scripts/submit_s3_bucket
```diff
@@ -0,0 +1,142 @@
...
+        "--prefix",
+        help="S3 prefix to narrow down which objects you want to submit.",
+        required=False,
+    )
+    parser.add_argument(
+        "--api_url",
+        help="The Base URL of API gateway that you want to submit to. eg. https://mc620fy2hf.execute-api.us-east-1.amazonaws.com/",
+        required=True,
+    )
+    parser.add_argument("--loop", action="store_true")
```
BarrettOlson, line 129: nit: help text

hasher-matcher-actioner/scripts/submit_s3_bucket
```diff
@@ -0,0 +1,142 @@
...
+                    Params={
+                        "Bucket": self.bucket,
+                        "Key": object["Key"],
+                    },
+                    ExpiresIn=3 * 60 * 60,  # 3 hours from now.
+                )
+                callable(presigned_url)
+
+    def run(self):
+        if self.loop:
```
BarrettOlson, line 81: Had to read through multiple times can you add a comment?

if loop=true does this just mean use multiple workers? Do they somehow share a  `paginator` how are repeats avoided?

BarrettOlson (approved):
I do think the script itself is a little hard to understand. Not sure I fully get 'loop.' but our scripts need consolidation and refactoring anyway so happy to stamp :)

hasher-matcher-actioner/scripts/submit_s3_bucket
schatten: Added a comment. Repeats are not avoided. Loop merely causes the process to be repeated until Ctrl-Cd. We can't do 4000 QPS with 5000 images, so we loop and submit the same images again and again.

BarrettOlson: got it, I'd probably recommends renaming loop to storm then but fine either way.

PR #707, opened 2021-07-01
Stop counting totals in DDB transactions

schatten:
Summary
---------
DDB does not block writes when a transaction is in conflict. It errors out. So, we can't do transactions for every write. Will figure something out later, but until them, remove all vestiges of using transactions to count things.

Removed from models and from dashboard too.

Test Plan
---------
```shell
$ python
>>> from hmalib.models import PipelinePDQHashRecord
>>> import boto3
>>> table = boto3.resource('dynamodb').Table('dipanjanm-HMADataStore')
>>> x = PipelinePDQHashRecord.get_from_time_range(table, "2020-02-02T00:00:00", "2022-02-02T00:00:00")[0]
>>> x.write_to_table_if_not_found(table)
False
```

Screen Shot 2021-07-01 at 15 11 53

```
$ python -m mypy hmalib
$ python -m black hmalib
$ python -m py.test
```

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/stats.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/stats.py
@@ -8,7 +8,7 @@ import typing as t
 from mypy_boto3_dynamodb.service_resource import Table
 
 from hmalib.common.logging import get_logger
-from hmalib import metrics, models
+from hmalib import metrics
 from hmalib.metrics import query as metrics_query
 from hmalib.metrics.query import is_publishing_metrics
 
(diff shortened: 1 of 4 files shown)


BarrettOlson (approved):
I to, do some of my best programing with the delete key

BarrettOlson:
nit: changes like this a test plan with before and after can be helpful

BarrettOlson:
This also made me realize we don't have a `Dashboard` wiki page.

schatten:
> nit: changes like this a test plan with before and after can be helpful

You're right. In this case, before and after were the same result. No regressions, despite the change.

BarrettOlson:
> > nit: changes like this a test plan with before and after can be helpful
> 
> You're right. In this case, before and after were the same result. No regressions, despite the change.

I meant a before picture to see that:
LARGE NUMBER
(since HMA is online.)
is now gone from the dashboard widget

The before and after I think you're suggesting (after minor stress test runs are closer to turn key) would be great too!

PR #711, opened 2021-07-07
Convert threatexchange settings typescript

mengyangwang:
Summary
---------

This PR will do two things:
1. convert ThreatExchangeSettingsTab, HolidaysDatasetInformationBlock, Settings file to tsx file and fix type errors
2. fix the issue that UI will display errors when updating a dataset which has string privacry group id

Future Action
---------
Need to also convert ThreatExchangePrivacyGroupCard to tsx file, but there is a known issue with ion-icon package and typescript, will spend little more time to look into this issue

Test Plan
---------
1. black hmalib
2. mypy hmalib
4. make docker
5. make upload_docker 
6. terraform apply
7.  test localhost for all test cases on ThreatExchange setting tab

https://user-images.githubusercontent.com/81996660/124833552-5c101b00-df4c-11eb-99f6-2189cab854a4.mov

### DIFF
--- a/hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.tsx
+++ b/hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.tsx
@@ -25,13 +25,30 @@ import {
   deleteDataset,
 } from '../../Api';
 
+type privacyGroupType = {
+  privacyGroupId: string;
+  localFetcherActive: boolean;
+  localWriteBack: boolean;
+  localMatcherActive: boolean;
+};
+type datasetType = {
+  privacy_group_id: string;
+  privacy_group_name: string;
+  description: string;
+  fetcher_active: boolean;
+  matcher_active: boolean;
+  in_use: boolean;
+  write_back: boolean;
+  hash_count: number;
+  match_count: number;
+};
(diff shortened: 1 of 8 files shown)


hasher-matcher-actioner/webapp/src/pages/settings/ThreatExchangeSettingsTab.tsx
```diff
@@ -25,13 +25,30 @@ import {
   deleteDataset,
 } from '../../Api';
 
+type privacyGroupType = {
```
schatten, line 28: Dunno whether we are following a specific convention or not, but Types should be UpperCamelCase, IMO.

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html seems to follow the same.

schatten: nit: And the 'Type' suffix is unnecessary. 

Because this is distinct from the actual PrivacyGroup domain objects, I understand you'd want to clearly make this separate. In previous projects, I have used ThreatExchangeSettingsPrivacyGroupState style type/class names. 

It is verbose, but explicit. YMMV.

schatten (approved):
small nits aside, looks good to me.

Type should be UpperCamelCase, and can use more verbose explicit type / class names.

PR #712, opened 2021-07-08
Various Changes to Cloudwatch Dashboard

schatten:
## Summary

Various changes to the dashboard have been done as part of the PR. If you keep this description open while you review, you'll find that the changes here are laid out in the same order as the diff.

**Changes across the dashboard**
- Add Titles to the dashboard, so the graph labels are not so long that they get ellipsized

**API Section Updates:**
- Switch from the regular lambda widget to a custom one for API. I want to show the response times in a separate graph.
    - Also because the concurrent executions were super-low in scale compared to the number of invocations once we started doing URL only submissions where the response times are fairly low.
- Add an API Response times widget
- In status codes graph, display %age 5xx and 4xx instead of absolute. I expect in production the number of non 2xx to be low enough that one graph can't show both without left-and-right axes. At which point, for me personally, it gets confusing. So, left is absolute, and right is %age based on that absolute.

**DynamoDB Section Updates:**
- Update Dynamodb capacity units. Use a shorter title and shorter legend labels so that it shows up in the graph even when not full screened.
    - Bugfix: Use "Sum" instead of average.
- For Dynamodb Errors widget, show Updates, puts and batch writes. We did not have updates earlier and they were not showing up.

## Test Plan

```
$ terraform apply
```

Saw the dashboard, took a screenshot and annotated.

07:08-HMA Dashboard Refresh PR

### DIFF
--- a/hasher-matcher-actioner/terraform/dashboard/main.tf
+++ b/hasher-matcher-actioner/terraform/dashboard/main.tf
@@ -1,6 +1,41 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 locals {
+  title_glance = jsonencode({
+    height = 1,
+    width  = 24,
+    type   = "text",
+    properties = {
+      markdown = "## Hashing, Matching &amp; Actioning at a Glance"
+    }
+  })
+
+  title_dynamodb = jsonencode({
+    height = 1,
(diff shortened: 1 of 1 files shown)


BarrettOlson (approved):
Nice Improvements! LGTM!

PR #713, opened 2021-07-08
Devcontainer for python-threatexchange

schatten:
Summary
---------
In some distant future, we might be able to do a single devcontainer for both python-threatexchange and hma, but for now, have a separate one.

Test Plan
---------
Booted the devcontainer using "Cmd+P" -> "Remote Container: Open Folder.."

The necessary extensions were installed. Once I selected the python interpreter, I could cmd-click into library functions.

### DIFF
--- a/python-threatexchange/.devcontainer/bashrc
+++ b/python-threatexchange/.devcontainer/bashrc
@@ -1,0 +1,10 @@
+source /etc/bash.bashrc
+
+# Make path respect local binaries from pip
+export PATH=$PATH:~/.local/bin
+
+# Add directory containing hmalib to PYTHONPATH. This allows vscode
+# testdiscovery to work
+export PYTHONPATH=$PYTHONPATH:/workspaces/ThreatExchange/hasher-matcher-actioner
+
+# Last line must have a newline char!
--- a/python-threatexchange/.devcontainer/bashrc
+++ b/python-threatexchange/.devcontainer/bashrc
@@ -5,6 +5,6 @@ export PATH=$PATH:~/.local/bin
 
 # Add directory containing hmalib to PYTHONPATH. This allows vscode
 # testdiscovery to work
-export PYTHONPATH=$PYTHONPATH:/workspaces/ThreatExchange/hasher-matcher-actioner
+export PYTHONPATH=$PYTHONPATH:/workspaces/ThreatExchange/python-threatexchange
 
 # Last line must have a newline char!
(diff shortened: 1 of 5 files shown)


BarrettOlson (approved):
Approving to unblock though this does feel like a bit overkill (especially if both are given 4GB of memory). I would think something like a script to create a python `venv` would be enough for this directory (and maybe a .vscode workspace settings file).

If we do go this route we should have both containers use the same base layer to reduce the point of maintenance because I'm sure there will be drift. 

Thanks!

python-threatexchange/.devcontainer/zshrc
```diff
@@ -0,0 +1,19 @@
...
+
+# Incrementally add to the history file.
+setopt INC_APPEND_HISTORY
+
+# Make path respect local binaries from pip
+export PATH=$PATH:~/.local/bin
+
+# Add directory containing hmalib to PYTHONPATH. This allows vscode
+# testdiscovery to work
+export PYTHONPATH=$PYTHONPATH:/workspaces/ThreatExchange/hasher-matcher-actioner
```
BarrettOlson, line 17: I think this needs to be updated...

Also could you have this also be added to `/home/$unixname/.profile` for those using https://code.visualstudio.com/docs/remote/containers#_personalizing-with-dotfile-repositories

python-threatexchange/.devcontainer/bashrc
```diff
@@ -0,0 +1,10 @@
+source /etc/bash.bashrc
+
+# Make path respect local binaries from pip
+export PATH=$PATH:~/.local/bin
+
+# Add directory containing hmalib to PYTHONPATH. This allows vscode
+# testdiscovery to work
+export PYTHONPATH=$PYTHONPATH:/workspaces/ThreatExchange/hasher-matcher-actioner
```
BarrettOlson, line 8: same note from the zshrc file

BarrettOlson (commented):
See note about PYTHONPATH before merge.
Also might be worth adding a post-create scripts that will do the`make local_install` step

python-threatexchange/.devcontainer/zshrc
schatten: Updating the path, but I'm not sure what .profile would do here. Feel free to move the files around. I only got things to a point where they work. I'm _certain_ I have not followed best practices. :)

BarrettOlson: I don't know that the best practice is but the way I use the `.profile` file is in my zshenv:
```
# for machine specific configuration
if [[ -f ~/.profile ]]; then
  source ~/.profile
fi
```
https://github.com/BarrettOlson/dotfiles/blob/master/zshenv#L31-L34

PR #714, opened 2021-07-09
Change confusing text on Action Settings Tab

slustbader:
Summary
---------

Changed confusing text on Action Settings Tab by making all the pronouns consistent.

Test Plan
---------

Ran web app locally, navigated to http://localhost:3000/settings/actions, confirmed updated text.


BarrettOlson (approved):
LGTM! :)

PR #715, opened 2021-07-09
convert action settings related files to typescript files

mengyangwang:
Summary
---------

This PR is to convert WebhookActioner, ActionPerformerColumns files to tsx files and fix type errors

Future Action
---------
Need to also convert ActionPerformerRows and ActionSettingsTab to tsx file, but there is a known issue with ion-icon package and typescript, will spend little more time to look into this issue

Test Plan
---------
test localhost for all test cases on Action setting tab

https://user-images.githubusercontent.com/81996660/125141707-c7d8bc00-e0e3-11eb-9094-50f2290a5130.mov

### DIFF
--- a/hasher-matcher-actioner/webapp/src/components/settings/ActionPerformer/WebhookActioner.tsx
+++ b/hasher-matcher-actioner/webapp/src/components/settings/ActionPerformer/WebhookActioner.tsx
@@ -11,12 +11,19 @@ export interface Map {
   [key: string]: string;
 }
 
+interface WebhookType extends Record<string, any> {
+  WebhookPostActionPerformer: string;
+  WebhookGetActionPerformer: string;
+  WebhookDeleteActionPerformer: string;
+  WebhookPutActionPerformer: string;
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/webapp/src/components/settings/ActionPerformer/WebhookActioner.tsx
```diff
@@ -5,15 +5,27 @@
 import React from 'react';
 import Card from 'react-bootstrap/Card';
 import Form from 'react-bootstrap/Form';
-import {PropTypes} from 'prop-types';
+import PropTypes from 'prop-types';
+
+export interface Map {
+  [key: string]: string;
+}
```
jeberl, line 12: Dont think we need this interface if the two uses are unrelated. Can we change to two `type`s instead?

jeberl (approved):
Looks good to me as long as you remove the Map interface. Dont need to come back to me. Thanks!

PR #716, opened 2021-07-14
Update dataset command to show more stats

slustbader:
Summary
---------

Add functionality to the existing `dataset` command to display stats about downloaded data. My proposed changes to the command:
`threatexchange dataset --type <content type>`:
```
 % threatexchange --config config.json dataset --type photo
pdq: 14023
photo_md5: 160
video_md5: 23
```

`threatexchange dataset --type <signal type>` (existing functionality):
```
% threatexchange --config config.json dataset --type video_md5
video_md5: 161124
photo_md5: 23
```

`threatexchange dataset --tag <tag name>`:
```
% threatexchange --config config.json dataset --tag media_priority_s1
video_md5: 14529
pdq: 657
photo_md5: 109
```

Test Plan
---------

See above for sample commands run to show filtering on a subset of the data for privacy group <>, app id <>. These commands demonstrate content type, indicator type, and tag filtering applied to the base data set; for reference, here are the stats for the full data set I have locally:
```
% threatexchange --config config.json dataset -s
video_md5: 161124
pdq: 14023
photo_md5: 160
```

To validate the actual counts, I used grep against the raw data to compare to the output from the dataset commands

For example:
```
% grep media_priority_s1 simple.HASH_MD5.te | wc -l
   14618
% threatexchange --config config.json dataset -t HASH_MD5 -g media_priority_s1
video_md5: 14529
photo_md5: 109
```
The total from `dataset` is 14,638, but `grep media_priority_s1 simple.HASH_MD5.te | grep media_type_photo | grep media_type_video` shows that there are 20 records tagged as both `media_type_photo` and `media_type_video`, which is why the count from dataset is 20 more than the initial grep search.

### DIFF
--- a/python-threatexchange/threatexchange/cli/dataset_cmd.py
+++ b/python-threatexchange/threatexchange/cli/dataset_cmd.py
@@ -107,9 +107,15 @@ class DatasetCommand(command_base.Command):
                 k: v
                 for k, v in indicators.items()
                 if v.indicator_type == self.only_type
-                or (signal_types
-                and any(sig_type.indicator_applies(v.indicator_type, v.rollup.labels) for sig_type in signal_types))
+                or (
+                    signal_types
+                    and any(
+                        sig_type.indicator_applies(v.indicator_type, v.rollup.labels)
+                        for sig_type in signal_types
+                    )
+                )
             }
+
         if self.rebuild_indices:
             generate_cli_indices(dataset, stores)
         if self.print_records:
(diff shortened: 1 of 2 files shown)


python-threatexchange/threatexchange/cli/dataset_cmd.py
```diff
@@ -43,7 +43,14 @@ def init_argparse(cls, ap) -> None:
...
             metavar="STR",
-            help="only process one type of indicators (or signals)",
+            help="only process one type of indicator (eg HASH_MD5), signal (eg photo_md5), or content (eg photo); specifying a signal or content type implies '--signal-summary'",
+        )
+        ap.add_argument(
+            "--tag",
+            "-g",
+            dest="only_tag",
+            metavar="STR",
+            help="only process a single tag; implies '--signal-summary'",
```
BarrettOlson, line 53: thanks for updating the docstrings! this strikes mea as more of a 'filter' than a 'process' but just a minor nit especially given this matches the wording of `--type`

BarrettOlson (approved):
LGTM!

PR #717, opened 2021-07-14
always mark sample test dataset id as in_use

mengyangwang:
Summary
---------

This PR will not mark sample test dataset as in_active when users click "sync" button in ThreatExchange settings UI.

Test Plan
---------

1. black hmalib
2. mypy hmalib
4. make docker
5. make upload_docker 
6. terraform apply
7. test in localhost

before : 

https://user-images.githubusercontent.com/81996660/125684911-1f9edc31-311d-489a-92fa-6e8129adb164.mov

after:

https://user-images.githubusercontent.com/81996660/125684940-1d4021af-be5a-4dd7-b9c1-cbc5cc12c3c6.mov

If sample test dataset is not created, click sync button will not add it

https://user-images.githubusercontent.com/81996660/125691572-c63b8f94-a453-4f81-85d6-ae0978571a45.mov

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/threatexchange_config.py
+++ b/hasher-matcher-actioner/hmalib/common/threatexchange_config.py
@@ -19,6 +19,7 @@ logger = get_logger(__name__)
 FETCHER_ACTIVE_DEFAULT = True
 WRITE_BACK_DEFAULT = True
 MATCHER_ACTIVE_DEFAULT = True
+SAMPLE_DATASET_PRIVACY_GROUP_ID = ["inria-holidays-test"]
 
 
 def create_privacy_group_if_not_exists(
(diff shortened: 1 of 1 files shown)


BarrettOlson (changes requested):
This means syncing with TE will always add the sample privacy groups. I do not think we want this.

BarrettOlson:
if they haven't hit the 'Create' button my understanding is this will create it? Given that those sample hashes are likely to be something you'd want to action in productions, at a minimum I think this should default matching to off. Thoughts?

mengyangwang:
> if they haven't hit the 'Create' button my understanding is this will create it? Given that those sample hashes are likely to be something you'd want to action in productions, at a minimum I think this should default matching to off. Thoughts?

No It will not add the 

> if they haven't hit the 'Create' button my understanding is this will create it? Given that those sample hashes are likely to be something you'd want to action in productions, at a minimum I think this should default matching to off. Thoughts?

mengyangwang:
> if they haven't hit the 'Create' button my understanding is this will create it? Given that those sample hashes are likely to be something you'd want to action in productions, at a minimum I think this should default matching to off. Thoughts?

No it will not add sample dataset if they haven't hit the "create" button, I added a video to show this. Adding sample test id to priavcy_group_id_in_use set will just be used to compare with current datasets in dynamoDB, it will not create anything

BarrettOlson (approved):
My mistake, thanks for the additional test case. Maybe someday I'll get used to the way github compressed code, I thought this was in `create_privacy_group_if_not_exists` at first.

LGTM. Apologies (though having this be passed to `priavcy_group_id_in_use` for all cases does feel weird)

schatten (approved):
LGTM.

PR #719, opened 2021-07-20
Convert files to TSX, fix errors and warnings

jeberl:
Summary
---------

Converting javascript files to typescript. Did some refactoring of ActionRules into an object which ended up taking longer than expected but I think is cleaner now.

Typescript doesn't play nice with Ion-Icons. Made a few attempts to fix
1. Add package `@ionic/react` eg: `npm install @ionic/react`. This got rid of the errors but the icons stopped showing up
2. Import directly from script - still led to errors
3. When all else fails, ignore the error since it compile correctly: 
```{/* @ts-expect-error TODO: ts doens't recognize that ion-icon has been imported */}```

Test Plan
---------

1. Ran locally (`npm start`)
2. tested the following functionality: 
    1. New ActionRule
    2. Delete Action Rule
    3. Rename ActionRule
    4. Add/Remove Classifications on ActionRule
    5. Edit classification options
    6. change Action

### DIFF
--- a/hasher-matcher-actioner/webapp/package-lock.json
+++ b/hasher-matcher-actioner/webapp/package-lock.json
@@ -8,7 +8,6 @@
       "version": "0.0.0",
       "dependencies": {
         "@aws-amplify/ui-react": "^1.2.4",
-        "@ionic/react": "^5.6.11",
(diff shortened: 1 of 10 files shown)


mengyangwang (approved):
LGTM!

PR #720, opened 2021-07-20
[video md5] Support Hashing for any content and signal type

schatten:
### [`d21f0912`](https://github.com/facebook/ThreatExchange/pull/720/commits/d21f0912a3429d6d708d8a4f1a23c316ab5f4755) [threatexchange][test][bugfix] Allow PDQ index results to have arbitrary order for same distance values

An annoying bug that only surfaced on the python version I am on. More details in commit message.

### [`84fa9c7e`](https://github.com/facebook/ThreatExchange/pull/720/commits/84fa9c7e3a0e87c72c48b77540cddfa60bf0d937) [hma] Add ContentType to URL submission requests

Instead of writing additional conditionals every time we want to support a new content type, make the submission message (for URLs) generic. It will support all content_types.

For brevity, this commit does not include all of this refactor. pdq_hasher must have changes to support the new submission messages, but they are removed in the next commits. URL Submissions messages are now supported by a unified lambda for all content types.

### [`03811043`](https://github.com/facebook/ThreatExchange/pull/720/commits/038110433830c3b95820af0c241b9ab06422cd3f) [hma] hmalib.ContentType enum replaced with threatexchange.ContentType classes

First, map ContentType enum to threatexchange.ContentType subclasses rather than hmalib content types.

Second, switch over string usage to enum usage wherever hard-coded.

### [`6cbd8f6a`](https://github.com/facebook/ThreatExchange/pull/720/commits/6cbd8f6a8f32cc960afb25ec81b2fdeddf8436a5) [hma] Webapp supports explicit content type

First, map ContentType enum to threatexchange.ContentType subclasses rather than hmalib content types.

Second, switch over string usage to enum usage wherever hard-coded.

### [`75d360ba`](https://github.com/facebook/ThreatExchange/pull/720/commits/75d360babb58a94c8d22a5b0e46c48660d4f0f97) [hma] Rename image_sources to content_sources

Re-use image sources for getting bytes of all kinds of content. They are capable of fetching bytes from URLs. Nothing fundamentally image specific about them.

Apologies. Some of the refactor for this also bleeds over into the next commit. Specifically, `pdq_hasher.py`

### [`92b6ce8b`](https://github.com/facebook/ThreatExchange/pull/720/commits/92b6ce8b1b55a01b75e635856efdb0206881cd0c) [hma] Add generic hasher which can handle any content and signal types; route submissions to it via a different SQS queue

As part of this, pdq_hasher.py's capability to handle URL only submissions is also being moved to the generic hasher. Once this diff lands, only S3 based submissions will go to pdq_hasher.

Also sets up terraform infrastructure.
- New SQS queue for URL submissions. Soon to include all submissions.

Does not yet publish matches, only prints them to logger. That will require generic-ifying matcher first.

Includes some pending refactors from previous commits. in submit.py and pdq_hasher.py. `URLImageSubmissionMessage` -> `URLSubmissionMessage` and `ImageSource` -> `ContentSource`.

`setup.py`: [tiny] so far, we were depending on threatexchange to provide requests implicitly. Made that explicit

### [`1fc52c5a`](https://github.com/facebook/ThreatExchange/pull/720/commits/1fc52c5a3158d0cd378c69701abf4ac28767c09c) Tiny, can be ignored: helper script for creating multi-part pull requests

I have now created enough multi-part pull requests to warrant automating somethings. 

# Test Plan

Lints:
```
$ python -m black hmalib 
$ python -m mypy hmalib
$ cd webapp; npx prettier src --check
```

Tests:
```
$ python -m py.test 
```

Deployed using terraform after tainting webapp.
- submitted a single image URL using the webapp. The generic hasher lambda printed to the logger.
- used the storm script to submit a large number of images via DIRECT upload. The matcher was triggered, indicating that the submission and the hashing went through.
  the matcher errored out due to a threatexchange issue (this PR required local_threatexchange which I had messed up)

# Post merge

1. Will need to cut a new release of threatexchange because hash_from_bytes is not yet available on pypy threatexchange.
2. Generic matching capabilities
3. Generic indexing capabilities, set() based index for MD5.
4. Fetch MD5 from threatexchange


schatten:
Test Failures are expected because threatexchange version bump needs to be merged in.

python-threatexchange/threatexchange/content_type/meta.py
```diff
@@ -28,3 +28,18 @@ def get_all_signal_types() -> t.Set[t.Type[signal_base.SignalType]]:
...
+
+
+_content_type_map = {
+    content_type.get_name(): content_type for content_type in get_all_content_types()
+}
+
+
+def get_content_type_for_name(name: str) -> t.Type[content_base.ContentType]:
+    """
+    Given a name, get the ContentTYpe which returns that name on calling
```
Dcallies, line 40: nit: TYpe

hasher-matcher-actioner/hmalib/hashing/__init__.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
Dcallies, line 1: Nit/YMMV:  I don't think we've been doing stuff in the `__init__` files, can sometimes make it hard to find the right file. I have a small preference for just using named files.

hasher-matcher-actioner/hmalib/hashing/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from threatexchange.signal_type.signal_base import SignalType, BytesHasher
+
+
+@dataclass
+class SignalValue:
+    signal_type: SignalType
+    signal_value: str
+
+
+class Hasher:
```
Dcallies, line 16: docstring plox, although if it does what I think it does, I like it!

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
...
+@functools.lru_cache(maxsize=None)
+def get_sqs_client() -> SQSClient:
+    return boto3.client("sqs")
+
+
+# OUTPUT_QUEUE_URL = os.environ["HASHES_QUEUE_URL"]
+DYNAMODB_TABLE = os.environ["DYNAMODB_TABLE"]
+
+# If you want to support additional content or signal types, they can be added
+# here.
```
Dcallies, line 41: One thing I was expecting was a kind of "wildcard" catch, that would use a generic implementation of any signal that didn't have explicit support.

However, I assume that means we need equivalent fetch/sync support which we don't have yet, so guessing this a project for the future.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
...
+        content_type = get_content_type_for_name(submission_message.content_type)
+        if not hasher.supports(content_type):
+            logger.warn(f"Unprocessable content type: {content_type}")
+            continue
+
+        with metrics.timer(metrics.names.hasher.download_file):
+            bytes_: bytes = URLContentSource().get_bytes(submission_message.url)
+            for signal in hasher.get_hashes(content_type, bytes_):
+                # TODO: Actually Publish to matcher.
+                print(f"{signal}")
```
Dcallies, line 85: 👍

I like that you are breaking up your commits! Great to create the pieces and connect them together first.

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -242,6 +242,154 @@ module "authentication" {
...
+ * Instead, we will have a single hashing lambda capable of handling all
+ * content_types. If the content, because of its size or because of compute
+ * complexity can't be handled by this "base" lambda, it will be routed to
+ * another specially capable lambda queue.
+ *
+ * - This should soon absorb the pdq_images queue + SNS topic as the only queue
+ *   that we will publish submissions on.
+ *   If we have proven that the generic lambda can generate PDQ signals, we can 
+ *   do away with the PDQ specific infrastructure altogether.
+*/
```
Dcallies, line 260: nit: misaligned

schatten:
Abandoning because github does not deal with the common ancestor changing in pull requests.

PR #721, opened 2021-07-20
[threatexchange][videomd5] Support hashing by bytes in Threatexchange SignalTypes

schatten:
### [`d21f0912`](https://github.com/facebook/ThreatExchange/pull/721/commits/d21f0912a3429d6d708d8a4f1a23c316ab5f4755) [threatexchange][test][bugfix] Allow PDQ index results to have arbitrary order for same distance values

An annoying bug that only surfaced on the python version I am on. More details in commit message.

### [`b6d54535`](https://github.com/facebook/ThreatExchange/pull/721/commits/b6d54535462e98a96a9e340fb4ed6ce0220a2780) [threatexchange] utility method to get ContentType from name

Test Plan
---------
```
python-threatexchange $ python -m py.test
python-threatexchange $ python -m black .
python-threatexchange $ python -m mypy .
< unrelated failure > 
threatexchange/__init__.py: error: Duplicate module named 'threatexchange' (also at './build/lib/threatexchange/__init__.py')
```


python-threatexchange/threatexchange/signal_type/tests/test_pdq_index.py
```diff
@@ -63,9 +65,41 @@ def assertEqualPDQIndexMatchResults(self, result, expected):
         self.assertEqual(
             len(result), len(expected), "search results not of expected length"
         )
-        for (r, e) in zip(result, expected):
-            assert r.distance == e.distance, "result distance does not match"
-            self.assertDictEqual(r.metadata, e.metadata)
+
+        # Between python 3.8.6 and 3.8.11, something caused the order of results
```
BarrettOlson, line 69: weird... but we didn't guarantee it so I tests shouldn't have depended on it.

python-threatexchange/threatexchange/signal_type/pdq.py
```diff
@@ -61,3 +62,16 @@ def match_hash(self, signal_str: str) -> t.List[signal_base.SignalMatch]:
...
             if pdq_match(pdq_hash, signal_str, self.PDQ_CONFIDENT_MATCH_THRESHOLD)
         ]
+
+    @classmethod
+    def hash_from_bytes(self, bytes_: bytes) -> str:
+        try:
+            from threatexchange.hashing.pdq_hasher import pdq_from_bytes
+        except:
+            warnings.warn(
+                "PDQ from file require Pillow and pdqhash to be installed; install threatexchange with the [pdq_hasher] extra to use them",
```
BarrettOlson, line 72: nit: update text or make it generic 'const' string  and reuse.

python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -11,7 +11,6 @@
 import typing as t
 
 from .. import common
-from .. import dataset
```
BarrettOlson, line 16: When the commit message is 25 times longer than the fix itself, mistakes have been made... haha

python-threatexchange/threatexchange/content_type/meta.py
```diff
@@ -28,3 +28,18 @@ def get_all_signal_types() -> t.Set[t.Type[signal_base.SignalType]]:
 @functools.lru_cache(1)
 def get_signal_types_by_name() -> t.Dict[str, signal_base.SignalType]:
     return {s.get_name(): s for s in get_all_signal_types()}
+
+
+_content_type_map = {
```
BarrettOlson, line 33: nit: the pattern in this files seems to be to use caching instead of import/runtime creation. 

Thoughts on this instead to match? 
```
@functools.lru_cache(1)
def get_content_map():
    return content_type.get_name(): content_type for content_type in get_all_content_types()
```

BarrettOlson (approved):
LGTM!
few nits that are probably worth fixing before merge.
(Anything I didn't love in the design was already there and fair amount of it was my doing haha)

python-threatexchange/threatexchange/content_type/meta.py
schatten: True for exported methods. For internal things like the cached map, there are no precedents.

PR #723, opened 2021-07-20
Bucket hasher

jeberl:
Summary
---------

This PR enables a user to connect HMA to a local bank such that all images uploaded to that bank will be processed by HMA without having to call a separate API.

Future work:
1. Enable user to filter to specific folders to upload

Test Plan
---------

```
```
3. `terraform -chdir=terraform apply`
5. Through logs watch them be processed by the local_bucket_hasher lambda, added to the sqs queue, and then loaded and hashed by the pdq_hasher: Screen Shot 2021-07-20 at 7 11 09 PM
6. Uploaded images using the API to ensure that path still works
7. Uploaded images to HMA s3 bucket directly through UI to ensure that path still works


hasher-matcher-actioner/hmalib/common/message_models.py
```diff
@@ -208,9 +220,12 @@ def could_be(cls, d: dict) -> bool:
         """
         return "EventType" in d
 
+    def get_bytes(self) -> bytes:
+        return URLImageSource().get_image_bytes(self.url)
+
 
 @dataclass
-class S3ImageSubmission:
+class S3ImageSubmission(ImageSubmission):
```
jeberl, line 228: Could probably combine the `Submission` and `Source` objects such as `S3ImageSubmission` and `S3BucketImageSource`

hasher-matcher-actioner/hmalib/lambdas/local_bucket_hasher.py
```diff
@@ -0,0 +1,44 @@
+import os
```
jeberl, line 1: Renamed this file from hasher_integrations

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -30,26 +30,6 @@
 ]  # Misnamed, this is a prefix, not a key, if renaming, use IMAGE_PREFIX
 
 
-def get_image_bytes(
```
jeberl, line 33: Moved this code into the `S3BucketImageSource`, `URLImageSource`, and `S3LocalBucketImageSource` objects to make use of patterns

hasher-matcher-actioner/hmalib/common/image_sources.py
```diff
@@ -80,7 +80,27 @@ class URLImageSource(ImageSource):
...
+    """
+
+    def get_image_bytes(self, content_id: str) -> bytes:
+        """
+        Get bytes for the image.
+
+        Assumes content_id is stored as <bucket_id>/key based on
+        S3LocalImageSubmissionBatchMessage.
+        """
+        bucket = content_id[: content_id.find("/")]
```
schatten, line 103: Easier to use split? 

```python
bucket, key = content_id.split('/', 1)
```

hasher-matcher-actioner/hmalib/lambdas/local_bucket_hasher.py
```diff
@@ -0,0 +1,44 @@
...
+logger = get_logger(__name__)
+sqs_client = boto3.client("sqs")
+
+PDQ_IMAGES_QUEUE_URL = os.environ["PDQ_IMAGES_QUEUE_URL"]
+
+
+def lambda_handler(event, context):
+    if is_s3_event(event):
+        event["local_bucket"] = True
+        sqs_client.send_message(
```
schatten, line 16: This can be made more generic. 

Instead of sending to the PDQ IMAGES QUEUE, if you make an API call to the submissions API, that would allow changes to the submissions API to automatically benefit this.

If we start supporting another image hashing algorithm, this will need to be reworked in its present state. However, if you send the event to the submissions API, there is a single point of maintenance.

Submission API needs the following attributes:
`content_id`: Which in this case will be the full s3_url
`content_type`: Which will be photo to begin with, but can be expanded to video easily.
`url`: This could be a pre-signed S3 URL. You can generate one from the bucket + key combo.

This removes the need for having custom events or code to handle new use-cases.

If we go as close to the main-line as possible when building a new feature, it reduces the points of maintenance.

What you are proposing:
```
local_bucket_hasher───────┐
                          │
                          │
s3Event_source────────────┤
                          │
                          │
submissions_api ──┬───────┴─────── pdq_hasher
                  │
                  │
                  ├─────────────── alternate_hasher
                  │
                  └─────────────── alternate_content_type
```

What it could be, thus benefiting automatically from new hashers:
```
local_s3_bucket│     ┼
               │
               │
api_calls──────┴──────────submissions_api ──┬─────────────── pdq_hasher
                                            │
                                            │
                                            ├─────────────── alternate_hasher
                                            │
                                            └─────────────── alternate_content_type
```

schatten (changes requested):
## "Local" Not quite getting the hang of it.
Considering users of this feature will be developers, what does local mean to them? When they are developing, local means on their system, not on the cloud. Eg. localhost, local dev environment. How does that translate here? 

Can we think of a better name? One that reflects that it is external to HMA and managed by partners?

Also, on the names note, "local_bucket_hasher" is not hashing. It is submitting for the purposes of hashing.

## Using the submissions API instead of sending directly to pdq_hasher

I have outlined my proposal in inline comments below. Essentially, if you are sending to pdq_hasher, you are losing the routing capability of the submissions API. Submit directly to the API from your new lambda.

Let the submissions API identify the kind of hashers it needs to apply to the content. Local S3 Buckets will automatically benefit from new hashing algorithms and content_types. eg. you can make a one-line change to enable processing videos from the local s3 buckets.

## Merge Conflicts, Yay!
Even though github may allow this to be merged in, image_sources file has been moved, so I'd recommend rebasing before you make changes.

jeberl:
Replaying to @schatten 's top-level comments:

## local_bucket_hasher
All good points, went back and forth on the name. I originally had `hashing_integrations`. How about, since "Partner" is a defined term in the glossary, `partner_bucket_integration`. Happy to workshop more

## Using the submissions API instead of sending directly to pdq_hasher
I agree that we will need one "meta-hasher" that should take all content types and route it to the correct hasher implementations (pdq_hasher, tmk_hasher, ...). I don't think that place should be the api_submissions for 2 reasons:
1. We already have other inputs to the pdq_hasher that need to be converted. I believe a better structure would 
local_bucket_hasher───────┐                                                        |── alternate_hasher
                                                               │                                                         |
                                                               │                                                         |
s3Event_source─────────-┤--------------meta-hasher -----┤---pdq_hasher
                                                                │                                                        |
                                                                │                                                        |
submissions_api ──┬──────-┴                                                       └── alternate_content_type



2. Latency

My proposal

hasher-matcher-actioner/hmalib/common/image_sources.py
jeberl: Wasn't aware of the _second_ split argument!

schatten:
@jeberl  and I discussed on VC. Here are the highlights.

Firstly, `partner_bucket_integrations`. Sooper like!

Also, to clarify, I am recommending calling the API via the API gateway rather than the python function.

If we make changes to the api handlers, it might also come with terraform policy changes. Unless there is a way to share terraform policies between partner_bucket_integrations lambda and the api_submissions_lambda, let's call the API via API Gateway.  Else, we end up with two points of maintenance. 

Note: If @jeberl  can figure out a way to share policies between api and partner_integrations lambdas though, great! We can call the functions directly.

### Why I prefer submissions_api doing the routing.
`submissions_api` does more than just route at this point. It creates the content object in dynamodb which may become useful to downstream systems. Bypassing it would mean the meta-hashing lambda needs to do the coordination work that is presently done by submissions API.

Meta-hashing lambda would need to be a higher tier lambda because it is going to download the contents. We've had to go from 128M -> 512M already because pdq_hasher refused to work on 10MB images.

The same operations done on hasher will be more expensive than anywhere else. And linearly so. So if the hashing lambda is at 1 GB and the submissions lambda is at 128 MB, the same ddb call will be 8x more costly.

Going forward, one of the ways to reduce costs for partners would be to optimize what the hashing lambdas are doing.

### Why the hashing-data/images s3Event is not external
Presently, for payload submission (where the bytes of the content arrive at the API), we write a content object and then upload to s3, but we don't send an event directly. This is to prevent race conditions where the image is not yet available on s3, but the event has already arrived on the hasher.

Partners should not upload directly to hashing-data/images. Their entry point should be submissions API alone.

schatten:
Oh, and I was mistaken. #720 is still far away from merge, so the image_sources file exists. You're good.

jeberl:
Spoke with @schatten offline. We came up with a plan to have the S3 upload invoke api_root directly to accomplish the same. Putting up a #726 since they are quite different

PR #725, opened 2021-07-21
[DO NOT MERGE WITHOUT MATCHER][video md5] Support Hashing for any content and signal type

schatten:
Summary
---------

Builds off of threatexchange 0.0.22 which introduces hashing in memory.

### [`3118f8dc`](https://github.com/facebook/ThreatExchange/pull/725/commits/3118f8dce6950349f6193937b2c9eb0644792373) [hma] Add ContentType to URL submission requests

Instead of writing additional conditionals every time we want to support a new content type, make the submission message (for URLs) generic. It will support all content_types.

For brevity, this commit does not include all of this refactor. pdq_hasher must have changes to support the new submission messages, but they are removed in the next commits. URL Submissions messages are now supported by a unified lambda for all content types.

### [`0098c1ed`](https://github.com/facebook/ThreatExchange/pull/725/commits/0098c1ed3e83951c0365191728d7ac8ec84bd1f5) [hma] hmalib.ContentType enum replaced with threatexchange.ContentType classes

First, map ContentType enum to threatexchange.ContentType subclasses rather than hmalib content types.

Second, switch over string usage to enum usage wherever hard-coded.

### [`5e569a5f`](https://github.com/facebook/ThreatExchange/pull/725/commits/5e569a5f8307fc12d1653f801bf1b9e44275961b) [hma] Webapp supports explicit content type

First, map ContentType enum to threatexchange.ContentType subclasses rather than hmalib content types.

Second, switch over string usage to enum usage wherever hard-coded.

### [`11cda7a6`](https://github.com/facebook/ThreatExchange/pull/725/commits/11cda7a6c609e912c3c25989f2d706b27794f3ae) [hma] Rename image_sources to content_sources

Re-use image sources for getting bytes of all kinds of content. They are capable of fetching bytes from URLs. Nothing fundamentally image specific about them.

Apologies. Some of the refactor for this also bleeds over into the next commit. Specifically, `pdq_hasher.py`

### [`269fdecf`](https://github.com/facebook/ThreatExchange/pull/725/commits/269fdecf1c7a617873e858157f35b52d38b51e73) [hma] Add generic hasher which can handle any content and signal types; route submissions to it via a different SQS queue

As part of this, pdq_hasher.py's capability to handle URL only submissions is also being moved to the generic hasher. Once this diff lands, only S3 based submissions will go to pdq_hasher.

Also sets up terraform infrastructure.
- New SQS queue for URL submissions. Soon to include all submissions.

Does not yet publish matches, only prints them to logger. That will require generic-ifying matcher first.

Includes some pending refactors from previous commits. in submit.py and pdq_hasher.py. `URLImageSubmissionMessage` -> `URLSubmissionMessage` and `ImageSource` -> `ContentSource`.

`setup.py`: [tiny] so far, we were depending on threatexchange to provide requests implicitly. Made that explicit

### [`ac990868`](https://github.com/facebook/ThreatExchange/pull/725/commits/ac99086874168bec601fa1d48ad0c9455356784f)Tiny, can be ignored: helper script for creating multi-part pull requests

I have now created enough multi-part pull requests to warrant automating somethings. 

### [`397e789b`](https://github.com/facebook/ThreatExchange/pull/725/commits/397e789b392616941b4d2e6467c37efb0461add5) Tiny, can be ignored: Review feedback and test fixes

# Test Plan

Lints:
```
$ python -m black hmalib 
$ python -m mypy hmalib
$ cd webapp; npx prettier src --check
```

Tests:
```
$ python -m py.test 
```

Deployed using terraform after tainting webapp.
- submitted a single image URL using the webapp. The generic hasher lambda printed to the logger.
- used the storm script to submit a large number of images via DIRECT upload. The matcher was triggered, indicating that the submission and the hashing went through.
  the matcher errored out due to a threatexchange issue (this PR required local_threatexchange which I had messed up)

# Post merge
1. Will need to cut a new release of threatexchange because hash_from_bytes is not yet available on pypy threatexchange.
2. Generic matching capabilities
3. Generic indexing capabilities, set() based index for MD5.
4. Fetch MD5 from threatexchange


hasher-matcher-actioner/webapp/src/utils/constants.jsx
```diff
@@ -1,5 +1,9 @@
 /**
  * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ *
+ * TODO: typescript supports enums. However, all files that use enums from here
+ * need to be converted to typescript before we can change this file into
+ * typescript.
```
jeberl, line 6: Seems like a good time to typescript these files? Can you add that in a new commit?

hasher-matcher-actioner/webapp/src/utils/constants.jsx
```diff
@@ -37,3 +41,12 @@ export const StatNames = Object.freeze({
   HASHES: 'hashes',
   MATCHES: 'matches',
 });
+
+// Matches threatexchange.content_type.* for each subclass of ContentType, get_name()
+// >>> from threatexchange.content_type.meta import get_all_content_types
+// >>> [content_type.get_name() for content_type in get_all_content_types()]
```
jeberl, line 47: Love this however it doesn't work for me on my devcontainer. get the following error:
```AttributeError: partially initialized module 'threatexchange.signal_type.signal_base' has no attribute 'SignalType' (most likely due to a circular import)```

File: hasher-matcher-actioner/hmalib/common/content_sources.py [hunk 3]
```diff
@@ -6,15 +6,15 @@
 import typing as t
 
 
-class ImageSource:
+class ContentSource:
```
jeberl, line 9: Naming nit: We use "Source" as a [defined term](https://github.com/facebook/ThreatExchange/wiki/Glossary#fetcher) elsewhere in HMA. Can we use another word here? "Origin" maybe?

hasher-matcher-actioner/setup.py
```diff
@@ -13,5 +13,6 @@
         "threatexchange[faiss,pdq_hasher]>=0.0.20",
         "bottle",
         "apig_wsgi",
+        "requests==2.25.1",
```
jeberl, line 16: 👍

hasher-matcher-actioner/terraform/api/variables.tf
```diff
@@ -109,3 +109,11 @@ variable "writebacks_queue" {
   })
 }
 
+variable "submissions_queue" {
+  description = "URL and ARN for submissions queue. Messages from the submission APIs will be dropped on this queue"
```
jeberl, line 113: nit s\submissions\content submissions

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
...
+
+# If you want to support additional content or signal types, they can be added
+# here.
+hasher = Hasher(
+    supported_content_types=[PhotoContent, VideoContent],
+    supported_signal_types=[PdqSignal, VideoMD5Signal],
+)
+
+
+def lambda_handler(event, context):
```
jeberl, line 48: Seems like we should move the content record keeping from the API submissions function here? If the plan is for this lambda to be small.

Then, potentially, Partner s3 bucket uploads could be connected to the hasher, rather than to the api_root

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
jeberl, line 1: Why `hashing.py` instead of `hasher.py`?

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
...
+    records_table = get_dynamodb().Table(DYNAMODB_TABLE)
+
+    for sqs_record in event["Records"]:
+        logger.info(f"Body: {sqs_record}")
+        message = json.loads(sqs_record["body"])
+
+        if message.get("Event") == "s3:TestEvent":
+            continue
+
+        if URLSubmissionMessage.could_be(message):
```
jeberl, line 71: Why not feed S3 events here as well? Why do they still go to the pdq_hasher?

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
...
+            submission_message = URLSubmissionMessage.from_sqs_message(message)
+        else:
+            logger.warn(f"Unprocessable Message: {message}")
```
jeberl, line 74: I think we want to `return` or `continue` after this or we'll get errors on L76

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -0,0 +1,85 @@
...
+        content_type = get_content_type_for_name(submission_message.content_type)
+        if not hasher.supports(content_type):
+            logger.warn(f"Unprocessable content type: {content_type}")
+            continue
+
+        with metrics.timer(metrics.names.hasher.download_file):
+            bytes_: bytes = URLContentSource().get_bytes(submission_message.url)
+            for signal in hasher.get_hashes(content_type, bytes_):
+                # TODO: Actually Publish to matcher.
+                print(f"{signal}")
```
jeberl, line 85: logger.info(f"{signal}")

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
```diff
@@ -42,21 +42,18 @@ def get_image_bytes(
...
     """
-    if isinstance(submission_message, URLImageSubmissionMessage):
-        return URLImageSource().get_image_bytes(submission_message.url)
-    else:
-        return S3BucketImageSource(
-            submission_message.bucket, default_s3_bucket_image_prefix
-        ).get_image_bytes(submission_message.content_id)
+    return S3BucketContentSource(
+        submission_message.bucket, default_s3_bucket_image_prefix
+    ).get_bytes(submission_message.content_id)
```
jeberl, line 47: See above. Why not move this to hashing.py as well?

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -242,6 +242,154 @@ module "authentication" {
...
+
+  timeout     = 300
+  memory_size = 512
+
+  environment {
+    variables = {
+      DYNAMODB_TABLE      = module.datastore.primary_datastore.name
+      METRICS_NAMESPACE   = var.metrics_namespace
+      MEASURE_PERFORMANCE = var.measure_performance ? "True" : "False"
+      # HASHES_QUEUE_URL    = module.pdq_signals.hashes_queue_url
```
jeberl, line 296: remove?

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -242,6 +242,154 @@ module "authentication" {
...
+  statement {
+    effect    = "Allow"
+    actions   = ["sqs:GetQueueAttributes", "sqs:ReceiveMessage", "sqs:DeleteMessage"]
+    resources = [aws_sqs_queue.submissions_queue.arn]
+  }
+  /*statement {
+    effect    = "Allow"
+    actions   = ["sqs:SendMessage"]
+    resources = [aws_sqs_queue.hashes_queue.arn]
+  }*/
```
jeberl, line 356: remove?

hasher-matcher-actioner/hmalib/hashing/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(
+        self,
+        supported_content_types: t.List[t.Type[ContentType]],
+        supported_signal_types: t.List[t.Type[SignalType]],
+    ):
+        self.supported_content_types = supported_content_types
+
+        # Not enforced in typing because python does not yet have t.Intersect,
+        # but all supported_signal_types must also implement BytesHasher
+        assert all([issubclass(t, BytesHasher) for t in supported_signal_types])
```
jeberl: nit: You can remove the square brackets here. `all` takes any Iterable, not just a list:

```
assert all(issubclass(t, BytesHasher) for t in supported_signal_types)
```

hasher-matcher-actioner/hmalib/hashing/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+        return content_type in self.supported_content_types
+
+    def get_hashes(
+        self, content_type: t.Type[ContentType], bytes_: bytes
+    ) -> t.Generator[SignalValue, None, None]:
+        for signal_type in content_type.get_signal_types():
+            if signal_type in self.supported_signal_types and issubclass(
+                signal_type, BytesHasher
+            ):
+                yield SignalValue(signal_type, signal_type.hash_from_bytes(bytes_))
```
jeberl: Nice use of a Generator!

hasher-matcher-actioner/hmalib/hashing/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        return content_type in self.supported_content_types
+
+    def get_hashes(
+        self, content_type: t.Type[ContentType], bytes_: bytes
+    ) -> t.Generator[SignalValue, None, None]:
+        for signal_type in content_type.get_signal_types():
+            if signal_type in self.supported_signal_types and issubclass(
+                signal_type, BytesHasher
+            ):
```
jeberl: If we have the assert above, is the second condition necessary?

hasher-matcher-actioner/hmalib/hashing/unified_hasher.py
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
jeberl, line 1: See comments on earlier commit

hasher-matcher-actioner/scripts/multi_part_pull_request_template
```diff
@@ -0,0 +1,52 @@
+#! /usr/bin/env python3
```
jeberl, line 1: This is very meta 🤣

jeberl (changes requested):
lmk when you've replied to my comments and I'll rereview or can discuss over vc

hasher-matcher-actioner/webapp/src/utils/constants.jsx
schatten: From my past experiences, hard/impossible to `import <typescript module>` in a js file. 

These leaf files (ones that do not import anything else) must be the last ones to be migrated. That includes constants.jsx.

For the others, yeah, sure I can give that a shot.

hasher-matcher-actioner/webapp/src/utils/constants.jsx
schatten: Yeah, you'll need to `pip install .` again. Threatexchange needs to be updated. I fixed that in 0.0.22.

hasher-matcher-actioner/hmalib/common/content_sources.py
schatten: Sure. How about `ContentStorage`? Subclasses would be `S3ContentStorage` and `URLContentStorage`.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: This is the main hashing lambda. I expect it to run on a higher tier of AWS Lambda functions. We will be decommissioning pdq_hasher lambda once this is proven out.

hashing need not be the only classification / labelling system used. eg. regex matching of text will not have a hashing stage. Submission is what's common across all content types. Better to keep common things (record keeping) there.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: Reducing the surface area of change. Once proven out, we can move all submission types here.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: This is temporary. Obviously, will replace with actual publishing of message. This only "enables" hashing. 

Working on matching next.

hasher-matcher-actioner/hmalib/lambdas/pdq/pdq_hasher.py
schatten: Answered above. Just reducing surface area of change until I have proven out matcher.

Right now, hashing.py just does the hashing. It does not forward that to any other system. Once it starts to do that, I'll connect these pieces.

hasher-matcher-actioner/terraform/main.tf
schatten: Will be required once hashes are published.

hasher-matcher-actioner/terraform/main.tf
schatten: will be required once hashes are published.

hasher-matcher-actioner/hmalib/hashing/__init__.py
schatten: Naah, I guess not.

schatten:
Hey @jeberl  I have replied to your comments. Note: that I do not plan on merging this as-is without building the matcher first. I'm branching off from this branch and working on the matcher / etc. 

Even if approved, this will be merged only after the matcher bits are complete. Updated PR title to reflect the same.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: No reason. Not held to either name.

hasher-matcher-actioner/webapp/src/utils/constants.jsx
jeberl: Works when I do it lol :D

1. Change file name to `constants.tsx`
2. Find and replace imports like: `s/constants';/constants.tsx';`

hasher-matcher-actioner/hmalib/lambdas/hashing.py
jeberl: Got it

hasher-matcher-actioner/hmalib/lambdas/hashing.py
jeberl: Not blocking: I kind of prefer hasher since its called hasher-matcher-actioner and we already have a matcher.py and an actioner.py (well technically an action_evaluator.py and an action_performer.py)

hasher-matcher-actioner/hmalib/lambdas/hashing.py
jeberl: Oops. Yeah duh

jeberl (approved):
Some in-line responses but nothing blocking. See the typescript and hashing/hasher specifically.

schatten:
Merged as part of integration branch #749

PR #726, opened 2021-07-21
Partner Bucket Hasher (v2)

jeberl:
This is a second version of  #723. See comments there for context

Test Plan
---------

```
```
3. `make upload_docker && terraform -chdir=terraform apply`
5. Through logs watch them be processed by the api_root lambda, added to the sqs queue, and then loaded and hashed by the pdq_hasher: Screen Shot 2021-07-21 at 5 47 28 PM


hasher-matcher-actioner/terraform/main.tf
```diff
@@ -73,21 +73,6 @@ module "hashing_data" {
   additional_tags       = merge(var.additional_tags, local.common_tags)
 }
 
-module "hashing_integrations" {
```
jeberl, line 76: No longer need a separate module! #deadcode

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -71,12 +78,72 @@ def root():
...
+                dynamodb_table=dynamodb.Table(DYNAMODB_TABLE),
+                images_topic_arn=IMAGES_TOPIC_ARN,
+            )
+            if (
+                isinstance(response, SubmitContentResponse)
+                and response.submit_successful
+            ):
+                logger.info(f"Sucessfully submitted s3 event record as url upload.")
+            else:
+                logger.info(
```
schatten, line 109: logger.error?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -71,12 +78,72 @@ def root():
...
+
+    2. Platforms can connect their AWS S3 Buckets directly to HMA so that uploads to those buckets are
+    fed directly into the system. When an upload occurs, this lambda is invoked with an s3 event. We then
+    convert the event into a URL which we submit to the hasher (via SNS)
     """
+    if is_s3_event(event):
+        logger.info(
+            "Lambda triggered with S3 event. Converting to submit content request."
+        )
+        for record in event["Records"]:
```
schatten, line 93: Can you extract this out into a separate function?

the body of `if is_s3_event()`... is too large compared to the `else`.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -71,12 +78,72 @@ def root():
...
+            if (
+                isinstance(response, SubmitContentResponse)
+                and response.submit_successful
+            ):
+                logger.info(f"Sucessfully submitted s3 event record as url upload.")
+            else:
+                logger.info(
+                    "Failed to submit s3 event record as url upload. "
+                    + response.message
+                    if isinstance(response, SubmitContentError)
```
schatten, line 112: We are swallowing the error cases here. That would prevent the queue from doing retries. Shouldn't we raise here instead?

If you are repurposing an HTTP handler or view for another objective, you'll keep converting between the view classes and your events.

Instead, extract the meat of submit_from_url into a method that raises exceptions when failed. 
```
def submit_from_url_or_a_better_name(url, queue, dynamodb) 
# use primitive arguments rather than a serializable. Serializability is helpful
# when doing remote calls or in HTTP, usually limiting when doing python - 
# python calls.
# This can raise exceptions. Or do pretty much anything the python way.

@submit_api.route
def submit_from_url_api_method():
  try: submit_from_url() except Exception e: return SubmitContentResponse
```

Keep the HTTP ness of the method hidden from the s3 event handing portions.

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -153,6 +160,28 @@ def record_content_submission(dynamodb_table: Table, request: SubmitContentReque
     ).write_to_table(dynamodb_table)
 
 
+def submit_from_url(
```
schatten, line 163: This is the method that is unnecessarily HTTP friendly. Make this pythonic. can raise exceptions and use primitive arguments. 

Restrict the http request / response payload / status code things to the `from_url` method below.

schatten: The benefit is that the function invocation from python becomes more natural. Right now, we have

`submit_from_url()` catches exception, returns SubmitContentError -> api_root's `s3_event_handler` checks for instance of SubmitContentError and then raises exception.

This is happening because submit_from_url() is trying to be HTTP friendly. It need not.

schatten (changes requested):
The architecture looks good. However, by my reading, retries will not be supported because we are swallowing errors in the handler. I may be wrong.

Have recommended slight reorg in the submit.py file.

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
jeberl: discussed offline. `submit_from_url` can never return an error (only throw exceptions) so simplifying logic

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
jeberl: see above

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -16,7 +16,14 @@
 from .datasets_api import get_datasets_api
 from .matches import get_matches_api
 from .stats import get_stats_api
-from .submit import get_submit_api
+from .submit import (
+    SubmitContentError,
```
schatten, line 20: Clean up imports?

schatten (approved):
LGTM. Small nit about cleaning up imports. 

Can you add an issue to the documentation project so that we don't miss out adding a wiki entry for this feature?

jeberl:
Issue: 731

PR #728, opened 2021-07-23
[Video MD5 support] PR 1/n: Fetch MD5 as well as PDQ from threatexchange and store in S3

schatten:
Summary
---------
`fetcher.py` initially pulled PDQ hashes alone. With this, it can pull any indicator type. For now, we continue to restrict it to PDQ and MD5, but more can be added by making one line changes.

Would prefer a commit by commit review.

### [`bdab484a`](https://github.com/facebook/ThreatExchange/pull/728/commits/bdab484a301b6028b06a29ce320da1bc787bd4b5) [md5-fetcher] Fetch indicators of type MD5 in fetcher.py

This is the feature addition. `ThreatUpdateS3PdqStore` ➡️ `ThreatUpdateS3Store` which accepts a list of supported signal types.

### [`7bfddb36`](https://github.com/facebook/ThreatExchange/pull/728/commits/7bfddb36ce0622002dbdb563314a43ca283db18e) Move ThreatUpdateS3Store to s3_adapters.py

It seemed like fetcher, indexer and some parts datasets_api were all interacting with the same files in s3. It makes sense to colocate the logic so we can more easily refactor.

The fetcher and indexer were using files ending in `.pdq.te` as an informal convention which must be eliminated. This is a move towards that.

### [`/5bfd877`](https://github.com/facebook/ThreatExchange/pull/728/commits/5bfd877dbe8478085a014e61b94c0631bd07733a) Call ThreatUpdateS3Store.supported_indicator_types only once

Based on @jeberl's review suggestion to remove supported_signal_types and replace with supported_indicator_types. This removes the need to cache results.

Test Plan
---------
```
$ python -m black hmalib
$ python -m mypy hmalib
$ python -m py.test .
```

Also, ran fetcher locally after clearing `threat_exchange_data/`. Used the following script

```python
import os

os.environ[
    "THREAT_EXCHANGE_DATA_BUCKET_NAME"
] = "dipanjanm-hashing-<>"
os.environ["THREAT_EXCHANGE_DATA_FOLDER"] = "threat_exchange_data/"
os.environ["CONFIG_TABLE_NAME"] = "<>"
os.environ["DYNAMODB_DATASTORE_TABLE"] = "<>"
os.environ["THREAT_EXCHANGE_API_TOKEN_SECRET_NAME"] = "threatexchange/dipanjanm_api_tokens"

from hmalib.lambdas.fetcher import lambda_handler

lambda_handler(None, None)

```


hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -182,38 +184,55 @@ def __init__(
...
+        self.supported_signal_types = supported_signal_types
+
+    @functools.lru_cache(maxsize=None)
+    def _get_supported_indicator_types(self):
+        """
+        For supported self.signal_types, get their corresponding indicator_types.
+        """
+        indicator_types = [
+            "HASH_VIDEO_MD5"
+        ]  # hardcoding this because md5 currently points to HASH_MD5
```
jeberl: Is this to make it backwards compatible? 

Why not change `VideoMD5Signal.INDICATOR_TYPE  = "HASH_VIDEO_MD5" `

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -182,38 +184,55 @@ def __init__(
...
+
+        for signal_type in self.supported_signal_types:
+            indicator_type = getattr(signal_type, "INDICATOR_TYPE", None)
+            if indicator_type:
+                indicator_types.append(indicator_type)
+            else:
+                warnings.warn(
+                    f"SignalType: {signal_type} does not provide an indicator type."
+                )
```
jeberl: nit: If this code is reached its like the result of a bug right? Feels like it should be a error/exception rather than warn

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -182,38 +184,55 @@ def __init__(
...
         super().__init__(privacy_group)
         self.app_id = app_id
         self._cached_state: t.Optional[t.Dict] = None
         self.s3_bucket = s3_bucket
         self.s3_te_data_folder = s3_te_data_folder
         self.data_store_table = data_store_table
+        self.supported_signal_types = supported_signal_types
+
+    @functools.lru_cache(maxsize=None)
+    def _get_supported_indicator_types(self):
```
jeberl: nit: Maybe add this to the constructor before setting supported_signal_types. That way we know its only called once and no one uses the raws `self.supported_signal_types` instead of this function

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -182,38 +184,55 @@ def __init__(
...
+        """
+        For supported self.signal_types, get their corresponding indicator_types.
+        """
+        indicator_types = [
+            "HASH_VIDEO_MD5"
+        ]  # hardcoding this because md5 currently points to HASH_MD5
+
+        for signal_type in self.supported_signal_types:
+            indicator_type = getattr(signal_type, "INDICATOR_TYPE", None)
```
jeberl: Why not just `indicator_type = signal_type.INDICATOR_TYPE` and raise any exception from missing params?

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -182,38 +184,55 @@ def __init__(
...
 
-    @property
-    def data_s3_key(self) -> str:
+    def get_s3_object_key(self, indicator_type) -> str:
         """
         Creates a 'file_name' structure that is dependent on downstream (see s3_adapters.py)
-        TODO replace ".pdq.te" with os.environ["THREAT_EXCHANGE_PDQ_FILE_EXTENSION"]
         """
-        return f"{self.s3_te_data_folder}{self.privacy_group}.pdq.te"
+        extension = f"{indicator_type.lower()}.te"
```
jeberl: <3

jeberl: Could maybe even make this a global function because `.pdq.te` or `os.environ["THREAT_EXCHANGE_PDQ_FILE_EXTENSION"]` is used all over the code

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -366,50 +399,59 @@ def post_apply(self, updated: t.Dict = {}):
...
-                tags=new_tags,
-                pending_opinion_change=new_pending_opinion_change,
-            )
-            # TODO: Combine 2 update functions into single function
-            if metadata.update_tags_in_table_if_exists(table):
-                logger.info(
-                    "Updated Signal Tags in DB for indicator id: %s source: %s for privacy group: %d",
-                    row[1],
+                # TODO: Once Signal storage is generified, remove this
+                # conditional and handle multiple indicator_types.
```
jeberl: Not quite understanding the roll out plan to deprecate (HASH_PDQ). Happy to follow up offline

jeberl (changes requested):
Looks great! Just RC to get an answer on the plan for HASH_PDQ. Can you add to the summary.

Some nits also

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
schatten: It is inside a loop, so I'd rather warn and process the next signal_type if one exists.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
schatten: Same as below. Do not want to raise an error inside a loop. `supported_signal_types` is editable outside of fetcher.py.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
schatten: Yeah, once matcher is migrated. (Next PR), I'll do a deep cleanse.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
schatten: Not deprecating HASH_PDQ. You'll notice that we have a record type called `PDQSignalMetadata`.  That obviously can't contain MD5 signals.

2 solutions:
1. Either we make a generic record `SignalMetadata`
2. Create another MD5SignalMetadata

We have to select a solution (outside the scope of this set of PRs). And then update this here. This is a reminder to generify signal storage and handle MD5 and PDQ updates at some point.

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
schatten: That is a great idea!

schatten: Pushed in https://github.com/facebook/ThreatExchange/pull/728/commits/5bfd877dbe8478085a014e61b94c0631bd07733a

hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -128,15 +115,17 @@ def lambda_handler(event, context):
...
             continue
 
-        indicator_store = ThreatUpdateS3PDQStore(
+        indicator_store = ThreatUpdateS3Store(
             int(collab.privacy_group_id),
             api.app_id,
             te_data_bucket,
             config.s3_te_data_folder,
             config.data_store_table,
+            supported_signal_types=[VideoMD5Signal, PdqSignal],
```
BarrettOlson, line 124: ~nit: I don't understand having this be a param instead of hardcoding this here if we are also going to have:
```      
         indicator_types = [
            "HASH_VIDEO_MD5"
        ]  # hardcoding this because md5 currently points to HASH_MD5
```
in the other files anyway. Either made that block conditional on this or just have them be class values.

BarrettOlson (approved):
LGTM 
Some of the nits I'll save for the next set of PRs. This one has gotten messy due to the file change of `ThreatUpdateS3Store` and comments getting lost in the shuffle. Any bugs should be exposed in the stacked PRs so I'd vote to merge unless @jeberl feels otherwise. Thanks!

BarrettOlson (commented):
One note: ideally we should test this end 2 end for each of these PRs: I can work on re-adding a make command for this to make it easier

schatten:
> One note: ideally we should test this end 2 end for each of these PRs: I can work on re-adding a make command for this to make it easier

This PR actually breaks master if merged alone. I plan to use the integration branch to test e2e. I could have tried to make every PR be backwards compatible, but decided against it in the interest of time.

There will be one giant PR composed of reviewed PRs. Only the giant one will be merged. Giant PR may have tip commits that  I will get reviewed separately.

schatten:
Merged as part of integration branch #749

PR #729, opened 2021-07-27
threatexchange: Allow subclasses to build pdq indexes

schatten:
Summary
---------
In doing MD5 indexing, I am extending PDQIndex and TrivialSignalTypeIndex and giving them s3 storage. However, I want to retain the build method from the super classes.

This change will allow a building sub class instances.

Test Plan
---------
```
$ python -m black threatexchange
$ python -m py.test .
```


BarrettOlson (approved):
Stamped

PR #734, opened 2021-07-28
Update default branch name in Docs and Workflows

BarrettOlson:
Summary
---------

Changes before updating the default branch name from `master` -> `main`

The rest of this is reported to happen automagically (I have my doubts)

Test Plan
---------
Create this PR and see things break until I actually change the branch name.


.github/workflows/python-threatexchange-release.yaml
```diff
@@ -29,13 +29,13 @@ jobs:
         run: |
           python ./setup.py sdist bdist_wheel
       - name: Publish distribution to Test PyPI
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@main
```
schatten, line 32: We're sure the gh-action are also on main now?

schatten (approved):
Approving to unblock. Unsure if the `pypa/gh-action-pypi-publish` and `pypa/gh-action-pypi-publish`action is available on main.

.github/workflows/python-threatexchange-release.yaml
BarrettOlson: nope, wanted to see what errored, looks like `@release/v1` is prefered so I updated. https://github.com/pypa/gh-action-pypi-publish

PR #735, opened 2021-07-28
fix go CI workflow

BarrettOlson:
Summary
---------

Attempt number 1 

Test Plan
---------

Ran commands locally and saw the same error. Not overly familiar with go but this seemed to fix it.


schatten (approved):
Let's see what happens!

PR #737, opened 2021-07-28
Suppress new GCC 8 warnings in pdq/cpp

BarrettOlson:
Summary
---------

GCC 8 added new warnings that cause `make` of `pdq/cpp` to fail due to `-Werror`

I don't think it is worth digging into the example library that is causing the errors so for now I have the Makefile ignore those.

Test Plan
---------

Locally installed gcc 8.5 have saw `make` complete successfully (failing with warnings before) after adding the `-Wno-stringop-truncation -Wno-class-memaccess` flags.

### DIFF
--- a/pdq/cpp/Makefile
+++ b/pdq/cpp/Makefile
@@ -1,8 +1,8 @@
 # ================================================================
 CC=g++ -g -std=c++11
 IFLAGS=-I../.. -I.
-# -Wno-char-subscripts is because of CImg.h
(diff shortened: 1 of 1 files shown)


Dcallies:
I don't know enough to know how to fix this properly or what the tradeoffs might be, but lets get CI working again.

Dcallies (approved):
Oh wait, I have to hit approve.

PR #738, opened 2021-07-29
Fix Content ID Bug In Partner Bucket Submissions

jeberl:
Summary
---------

Partner Bucket uploads formerly created content IDs as `bucket-name/path/to/content/key.extension`. This led to two issues:

1. The matches Page in the API doesn't work correctly if content ID has `/` in it
2. The path could potentially lead to collisions with other submission methods if the same string was used

This diff changes the content IDs to a string like `bucket-name:path.to.content.key.extension` which adresses problem 1 and mitigates problem 2 since the second string is less likely to be random. We also decided to tell partners that should not use Partner Bucket Submissions in conjunction with another submission method.

I have also updated the documentation here to reflect this change:

>If you upload content from s3 buckets in this manner, ContentIDs will be generated automatically from the bucket and key. Specifically, if you upload a photo with file name `xkcd.jpg` to the `images/important/` directory of bucket `my-separate-aws-bucket`, the content ID will be:

>`my-separate-aws-bucket:images.important.xkcd.jpg`

>Because ContentIDs are unique across HMA, and the above string could be submitted as a ContentID for submissions via the UI or API, if you are using partner bucket uploads in prod we recommend only submitting content via this method unless you are absolutely confident you wont cause collisions.

Test Plan
---------

Upload photo to bucket in folder. Show that the new ContentID is correct (jesses-separate-aws-bucket:folder.200600.jpg) and that the matches page work:

<br class="Apple-interchange-newline">
Screen Shot 2021-07-29 at 12 57 50 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -118,12 +118,10 @@ def submit_content_request_from_s3_event_record(
     """
     Converts s3 event into a SubmitContentRequestBody object with a URL to the content
     """
-    # For partner buckets we use the full bucket name and key as the content ID to avoid collisions with
-    # existing objects
     bucket = record["bucket"]["name"]
     key = record["object"]["key"]
-    content_id = bucket + "/" + key
 
+    content_id = get_content_id_from_bucket_and_key(bucket, key)
     url = create_presigned_url(bucket, key, None, 3600, "get_object")
 
     return SubmitContentRequestBody(
@@ -133,6 +131,22 @@ def submit_content_request_from_s3_event_record(
         content_bytes_url_or_file_type=url,
         additional_fields=None,
     )
+
+
+def get_content_id_from_bucket_and_key(bucket: str, key: str) -> str:
+    """
+    For partner buckets, we use the full bucket name and key as the content ID to avoid
+    ContentID collisions between buckets.
+
+    There is still a threat of collision between partner bucket submissions and API/UI
+    submissions which is why we encourage (but dont enforce) partners to only use a single
+    submission method in production
+
+    The Matches UI page will work incorrectly if the content id includes / or ? so we replace
+    These chars with . when generating content ids.
+    """
+
+    return bucket + ":" + key.replace("/", ".").replace("?", ".")
 
 
 class SignalSourceType(t.TypedDict):
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -7,6 +7,7 @@ import json
 import typing as t
 from apig_wsgi import make_lambda_handler
 from bottle import response, error
+from uuid import UUID
 
 from hmalib.common.logging import get_logger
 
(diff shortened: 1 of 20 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -135,6 +133,22 @@ def submit_content_request_from_s3_event_record(
...
+
+    There is still a threat of collision between partner bucket submissions and API/UI
+    submissions which is why we encourage (but dont enforce) partners to only use a single
+    submission method in production
+
+    The Matches UI page will work incorrectly if the content id includes / or ? so we replace
+    These chars with . when generating content ids.
+    """
+
+    return bucket + ":" + key.replace("/", ".").replace("?", ".")
```
schatten, line 149: When actions are called for s3 objects matched, we will have to reverse this.

How about fixing the matches UI page instead? There is nothing inherently unsupportable about content_ids containing special characters.

schatten: Also, reversing might become impossible if you are replacing with "." because file names will possibly contain ".".

The use case I am thinking of:
- partner processes a bucket.
- receives some matches on their webhook. webhook mentions content_id as partner_bucket:directory1.directory2.filename.filename.png
- how do they map back to the actual s3 key if we are 'destroying' the s3 key and not storing a copy

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -118,12 +118,10 @@ def submit_content_request_from_s3_event_record(
...
-    # For partner buckets we use the full bucket name and key as the content ID to avoid collisions with
-    # existing objects
     bucket = record["bucket"]["name"]
     key = record["object"]["key"]
-    content_id = bucket + "/" + key
 
+    content_id = get_content_id_from_bucket_and_key(bucket, key)
     url = create_presigned_url(bucket, key, None, 3600, "get_object")
 
     return SubmitContentRequestBody(
```
schatten, line 127: Would love to see the full S3 URL (s3://...) being added as an additional field. That way we are not losing much information in the process.

schatten (changes requested):
Would much rather this gets solved at the Matches UI, the source of the problem.

Blocking: Destroying the S3 key means actions become less valuable. Partners would need to understand the inner details to know how to go from webhook call to original S3 key.

jeberl:
> Would much rather this gets solved at the Matches UI, the source of the problem.

The "/" and "?" constraint, according to @BarrettOlson, is not just on the Matches page but lots of different API endpoints and would be a pretty large refactor. its unclear if this refactor is necessary at this point so I'd rather more of a bandaid solution

> Blocking: Destroying the S3 key means actions become less valuable. Partners would need to understand the inner details to know how to go from webhook call to original S3 key.

Would including original s3 url in the additional_fields of the Content address this?

> Also, reversing might become impossible if you are replacing with "." because file names will possibly contain ".".

Could replace with ":" instead. Not perfect but probably addresses the most common issue. Or just encourage people to use the url in additional_fields

schatten:
> > Blocking: Destroying the S3 key means actions become less valuable. Partners would need to understand the inner details to know how to go from webhook call to original S3 key.
> 
> Would including original s3 url in the additional_fields of the Content address this?

If at present, webhooks also add `additional_fields` to the payload, then yes. But if we are doing that, then the `content_id` is only providing a uniqueness guarantee. Because you can't do a string transform to go from hma content_id to partner s3_key. 

In that case, why not use a uuid.uuid4() instead? They are proven [1] to have sooper low chances of collisions and suffice for this case.

1: http://www.h2database.com/html/advanced.html#uuid

BarrettOlson:
> > Would including original s3 url in the additional_fields of the Content address this?
>
> In that case, why not use a uuid.uuid4() instead? They are proven [1] to have sooper low chances of collisions and suffice for this case. And would require no special handling to be useful as content_ids across the board.

Agree with recommendation. Maybe `f"{uuid.uuid4()}-<filename between / and .xyx>"` to make debugging a little easier/traceable?

jeberl:
Changed to uui4 + key per @BarrettOlson's suggestion. Re-tested

Screen Shot 2021-08-09 at 2 30 40 PM
tested:

jeberl:
Diff got overly complex with rebase so abandoning in favor of 746

PR #742, opened 2021-08-03
Send additional fields in Action responses

jeberl:
Summary
---------

We currently accept "AdditionalFields" when partners submit content but don't do anything with the fields they provide. This PR includes additional_fields in Webhooks when sent.

A future PR will add the ability to filter by additional_fields in actionRules 

Note: Also did some clean up of imports in files I touched

Test Plan
---------

1. Create Action and ActionRule that sends all TE matches to webhook https://webhook.site/81bf390d-d2e3-4757-989d-9822a81d85dd/4abcd206-e1cd-40f2-8514-1eb5511abe4f/1
1.  Submit matching Content to HMA with 2 additional_fields set 
2. Confirm that additional fields are included in Webhook response here: https://webhook.site/#!/81bf390d-d2e3-4757-989d-9822a81d85dd/4abcd206-e1cd-40f2-8514-1eb5511abe4f/1

Screen Shot 2021-08-03 at 3 54 04 PM
Screen Shot 2021-08-03 at 3 54 16 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/message_models.py
+++ b/hasher-matcher-actioner/hmalib/common/message_models.py
@@ -16,6 +16,7 @@ from hmalib.common.classification_models import (
     PendingOpinionChange,
 )
 from hmalib.common.evaluator_models import ActionLabel, ActionRule
+from hmalib.common.content_models import ContentObject
 from hmalib.common.aws_dataclass import HasAWSSerialization
 from hmalib.common.image_sources import S3BucketImageSource
 from hmalib.common.logging import get_logger
@@ -84,12 +85,16 @@ class ActionMessage(MatchMessage):
     action_label: ActionLabel = ActionLabel("UnspecifiedAction")
     action_rules: t.List[ActionRule] = field(default_factory=list)
 
+    # from content
+    additional_fields: t.List[str] = field(default_factory=list)
+
     @classmethod
     def from_match_message_action_label_and_action_rules(
         cls,
         match_message: MatchMessage,
         action_label: ActionLabel,
         action_rules: t.List[ActionRule],
+        additional_fileds=t.List[str],
     ) -> "ActionMessage":
         return cls(
             match_message.content_key,
@@ -97,6 +102,7 @@ class ActionMessage(MatchMessage):
             match_message.matching_banked_signals,
             action_label,
             action_rules,
+            additional_fileds,
         )
 
 
--- a/hasher-matcher-actioner/hmalib/common/message_models.py
+++ b/hasher-matcher-actioner/hmalib/common/message_models.py
@@ -86,7 +86,7 @@ class ActionMessage(MatchMessage):
     action_rules: t.List[ActionRule] = field(default_factory=list)
 
     # from content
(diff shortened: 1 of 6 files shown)


hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -78,15 +86,29 @@ def lambda_handler(event, context):
...
-        action_label_to_action_rules = get_actions_to_take(match_message, action_rules)
+        submitted_content = ContentObject.get_from_content_id(
+            config.dynamo_db_table,
+            match_message.content_key,
+        )
+
+        # Convert to dict becuase JSON can't serialize Sets for Action Message
+        # plus we think of these fields as key: value anyways
+        additional_fields: t.Dict[str, str] = dict(
+            field.split(":", 1) for field in submitted_content.additional_fields
```
BarrettOlson, line 97: I'm worried about adding this dependency without changes to the API, nothing requires there be a `:` as part of this set nor that the prefix have a unique property. I viewed these fields as more of a set of tags/labels on a piece of content not a key-value pair, this is a disconnect between the UI and the API/underlying data model we should fix. 

Also, unfortunately as it is written I think this will break the action evaluator for content without any additional_fields

jeberl: I see. I thought the Submit API took a dict but I'm seeing now its a List. I'll change this to convert the set to a list as well.

I do think we should change the UI to match this data model

BarrettOlson: Done: https://github.com/facebook/ThreatExchange/pull/745

jeberl: @BarrettOlson Updated to make a list of strings

hasher-matcher-actioner/hmalib/common/message_models.py
```diff
@@ -84,19 +85,24 @@ class ActionMessage(MatchMessage):
     action_label: ActionLabel = ActionLabel("UnspecifiedAction")
     action_rules: t.List[ActionRule] = field(default_factory=list)
 
+    # from content
+    additional_fields: t.List[str] = field(default_factory=list)
+
     @classmethod
-    def from_match_message_action_label_and_action_rules(
+    def from_match_message_action_label_action_rules_and_additional_fields(
```
BarrettOlson, line 92: nit/temporarily ignorable: this has a smell to me (beyond nit on the name), taking an object you inherit from directly and accessing its fields manually to initialize feels off. Flagging as this is making me question the interface/abstraction but that is likely beyond the scope of this PR.

BarrettOlson (approved):
Stamped. I don't love this serialize and send everything choice for the long term but like I noted for my nit inline that refactor is beyond the scope of the changes this PR.

hasher-matcher-actioner/hmalib/common/message_models.py
jeberl: Yep I agree. Open to refactor later

PR #743, opened 2021-08-03
Use Additional Fields In ActionRules

jeberl:
## THIS IS A STCKED PR. Only review commit a25c0c0

The other commits are included in PR #742

Summary
---------

This diff enables users to add conditions to their action rules based on Additional Fields included on the content. A use case for this might be a different queue for mobile uploads.

To enable this, we create SubmittedContentClassifications for each AdditionalField. We then add a new Classification type to the UI

Additional Fields are submitted as key-value pairs. To specify these in ActionRules, users must write `<key>:<value>`. I will add this to the documentation.

I considered adding validation rules to ensure the value contains a colon but decided against as this may change in the future and users will likely have to look at the documentation anyway to use this feature

Test Plan
---------

1. Create a new ActionRule that looks for Submitted Content classified as `mobile_upload:true`
2. Submit a photo to HMA with Additional Fields `[{key: mobile_upload, value:true}, {key: another field, value : also true}]`
3. Watch webhook get called: https://webhook.site/#!/81bf390d-d2e3-4757-989d-9822a81d85dd/c28766d5-ddb5-47d8-9d40-423221ba47c8/1
4. Submit a photo to HMA with Additional Fields `[{key: mobile_upload, false}]`
5. Watch webhook NOT get called: https://webhook.site/#!/81bf390d-d2e3-4757-989d-9822a81d85dd/c28766d5-ddb5-47d8-9d40-423221ba47c8/1

Screen Shot 2021-08-03 at 6 12 35 PM

Screen Shot 2021-08-03 at 6 13 00 PM

Screen Shot 2021-08-03 at 6 12 22 PM

Screen Shot 2021-08-09 at 1 25 02 PM

Screen Shot 2021-08-09 at 1 25 16 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/classification_models.py
+++ b/hasher-matcher-actioner/hmalib/common/classification_models.py
@@ -51,6 +51,11 @@ class BankedContentIDClassificationLabel(ClassificationLabel):
 
 
 @dataclass(unsafe_hash=True)
+class SubmittedContentClassification(ClassificationLabel):
+    key: str = field(default="SubmittedContent", init=False)
+
+
+@dataclass(unsafe_hash=True)
 class ActionLabel(Label):
     key: str = field(default="Action", init=False)
 
--- a/hasher-matcher-actioner/hmalib/common/classification_models.py
+++ b/hasher-matcher-actioner/hmalib/common/classification_models.py
@@ -1,13 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
-from hmalib.common.signal_models import PendingOpinionChange
-import hmalib.common.config as config
-import json
-import typing as t
-
-from dataclasses import dataclass, field, fields
+from dataclasses import dataclass, field
 from hmalib.common.logging import get_logger
-from requests import get, post, put, delete, Response
 
 logger = get_logger(__name__)
 
@@ -51,7 +45,7 @@ class BankedContentIDClassificationLabel(ClassificationLabel):
 
 
 @dataclass(unsafe_hash=True)
-class SubmittedContentClassification(ClassificationLabel):
+class SubmittedContentClassificationLabel(ClassificationLabel):
     key: str = field(default="SubmittedContent", init=False)
 
 
--- a/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
@@ -41,7 +41,7 @@ class ActionEvaluatorConfig:
 
     actions_queue_url: str
     sqs_client: SQSClient
-    dynamdo_db_table: Table
+    dynamo_db_table: Table
 
     @classmethod
     @lru_cache(maxsize=None)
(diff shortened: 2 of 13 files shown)


hasher-matcher-actioner/hmalib/common/classification_models.py
```diff
@@ -50,6 +50,11 @@ class BankedContentIDClassificationLabel(ClassificationLabel):
     key: str = field(default="BankedContentIDClassification", init=False)
 
 
+@dataclass(unsafe_hash=True)
+class SubmittedContentClassification(ClassificationLabel):
```
BarrettOlson, line 54: namings is hard but we should still try to have consistency here. It looks like all the other `ClassificationLabel` have that in their name not just `Classification`. Also, this is used to check against `additional_fields` yes? can we include the word 'fields` in the name at a minimum?

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
```diff
@@ -104,16 +125,27 @@ def lambda_handler(event, context):
...
+
+    logger.info(
+        "Adding SubmittedContentClassification Labels: %s", content_classifications
+    )
+
     for banked_signal in match_message.matching_banked_signals:
         for action_rule in action_rules:
             if action_rule_applies_to_classifications(
-                action_rule, banked_signal.classifications
+                action_rule, banked_signal.classifications | content_classifications
```
BarrettOlson, line 148: nit: the typing was not clear to me and I was confused by the `|`. I prefer we use `.union()` for future readers unless you feel strongly about it

hasher-matcher-actioner/webapp/src/components/settings/ActionRulesTableRow.tsx
```diff
@@ -94,19 +94,24 @@ export default function ActionRulesTableRow({
...
           default:
             ret += ` ${classification.classificationType}`;
             break;
         }
 
         if (classification.equalTo) {
           ret +=
-            classification.classificationType === 'Classification'
+            classification.classificationType === 'Classification' ||
+            classification.classificationType === 'SubmittedContent'
```
BarrettOlson, line 108: Could you include a screenshot of this change if it is visbal on the setting UI? I have the same nit about naming as well. This one is worse imo because it is a Classification just a different kind right?

BarrettOlson (approved):
Stamping to unblock but please do not merge without first updating the name to be consistent with other `ClassificationLabel`s  and reflect the value used for actioning both in models and the UI. 
Strong preference for an overly verbose name than a non-specific one.

hasher-matcher-actioner/hmalib/lambdas/actions/action_evaluator.py
jeberl: Lol. Syntactic sugar isn't so sugary if no one knows what's happening.

Will change

hasher-matcher-actioner/webapp/src/components/settings/ActionRulesTableRow.tsx
jeberl: Added screenshot.

Discussed offline

hasher-matcher-actioner/hmalib/common/classification_models.py
BarrettOlson: @jeberl, why did we decided not to add `field` here? I was under the impression we were going to but it is possible I am misremembering.

PR #745, opened 2021-08-04
[hma][ui] Update additional fields in submit form to reflect API backend.

BarrettOlson:
Summary
---------

Additional fields on content objects are currently a set of strings. If we want to make them a key-value pair in the future we can however the UI should match the current backend as it exists now to avoid confusion. 
  
Test Plan
---------
Did a full submit to verify functionality. 
Old form: 
Screen Shot 2021-08-04 at 2 16 33 PM

Form now looks like this:
Screen Shot 2021-08-04 at 2 12 07 PM


jeberl (approved):
Thanks for knocking this out so quick!

PR #746, opened 2021-08-09
Fix Content ID Bug In Partner Bucket Uploads

jeberl:
See #738 for context and test plan. Will be easier to review with a new PR and rebase

Test Plan
---------

```
```
3. `make upload_docker && terraform -chdir=terraform apply`
5. Through logs watch them be processed by the api_root lambda, added to the sqs queue, and then loaded and hashed by the pdq_hasher: image

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -7,6 +7,7 @@ import json
 import typing as t
 from apig_wsgi import make_lambda_handler
 from bottle import response, error
+from uuid import uuid4
 
 from hmalib.common.logging import get_logger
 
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -117,21 +118,33 @@ def submit_content_request_from_s3_event_record(
...
     """
-    # For partner buckets we use the full bucket name and key as the content ID to avoid collisions with
-    # existing objects
-    bucket = record["bucket"]["name"]
-    key = record["object"]["key"]
-    content_id = bucket + "/" + key
+    bucket: str = record["bucket"]["name"]
+    key: str = record["object"]["key"]
+
+    readable_key = key.replace("/", ".").replace("?", ".").replace("&", ".")
```
BarrettOlson, line 136: nit: I feel like the full path is a bit overkill here, maybe just the name after all the prefixes?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -117,21 +118,33 @@ def submit_content_request_from_s3_event_record(
 ) -> SubmitContentRequestBody:
     """
     Converts s3 event into a SubmitContentRequestBody object with a URL to the content
+
+    For partner bucket uploads the content IDs are unique and human understandable but
```
BarrettOlson, line 122: nit: human understandable seems like a weird phrase here. Depends on completely on file name conventions, maybe just "Part of the original file name is include in ID to help with possible debugging but do not depend on this value, actual reversible look up should refer to the objects additional fields for a URL corresponding to the bucket and object name."

BarrettOlson (approved):
Looks fine to me. I think this is a good patch to avoid having `/`s in content Id. 

I am going to have to do some refactor here as I break up the submit API anyway. If you can update the docs somewhere at least on how to test this, that would be helpful.

PR #748, opened 2021-08-10
[hma][ui] Remove old direct upload option

### LINKED ISSUE #598: [HMA] Upload pages do not work on firefox
**Repro:**
1. Open the submit page on firefox eg. https://d3jelb9xcmkers.cloudfront.net/submit
2. Open the console using <cmd>+<shift>+<c> on mac
2. Select a file to submit

**Stacktrace:**
```
Uncaught TypeError: e.nativeEvent.path is undefined
    handleInputChangeUpload SubmitContent.jsx:55
    React 11
    unstable_runWithPriority scheduler.production.min.js:18
    React 3
SubmitContent.jsx:55:17
    handleInputChangeUpload SubmitContent.jsx:55
    React 11
    bind_applyFunctionN self-hosted:1371
    Xt self-hosted:1334
    unstable_runWithPriority scheduler.production.min.js:18
    React 3
    bind_applyFunctionN self-hosted:1371
    Jt self-hosted:1334
```

**Screenshot:**
Screen Shot 2021-05-19 at 13 13 55

BarrettOlson:
Summary
---------

Having both an 'upload' and a 'direct upload' option is confusing and unnecessary for the submit UI.  It is still possible the  soon to be renamed 'direct submit' endpoint could be useful outside of the UI context for sending bytes directly in one request, so leaving that for now. It fails for large images anyway...

One of  the reasons I had kept it around up until now was actually because of #598 so folks could still ~demo HMA using Firefox. As a part of this removal I also tried to fix #598.  Which was easier to fix now that we only have one upload flow as I was able to just remove the need to get the path which is not supported the same way in all browsers. 

Test Plan
---------

Made sure the remaining 'upload' flow works for both Firefox and Chrome. If you'd like a demo video and/or screenshots let me know.

### DIFF
--- a/hasher-matcher-actioner/webapp/src/Api.jsx
+++ b/hasher-matcher-actioner/webapp/src/Api.jsx
@@ -158,27 +158,6 @@ export async function submitContentPostURLUpload(
   return result;
 }
 
-export async function submitContentDirectUpload(
-  submissionType,
-  contentId,
-  contentType,
-  content,
-  additionalFields,
-) {
-  const fileReader = new FileReader();
-  fileReader.onload = () => {
-    const fileContentsBase64Encoded = encode(fileReader.result);
-    apiPost('/submit/', {
-      submission_type: submissionType,
-      content_id: contentId,
-      content_type: contentType,
-      content_bytes_url_or_file_type: fileContentsBase64Encoded,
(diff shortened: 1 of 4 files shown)


schatten (approved):
LGTM!

PR #757, opened 2021-08-15
`ContentObject` queries no longer require `content_type` #756

### LINKED ISSUE #756: [hma] Content storage and retrieval should not require content type
Given a content_id, we only expect a single content_object for that type. The SK "content_type#<content_type>" means we need to have content_type as well as content_type. This is not readily available in later stages 'action-eval' and 'action-perform' because pipeline records (HashRecord, MatchRecord) store the SignalType <pdq, md5> rather than the original content_type.

I'm proposing we replace the dynamic SK with something static like "content_type#universal". This way, the ContentObject.* APIs can stop requiring content_types.

schatten:
### Fixes #756 

Summary
---------
ContentObject queries used to require content_type. This is problematic because
a) It may be taken to mean `content_type` + `content_id` is unique. But we would like content_id to be globally unique.
b) later pipeline stages (actioning) do not have access to content_type. With additional data now being required in actioning, the content_type was impossible to obtain.

This PR removes the need for content_type in ContentObject.** queries. This will also allow the UI to link to a single page for a content_id and then support different previews based on the content_type returned by the API.

Test Plan
---------
Ran mypy, black and py.test.

Deployed docker image and webapp after tainting.

1. Verified that a URL submission flow works. (including calling webhook)
2. Verified that the content details page show last_submitted at and all APIs return responses (image failed because of a CORS issue, but need to figure that out separately).

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/content_models.py
+++ b/hasher-matcher-actioner/hmalib/common/content_models.py
@@ -135,9 +135,9 @@ class ContentObject(ContentObjectBase, JSONifiable):
     is likely better.
     """
 
-    CONTENT_TYPE_PREFIX = "content_type#"
+    CONTENT_STATIC_SK = "#content_object"
 
-    # Used to create the SK in DDB.
+    # Is this a photo, a video, a piece of text? etc.
     content_type: t.Type[ContentType]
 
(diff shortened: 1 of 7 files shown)


BarrettOlson (approved):
LGTM

PR #758, opened 2021-08-16
[hma][hotfix] fix non-url submissions

BarrettOlson:
Summary
---------

#749 made a noble attempt to keep `pdq_hasher` and `pdq_matcher` working while adding the unified versions however a few things were missed/not yet handled.

This PR fixes those errors and migrates recent changes to the submit API that were skipped (likely because #750 merged after the implementation and review of the MD5 additions in 749). 
 
More clean up and migration to the architecture changes that #749 made are needed in follow ups but this should at least get prod/main working e2e again for all submission methods.

Test Plan
---------
Checked that both `upload` and `from url` submit work using `make dev_test_instance` for upload submissions e2e and `./scripts/submit_s3_bucket ...` for url submissions and well as both flows on the UI


schatten (approved):
Thanks!

image

PR #759, opened 2021-08-17
[hma][scripts] refactor helper script libs and add hma_shell cli

BarrettOlson:
Summary
---------

The scripts are a ~hodgepodge of tools at the moment. This is an attempt at starting to standardize around `hma_script_utils.py` and clean up how the values needed to access HMA are passed to the testing scripts. 

It also added a prototype of a shell wrapper for HMA utils for interactive  testing.

Test Plan
---------

```
export HMA_USER_PWD='<redacted>' 
$ python3 scripts/hma_script_utils.py 
Manual Test of API Request Methods:
{'message': 'Hello World, HMA'} 

$ python3 scripts/hma_client_lib.py
This simple tests should take a little over 2 minutes to complete (due to sqs timeout).

Added configurations to HMA instance for test
Submitted content_id e2e-test-2021-08-17-d9a51f00-3d82-469d-a03c-6b6e4cb23e9e
Waiting for action history of submitted content_id
Checking every 5 seconds; maximum tries = 25
Success action event found in history!
Removed actions configurations used in test
Test completed in 161 seconds
```

hma_shell is still a work in program, will update docs with usage once functionality is expanded to a level that is useful.
Right now it can be used to run the same test as `hma_client_lib` and query the HMA API for its set of:
- action configs (actions and action_rules)
- dataset configs
- matches 

And can query for the following using a content_id:

- hash_details
- matches
- action_history

Usage:
```
$ python3 scripts/hma_shell 
Welcome! Type help or ? to list commands.

> ?

Documented commands (type help <topic>):
========================================
action_history_for_id  dataset_configs      help            refresh       
action_rules           exit                 matches         run_basic_test
actions                hash_details_for_id  matches_for_id
```


hasher-matcher-actioner/scripts/hma_shell
```diff
@@ -0,0 +1,165 @@
+#! /usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Prototype of shell wrapper for HMA utils for interactive manual testings
+
+```
+python3 scripts/hma_shell --pwd <password-of-user>
```
schatten, line 8: I tried running the shell when testing my MD5 changes. It was not clear to me which password you are talking about here.

hasher-matcher-actioner/scripts/hma_script_utils.py
```diff
@@ -269,51 +279,167 @@ def delete_action_rule(
...
+    return json.loads(out)
 
-    # Testing HasherMatcherActionerAPI
-    #   Since the API requries a deployed instance the majority of
-    #   testing needs to be done manually. See below.
 
-    # Add init values for the api:
-    # `os.environ.get(...)` will enable us pass values from `terraform output` in sh
+def get_terraform_outputs_from_file(
+    path: str = "/workspaces/ThreatExchange/hasher-matcher-actioner/tmp.out",
```
schatten, line 292: Who would be populating this file?

schatten (approved):
Love the steps you are taking here. 

For promoting reuse and creating a single entrypoint into all scripts, we could make scripts into its own python module. That way, you can create abstractions for script-type commands.

eg. AbstractScriptThatNeedsRefreshToken, AbstractScriptThatUsesLambdaEnvironmentVariables.

We can also move the makefile commands that require pre-existing configs as scripts that are interactive.

This will promote a unified script invocation, simpler for partners eg.
```
$ hmacli init
$ hmacli deploy
$ hmacli upgrade
$ hmacli benchmark
$ hmacli testshell
```

Etc. All of these could potentially use the same option and argument definitions which would greatly improve partner engineer experience and simplify the readme. If we can get rid of the `make` dependency, even better!

Obviously, this is not a one PR ask. I'm suggesting you start laying the groundwork (__init__.py, BaseHMACLISubCommand etc. in forthcoming touches to the scripts.

What we have here is good to land!

hasher-matcher-actioner/scripts/hma_shell
BarrettOlson: fair, it is meant to be the password given to a pervious call to hma_shell that used, `--create_user`

hasher-matcher-actioner/scripts/hma_script_utils.py
BarrettOlson: yeah, that path specifically was to test that it works. 
The idea was that someone would want to use the CLI somewhere besides where they deployed hma using tf (i.e. on an ec2 instance for example)

PR #762, opened 2021-08-18
[hma][api] add get matches for raw hash endpoint

BarrettOlson:
Summary
---------

Instead of submitting content to HMA right away, if the user just want to see if a signal they have ~would~ match they can now submit that signal to `/matches/for-hash/` to get what signal on the live indexes should have found.

Test Plan
---------

Screen Shot 2021-08-18 at 4 07 25 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
@@ -286,6 +286,8 @@ def get_matches_api(
     def for_hash(request: MatchesForHashRequest) -> MatchesForHashResponse:
         """
         For a given hash/signal check the index(es) for matches and return the details
+        NOTE: currently metadata returned will not be written to the dynamodb table
+        unlike in the case of a pipeline match based on submissions.
         """
         signal_type = None
         if request.signal_type == "pdq":
--- a/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
@@ -2,6 +2,8 @@
 
 import bottle
 import datetime
+import functools
+import dataclasses
 
 from dataclasses import dataclass, asdict
 from mypy_boto3_dynamodb.service_resource import Table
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -240,4 +282,32 @@ def request_signal_opinion_change() -> ChangeSignalOpinionResponse:
...
+
+        matches = _get_matcher(indexes_bucket_name).match(
+            signal_type, request.signal_value
+        )
+
+        match_objects = []
+
+        for match in matches:
+            match_objects.extend(
+                Matcher.get_metadata_objects_from_match(signal_type, match)
```
schatten, line 308: AFAIK, this will not write signal metadata. Is that a conscious choice?

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -159,17 +160,35 @@ def write_signal_if_not_found(
         their own store. Perhaps the API could be useful when building local
         banks. Who knows! :)
         """
+        for signal in cls.get_metadata_objects_from_match(signal_type, match):
+            signal.write_to_table_if_not_found(table)
+
+    @classmethod
+    def get_metadata_objects_from_match(
```
schatten, line 167: ❤️

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -218,3 +237,10 @@ def publish_match_message(
         )
 
         sns_client.publish(TopicArn=topic_arn, Message=match_message.to_aws_json())
+
+    def flatten_matches(
```
schatten, line 241: nit: unused.

schatten (approved):
LGTM. Except for whether we want to write signal metadata for these requests. Can lean either way, but need that to be a conscious, documented decision.

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
BarrettOlson: for now yes, long term we probably will want to but for the time being I'm preferring to leave hitting this endpoint side effect free.

BarrettOlson:
> LGTM. Except for whether we want to write signal metadata for these requests. Can lean either way, but need that to be a conscious, documented decision.

Electing to not record for now, updated the endpoint methods docstring to document the choice.

PR #764, opened 2021-08-18
[hma][small] Update `post_url` to `put-url` to be clearer and match convention

BarrettOlson:
Summary
---------

Small change. The submit API for upload returns a signed url to the client. Code calls this a 'post_url' but the url is actually designed to be used in a put request. Also I used an `_` instead of a `-` in the path (we agreed to use the dash awhile ago and should stick to it). 

Test Plan
---------

Made sure both manual submission and `python scripts/hma_client_lib.py` still worked.


schatten (approved):
I can't begin to tell you how much I love janitorial stuff like this!

PR #767, opened 2021-08-20
[hma] add resubmit arg to submission requests and require it if content_id is found.

BarrettOlson:
Summary
---------

Right now the system pipeline depends on `content_id` given by the client. 
Submitting again with the same `content_id` is possible for things like testing but should not be the default.

Additionally trying to submit different content for the same id is likely to mess with the system and result in undefined behavior.  We have some protections to avoid this but more couldn't hurt.

Once merged if a client wants to submit a content object with the same id they must include `force_resubmit=True` in their request. 

I also added the ability to add this flag on the submit UI using a textbox and improve our error handling. 
- full disclosure I took a short cut on the UI work because it is annoying to work in jsx files right now. I will clean up when changing the file to .tsx (soon(tm)).

Test Plan
---------
Made sure regular submission still works and the resubmit for the same submission media and method went through on the UI.
 
Screen Shot 2021-08-20 at 12 02 19 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/content_models.py
+++ b/hasher-matcher-actioner/hmalib/common/content_models.py
@@ -215,6 +215,14 @@ class ContentObject(ContentObjectBase, JSONifiable):
         )
 
     def write_to_table_if_not_found(self, table: Table) -> bool:
+        """
+        Write operations for this object need to be special cased (to avoid overwritting)
+        Therefore we do not implement `to_dynamodb_item` however basically the body of that
+        method is used here
+
+        Returns false if a content object with that Id is already present
+        and does not write to table. True is write was successful.
+        """
         try:
             table.put_item(
                 Item={
--- a/hasher-matcher-actioner/hmalib/common/content_models.py
+++ b/hasher-matcher-actioner/hmalib/common/content_models.py
@@ -187,20 +187,8 @@ class ContentObject(ContentObjectBase, JSONifiable):
     def write_to_table(self, table: Table):
         """
         Write operations for this object need to be special cased (to avoid overwritting)
-        Therefore we do not implement `to_dynamodb_item`
-
-        If you're curious it would ~look like this:
-        def to_dynamodb_item(self) -> dict:
-            return {
-                "PK": self.get_dynamodb_content_key(self.content_id),
-                "SK": self.get_dynamodb_content_type_key(),
-                "ContentRef": self.content_ref,
-                "ContentRefType": self.content_ref_type,
-                "AdditionalFields": self.additional_fields,
-                "SubmissionTimes": [s.isoformat() for s in self.submission_times],
-                "CreatedOn": self.created_at.isoformat(),
-                "UpdatedAt": self.updated_at.isoformat(),
-            }
+        Therefore we do not implement `to_dynamodb_item` however basically the body of that
+        method is used in this class's impl of `write_to_table_if_not_found`
         """
         # put_item does not support UpdateExpression
         table.update_item(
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -72,6 +72,7 @@ class SubmitRequestBodyBase(DictParseable):
     content_id: str
     content_type: t.Type[ContentType]
     additional_fields: t.Optional[t.List]
+    resubmit: bool = False
```
schatten, line 75: nit/personal-pref: do you want to make this more explicit like `force_resubmit`?

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -226,6 +214,37 @@ def to_dynamodb_item(self) -> dict:
...
+                    if self.additional_fields
+                    else {self.ADDITIONAL_FIELDS_PLACE_HOLDER},
+                    "SubmissionTimes": [s.isoformat() for s in self.submission_times],
+                    "CreatedAt": self.created_at.isoformat(),
+                    "UpdatedAt": self.updated_at.isoformat(),
+                },
+                ConditionExpression="attribute_not_exists(PK) AND attribute_not_exists(SK)",
+            )
+        except ClientError as client_error:
+            # boto3 exception handling https://imgflip.com/i/5f5zfj
```
schatten, line 236: Look what we've been reduced to! I actually followed the link!

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -226,6 +214,37 @@ def to_dynamodb_item(self) -> dict:
...
+            if (
+                client_error.response.get("Error", {"Code", "Unknown"}).get(
```
schatten, line 238: Shouldn't it be `{"Code": "Unknown"}`? You're creating a set here, not a dict.

hasher-matcher-actioner/hmalib/common/content_models.py
```diff
@@ -226,6 +214,37 @@ def to_dynamodb_item(self) -> dict:
             },
         )
 
+    def write_to_table_if_not_found(self, table: Table) -> bool:
```
schatten, line 217: docstring? How is the return value to be interpreted?

hasher-matcher-actioner/hmalib/lambdas/api/submit.py
```diff
@@ -233,9 +244,15 @@ def get_submit_api(
     submit_api = bottle.Bottle()
     s3_bucket_image_source = S3BucketContentSource(image_bucket, image_prefix)
 
+    def _content_exist_error(content_id: str):
+        return bottle.abort(
+            500,
```
schatten, line 249: Not that we are super sensitive about status codes, but this is a 400 Bad Request or a 409 Conflict. Upto you, but a 4xx is more appropriate here.

hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
```diff
@@ -75,26 +76,39 @@ export default function SubmitContent() {
...
       submitContentViaPutURLUpload(
         inputs.contentId,
         inputs.contentType,
         packageAdditionalFields(),
         inputs.content.raw,
-      ).then(() => {
-        setSubmitting(false);
-        setSubmittedId(inputs.contentId);
-      });
+        inputs.resubmit,
```
schatten, line 86: nit: Does not exist in FORM_DEFAULTS. That means inadvertent type coercion of undefined into false.

schatten (approved):
General inline nits

blocking: Use of a set in content_models.py seems out of place. Inline comment for details.

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: you did this to yourself: https://github.com/facebook/ThreatExchange/blame/main/hasher-matcher-actioner/hmalib/models.py#L39

hasher-matcher-actioner/hmalib/common/content_models.py
BarrettOlson: it's worse than you'd think, the key is a set...

hasher-matcher-actioner/webapp/src/pages/SubmitContent.jsx
BarrettOlson: good catch! thanks!

PR #768, opened 2021-08-20
Combine data from all datafiles for privacy group in fetcher.load_state()

schatten:
Summary
---
During the MD5 feature additions, the fetcher flow was tested, but the syncer
wasn't. It was failing because load_state() called only during sync updates did
not yet support multiple files per privacy group.

This has been fixed. When load_state() is called, it gets data for all datafiles
for a privacy group.

Test Plan
---
Using 
```python
import os

os.environ[
    "THREAT_EXCHANGE_DATA_BUCKET_NAME"
] = "..."
os.environ["THREAT_EXCHANGE_DATA_FOLDER"] = "threat_exchange_data/"
os.environ["CONFIG_TABLE_NAME"] = "..."
os.environ["DYNAMODB_DATASTORE_TABLE"] = "..."
os.environ[
    "THREAT_EXCHANGE_API_TOKEN_SECRET_NAME"
] = "..."

from hmalib.lambdas.fetcher import lambda_handler

lambda_handler(None, None)
```

The fetcher and syncer was run locally. The syncer run fails with
`AttributeError: 'ThreatUpdateS3Store' object has no attribute 'data_s3_key'`
before the fix and successfully runs after.

First run produced the requisite files for privacy groups on my instance. On
adding more hashes to the privacy group "HMA Test Collaboration (Superset)", the
second fetcher run added all hashes to the datafiles.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/s3_adapters.py
+++ b/hasher-matcher-actioner/hmalib/common/s3_adapters.py
@@ -16,6 +16,7 @@ from dataclasses import dataclass, field
 import typing as t
 
 from botocore.errorfactory import ClientError
+from mypy_boto3_s3 import Client as S3Client
 from mypy_boto3_s3.service_resource import Bucket
 from threatexchange import threat_updates as tu
 from threatexchange.cli.dataset.simple_serialization import HMASerialization
@@ -209,11 +210,14 @@ class ThreatUpdateS3Store(tu.ThreatUpdatesStore):
     ThreatUpdatesStore, but stores files in S3 instead of local filesystem.
     """
 
+    CHECKPOINT_SUFFIX = ".checkpoint"
+
     def __init__(
         self,
         privacy_group: int,
         app_id: int,
-        s3_bucket: Bucket,  # Not typable?
+        s3_client: S3Client,
+        s3_bucket_name: str,
         s3_te_data_folder: str,
         data_store_table: str,
         supported_signal_types: t.List[SignalType],
@@ -221,12 +225,26 @@ class ThreatUpdateS3Store(tu.ThreatUpdatesStore):
         super().__init__(privacy_group)
         self.app_id = app_id
         self._cached_state: t.Optional[t.Dict] = None
-        self.s3_bucket = s3_bucket
         self.s3_te_data_folder = s3_te_data_folder
         self.data_store_table = data_store_table
         self.supported_indicator_types = self._get_supported_indicator_types(
             supported_signal_types
         )
+        self.s3_client = s3_client
+        self.s3_bucket_name = s3_bucket_name
+
+    @classmethod
+    def indicator_type_str_from_signal_type(
+        cls, signal_type: t.Type[SignalType]
+    ) -> str:
+        """
+        This mapping is only necessary for types that are in the process of
+        being migrated. eg. VideoMD5.
+        """
+        if signal_type == VideoMD5Signal:
+            return "HASH_VIDEO_MD5"
+
+        return getattr(signal_type, "INDICATOR_TYPE", None)
 
     def _get_supported_indicator_types(
         self, supported_signal_types: t.List[t.Type[SignalType]]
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -251,7 +267,13 @@ def _get_supported_indicator_types(
 
     @property
     def checkpoint_s3_key(self) -> str:
-        return f"{self.s3_te_data_folder}{self.privacy_group}.checkpoint"
+        return f"{self.s3_te_data_folder}{self.privacy_group}{self.CHECKPOINT_SUFFIX}"
+
+    def get_privacy_group_prefix(self) -> str:
```
BarrettOlson, line 272: The '.' is pretty easy to miss here: without it you could also use this on line 270.

hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -338,38 +362,68 @@ def _store_checkpoint(self, checkpoint: tu.ThreatUpdateCheckpoint) -> None:
...
+            for datafile in all_datafile_keys:
+                txt_content = read_s3_text(
+                    self.s3_client, self.s3_bucket_name, datafile
+                )
+                signal_type = self.get_signal_type_from_object_key(datafile)
+
+                if txt_content is None:
+                    logger.warning("No TE state for %d. First run?", self.privacy_group)
+                else:
+                    csv.field_size_limit(65535)  # dodge field size problems
```
BarrettOlson, line 407: not your addition but... what?

hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -209,37 +210,52 @@ class ThreatUpdateS3Store(tu.ThreatUpdatesStore):
...
+    @classmethod
+    def indicator_type_str_from_signal_type(
+        cls, signal_type: t.Type[SignalType]
+    ) -> str:
+        """
+        This mapping is only necessary for types that are in the process of
+        being migrated. eg. VideoMD5.
+        """
+        if signal_type == VideoMD5Signal:
+            return "HASH_VIDEO_MD5"
```
BarrettOlson, line 245: nit: I think we're going to forget about this here. Can we address this in py threatexchange?

hasher-matcher-actioner/hmalib/common/s3_adapters.py
```diff
@@ -338,38 +362,68 @@ def _store_checkpoint(self, checkpoint: tu.ThreatUpdateCheckpoint) -> None:
...
+
+                if txt_content is None:
+                    logger.warning("No TE state for %d. First run?", self.privacy_group)
+                else:
+                    csv.field_size_limit(65535)  # dodge field size problems
+                    for row in csv.reader(txt_content):
+                        items.append(
+                            HMASerialization(
+                                row[0],
+                                self.indicator_type_str_from_signal_type(signal_type),
```
BarrettOlson, line 412: nit: check if this returns None and error if so (could be in helper)?

hasher-matcher-actioner/hmalib/common/s3_adapters.py
schatten: I'll add more in the docstring, but this is necessary here. Imagine two privacy groups 123 and 1234. Unless we specify the . here, the `list_objects()` call later for a pg 123 would return datafiles for both 123 and 1234.

hasher-matcher-actioner/hmalib/common/s3_adapters.py
schatten: We must. But I do not have a clean solution. Partners need both HASH_MD5 and HASH_VIDEO_MD5 to be INDICATOR_TYPEs for VideoMD5 depending on:
a) whether they have started using threat_updates
b) their particular collaboration has been backfilled

hasher-matcher-actioner/hmalib/common/s3_adapters.py
BarrettOlson: [ignore-able] nit/to clarify: not questioning the need for the dot at all just having this helper. The subtle nature of that case means I'd almost not want this helper at all or handle it directly where the call is made. 
I don't know from reading the code that a prefix OR a suffix is going to have a '.' or a '/'

PR #770, opened 2021-08-23
Consolidate HMA GitHub actions

schatten:
Summary
---------
Consolidate main and pull request into a single workflow:
- remove the master/main push flow, add the triggers for this into the `hma-ci-pull-request` flow and rename to `hma-ci`
In a previous version, the main and pull_request workflows were kept separate because I had imagined that we can deploy to AWS from Github. This no longer seems plausible. instead, reduce the points of maintenance.

- reuse the same step for types and formatting. Building a new image is unnecessary and potentially eats away our org's minutes. 
- do terraform validation, tests, type and lint checking as part of the same workflow
- with this, we'll have 2 workflows for HMA. One for webapp and one for everything else.

Test Plan
---------
CI!


BarrettOlson (approved):
Sure! we can break it up again if needed anyway.

PR #773, opened 2021-08-24
Make PDQSignalMetadata signal-type agnostic -> ThreatExchangeSignalMetadata

schatten:
NOTE: All current signal metadata objects will be unusable after this. On matches on your instance, new signal metadata objects will be created. If your opinions have synced with threatexchange, you've nothing to worry about.

Summary
---------
This is in the lead-up to supporting video-md5 signal metadata.
Be sure to go through the module docstring for `hmalib.common.models.signal`

To reviewer: There may be questions on choice of PK / SK for signal metadata. Happy to respond in inline comments. 

Test Plan
---------
```
$ python -m black hmalib
$ python -m mypy hmalib
$ python -m py.test
```

A URL upload for an image in our test privacy groups was done. This resulted in matches and actions. These were visually verified on the pipeline progress page.

On the content details page as shown in a series of screenshots, I removed the opinion and added it back again. After every change, I manually triggered fetcher because I did not want to wait for 15 minutes to have passed. ;) 

### Original: Has opinion
Screen Shot 2021-08-24 at 16 59 52

### UI used to remove opinion
Screen Shot 2021-08-24 at 17 00 02

\< Runs fetcher manually \>

### Opinion was removed.
Screen Shot 2021-08-24 at 17 01 00

I also added the opinion back and got it to reflect in the UI.
Not adding screenshots. You get the drift..

### Before the fetcher was run, refreshing the content details page gave me the interim state
Screen Shot 2021-08-24 at 14 31 53

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/signal.py
+++ b/hasher-matcher-actioner/hmalib/common/models/signal.py
@@ -63,18 +63,24 @@ class ThreatExchangeSignalMetadata(DynamoDBItem):
     Once the opinion is written to threatexchange and synced back, the 'pending'
     opinion is cleared.
 
+    As clarified in the module's doc, it is okay to expose the threatexchange
+    specific attribute 'privacy_group_id' because the capabilities that an
+    exchange offers is going to be varied enough that abstracting might be
+    impossible.
+
     Storage
     ---
     PK: s#{source_short_code}#{signal_id}
-    SK: #threatexchange-signal-metadata
+    SK: pg#{privacy_group_id}
 
     Where
     {source_short_code} = "te"
     and signal_id = {indicator_id} in threatexchange.
+    and privacy_group_id is privacy_group_id in threatexchange.
     """
 
     SIGNAL_SOURCE_SHORTCODE = "te"
-    DATASET_PREFIX = "ds#"
+    PRIVACY_GROUP_PREFIX = "pg#"
 
     signal_id: str
     privacy_group_id: str
@@ -88,9 +94,9 @@ class ThreatExchangeSignalMetadata(DynamoDBItem):
 
     PROJECTION_EXPRESSION = "PK, SignalHash, SignalSource, SignalType, PrivacyGroup, Tags, PendingThreatExchangeOpinionChange, UpdatedAt"
 
-    @staticmethod
-    def get_sort_key() -> str:
-        return "#threatexchange-signal-metadata"
+    @classmethod
+    def get_sort_key(self, privacy_group_id: str) -> str:
+        return f"pg#{privacy_group_id}"
 
     def to_json(self) -> t.Dict:
         """
@@ -108,7 +114,7 @@ class ThreatExchangeSignalMetadata(DynamoDBItem):
             "PK": self.get_dynamodb_signal_key(
                 self.SIGNAL_SOURCE_SHORTCODE, self.signal_id
             ),
-            "SK": self.get_sort_key(),
+            "SK": self.get_sort_key(self.privacy_group_id),
             "SignalHash": self.signal_hash,
             "SignalSource": self.SIGNAL_SOURCE_SHORTCODE,
             "UpdatedAt": self.updated_at.isoformat(),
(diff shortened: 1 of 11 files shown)


hasher-matcher-actioner/hmalib/common/models/signal.py
```diff
@@ -1,45 +1,96 @@
...
+
+    User's registration of an opinion is a multi-stage process. The opinion is
+    first recorded on this object as a 'pending' opinion directly by the UI.
+    Once the opinion is written to threatexchange and synced back, the 'pending'
+    opinion is cleared.
+
+    Storage
+    ---
+    PK: s#{source_short_code}#{signal_id}
+    SK: #threatexchange-signal-metadata
```
BarrettOlson, line 69: Unfortunately SK as ds_id (dataset i.e. privacy_group_id) is currently used as to determine uniqueness so I don't think this change will work. Because ThreatExchangeOpinionChange has that level of the granularity: e.g. users could agree this signal belongs in the `cat` dataset but not the `dog` dataset.

BarrettOlson (changes requested):

hasher-matcher-actioner/hmalib/common/models/signal.py
schatten: Aah! Because the same indicator can be in multiple privacy groups? Gotcha!

Well, back to the drawing board. Thanks for the flag!

schatten:
@BarrettOlson After making `privacy_group` essential to uniquely identifying a signal metadata object, I ran the test plan again. Surprise, surprise, the image I'm testing actually matches two separate privacy groups on threatexchange. So we have a testcase close to reality.

In the first screenshot, you'll see that I have tried to change opinions on both PGs.
Screen Shot 2021-08-25 at 10 30 41

Then, I waited for a couple minutes and ran the fetcher manually, syncing the new opinions.
Screen Shot 2021-08-25 at 10 37 07

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -133,21 +136,20 @@ def get_match_details(table: Table, content_id: str) -> t.List[MatchDetail]:
...
     table: Table, signal_id: t.Union[str, int], signal_source: str
 ) -> t.List[MatchDetailMetadata]:
+    # TODO: Remove need for signal_source. That's part of the *SignalMetadata class now.
     if not signal_id or not signal_source:
         return []
 
     return [
         MatchDetailMetadata(
-            dataset=metadata.ds_id,
+            dataset=metadata.privacy_group_id,
```
BarrettOlson, line 145: nit: finish the job and update `MatchDetailMetadata` dataset to privacy_group as well.

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -272,20 +274,21 @@ def request_signal_opinion_change() -> ChangeSignalOpinionResponse:
...
-        signal = PDQSignalMetadata(
-            signal_id=signal_id,
-            ds_id=ds_id,
-            updated_at=datetime.datetime.now(),
-            signal_source=signal_source,
-            signal_hash="",  # SignalHash not needed for update
-            tags=[],  # Tags not needed for update
-            pending_opinion_change=pending_opinion_change,
+        signal = ThreatExchangeSignalMetadata.get_from_signal_and_privacy_group(
+            dynamodb_table, signal_id=signal_id, privacy_group_id=ds_id
```
BarrettOlson, line 278: nit: cut out `ds_id` here too (okay if you want to leave `bottle.request.query.dataset_q` for now to avoid touching the client)

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -167,23 +167,27 @@ def get_metadata_objects_from_match(
...
-                    privacy_group_id,
-                    datetime.datetime.now(),
-                    match.metadata["source"],
-                    match.metadata["hash"],
-                    match.metadata["tags"].get(privacy_group_id, []),
+                ThreatExchangeSignalMetadata(
+                    signal_id=str(match.metadata["id"]),
+                    privacy_group_id=privacy_group_id,
+                    updated_at=datetime.datetime.now(),
+                    signal_type=PdqSignal,
```
BarrettOlson, line 188: nit: give the check it above on like 179 just pass in `signal_type` to make it even easier to generalize.

BarrettOlson (approved):
LGTM!
Some inline nits around fully removing the poorly named 'ds_id' 
(Also not blocking: this is getting complex enough that more unit tests around this would be nice, ideally something that would have caught the essential to uniquely identify requirement in CI instead of my review)

PR #774, opened 2021-08-25
[pytx] Rename experimental fetch => fetch

Dcallies:
Summary
---------

Experimental fetch is the new hotness. However, it doesn't quite yet have all the wizziwangs and bimblibops that the old fetch has. The old fetch has the small problem in that the implementation isn't correct in many cases. Increase the pain and suffering of users by forcing them to wait for the blisteringly slow threat_updates fetch, at the small benefit of provable correctness.

Test Plan
---------
Look at helps, because everything is broken at the moment because of a requests version change

```
$ threatexchange --help
usage: threatexchange [-h] [--config CONFIG] [--app-token TOKEN]
                      [--state-dir DIR]
                      {fetch,match,label,dataset,hash,tag-fetch} ...

A wrapper around multi-stage ThreatExchange operations.

Includes simple matching and writing back. Useful for quickly validating new
sources of ThreatExchange data. A possible template for a native
implementation in your own architecture.

This helper heavily relies on a config file to provide consistent behavior
between stages, and a state file to store hashes.

optional arguments:
  --config CONFIG, -c CONFIG
                        a ThreatExchange collaboration config
  --app-token TOKEN, -a TOKEN
                        the App token for ThreatExchange
  --state-dir DIR, -s DIR
                        the directory with the config state

verbs:
  {fetch,match,label,dataset,hash,tag-fetch}
                        which action to do
    fetch               fetch content from ThreatExchange to disk
    match               match content to items in ThreatExchange
    label               apply labels to items in ThreatExchange
    dataset             introspect the local ThreatExchange dataset
    hash                hash content into signatures (aka hashes)
    tag-fetch           fetch the old-fashioned way

$ threatexchange fetch --help
usage: threatexchange fetch [-h] [--full] [--skip-index-rebuild]
                            [--limit LIMIT] [--stop-time STOP_TIME]

    Fetch content from ThreatExchange to disk.

    Using the CollaborationConfig, download signals that
    correspond to a single collaboration, and store them in the state
    directory.

    This endpoint uses /threat_updates to fetch content sequentially, and in
    theory can be interrupted without issues.


optional arguments:
  --full                force a refetch from the beginning of time (this is
                        almost certainly not needed)
  --skip-index-rebuild  don't rebuild indices after fetch
  --limit LIMIT         stop after fetching this many items
  --stop-time STOP_TIME
                        only fetch until this point

$ threatexchange tag-fetch --help
usage: threatexchange tag-fetch [-h] [--sample] [--clear] [--full]
                                [--until-timestamp UNTIL_TIMESTAMP]
                                [--only-signals signal [signal ...]]
                                [--not-signals signal [signal ...]]

    Fetch the old-fashioned way.

    This function does something very similar to `fetch`. However, while
    fetch is powered by the `threat_updates` API, this command is powered by
    the legacy `<threat_tag>/tagged_objects` API. In nearly all cases you
    want regular old `fetch`, but for some older collaborations only
    `tag-fetch` will work.

    For some additional historical baggage, the implementation of this
    command stores things slightly differently than `fetch` does.


optional arguments:
  --sample              Only fetch a sample of data instead of the whole
                        dataset
  --clear               Don't fetch anything, just clear the dataset
  --full                Force a full refresh of data.
  --until-timestamp UNTIL_TIMESTAMP
                        Instead of fetching all the data, incrementally fetch
                        up to this time
  --only-signals signal [signal ...], -o signal [signal ...]
                        Only download the specified signal types.
  --not-signals signal [signal ...], -n signal [signal ...]
                        Don't fetch the specified signal types.

```


Dcallies:
Missed a reference, will fix.

BarrettOlson (changes requested):
LGTM: fetch is dead, long live fetch
RC: for CI fixes

PR #776, opened 2021-08-25
[hma][api] create configurable integration_api_access_token

BarrettOlson:
Summary
---------

For integration it is good to have permanent access token disconnected from any specific users. 
This PR updates the APIGateway authorizer to be a custom lambda that is able to validate cognito JWTs as well as check for a specific token given at deployment as a part of `terraform/terraform.tfvars` -> `integration_api_access_token` 

This token can be user everywhere the ephemeral access token from cognito was used. Warning: this effectively give you the ability to have a skeleton key for the API, use with care. 

Test Plan
---------

Made sure HMA still worked e2e and used postman to query the api with the value used for `integration_api_access_token` in the header as { "Authorization" : "<token>"} instead of using OAuth 2.0 (and that non-valid and expired tokens failed).

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
@@ -1,0 +1,92 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import os
+import json
+import time
+import jwt
+import requests
+import base64
+from jwt.algorithms import RSAAlgorithm
+
+from hmalib.common.logging import get_logger
+
+ACCESS_TOKEN = os.environ["ACCESS_TOKEN"]
+USER_POOL_URL = os.environ["USER_POOL_URL"]
+CLIENT_ID = os.environ["CLIENT_ID"]
+
+keys_url = f"{USER_POOL_URL}/.well-known/jwks.json"
+
+response = requests.get(keys_url)
+key_list = json.loads(response.text).get("keys", [])
+keys = {key["kid"]: json.dumps(key) for key in key_list}
+
+
+def is_jwt(token: str) -> bool:
+    jwt_split = token.split(".")
+    try:
+        if len(jwt_split) != 3:
+            return False
+        header = json.loads(base64.b64decode(jwt_split[0] + "=="))
+        if header["alg"] != "RS256":
+            return False
+    except Exception as e:
+        print("INFO: Provisioning Data is not in JWT format\n {0}".format(e))
+        return False
+    return True
+
+
+def validate_jwt(token: str):
+    logger = get_logger()
+
+    try:
+        if not keys:
+            logger.error("No JWT public keys found. User auth will always fail.")
+
+        kid = jwt.get_unverified_header(token)["kid"]
+        key = keys[kid]
+
+        public_key = RSAAlgorithm.from_jwk(key)
+
+        # Decode does verify_signature
+        decoded = jwt.decode(
+            token, public_key, algorithms=["RS256"], issuer=USER_POOL_URL
+        )
+
+        # Congnito JWT use 'client_id' instead of 'aud' e.g. audience
+        if decoded["client_id"] == CLIENT_ID and decoded["token_use"] == "access":
+            # Because we don't require username as part of the request,
+            # we don't check it beyond making sure it exists.
+            username = decoded["username"]
+            logger.info(f"User: {username} JWT verified.")
+            return {"isAuthorized": True, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+    except Exception as e:
+        logger.exception(e)
+
+    return {"isAuthorized": False, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+
+def validate_access_toke(token: str):
+    response = {"isAuthorized": False, "context": {"AuthInfo": "ServiceAccessToken"}}
+
+    if token == ACCESS_TOKEN:
+        get_logger().info("Access token approved")
+        response["isAuthorized"] = True
+
+    return response
+
+
+def lambda_handler(event, context):
+
+    token = event["identitySource"][0]
+
+    if is_jwt(token):
+        return validate_jwt(token)
+
+    return validate_access_toke(token)
+
+
+if __name__ == "__main__":
+    token = "text_token"
+    event = {"identitySource": [token]}
+    lambda_handler(event, None)
--- a/hasher-matcher-actioner/terraform/api/main.tf
+++ b/hasher-matcher-actioner/terraform/api/main.tf
@@ -160,6 +160,82 @@ resource "aws_iam_role_policy_attachment" "api_root" {
   policy_arn = aws_iam_policy.api_root.arn
 }
 
+# Authorizer API Lambda
+
+locals {
+  user_pool_url = "https://cognito-idp.${var.region}.amazonaws.com/${var.api_and_webapp_user_pool_id}"
+}
+
+resource "aws_lambda_function" "api_auth" {
+  function_name = "${var.prefix}_api_auth"
+  package_type  = "Image"
+  role          = aws_iam_role.api_auth.arn
+  image_uri     = var.lambda_docker_info.uri
+  image_config {
+    command = [var.lambda_docker_info.commands.api_auth]
+  }
+  timeout     = 300
+  memory_size = 512
+  environment {
+    variables = {
+      ACCESS_TOKEN  = var.integration_api_access_token
+      USER_POOL_URL = local.user_pool_url
+      CLIENT_ID     = var.api_authorizer_audience
+    }
+  }
+  tags = merge(
+    var.additional_tags,
+    {
+      Name = "AuthAPIFunction"
+    }
+  )
+}
+
+resource "aws_cloudwatch_log_group" "api_auth" {
+  name              = "/aws/lambda/${aws_lambda_function.api_auth.function_name}"
+  retention_in_days = var.log_retention_in_days
+  tags = merge(
+    var.additional_tags,
+    {
+      Name = "AuthAPILambdaLogGroup"
+    }
+  )
+}
+
+resource "aws_iam_role" "api_auth" {
+  name_prefix        = "${var.prefix}_api_auth"
+  assume_role_policy = data.aws_iam_policy_document.lambda_assume_role.json
+  tags = merge(
+    var.additional_tags,
+    {
+      Name = "AuthAPILambdaRole"
+    }
+  )
+}
+
+data "aws_iam_policy_document" "api_auth" {
+  statement {
+    effect = "Allow"
+    actions = [
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "logs:DescribeLogStreams"
+    ]
+    resources = ["${aws_cloudwatch_log_group.api_auth.arn}:*"]
+  }
+}
+
+resource "aws_iam_policy" "api_auth" {
+  name_prefix = "${var.prefix}_api_auth_role_policy"
+  description = "Permissions for Auth API Lambda"
+  policy      = data.aws_iam_policy_document.api_auth.json
+}
+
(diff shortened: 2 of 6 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+            return {"isAuthorized": True, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+    except Exception as e:
+        logger.exception(e)
+
+    return {"isAuthorized": False, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+
+def validate_access_toke(token: str):
+    response = {"isAuthorized": False, "context": {"AuthInfo": "ServiceAccessToken"}}
```
jeberl, line 70: Hmm. When building the workplace integration I was getting errors when `"isAuthorized"` in my response was a bool. APIGateway was expecting a string (`"true"|"false"`). You've tested this e2e?

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+            logger.info(f"User: {username} JWT verified.")
+            return {"isAuthorized": True, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+    except Exception as e:
+        logger.exception(e)
+
+    return {"isAuthorized": False, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+
+def validate_access_toke(token: str):
```
jeberl, line 69: s\toke\token

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+import json
+import time
+import jwt
+import requests
+import base64
+from jwt.algorithms import RSAAlgorithm
+
+from hmalib.common.logging import get_logger
+
+ACCESS_TOKEN = os.environ["ACCESS_TOKEN"]
```
jeberl, line 13: Blocking: I think we should store this in an AWS secret rather than an environment variable

Pros:
- Users can control access easier with IAM permissions
- Its encrypted
- Users don't have to re-depoly terraform every time they want to change the token and the token can be changed by users who might not have access to the terraform deployment. If we eventually enable users to create/generate API ACCESS_TOKENs from the UI, this lambda would have to be redeployed on every change

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+        if header["alg"] != "RS256":
+            return False
+    except Exception as e:
+        print("INFO: Provisioning Data is not in JWT format\n {0}".format(e))
+        return False
+    return True
+
+
+def validate_jwt(token: str):
+    logger = get_logger()
```
jeberl, line 39: Don't we usually put the logger at the top level because we only want to call `get_logger()` once

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+            # Because we don't require username as part of the request,
+            # we don't check it beyond making sure it exists.
+            username = decoded["username"]
+            logger.info(f"User: {username} JWT verified.")
+            return {"isAuthorized": True, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+    except Exception as e:
+        logger.exception(e)
+
+    return {"isAuthorized": False, "context": {"AuthInfo": "JWTTokenCheck"}}
```
jeberl, line 66: Nit: log not verified?

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+        decoded = jwt.decode(
+            token, public_key, algorithms=["RS256"], issuer=USER_POOL_URL
+        )
+
+        # Congnito JWT use 'client_id' instead of 'aud' e.g. audience
+        if decoded["client_id"] == CLIENT_ID and decoded["token_use"] == "access":
+            # Because we don't require username as part of the request,
+            # we don't check it beyond making sure it exists.
+            username = decoded["username"]
+            logger.info(f"User: {username} JWT verified.")
```
jeberl, line 60: 👍

hasher-matcher-actioner/terraform/api/variables.tf
```diff
@@ -136,3 +136,9 @@ variable "partner_image_buckets" {
     params = map(string)
   }))
 }
+
+variable "integration_api_access_token" {
```
jeberl, line 140: Can users disable access via access token if they only want to use a cognito pool?

Maybe set a default here of empty string and in the auth lambda check for empty string before allowing?

hasher-matcher-actioner/terraform/api/variables.tf
```diff
@@ -17,8 +17,8 @@ variable "prefix" {
   type        = string
 }
 
-variable "api_authorizer_jwt_issuer" {
-  description = "A URL to the JWT issuer (used by the api gateway authorizer)"
+variable "api_and_webapp_user_pool_id" {
+  description = "user pool id that can be used to create a URL to the JWT issuer (used by the api gateway authorizer)"
```
jeberl, line 21: Once this lands can you send in chat an example of what we should include for this variable and maybe update `terraform.tfvars.example` also

jeberl (changes requested):

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
BarrettOlson: These changes are incremental, I like your suggestion and have thought of it as well but is adds another layer complexity and would have greatly increased the size of the PR. More over it does not actually change the way one would integrate just how the token value is set.    

As is this addresses the need for a permanent token right now and keeps the door open for this: all we'd need to do is change the impl of `validate_access_token` to query aws secret. 

Aside: an apply is not a full re-depoly, and is very easy to update without require any changes or use of the aws console.

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+
+from hmalib.common.logging import get_logger
+
+ACCESS_TOKEN = os.environ["ACCESS_TOKEN"]
+USER_POOL_URL = os.environ["USER_POOL_URL"]
+CLIENT_ID = os.environ["CLIENT_ID"]
+
+keys_url = f"{USER_POOL_URL}/.well-known/jwks.json"
+
+response = requests.get(keys_url)
```
schatten, line 19: Put behind an lru_cache? Avoid IO statements in module scope.

hasher-matcher-actioner/terraform/api/variables.tf
BarrettOlson: `integration_api_access_token = "super-secret-string-I-made-by-hitting-my-head-on-the-keyboard"`

hasher-matcher-actioner/terraform/api/variables.tf
BarrettOlson: yes, that happens automatically based on my testing but incase it is system dependent (i've been burned by tf null checks before) I add that logic explicitly.

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+key_list = json.loads(response.text).get("keys", [])
+keys = {key["kid"]: json.dumps(key) for key in key_list}
+
+
+def is_jwt(token: str) -> bool:
```
schatten, line 24: nit: this could be done by pyjwt itself, no?

pyjwt.decode throws a DecodeError if it can't process the JWT.

https://pyjwt.readthedocs.io/en/stable/api.html#jwt.exceptions.DecodeError

hasher-matcher-actioner/terraform/api/variables.tf
BarrettOlson: not going to add it to `terraform.tfvars.example` because if you miss it the example value might work and give someone unintentional access

hasher-matcher-actioner/terraform/api/main.tf
```diff
@@ -160,6 +160,82 @@ resource "aws_iam_role_policy_attachment" "api_root" {
...
+resource "aws_lambda_function" "api_auth" {
+  function_name = "${var.prefix}_api_auth"
+  package_type  = "Image"
+  role          = aws_iam_role.api_auth.arn
+  image_uri     = var.lambda_docker_info.uri
+  image_config {
+    command = [var.lambda_docker_info.commands.api_auth]
+  }
+  timeout     = 300
+  memory_size = 512
```
schatten, line 178: Quite large for a compute-only lambda.

schatten: May want to reduce the timeout while you're at it.

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+    logger = get_logger()
+
+    try:
+        if not keys:
+            logger.error("No JWT public keys found. User auth will always fail.")
+
+        kid = jwt.get_unverified_header(token)["kid"]
+        key = keys[kid]
+
+        public_key = RSAAlgorithm.from_jwk(key)
```
schatten, line 48: Compute and cache this once per key. I see that [`from_jwk`](https://github.com/jpadilla/pyjwt/blob/master/jwt/algorithms.py#L301) method does some crypto work which can be avoided.

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
schatten: If used via an automated client, this could do a deluge of logs. logger.debug instead?

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+        logger.exception(e)
+
+    return {"isAuthorized": False, "context": {"AuthInfo": "JWTTokenCheck"}}
+
+
+def validate_access_toke(token: str):
+    response = {"isAuthorized": False, "context": {"AuthInfo": "ServiceAccessToken"}}
+
+    if token == ACCESS_TOKEN:
+        get_logger().info("Access token approved")
```
schatten, line 73: build logger once and use debug level logging for success events.

schatten (changes requested):
a) Would really prefer this be done via secrets as @jeberl requested. And that it supports multiple access keys.
b) that we cache compute heavy operations like from_jwk
c) that the lambda size be the minimum necessary

Optional:
d) think about this being done in api_root itself. What are the pros and cons?

hasher-matcher-actioner/hmalib/lambdas/api/api_auth.py
```diff
@@ -0,0 +1,92 @@
...
+import base64
+from jwt.algorithms import RSAAlgorithm
+
+from hmalib.common.logging import get_logger
+
+ACCESS_TOKEN = os.environ["ACCESS_TOKEN"]
+USER_POOL_URL = os.environ["USER_POOL_URL"]
+CLIENT_ID = os.environ["CLIENT_ID"]
+
+keys_url = f"{USER_POOL_URL}/.well-known/jwks.json"
```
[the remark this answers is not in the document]
BarrettOlson: Seems to be the pattern generally but I can still add refs to docs:
https://auth0.com/docs/tokens/json-web-tokens/json-web-key-sets 
https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html

PR #781, opened 2021-08-30
Parse URLs and headers in Actions

jeberl:
Summary
---------

Enable users to include the content ID and content hash in webhook actions.

When we send webhooks in the action framework, the context of the match is included in the data of the HTTP request. However, for the workplace integration, and potentially other integrations as well, including content ID and/or hash in the url or the headers of the request is required.

This diff enables that by parsing the urls and headers of the action rule and replacing the strings `<content-id>` with the matched content ID and `<content-hash>` with the matched content hash

Test Plan
---------

Run `pytest` which will send real requests to the webhook monitoring page
 
 Navigate to [the monitoring page](https://webhook.site/#!/d0dbb19d-2a6f-40be-ad4d-fa9c8b34c8df/11074918-20e3-4f8f-8d72-9b77a3a6f6ae/1) and confirm the headers are correct
Screen Shot 2021-08-30 at 12 08 56 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/configs/actioner.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/actioner.py
@@ -6,7 +6,7 @@ import typing as t
 
 from dataclasses import dataclass
 from hmalib.common.logging import get_logger
-from hmalib.common.messages.match import BankedSignal, MatchMessage
+from hmalib.common.messages.match import MatchMessage
 from hmalib.common.aws_dataclass import HasAWSSerialization
 from requests import get, post, put, delete, Response
 
@@ -78,7 +78,7 @@ class WebhookPostActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a POST"""
 
     def call(self, url: str, headers: str, data: str) -> Response:
-        logger.info(f"Performing a POST request to URL. {self.url}")
+        logger.info(f"Performing a POST request to URL. {url}")
         return post(url, data, headers=json.loads(headers))
 
 
@@ -87,8 +87,8 @@ class WebhookGetActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a GET"""
 
     def call(self, url: str, headers: str, data: str) -> Response:
-        logger.info(f"Performing a GET request to URL. {self.url}")
-        return get(url, data, headers=json.loads(headers))
+        logger.info(f"Performing a GET request to URL. {url}")
+        return get(url, headers=json.loads(headers))
 
 
 @dataclass
@@ -96,7 +96,7 @@ class WebhookPutActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a PUT"""
 
     def call(self, url: str, headers: str, data: str) -> Response:
-        logger.info(f"Performing a PUT request to URL. {self.url}")
+        logger.info(f"Performing a PUT request to URL. {url}")
         return put(url, data, headers=json.loads(headers))
 
 
@@ -104,33 +104,6 @@ class WebhookPutActionPerformer(WebhookActionPerformer):
 class WebhookDeleteActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a DELETE"""
 
-    def call(self, url: str, headers: str, _data: str) -> Response:
-        logger.info(f"Performing a DELETE request to URL. {self.url}")
+    def call(self, url: str, headers: str, data: str) -> Response:
+        logger.info(f"Performing a DELETE request to URL. {url}")
         return delete(url, headers=json.loads(headers))
-
-
-if __name__ == "__main__":
-    content_id = "cid1"
-    content_hash = "0374f1g34f12g34f8"
-
-    banked_signal = BankedSignal(
-        banked_content_id="4169895076385542",
-        bank_id="303636684709969",
-        bank_source="te",
-    )
-
-    action_performer = WebhookPostActionPerformer(
-        name="EnqueueForReview",
-        url="https://webhook.site/d0dbb19d-2a6f-40be-ad4d-fa9c8b34c8df",
-        headers='{"Connection":"keep-alive", "content-id" : "<content-id>", "content-hash" : "<content-hash>"}',
-        # monitoring page:
-        # https://webhook.site/#!/d0dbb19d-2a6f-40be-ad4d-fa9c8b34c8df
-    )
-
-    match_message = MatchMessage(
-        content_id,
-        content_hash,
-        [banked_signal],
-    )
-
-    action_performer.perform_action(match_message)
--- a/hasher-matcher-actioner/hmalib/common/configs/actioner.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/actioner.py
@@ -44,7 +44,6 @@ class ActionPerformer(config.HMAConfigWithSubtypes, HasAWSSerialization):
 # function on the match message
 WEBHOOK_ACTION_PERFORMER_REPLACEMENTS: t.Dict[str, t.Callable[[MatchMessage], str]] = {
     "<content-id>": lambda match_message: match_message.content_key,
-    "<content-hash>": lambda match_message: match_message.content_hash,
 }
 
 
@@ -56,7 +55,7 @@ class WebhookActionPerformer(ActionPerformer):
     headers: str
 
     def perform_action(self, match_message: MatchMessage) -> None:
-        parsed_url, parsed_headers = self.url, self.headers
+        parsed_url = self.url
         for (
             replacement_str,
             replacement_func,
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -48,43 +56,81 @@ class WebhookActionPerformer(ActionPerformer):
...
+            replacement_str,
+            replacement_func,
+        ) in WEBHOOK_ACTION_PERFORMER_REPLACEMENTS.items():
+            parsed_url = parsed_url.replace(
+                replacement_str, replacement_func(match_message)
+            )
+            parsed_headers = parsed_headers.replace(
+                replacement_str, replacement_func(match_message)
+            )
+        self.call(parsed_url, parsed_headers, data=json.dumps(match_message.to_aws()))
```
BarrettOlson, line 70: if content_id/key and content_hash are already a part of data why add them to the header?

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -40,6 +40,14 @@ def perform_action(self, match_message: MatchMessage) -> None:
         raise NotImplementedError
 
 
+# at match time, these strings are replaced by the result of running the associated
+# function on the match message
+WEBHOOK_ACTION_PERFORMER_REPLACEMENTS: t.Dict[str, t.Callable[[MatchMessage], str]] = {
+    "<content-id>": lambda match_message: match_message.content_key,
+    "<content-hash>": lambda match_message: match_message.content_hash,
```
BarrettOlson, line 47: I don't see how `<content-hash>` is useful, especially where the usual expectation is that it will be HMA generated

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -48,43 +56,81 @@ class WebhookActionPerformer(ActionPerformer):
...
         raise NotImplementedError()
 
 
 @dataclass
 class WebhookPostActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a POST"""
 
-    def call(self, data: str) -> Response:
+    def call(self, url: str, headers: str, data: str) -> Response:
         logger.info(f"Performing a POST request to URL. {self.url}")
```
BarrettOlson, line 81: nit: fix logging

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -48,43 +56,81 @@ class WebhookActionPerformer(ActionPerformer):
...
 
 @dataclass
 class WebhookGetActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a GET"""
 
-    def call(self, data: str) -> Response:
+    def call(self, url: str, headers: str, data: str) -> Response:
         logger.info(f"Performing a GET request to URL. {self.url}")
-        return get(self.url, headers=json.loads(self.headers))
+        return get(url, data, headers=json.loads(headers))
```
BarrettOlson, line 91: Did you test get or delete? I don't think either requests.get or .delete expects 'data' as the requests don't have bodies at least usually.

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -48,43 +56,81 @@ class WebhookActionPerformer(ActionPerformer):
...
-        return put(self.url, data, headers=json.loads(self.headers))
+        return put(url, data, headers=json.loads(headers))
 
 
 @dataclass
 class WebhookDeleteActionPerformer(WebhookActionPerformer):
     """Hit an arbitrary endpoint with a DELETE"""
 
-    def call(self, data: str) -> Response:
+    def call(self, url: str, headers: str, _data: str) -> Response:
```
BarrettOlson, line 107: This will break: see https://github.com/facebook/ThreatExchange/pull/673/files where it was fixed before.

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -48,43 +56,81 @@ class WebhookActionPerformer(ActionPerformer):
...
         logger.info(f"Performing a DELETE request to URL. {self.url}")
-        return delete(self.url, headers=json.loads(self.headers))
+        return delete(url, headers=json.loads(headers))
+
+
+if __name__ == "__main__":
```
BarrettOlson, line 112: These "local test pieces"  become outdate if we make any changes and then become an unnecessary layer to debug on top of other things. It also causes you to expand imports when it isn't used anywhere else in the file  (e.g. BankedSignal here) 
What I do and what I would prefer instead of this being committed is having another python file (outside of source control) that can just do:
```
# in no_commit.py for example
from hmalib.common.configs.actioner import  WebhookPostActionPerformer
from hmalib.common.messages.match import BankedSignal, MatchMessage

# the test I want to run manually while making changes
```

if you end up having code use use regularly we can then add it to a unit test or even script/shell command.

BarrettOlson (changes requested):
Having dynamic URLs make sense to me and are a good addition. 
RCing because of breaking changes on delete method and for more context on header and <content_hash> additions, as well as moving test code outside of the module. 

Thanks!

hasher-matcher-actioner/hmalib/common/configs/actioner.py
jeberl: I can imagine a use case where users might want this. It doesn't seem to add any more complexity so I figure why not include it

jeberl:
Addressed and retested all 4 action performers

hasher-matcher-actioner/hmalib/common/configs/actioner.py
jeberl: A lot of this is hard to test using the unit test framework. I get that its not super clean but I actually find it quite useful to store example executions along with the code I'm testing.

Having them linted and checked by mypy has helped me catch bugs as well. If you feel strongly I can delete it but I would prefer to keep it

hasher-matcher-actioner/hmalib/common/configs/actioner.py
jeberl: Same comment as on headers. I don't think its common but don't see any downside to including it

BarrettOlson: The downside imo is the added complexity and the possibility of it breaking something unexpected

hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
```diff
@@ -144,3 +145,33 @@ def test_get_action_labels(self):
...
+        match_message = MatchMessage(
+            content_id,
+            content_hash,
+            [banked_signal],
+        )
+
+        action_performers = [
+            performer_class(
+                name="EnqueueForReview",
+                url="https://webhook.site/d0dbb19d-2a6f-40be-ad4d-fa9c8b34c8df",
```
BarrettOlson: not blocking: Can you just use something like http://httpstat.us/404 instead to make it clear that this link won't stick around? feels a bit weird to have the test make an external request at all tbh

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -40,6 +40,13 @@ def perform_action(self, match_message: MatchMessage) -> None:
         raise NotImplementedError
 
 
+# at match time, these strings are replaced by the result of running the associated
+# function on the match message
+WEBHOOK_ACTION_PERFORMER_REPLACEMENTS: t.Dict[str, t.Callable[[MatchMessage], str]] = {
```
BarrettOlson, line 45: nit/ignorable: this could be inline now that we are down to one value. Especially imo as for changes beyond this I'd actually like to see the replacement be part of the action config, e.g. provide specifics of how to parse the values added to 'additional_fields' etc.

BarrettOlson (approved):
LGTM

hasher-matcher-actioner/hmalib/common/tests/test_action_rule_evaluation.py
jeberl: I want to use a real url for the same reason I included a real URL in the `if __name__ == "__main__":`. This is more of an integration test than a unit test. We should check that the correct url shows up at https://webhook.site/d0dbb19d-2a6f-40be-ad4d-fa9c8b34c8df.

Also, fwiw, this url shows up in other test cases already and is non permanent so after a few days it will not be visible

Going to merge since you marked as non blocking

hasher-matcher-actioner/hmalib/common/configs/actioner.py
jeberl: I like this as separate so its clear where to make future additions which I think we may want

PR #782, opened 2021-08-30
Support S3 ObjectCreate events in unified hasher; cleanup pdq-* components

### LINKED ISSUE #755: Modules should not declare their own providers. They should reuse one maintained in the root module.
https://www.terraform.io/docs/language/meta-arguments/module-providers.html

Recommends to have a single provider definition in the root module. This way we can remove / add modules without an issue. If a future release were to rename or remove a module, partners would have a tough time. I know I am when trying to resolve a merge conflict.

schatten:
Summary
---------
### [`422e23ec`](https://github.com/facebook/ThreatExchange/pull/782/commits/422e23ecaf60c1e5a1b612a16d3e6a2ae5150074) nitty: Make usage of prefix consistent across images/
The folder construct is inaccurate and potentially confusing. Using the more obvious and "known" prefix terminology.

### [`95e8f120`](https://github.com/facebook/ThreatExchange/pull/782/commits/95e8f12077ffc607094606ca37267ce665117dd5) Extract primary-bucket out of hashing-data
It is used for so much more, (indexing, fetching etc). It might also be time to reconsider the name for hashing-data module.
    
This was prompted by a weird dependency cycle in trying to get submission_queue and the bucket refer to each other. This was not strictly necessary, but is a step in the direction.

### [`786fc051`](https://github.com/facebook/ThreatExchange/pull/782/commits/786fc051b4a9490edf9d3e320c99345474810b08) Cleanup objects in the pdq-signals module.
It still exists so that terraform can identify the missing artifacts and clean them up.
    
Once we fix #755, this should not be necessary when removing a module.

Test Plan
---------
```
$ terraform apply
```

Uploaded an image via URL and from the file-system. Verified that they got all the way to the action performed stage.


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -45,9 +45,7 @@
 HMA_CONFIG_TABLE = os.environ["HMA_CONFIG_TABLE"]
 DYNAMODB_TABLE = os.environ["DYNAMODB_TABLE"]
 IMAGE_BUCKET_NAME = os.environ["IMAGE_BUCKET_NAME"]
-IMAGE_FOLDER_KEY = os.environ[
-    "IMAGE_FOLDER_KEY"
-]  # Misnamed, this is a prefix, not a key, if renaming, use IMAGE_PREFIX
+IMAGE_PREFIX = os.environ["IMAGE_PREFIX"]
```
BarrettOlson, line 48: :)

hasher-matcher-actioner/terraform/datastore/main.tf
```diff
@@ -71,4 +71,8 @@ resource "aws_dynamodb_table" "hma_datastore" {
       Name = "HMADataStore"
     }
   )
+
+  lifecycle {
```
BarrettOlson, line 75: nit: comment with context/why

hasher-matcher-actioner/terraform/hashing-data/main.tf
```diff
@@ -11,47 +11,45 @@ provider "aws" {
...
   block_public_acls       = true
   block_public_policy     = true
   ignore_public_acls      = true
   restrict_public_buckets = true
 }
 
+resource "aws_sqs_queue_policy" "allow_create_events_from_primary_bucket" {
+  queue_url = var.submissions_queue.queue_url
+
+  policy = <<POLICY
```
BarrettOlson, line 26: nit: strongly prefer using: 
```
data "aws_iam_policy_document" "doc_name" {
...
}
...
  policy    = data.aws_iam_policy_document.doc_name.json
...
```

hasher-matcher-actioner/terraform/hashing-data/main.tf
```diff
@@ -11,47 +11,45 @@ provider "aws" {
   profile = var.profile
 }
 
-# S3 Bucket
```
BarrettOlson, line 14: +1 on the rename now that this is removed from the file

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -255,6 +228,40 @@ module "authentication" {
   webapp_and_api_shared_user_pool_client_id = var.webapp_and_api_shared_user_pool_client_id
 }
 
+
+/**
+ * # Primary S3 Bucket:
+ * Jack-of-all-trades S3 bucket. Used for storing raw data from threatexchange,
+ * checkpoints, and upload-type media submissions.
+ *
```
BarrettOlson, line 236: nit: acknowledge the existence of `"aws_s3_bucket_notification" "bucket_notifications" {` and what it does here as it is now harder to trace and easier to miss, I would almost recommend pull that out of `hashing-data` because of this but am okay with you punting refactor changes like that for a future PR.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -71,30 +78,46 @@ def lambda_handler(event, context):
         if message.get("Event") == "s3:TestEvent":
             continue
 
+        media_to_process: t.List[t.Union[S3ImageSubmission, URLSubmissionMessage]] = []
+
         if URLSubmissionMessage.could_be(message):
-            submission_message = URLSubmissionMessage.from_sqs_message(message)
+            media_to_process.append(URLSubmissionMessage.from_sqs_message(message))
+        elif S3ImageSubmissionBatchMessage.could_be(message):
```
BarrettOlson, line 85: same (not blocking) nit about could_be from earlier PRs: I find this flow harder to follow and refactoring will be ~tricker than having, could_be and from be linked in Message class and treated as a single call here.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -71,30 +78,46 @@ def lambda_handler(event, context):
...
-                signal_type=signal.signal_type,
-                content_hash=signal.signal_value,
-                updated_at=datetime.datetime.now(),
-            )
-
-            hasher.write_hash_record(records_table, hash_record)
-            hasher.publish_hash_message(sqs_client, hash_record)
+        for media in media_to_process:
+            if not hasher.supports(media.content_type):
+                logger.warn(f"Unprocessable content type: {media.content_type}")
```
BarrettOlson, line 97: I think we want this to be an error to it'll show up red on the dashboard? (unless doing so results in batch retries)

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -71,30 +78,46 @@ def lambda_handler(event, context):
...
+                continue
+
+            with metrics.timer(metrics.names.hasher.download_file):
+                if hasattr(media, "key") and hasattr(media, "bucket"):
```
BarrettOlson, line 101: confused: you already did/had disambiguation above? this check feels like more work due to losing some context.

BarrettOlson (approved):
Approving to unblock because I think this change is important to get in sooner than later. (start the clock on the removal of pdq_signals/.tf)

Notes I'd like addressed soon if not pre-merge:
- It feels like there is extra work being done in hasher lambda in the name of generalization that could be simplified/optimized.
- Where we can avoid hardcoding things like `<<POLICY` we should
- +1 Rename or removal of hashing-data module to fit what it does now 
- Acknowledge the existence of  "bucket_notifications" in doc string or s3 bucket move to the root main.tf. 

Thanks!

hasher-matcher-actioner/terraform/main.tf
schatten: I am kind of getting confused with what hashing-data intends to do. Perhaps in a future commit, I can move the SQS events to the hasher submodule, because that's what it is: an entry point to the hasher.

note: I have added an acknowledgement in the feedback commit.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: It does result in retries.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: Nope, the disambiguation was between `URLSubmissionMessage` and `S3ImageSubmissionBatchMessage`. For S3ImageSubmissionBatchMessage, we end up adding a property `.image_submissions` to the `media_to_process` list. 

Here, we are differentiating between `URLSubmissionMessage` and `S3ImageSubmission` which is a different class.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: I forget. Are you suggesting instead of `could_be`, we simply try to deserialize and use a None value as our hint that it was not successful?

eg. 
```python
message = ClassA.from_sqs_message(raw_message)
if message is None:
  message = ClassB.from_sqs_message(raw_message)
if message is None:
  message = ClassC.from_sqs_message(raw_message)
```

BarrettOlson: ignore-able: effectively, maybe with prefix `.try_from_sqs_message(...)` and not as a hint that waterfalls just as a quick check for each:

```python
if m := ClassA.try_from_sqs_message(raw_message): media_to_process.append(m)
if m := ClassB.try_from_sqs_message(raw_message): media_to_process.append(m)
if m := ClassC.try_from_sqs_message(raw_message): media_to_process.append(m)
```
or even loop through possible message types.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
BarrettOlson: So I could be missing something but if I know something is an `URLSubmissionMessage` or a `S3ImageSubmission` (from `S3ImageSubmissionBatchMessage`) I then know how to get its bytes right there and then but instead we add them to the same list and do another check?

The two different message object are already disambiguated by their type, right?

Meta question that will explain my point or make it clear I am misunderstanding: 
Could media_to_process be created as a  list of tuples[message, get_bytes_method] ?

@schatten Thanks!

schatten: That does make sense. We could reduce the checks!

Action item: @schatten don't let this slip your mind. :)

PR #784, opened 2021-08-31
[hma][api][auth] store API access token in aws_secret

BarrettOlson:
Summary
---------

Builds off of #776 
Instead of passing the access token value to api_auth as an raw environment variable, tf will store it in aws_secret and can also be adjusted manually.
This also adds support for multiple access tokens instead of just one therefore the tfvars value defaults to empty list instead of string (though api_auth still has null checks to be safe). e.g. `integration_api_access_tokens  = ['<token1>', '<token2>']`

Test Plan
---------

Made sure all auth flows still worked for failed as expected in the same way they did in #776


schatten (approved):
LGTM!

PR #785, opened 2021-08-31
[hma][api] generalize for-hash and fix to_json for match data

BarrettOlson:
Summary
---------

 Generalize `/matches/for-hash/` (i.e. the "M") endpoint so that it can support both 'pdq' and 'video_md5'

Also fixes a tiny bug in `ThreatExchangeSignalMetadata.to_json` from changes  #784

Test Plan
---------

Used postman to hit the endpoint using both signal types:

exact pdq:
Screen Shot 2021-08-31 at 4 13 12 PM

exact md5:
Screen Shot 2021-08-31 at 4 17 18 PM

near pdq:
Screen Shot 2021-08-31 at 4 17 38 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
@@ -188,7 +188,7 @@ class MatchesForHashRequest(DictParseable):
 
 @dataclass
 class MatchesForHashResponse(JSONifiable):
-    matches: t.List[t.Union[ThreatExchangeSignalMetadata]]
+    matches: t.List[ThreatExchangeSignalMetadata]
     signal_value: str
 
     def to_json(self) -> t.Dict:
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -306,25 +309,15 @@ def for_hash(request: MatchesForHashRequest) -> MatchesForHashResponse:
...
-            # only support PDQ at the moment
-            bottle.response.status = 400
-            return MatchesForHashResponse([], request.signal_value)
 
         matches = _get_matcher(indexes_bucket_name).match(
-            signal_type, request.signal_value
+            request.signal_type, request.signal_value
         )
 
         match_objects = []
```
schatten, line 317: nit: type this? It is not immediately apparent what the type for this will be. Esp with the `.extend()` thing we're doing a few lines down.

schatten (approved):

schatten (approved):
Approved again. nit about adding to the docstring. 

But yeah, looks good!

PR #786, opened 2021-08-31
[hma][scripts] have test scripts use custom access tokens instead of user_pool

BarrettOlson:
Summary
---------
Remove the confusing and overly complex cognito work around/flow for the test scripts as we now have access to the kind of access token added in #776 

Test Plan
---------
`python scripts/hma_client_lib.py` and more.


schatten (approved):
Scripts, so only did a cursory look!

LGTM.

PR #787, opened 2021-08-31
Workplace Integration

jeberl:
We decided not to ship the integration but to save the code as an example


hasher-matcher-actioner/hmalib/lambdas/integrations.py
```diff
@@ -0,0 +1,200 @@
...
+        url=HMA_API_GATEWAY_URL + "/submit/url/",
+        headers={
+            "Content-Type": "application/json",
+            "Authorization": API_TOKEN,
+        },
+        data=payload_bytes,
+    )
+    response.raise_for_status()
+
+    logger.info(response.json())
```
jeberl, line 63: This is potentially the only code that would be a helpful example for a partner. Potentially just the `upload_to_HMA` function

hasher-matcher-actioner/hmalib/lambdas/integrations.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def get_bytes_from_url(url: str):
+    r = requests.get(url)
+    r.raise_for_status()
+    return r.content
```
jeberl, line 69: Dont actually need this anymore

hasher-matcher-actioner/hmalib/lambdas/integrations.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def process_request(
```
jeberl, line 72: All pretty workplace specific but the idea of having different content_ids and image_ids my be a useful distinction to highlight in an example

hasher-matcher-actioner/terraform/authentication/main.tf
```diff
@@ -73,3 +73,83 @@ resource "aws_cognito_user_pool_client" "webapp_and_api_user_pool_client" {
...
+  timeout     = 300
+  memory_size = 512
+  # hard coded for now in lambda. Should eventually becoem terraform variables
+  #environment {
+    #variables = {
+    #  API_TOKEN = var.api_token
+    #  HMA_API_GATEWAY_URL  = var.api_url
+    #  WORKPLACE_APP_SECRET = var.workplace_app_secret
+    #}
+  #}
```
jeberl, line 106: See comment

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -226,6 +226,15 @@ module "authentication" {
   use_shared_user_pool                      = var.use_shared_user_pool
   webapp_and_api_shared_user_pool_id        = var.webapp_and_api_shared_user_pool_id
   webapp_and_api_shared_user_pool_client_id = var.webapp_and_api_shared_user_pool_client_id
+
+  lambda_docker_info = {
+    uri = var.hma_lambda_docker_uri
+    commands = {
+      integrations = "hmalib.lambdas.integrations.lambda_handler"
```
jeberl, line 233: if we move to different folder will need to change this

hasher-matcher-actioner/hmalib/lambdas/integrations.py
schatten: That does sound right. Unless partners are using lambdas to power their APIs, there is limited value-add for these pieces of code. 

`upload_to_HMA` is something we could provide. Even that is limited to being used by those who use python, but is a reference impl. @BarrettOlson already wrote one when he was writing an e2e test. https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/scripts/hma_script_utils.py#L110-L131

That should IMO suffice as a reference impl. @BarrettOlson correct me if I am wrong, but you've also ported the same library to use access tokens instead of the cognito mechanism.

What we _could_ do, to make this more usable, is to provide it in form of a python package. Something like `hmaclient??`. 

Partners could do:
```python
from hmaclient import HMAClient

..
  client = HMAClient(endpoint="https://***.execute.aws", access_token="SDFSF......S"
  client.submit_url(...)
..
```

This way, your integration process might have been smoother. You'd not have to write the upload_to_HMA() function from scratch.

hasher-matcher-actioner/terraform/authentication/main.tf
```diff
@@ -73,3 +73,83 @@ resource "aws_cognito_user_pool_client" "webapp_and_api_user_pool_client" {
   access_token_validity  = 60
   id_token_validity      = 60
 }
+
+## Integrations Lambda - maybe move to its own module
```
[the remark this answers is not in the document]
schatten: As we discussed at standup, my belief is that partners will already have some event processor which would then submit to HMA. 

This appears too specific to be generalizable (even if made optional) for multiple partners. Vote for not adding integration lambdas to the main terraform setup. 

If partners want to receive other payloads from their own systems, my suggestion would be for them to expand on the submit APIs in submit.py. Since we intend on providing support, we could nudge them towards that.

hasher-matcher-actioner/hmalib/lambdas/integrations.py
jeberl: I like this. I wonder if its worth the effort however

BarrettOlson (changes requested):
can you instead make `/hasher-matcher-actioner/examples/workplace_integration/` and put both files there?

hmalib gets copied to the docker image each time we deploy.

schatten:
> can you instead make `/hasher-matcher-actioner/examples/workplace_integration/` and put both files there?
> 
> hmalib gets copied to the docker image each time we deploy.

+1. blocking.

BarrettOlson (approved):

PR #789, opened 2021-09-02
package lock

jeberl:
Summary
---------

After all of @BarrettOlson's type scripting work I needed to update my package.json file. This should ensure no one else has to

Test Plan
---------

UI Runs now


BarrettOlson (approved):
LGTM (I'm not sure this is needed but it can't hurt.)

PR #792, opened 2021-09-08
[hma][api] remove fields from MatchesForHashResponse matches

BarrettOlson:
Summary
---------

MatchesForHashResponse object based it's reply on ThreatExchangeSignalMetadata however (also added to docstring):

        ThreatExchangeSignalMetadata is used to store metadata in dynamodb
        and handle opinion changes on said signal. However the request this object
        responds to only handles directly accessing the index. Because of this
        not all fields of the object are relevant or accurate.

This PR removes the irrelevant 'pending_opinion_change' and inaccurate 'updated_at' fields from the reply. 

Unfortunately "updated_at" would required either an indexer refactor or a ddb call (likely both) to be accurate so it is likely worth holding off on such changes until there is a clear use case for it.   

There is a lot of ways to handle this removal, I choose the API layer to keep the other interface clean and untouched. This has a slight inefficiency tradeoff and I am open to feedback/alternatives. 

Test Plan
---------

Made sure the endpoint still worked and returned as expected.

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
@@ -191,8 +191,30 @@ class MatchesForHashResponse(JSONifiable):
     matches: t.List[ThreatExchangeSignalMetadata]
     signal_value: str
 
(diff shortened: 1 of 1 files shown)


schatten (approved):
LGTM!

Issue #779, opened 2021-08-27
[hma][ui][bug] Action and ActionFramework page dont update


jeberl:
When making an update on the Actions Settings page then navigating to the ActionRules page, the changes are not reflected. This can have issues on name changes, creations, and deletions.

The same is true in reverse

Workaround: Refresh the page after saving

PR #793, opened 2021-09-08
Make Actions and ActionRules Settings pages update automatically

jeberl:
Issue #779

Summary
---------

OK. Bear with me here...

I went about trying to make the Actions and ActionRules Settings pages update automatically. Formerly, if you added, deleted, or updated an Action, the change would not automatically show up on the ActionRules page until you refreshed. Similarly, changes to actionrules wouldn't show up on the actions page. The solution here was to move the React State components from the child tab classes to the parent settings class so they could be shared.

In making this change however I started doing some refactoring and adding additional tying. These changes grew until I was no longer sure which changes were for the initial feature and which were simply refactors.

All together the new changes are:
- Adding stricter typing across Actions and ActionRules specifically in interacting with the API
- Pass around Action and ActionRule objects rather than individual fields
- Using the same classes for Actions and ActionRules throughout the UI
- Handle API errors at the API level in `API.tsx`
- better function names
- remove old typing structure (`propTypes`)
- add missing key params to option lists

Test Plan
---------

### Video of new feature
1. ActionRule is set up to use Action 1
2. Change ActionRule 1 to use Action 2
3. Show that Action 1 can now be deleted without refreshing the page
4. Change name of Action 1
5. Show that new name appears without refresh when editing ActionRule

https://user-images.githubusercontent.com/24800630/132585681-4ac16600-55ea-45fc-a8c5-cde8c180e822.mov

### Video demonstrating all other functionality still works
1. Add, Delete, and Update Actions
2. Add, Delete, and Update ActionRules
3. Refresh page to demonstrate that changes are stored in DB

https://user-images.githubusercontent.com/24800630/132586372-af4e86b7-7fbf-4059-a937-a00ef5854384.mov

### DIFF
--- a/hasher-matcher-actioner/webapp/src/Api.tsx
+++ b/hasher-matcher-actioner/webapp/src/Api.tsx
@@ -347,9 +347,13 @@ export async function fetchAllActions(): Promise<Action[]> {
 }
 
 export async function createAction(
-  newAction = {},
+  newAction: Action,
 ): Promise<{response: string}> {
-  return apiPost('actions/', newAction);
+  return apiPost('actions/', {
+    name: newAction.name,
+    config_subtype: newAction.config_subtype,
+    fields: newAction.params,
+  });
 }
 
 export async function updateAction(
@@ -357,7 +361,11 @@ export async function updateAction(
   old_config_subtype: string,
   updatedAction: Action,
 ): Promise<{response: string}> {
-  return apiPut(`actions/${old_name}/${old_config_subtype}`, updatedAction);
+  return apiPut(`actions/${old_name}/${old_config_subtype}`, {
+    name: updatedAction.name,
+    config_subtype: updatedAction.config_subtype,
+    fields: updatedAction.params,
+  });
 }
 
 export async function deleteAction(name: string): Promise<{response: string}> {
--- a/hasher-matcher-actioner/webapp/src/Api.tsx
+++ b/hasher-matcher-actioner/webapp/src/Api.tsx
@@ -372,6 +372,8 @@ export async function deleteAction(name: string): Promise<{response: string}> {
   return apiDelete(`actions/${name}`);
 }
 
+// We need two different ActionRule types because the mackend model (must (not) have labels) is different
+// from the Frontend model (classification conditions)
 type APIActionRule = {
   name: string;
   must_have_labels: Label[];
--- a/hasher-matcher-actioner/webapp/src/Api.tsx
+++ b/hasher-matcher-actioner/webapp/src/Api.tsx
@@ -3,6 +3,8 @@
  */
 
 import {Auth, API} from 'aws-amplify';
+import {ActionRule, Label} from './pages/settings/ActionRuleSettingsTab';
+import {Action} from './pages/settings/ActionSettingsTab';
 
 async function getAuthorizationToken(): Promise<string> {
   const currentSession = await Auth.currentSession();
@@ -324,11 +326,24 @@ export async function fetchHashCount(): Promise<Response> {
 
 type AllActions = {
   error_message: string;
-  actions_response: Array<any>;
+  actions_response: Array<{
+    name: string;
+    config_subtype: string;
+    url: string;
+    headers: string;
+  }>;
 };
 
-export async function fetchAllActions(): Promise<AllActions> {
-  return apiGet('actions/');
+export async function fetchAllActions(): Promise<Action[]> {
+  return apiGet<AllActions>('actions/').then(response => {
+    if (response && !response.error_message && response.actions_response) {
+      return response.actions_response.map(
+        ({name, config_subtype, url, headers}) =>
+          new Action(name, config_subtype, url, headers),
+      );
+    }
+    return [];
+  });
 }
 
 export async function createAction(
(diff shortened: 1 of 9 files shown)


hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -324,38 +326,83 @@ export async function fetchHashCount(): Promise<Response> {
...
+    config_subtype: updatedAction.config_subtype,
+    fields: updatedAction.params,
+  });
 }
 
 export async function deleteAction(name: string): Promise<{response: string}> {
   return apiDelete(`actions/${name}`);
 }
 
+type APIActionRule = {
```
jeberl, line 375: We need two different ActionRule types because the Backend model is different from the front end model

jeberl: We could (and probably should) change this in the future but I would rather first type the current model and then change it

BarrettOlson: nit: if you feel the need to add this clarification/context, why not add it as a comment? Especially if you think it is worth of a ToDo?

jeberl: Agreed

BarrettOlson: did I miss you adding this somewhere?

hasher-matcher-actioner/webapp/src/components/settings/ActionPerformer/WebhookActioner.tsx
```diff
@@ -14,45 +14,45 @@ interface WebhookTypeInterface extends Record<string, any> {
...
-    },
-  };
-  const WebhookType: WebhookTypeInterface = {
-    WebhookPostActionPerformer: 'POST',
-    WebhookGetActionPerformer: 'GET',
-    WebhookDeleteActionPerformer: 'DELETE',
-    WebhookPutActionPerformer: 'PUT',
-  };
-  const actionerDetails = ActionerTypes.WebhookActioner;
+  const [, setURL] = useState(action.params.url);
```
BarrettOlson, line 53: nit: what does this state and the corresponding calls do?

hasher-matcher-actioner/webapp/src/pages/settings/ActionSettingsTab.tsx
```diff
@@ -2,123 +2,119 @@
...
+          }
+          return action;
+        });
+        setActions(updatedActions);
         displayToast(response.response);
-        refreshActions();
       })
-      .catch(() => {
+      .catch(e => {
+        /* eslint-disable-next-line no-console */
```
BarrettOlson, line 86: comment out or delete

hasher-matcher-actioner/webapp/src/pages/settings/ActionRuleSettingsTab.tsx
```diff
@@ -136,66 +132,32 @@ export class ActionRule {
...
-  name: string;
-  id: string;
-};
-
 const defaultActionRule = new ActionRule(
   '',
   [{key: classificationTypeTBD, value: ''}],
   [],
-  '0',
+  {key: 'ActionLabel', value: '0'},
```
BarrettOlson, line 139: `'0'` feel pretty dubious (not blocking as it already exists but having this work with null or undefined would be preferred)

hasher-matcher-actioner/webapp/src/pages/settings/ActionRuleSettingsTab.tsx
```diff
@@ -44,20 +39,23 @@ export class ActionRule {
 
   must_not_have_labels: Label[];
 
-  action_id: string;
-
   classification_conditions: ClassificationCondition[];
 
+  action: string;
```
BarrettOlson, line 44: Why the rename? `action_id` is not accurate? I think this adds more ambiguity personally.

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -63,14 +63,29 @@ export default function ActionRuleFormColumns({
             showErrors &&
             classification.classificationType === classificationTypeTBD
           }>
-          <option value={classificationTypeTBD}> Select... </option>
-          <option value="BankSourceClassification">Dataset Source</option>
-          <option value="BankIDClassification">Dataset ID</option>
-          <option value="BankedContentIDClassification">
+          <option key="tbd" value={classificationTypeTBD}>
```
BarrettOlson, line 66: nit: can you make these 'classificationTypes' an enum-type thing like we did for `ActionerTypes` ?

BarrettOlson (approved):
Nice! nits and questions in line but this is definitely an improvement and step in the right direction.

hasher-matcher-actioner/webapp/src/Api.tsx
jeberl: Nope. Just waiting to adress all your changes at once

hasher-matcher-actioner/webapp/src/components/settings/ActionPerformer/WebhookActioner.tsx
jeberl: True. I don't think we need this. Will remove and retest

jeberl: Updated state

hasher-matcher-actioner/webapp/src/pages/settings/ActionRuleSettingsTab.tsx
jeberl: Looking at more refactoring of action rules later and will incorporate this then

jeberl:
Added comment to APIActionRule and changed state.

ActionRules Refactor coming in separate PR

PR #794, opened 2021-09-08
Bank create view apis

schatten:
Summary
---------
### [`f42cf4e0`](https://github.com/facebook/ThreatExchange/pull/794/commits/f42cf4e0ca8ce96076c8113c83623988c61e8e8a) API to create, get-all and get banks

Test Plan
---------
```
$ python -m mypy hmalib
$ python -m black hmalib
$ python -m py.test
```

Applied terraform and tested out the new APIs using WIP UI code.
✅ Create Bank
✅ View all banks
✅ View bank by id


hasher-matcher-actioner/hmalib/common/messages/banks_api.py
```diff
@@ -0,0 +1,15 @@
...
+
+import typing as t
+from dataclasses import dataclass
+
+from hmalib.lambdas.api.middleware import JSONifiable
+from hmalib.common.models.bank import Bank
+
+
+@dataclass
+class AllBanksEnvelope(JSONifiable):
```
BarrettOlson, line 11: nit: it feels like this could stay in `hasher-matcher-actioner/hmalib/lambdas/api/bank.py` as we've done with most others; messages seemed to me primarily to be for lambda to lambda via sqs or sns. 

super nit: we've used `response` instead of `envelope` don't have a strong preference in naming as long as we can trend towards consistency though

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -403,7 +403,8 @@ module "api" {
       api_auth = "hmalib.lambdas.api.api_auth.lambda_handler"
     }
   }
-  datastore = module.datastore.primary_datastore
+  datastore       = module.datastore.primary_datastore
```
BarrettOlson, line 406: nit: also add usage widget to dashboard?

hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -0,0 +1,158 @@
...
+Stores banks and bank-members & bank-member signals.
+
+BankMemberSignals are signals extracted from the content in the bank member
+object. However, <signal_id> is unique for a signal_type, signal_value tuple.
+This is enforced by the 4th type of entry in the default table index.
+
+## Indices
+
+1. Default Table Index
+PK                        SK                             Item
```
BarrettOlson, line 28: nit: table is hard to read. recommendation: make `Item` the first column and add a `----------` line between column headers and values.

hasher-matcher-actioner/.dockerignore
```diff
@@ -0,0 +1,14 @@
+webapp/
```
BarrettOlson, line 1: nit: ELI5 why we want this (preferably inline). the set of dot/config files for hma has grown to the point where it might be work an appendix in CONTRIBUTING.md to keep track

hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -0,0 +1,158 @@
...
+<bank_id>                 member#<bank_member_id>        BankMember
+<bank_id>                 signal#<signal_id>             BankMemberSignal
+signal_type#<signal_type> signal_value#<signal_value>    <signal_id>
+
+Bank objects are all stored under the same partition key because all bank
+information is expected to be less than 10GB. Also, when a bank is deleted, it
+is removed from the (2) BankNameIndex. To be able to find it, there must be
+another way. That is provided by the "known" PK of #bank_object
+
+2. BankNameIndex
```
BarrettOlson, line 39: nit: make it clear what does and doesn't exist yet/post merge (could be mistaken but I am not seeing all these indexes in tf just yet). also if we are just checking if the name is repeated we could use a conditional write.

BarrettOlson (approved):
Cool, cool, cool, LGTM.
worth adding a test file here or in a future PR, I see you ran pytest in your test plan but I don't see one added. 
other nits inline.

hasher-matcher-actioner/terraform/main.tf
schatten: No, the present dashboard is all about pipeline metrics and latencies. Would like to keep it at that. Once we understand enough about the banks flow, we can see if there is merit in graphing it. Even if there is, should be in a separate dashboard.

hasher-matcher-actioner/hmalib/common/models/bank.py
schatten: Blaargh! I need stacked PRs! Github, are you listening!

By the time all my upcoming PRs land, this should be the state of the world. I assure you that's top of mind.

hasher-matcher-actioner/hmalib/common/messages/banks_api.py
schatten: I had a different reason for moving this here, but unless a lot more work is done, that does not make sense. You're right, I'll move it to api/bank.py instead.

I wanted the API representations in a separate module that could be made available to `hmaclient` should that ever come to life. Also, for my gencode experiments. But we can tackle both later. For now, moving back and renaming Envelope -> Response.

PR #795, opened 2021-09-09
Refactor ActionRules

jeberl:
Summary
---------

Add better comments and use of Classes to clearly tie Frontend and backend

Stop using `'0'` as the placeholder for "no action specified" and instead use `''`

also print error messages when saving changes leads to error

Test Plan
---------

Tested in UI that action rules can be created updated and deleted

Tested that Action Rules stay in sync with actions

Tested that if put create action request fails error is output:

Uploading Screen Recording 2021-09-09 at 3.05.10 PM.mov…

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/configs/actioner.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/actioner.py
@@ -15,7 +15,7 @@ logger = get_logger(__name__)
 
 TUrl = t.Union[t.Text, bytes]
 
-# This class should be kept in sync with typescript class APIActionPerformer
+# This class should be kept in sync with typescript class BackendActionPerformer in API.tsx
 class ActionPerformer(config.HMAConfigWithSubtypes, HasAWSSerialization):
     """
     An ActionPerfomer is the configuration + the code to perform an action.
--- a/hasher-matcher-actioner/hmalib/common/configs/actioner.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/actioner.py
@@ -15,7 +15,7 @@ logger = get_logger(__name__)
 
 TUrl = t.Union[t.Text, bytes]
 
-
+# This class should be kept in sync with typescript class APIActionPerformer
 class ActionPerformer(config.HMAConfigWithSubtypes, HasAWSSerialization):
     """
     An ActionPerfomer is the configuration + the code to perform an action.
--- a/hasher-matcher-actioner/hmalib/common/configs/evaluator.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/evaluator.py
@@ -15,6 +15,7 @@ class Action:
     superseded_by: t.List[ActionLabel]
 
 
+# This class should be kept in sync with TypeScript type APIActionRule
 @dataclass
 class ActionRule(HMAConfig, HasAWSSerialization):
     """
--- a/hasher-matcher-actioner/hmalib/common/configs/evaluator.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/evaluator.py
@@ -15,7 +15,7 @@ class Action:
     superseded_by: t.List[ActionLabel]
 
 
-# This class should be kept in sync with TypeScript type APIActionRule
+# This class should be kept in sync with TypeScript type BackendActionRule in API.tsx
 @dataclass
 class ActionRule(HMAConfig, HasAWSSerialization):
     """
--- a/hasher-matcher-actioner/webapp/src/Api.tsx
+++ b/hasher-matcher-actioner/webapp/src/Api.tsx
@@ -324,7 +324,7 @@ export async function fetchHashCount(): Promise<Response> {
 
 // TODO remove the trailing slash from the API URL, then add back the leading slash for /actions/ and /action-rules/ endpoints.
 
-// This class should be kept in sync with python class ActionPerformer
+// This class should be kept in sync with python class ActionPerformer (hmalib.common.configs.actioner.ActionPerformer)
 type BackendActionPerformer = {
   name: string;
   config_subtype: string;
@@ -381,7 +381,7 @@ export async function deleteAction(name: string): Promise<{response: string}> {
 
 // We need two different ActionRule types because the mackend model (must (not) have labels) is different
 // from the Frontend model (classification conditions).
-// This class should be kept in sync with python class ActionRule
+// This class should be kept in sync with python class ActionRule (hmalib.common.configs.evaluator.ActionRule)
 type BackendActionRule = {
   name: string;
   must_have_labels: Label[];
@@ -399,7 +399,7 @@ const convertToBackendActionRule = (frontend_action_rule: ActionRule) =>
     must_not_have_labels: frontend_action_rule.must_not_have_labels,
     action_label: {
       key: 'Action',
-      value: frontend_action_rule.action,
+      value: frontend_action_rule.action_name,
     },
   } as BackendActionRule);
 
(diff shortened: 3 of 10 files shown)


hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -324,23 +324,23 @@ export async function fetchHashCount(): Promise<Response> {
 
 // TODO remove the trailing slash from the API URL, then add back the leading slash for /actions/ and /action-rules/ endpoints.
 
+// This class should be kept in sync with python class ActionPerformer
```
BarrettOlson, line 327: nit/unsure: include file path

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -324,23 +324,23 @@ export async function fetchHashCount(): Promise<Response> {
...
+export type APIActionPerformer = {
```
BarrettOlson, line 328: nit: why API prefix?

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -324,23 +324,23 @@ export async function fetchHashCount(): Promise<Response> {
...
 };
 
 export async function fetchAllActions(): Promise<Action[]> {
   return apiGet<AllActions>('actions/').then(response => {
     if (response && !response.error_message && response.actions_response) {
-      return response.actions_response.map(
-        ({name, config_subtype, url, headers}) =>
-          new Action(name, config_subtype, url, headers),
-      );
+      return response.actions_response.map(action => new Action(action));
```
BarrettOlson, line 343: did you mean for this do be `APIActionPerformer` if not why does it exist compared to `Action`

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -373,8 +373,9 @@ export async function deleteAction(name: string): Promise<{response: string}> {
 }
 
 // We need two different ActionRule types because the mackend model (must (not) have labels) is different
-// from the Frontend model (classification conditions)
-type APIActionRule = {
+// from the Frontend model (classification conditions).
+// This class should be kept in sync with python class ActionRule
+export type APIActionRule = {
```
BarrettOlson, line 378: same feedback for this change as the one with `APIActionPerformer` 

Given `ActionRule` exists why are we using this one especially give what we return...

hasher-matcher-actioner/webapp/src/Api.tsx
jeberl: I think thats more likely to change than the class name. Changing the class name you would see the comment saying these should be kept in sync

hasher-matcher-actioner/webapp/src/Api.tsx
jeberl: Changed to `Backend` since I think thats more representative

hasher-matcher-actioner/webapp/src/Api.tsx
jeberl: Moved `APIActionPerformer` (now called `BackendActionPerformer`) so that it is only referenced in this file which does all the translation `Action <-> BackendActionPerformer`

hasher-matcher-actioner/hmalib/common/configs/actioner.py
```diff
@@ -15,7 +15,7 @@
 
 TUrl = t.Union[t.Text, bytes]
 
-
+# This class should be kept in sync with typescript class APIActionPerformer
```
BarrettOlson, line 18: Outdated

hasher-matcher-actioner/hmalib/common/configs/evaluator.py
```diff
@@ -15,6 +15,7 @@ class Action:
     superseded_by: t.List[ActionLabel]
 
 
+# This class should be kept in sync with TypeScript type APIActionRule
```
BarrettOlson, line 18: .

hasher-matcher-actioner/webapp/src/Api.tsx
BarrettOlson: haha the python file comments in this class disagree :P 
I would include both because then it is easy to trace the history of changes if it is missed.

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -324,22 +324,29 @@ export async function fetchHashCount(): Promise<Response> {
...
+
 type AllActions = {
   error_message: string;
-  actions_response: Array<{
-    name: string;
-    config_subtype: string;
-    url: string;
-    headers: string;
-  }>;
+  actions_response: Array<BackendActionPerformer>;
```
BarrettOlson, line 337: Just define this inline because `BackendActionPerformer` isn't referenced anywhere else and it'll make it harder for someone to just export it later, defeating the point of your most recent changes.

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -373,8 +380,9 @@ export async function deleteAction(name: string): Promise<{response: string}> {
 }
 
 // We need two different ActionRule types because the mackend model (must (not) have labels) is different
-// from the Frontend model (classification conditions)
-type APIActionRule = {
+// from the Frontend model (classification conditions).
```
BarrettOlson, line 383: .

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -396,9 +415,9 @@ export async function fetchAllActionRules(): Promise<ActionRule[]> {
         action_rule =>
```
BarrettOlson, line 415: nit: `backed_action_rule` instead to match

hasher-matcher-actioner/webapp/src/pages/settings/ActionRuleSettingsTab.tsx
```diff
@@ -45,15 +45,12 @@ export class ActionRule {
 
   constructor(
     name: string,
+    action: string,
```
BarrettOlson, line 48: nit/ignorable: I'd prefer this to have more context than just 'action'

hasher-matcher-actioner/webapp/src/Api.tsx
```diff
@@ -407,18 +426,20 @@ export async function fetchAllActionRules(): Promise<ActionRule[]> {
   });
 }
 
-export async function addActionRule(actionRule = {}): Promise<Response> {
+export async function addActionRule(actionRule: ActionRule): Promise<Response> {
```
BarrettOlson, line 429: can we call this "create..." to match what is going on and the backend method?

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
```diff
@@ -213,18 +213,15 @@ export default function ActionRuleFormColumns({
...
-            {' '}
-            (required)
-          </span>
+          <span hidden={!showErrors || !!actionRule.action}> (required)</span>
         </Form.Label>
         <Form.Control
           as="select"
           required
           value={actionRule.action}
           onChange={e => onChange('action_id', e.target.value)}
```
BarrettOlson, line 222: unsure: feels like this might now break something (or at least result in some confusion) based on the fact that it isn't called 'action_id' anymore?

BarrettOlson (approved):
yay! more steps toward improved typing :) 
nits inline

hasher-matcher-actioner/webapp/src/Api.tsx
jeberl: I want to keep it separate so its clear what needs to be kept in sync

hasher-matcher-actioner/webapp/src/components/settings/ActionRuleFormColumns.tsx
jeberl: Changed to action name

PR #796, opened 2021-09-09
[hma][api] support submitting direct video_md5 signal 

BarrettOlson:
Summary
---------

Updates the submit api to support the direct submission of md5 hashes (for videos) instead of only pdq.
Also updates all import to be absolute.

Only logic changes can be found in`/hmalib/lambdas/api/submit.py`

ALSO removed the _api suffix from the 3 files that had it.

Test Plan
---------

Made sure submitting worked for both signal types and e2e via updates to `/scripts/hma_client_lib.py`


schatten (approved):
Reviewed as part of #798. Feel free to merge that in. 

Curious why you decided to stop using relative imports. Also, can we add that as a linting rule?

BarrettOlson:
merged changes as part of #798

>Curious why you decided to stop using relative imports. Also, can we add that as a linting rule?

Early version of the script in that PR would error given relative imports when trying to find the function name. lint rule is a good call out!

PR #797, opened 2021-09-09
[hma][api] update root reply to include LastModified

BarrettOlson:
Summary
---------

See title. To enable this I had to allow the lambda to query for its own Function Configuration because `LastModified` is included in that [result](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.get_function_configuration) but not part of the [basic context](https://docs.aws.amazon.com/lambda/latest/dg/python-context.html):

Test Plan
---------
Screen Shot 2021-09-09 at 7 41 18 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
@@ -89,7 +89,8 @@ def root():
     )["LastModified"]
 
     return {
-        "message": f"Welcome to the HMA API! Last Modified/Deployed: {last_modified}",
+        "message": "Welcome to the HMA API!",
+        "last_modified": last_modified,
     }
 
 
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -78,8 +78,18 @@ def error500(e):
...
+    context = bottle.request.environ.get("apig_wsgi.context")
+    invoked_function_arn = context.invoked_function_arn
+    client = boto3.client("lambda")
+    last_modified = client.get_function_configuration(
+        FunctionName=invoked_function_arn
+    )["LastModified"]
+
     return {
-        "message": "Hello World, HMA",
+        "message": f"Welcome to the HMA API! Last Modified/Deployed: {last_modified}",
```
schatten, line 92: Can you add last modified to a separate field in the response? That way, we can automate easily. I was planning on scheduling an alarm for when the Continuous deploy script fails.

schatten (approved):
Approve to unblock. This is helpful, but not automation friendly. Move the last modified to another json field, and we are gold!

PR #800, opened 2021-09-14
bug: Get progress page to auto-refresh

schatten:
Summary
---------
bug: unreached stage's attribute has a value of null and not undefined

bonus: add a manual refresh button

Test Plan
---------
Verified that:
* UI self refreshes until action is done
* Refresh button clicks cause refresh no matter what


BarrettOlson (approved):
Cool cool cool

PR #805, opened 2021-09-21
Bank video members

schatten:
Summary
---------
### [`e4f7240d`](https://github.com/facebook/ThreatExchange/pull/805/commits/e4f7240d18206b1b7ad10d6f52bcb8731f0f1629) Use fixed aspect ratio for image / video previews
This allows the members page to appear like a grid even if all images are different aspect ratios.  This also fixes the problem where the blur filter would overflow into the actual area. Screenshot attached.  
### [`5eb37807`](https://github.com/facebook/ThreatExchange/pull/805/commits/5eb378074f8c6e23caf1d2799258d57d2b67178c) Replace blur on hover with explicit eye button clicks
This could have been its own PR, and I tried, but rebase fails, and I don't have the energy to fix it. If that is important to reviewer, can be done.  
### [`5fd20254`](https://github.com/facebook/ThreatExchange/pull/805/commits/5fd20254326bd7d087b9dabb19bf9a4a583fc0e8) Use ResponsiveEmbed in previewable dropzone
Use aspect ratio instead of height locking  
### [`004c5ece`](https://github.com/facebook/ThreatExchange/pull/805/commits/004c5ece16acf7ab3691266629f0559ecaf32da5) Support video previews in BankMembers

Test Plan
---------
The usual in python (black, mypy, py.test)

### Note how the image's blur does not bleed into the page. Top is after, bottom is before.
Screen Shot 2021-09-20 at 16 20 52

### See how videos can be previewed now.
0242f7f0f122f1b14c5bc574a385f3ef

### Borrowed the eye button from 
56d6b94279135bf103aa0fe9ca16d25c


hasher-matcher-actioner/webapp/src/forms/PreviewableDropzone.tsx
```diff
@@ -84,17 +86,23 @@ export default function PreviewableDropzone({
 
   const {getRootProps, getInputProps, open, isDragActive} = useDropzone({
     onDrop,
-    accept: type === 'video' ? 'video/*' : 'image/*',
+    accept: type === ContentType.Video ? 'video/*' : 'image/*',
   });
 
+  const fileType = type === ContentType.Video ? 'video' : 'photo';
```
BarrettOlson, line 92: Ignorable: why does `fileType` need to exist if it is 1-1 with ContentType, just use ContentType in PreviewPane?

hasher-matcher-actioner/webapp/src/forms/PreviewableDropzone.tsx
```diff
@@ -39,13 +40,14 @@ function PreviewPane({type, handleChange, file}: PreviewPaneProps) {
...
-        <Col
-          style={{textAlign: 'center', maxHeight: '300px', overflow: 'scroll'}}>
-          {type === 'video' ? (
-            <BlurVideo src={URL.createObjectURL(file)} override={!revealed} />
-          ) : (
-            <BlurImage src={URL.createObjectURL(file)} override={!revealed} />
-          )}
+        <Col>
+          <ResponsiveEmbed aspectRatio="16by9">
+            {type === 'video' ? (
```
BarrettOlson, line 45: why not just do this https://github.com/facebook/ThreatExchange/pull/805/files#diff-557a7cfcc6b8260e93c3cea9b7f62ea2a4ac0dedb89b129ffa7924df6c6cc005R79?

hasher-matcher-actioner/webapp/src/pages/bank-management/Members.tsx
```diff
@@ -106,37 +108,42 @@ export function PhotoMembers({bankId}: BankTabProps): JSX.Element {
...
+        setNoMoreMembers(true);
+      }
+    });
 
     fetchBank(bankId).then(resp => setBankName(resp.bank_name));
   }, [refetchCounter]);
 
   const empty = members.length === 0 && noMoreMembers === true;
 
+  let title = 'Photos in this bank.';
```
BarrettOlson, line 127: nit/person pref: `:` instead of `.` (or no punctuations)

hasher-matcher-actioner/webapp/src/utils/MediaUtils.tsx
```diff
@@ -74,16 +70,12 @@ export const BlurImage = BlurMedia(
  * Show a blurred preview of a video at {src}.
  */
 export const BlurVideo = BlurMedia(
-  (handleMouseLeave, handleMouseEnter, blur, src): JSX.Element => (
-    <div className="video-preview-container">
-      {/* eslint-disable jsx-a11y/media-has-caption */}
-      <video
-        onMouseLeave={handleMouseLeave}
```
BarrettOlson, line 81: no more hover option?

BarrettOlson (approved):
LGTM

(Aside: couldn't find it in the code so it might just be out of date but "Reveal!" with a `!` feels like a bit much, given what might be blurred)

hasher-matcher-actioner/webapp/src/utils/MediaUtils.tsx
schatten: Yeah, I found myself revealing too many images as I scanned through the grid view on the bank.

Hypothesis: User is more likely to not want to view the content. If they do, they'll click the button. Pattern matching from "<internal_system>" UI.

hasher-matcher-actioner/webapp/src/forms/PreviewableDropzone.tsx
schatten: Because type is of type `FileType` defined above.

Not using ContentType because we may (some day far in the future) use images for previews of videos. So, content_type != preview.file_type.

schatten: On second thought, we are not doing that now, so vote for consistency. I'll make the change, use ContentType through out.

PR #807, opened 2021-09-22
Remove local runners and a sample image

schatten:
Summary
---------
This was mistakenly checked in as part of #805

Test Plan
---------
These files are neither part of hmalib, nor webapp, so testing was unnecessary.


BarrettOlson (approved):
Stamp

PR #809, opened 2021-09-23
Run API locally + instructions on using the scripts

schatten:
Summary
---------
### [`9fb1a696`](https://github.com/facebook/ThreatExchange/pull/809/commits/9fb1a696a590a8b810c98ea374086793ea8ea0e9) Dedicated script to run the API
Could not re-use run_lambda because you'd want the API to keep running and not die after one request is answered.  
### [`e3445838`](https://github.com/facebook/ThreatExchange/pull/809/commits/e3445838e93dbe09c3d851e46c3e2da0eb6e62c7) Instructions on using run_lambda, run_api from hmacli
Also minor corrections to Correcting.md. 

Test Plan
--------
Screen Shot 2021-09-23 at 13 58 34

### DIFF
--- a/hasher-matcher-actioner/CONTRIBUTING.md
+++ b/hasher-matcher-actioner/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 Getting Started with Contributing
 
-In it's rough state, the prototype needs some prior configuration / setup before you are ready to start developing on it. Below are some steps and tips for getting started. If anything needs correction, PRs are always welcome.
+In its rough state, the prototype needs some prior configuration / setup before you are ready to start developing on it. Below are some steps and tips for getting started. If anything needs correction, PRs are always welcome.
 
 ## Prerequisites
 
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/CONTRIBUTING.md
```diff
@@ -86,3 +93,69 @@ Until a Syncer module is created, test threatexchange data needs to be manually
 3. Upload some target and no target images.
 
 This should run the images through the existing PDQ hasher and matcher lambda functions.
+
+
+## Running Lambdas locally
```
BarrettOlson, line 98: worth noting that both this and the run api commands not only require tf outputs but also need local aws admin auth to work.

BarrettOlson (approved):
Fancy! super convenient, like RunLambdaCommand it has limited portably (needs to be in the same place as deploy & aws certs) compared to the other commands but the utility will be great!

PR #810, opened 2021-09-23
Hash and write BankMemberSignal objects when a member is added

schatten:
Summary
---------
### [`7e570392`](https://github.com/facebook/ThreatExchange/pull/810/commits/7e570392870ce061d985ed8f73882f67004f69e7) Add model for enforcing signal_value uniquness in a signal_type
Tests for the same.  
### [`a2f5d103`](https://github.com/facebook/ThreatExchange/pull/810/commits/a2f5d10303f029e62285c390510b80a866785532) Support adding BankMemberSignals to the table and to the processing index
- add signals to the index - remove signals from the index  
### [`bc4fb6cb`](https://github.com/facebook/ThreatExchange/pull/810/commits/bc4fb6cb05f4778ca301d255c48f1b606872407d) Send custom message to hashing lambda when member is added to bank
* Defines new message type: `BankSubmissionMessage` 
* Publishes this message to the hashing queue with a presigned_url when a bank member is added 
* The hashing lambda hashes this media into  appropriate signal types and calls `bank_operations.add_bank_member_signal` 
* Removed content_id from ContentSignal class because it was not getting used and would have conflicted with bank submissions. Bank submissions don't have content_Id.  

> Note: Decided to add `bank_operations.add_bank_member_signal` because it is possible that partners may add hashes directly to bank in case they delete the original media because of legal requirements.  

Test Plan
---------
Deployed webapp and lambdas to my instance.

Uploaded a bank member on a bank.

3 records were created. a) bank_member b) signal uniqueness entry c) bank_member_signal

The md5 was the same as expected. Can be seen in the terminal window in the screenshot.
Screen Shot 2021-09-23 at 17 37 17

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/bank.py
+++ b/hasher-matcher-actioner/hmalib/common/models/bank.py
@@ -3,10 +3,11 @@
 import typing as t
 from datetime import datetime
 from dataclasses import asdict, dataclass, field
+from enum import Enum
 import uuid
+
 from boto3.dynamodb.conditions import Key
 from mypy_boto3_dynamodb.service_resource import Table
-from threatexchange import content_type
 
 from threatexchange.content_type.content_base import ContentType
 from threatexchange.content_type.meta import (
@@ -217,6 +218,26 @@ class BankMember(DynamoDBItem):
         return result
 
 
+class BankMemberSignalStatus(Enum):
+    """
+    Bank member signals are processed into the respective signal_type indexes.
+    This enum governs the possible states the BankMemberSignal object can be
+    in. This influences whether or not a BankMemberSignal object is present in
+    the PendingBankMemberSignalsIndex.
+    """
+
+    # Implies that this needs to be processed now.
+    PENDING = "PENDING"
+
+    # Has been processed into an index.
+    PROCESSED = "PROCESSED"
+
+    # State machine purists will note that we do not have a PROCESSING stage
+    # which will take a lock and prevent double processing for records. However,
+    # in our case, there is no cost to double processing as long as records are
+    # processed in order. i.e chronologically.
+
+
 @dataclass
 class BankMemberSignal(DynamoDBItem):
     """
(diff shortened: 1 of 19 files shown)


hasher-matcher-actioner/hmalib/hashing/unified_hasher.py
```diff
@@ -72,7 +71,7 @@ def supports(self, content_type: t.Type[ContentType]) -> bool:
         return content_type in self.supported_content_types
 
     def get_hashes(
-        self, content_id: str, content_type: t.Type[ContentType], bytes_: bytes
```
BarrettOlson, line 75: lol at us just handing the value right back. 
I might be nice to have if we wanted to try/catch for hashing errors and log them.

hasher-matcher-actioner/hmalib/lambdas/api/bank.py
```diff
@@ -79,7 +80,14 @@ def with_preview_urls(
     return list(map(with_preview_url, bank_members))
 
 
-def get_bank_api(bank_table: Table, bank_user_media_bucket: str) -> bottle.Bottle:
+@lru_cache(maxsize=None)
```
BarrettOlson, line 83: can we move this stuff to a common file or does it lose its value then?

hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -107,9 +114,19 @@ def lambda_handler(event, context):
...
+                    # route signals to bank datastore only.
+                    bank_operations.add_bank_member_signal(
+                        banks_table=banks_table,
+                        bank_id=media.bank_id,
+                        bank_member_id=media.bank_member_id,
+                        signal_type=signal.signal_type,
+                        signal_value=signal.signal_value,
+                    )
+                    # don't write hash records etc.
+                    continue
```
BarrettOlson, line 128: nit: when not just use else?

hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -240,6 +350,57 @@ class BankedSignalEntry(DynamoDBItem):
...
+    def get_pk(cls, signal_type: t.Type[SignalType]) -> str:
+        return f"signal_type#{signal_type.get_name()}"
+
+    @classmethod
+    def get_sk(self, signal_value: str) -> str:
+        return f"signal_value#{signal_value}"
+
+    def to_dynamodb_item(self) -> t.Dict:
+        # Raise an error so that write_to_table() fails. We never want to do that.
+        raise Exception("Do not write BankedSignalEntry to DDB directly!")
```
BarrettOlson, line 363: miss java's `final` keyword? haha

hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -240,6 +350,57 @@ class BankedSignalEntry(DynamoDBItem):
...
+                # expect it to get written. AFAIK, uuids are inexhaustible, and
+                # generation performance is good enough to not worry about it.
+                ":signal_id": str(uuid.uuid4()),
+                ":signal_type": signal_type.get_name(),
+                ":signal_value": signal_value,
+            },
+            ReturnValues="ALL_NEW",
+        ).get("Attributes")
+
+        assert result is not None
```
BarrettOlson, line 396: nit: add comment for point of assert or remove e.g. "this should never happen"

hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -378,3 +539,59 @@ def remove_bank_member(self, bank_id: str, bank_member_id: str):
...
+        member_signal.write_to_table(self._table)
+        return member_signal
+
+    def unmark_bank_member_signal(self, bank_id: str, signal_id: str):
+        """
+        Remove the bank-member-signal from the pending index.
+        """
+        BankMemberSignal.unmark(self._table, bank_id=bank_id, signal_id=signal_id)
+
+    def get_bank_member_signals_to_process(
```
BarrettOlson, line 587: nit: docstring or at least a comment for the reader how this might not be full set and value wont change unless some are 'processed' ?

BarrettOlson (approved):
LGTM!

PR #812, opened 2021-09-28
tmk: handle input filenames with shell metacharacters

Hello71:
Summary
---------

otherwise, filenames with spaces fail, and malicious filenames may
potentially cause arbitrary code execution.

Test Plan
---------

ran `make` which succeeded and printed "REGRESSION PASSED"


schatten (approved):
Thanks! @Hello71

PR #815, opened 2021-09-28
clarify cimg license

Hello71:
probably the "cleanest" method would be to include it using submodules,
but that is more technically complicated. however, it is important to
notify the user that some components are under separate license, and not
imply that whole repo is using BSD license.

Test Plan
---------

IANAL, but this is obviously superior to the existing disclosure which is clearly both insufficient (LICENSE does not apply to all files in repo) and overreaching (CImg author explicitly claims that at least *some* commercial use is permitted, pursuant to the license).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD License
 
-For ThreatExchange software
+For ThreatExchange software, except pdq/cpp/CImg.h (see pdq/LICENSE):
 
 Copyright (c) 2014-present, Facebook, Inc. All rights reserved.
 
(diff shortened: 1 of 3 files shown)


Hello71:
why was this closed? if you are concerned about offering legal advice, the current text ("its license does not allow commercial use") is already legal advice by any reasonable definition, and further, it is also incorrect legal advice according to the public statements of the CImg author and a plain English reading of the CeCILL and CeCILL-C licenses. the LICENSE and pdq/LICENSE.txt files are also unambiguously incorrect or at least incomplete, as they do not properly cover all the files contained in the referent directories.

PR #816, opened 2021-09-28
Remove providers from all but the root main.tf file

### LINKED ISSUE #755: Modules should not declare their own providers. They should reuse one maintained in the root module.

schatten:
Summary
---
In the title. Meant to fix #755. But I couldn't replicate the issue that prompted me to raise the issue in the first place. However, this is a good cleanup. :) 

Bonus:
- remove pdq_signals module

Test Plan
---
Ran terraform apply. Terrafrom said no change to infrastructure so would have been a no-op apply.


BarrettOlson (approved):
Provided this works :)

schatten:
I believe it does. :)

schatten:
Will run the CI manually just to be sure.

PR #821, opened 2021-10-01
[hma][minor] Move hma-lite and example_integrations to example folder to clean up root of hma 

BarrettOlson:
Summary
---------

Just moved some files around and removed the unused `collabs_example.json` from terraform 

hma-lite is old and mostly unused but easy to confuse with hmalib so I think moving it makes sense.

Test Plan
---------

`terraform apply`


schatten (approved):
Thanks for this. I believe this is simpler. We do not want folks to start using hma-lite at all and clarifying that it is an example is the way to go.

Also, thanks for the while-I'm-here deletes!

PR #831, opened 2021-10-11
Parse variables.tf into a neat little .tfvars example friendly output

schatten:
Summary
---
Makes it easy to have single point of maintenance. Will defer to @BarrettOlson on actually using this script in CI / during release.

Test
---
```
/hasher-matcher-actioner$ hmacli write-terraform-example
```

Output
```
#  The URI for the docker image to use for the hma lambdas
hma_lambda_docker_uri = None

#  Prefix to use for resource names
prefix = hma

#  The name / acronym to use for resource names that must be globally unique (use only
#  lower case alpha a-z, and, optionally, hyphens)
organization = None

#  How long to retain cloudwatch logs for lambda functions in days
log_retention_in_days = 14

#  Send metrics to cloudwatch and build a dashboard. Useful for benchmarking, but can
#  incur costs. Set to string True for this to work.
measure_performance = False

#  Cloudwatch namespace for metrics.
metrics_namespace = ThreatExchange/HMA

#  Additional resource tags. Will be applied to ALL resources created.
additional_tags = {}

#  Indicates whether a CloudFront distribution is included
include_cloudfront_distribution = False

#  The secret token used to authenticate your access to ThreatExchange. You can find
#  this by navigating to https://developers.facebook.com/tools/accesstoken/. Leave blank
#  if you would not like to fetch from ThreatExchange
te_api_token = None

#  How long to wait between calls to ThreatExcahnge. Must be an AWS Rate Expression.
#  See here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html

fetch_frequency = 15 minutes

#  Indicates if the web app and api will use a shared user pool (generally true for
#  developers / engineers sandbox environments, otherwise false)
use_shared_user_pool = False

#  The id of the shared user pool. Used in conjunction with use_shared_user_pool set
#  to true. Generate by running terraform init & apply from /authentication-shared.

webapp_and_api_shared_user_pool_id =

#  The id of the shared user pool app client. Used in conjunction with use_shared_user_pool
#  set to true. Generate by running terraform init & apply from /authentication-shared.

webapp_and_api_shared_user_pool_client_id =

#  The system's SQS Queues have a batch_size and timeout window configured for a production
#  use case. If this var is set to true those values will be overridden and set to the
#  minimum (helpful for fast one off testing).
set_sqs_windows_to_min = False

#  Names and arns of s3 buckets to consider as inputs to HMA. All images uploaded to
#  these buckets will be processed by the hasher
partner_image_buckets = []

#  Enable the upload notfication of partner buckets if given.
enable_partner_upload_notification = False

#  Access tokens checked for in authorizer api as an alternative to cognito user tokens.

integration_api_access_tokens = []

#  Should the API gateway used with HMA be made private behind a VPC. (Either way API
#  will also require authorization)
api_in_vpc = None

#  vpc that locks down the API and UI to the specfic vpc_subnets and security_groups.
#  Required if api_in_vpc = true. Note VPC must be in the same region that HMA is deployed
#  in.
vpc_id =

#  Subnet ids of the vpc given in for vpc_id. Required if api_in_vpc = true
vpc_subnets = []

#  Security group ids to be used with the vpc given in for vpc_id. Required if api_in_vpc
#  = true
security_groups = []
```

### DIFF
--- a/hasher-matcher-actioner/hmalib/scripts/cli/write_terraform_example.py
+++ b/hasher-matcher-actioner/hmalib/scripts/cli/write_terraform_example.py
@@ -1,0 +1,75 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import argparse
+import hcl2
+
+import hmalib.scripts.cli.command_base as base
+
+MAX_LINE_LENGTH = 80
+
+
+class WriteTerraformExampleConfigCommand(base.Command):
+    """
+    Takes the root input variables declaration file in
+    hasher-matcher-actioner/teraform directory and writes an example file
+    complete with description and defaults.
+
+    Input file is hardcoded. Output is to stdin.
+    """
+
+    @classmethod
+    def init_argparse(cls, ap: argparse.ArgumentParser) -> None:
+        return super().init_argparse(ap)
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "write-terraform-example"
+
+    @classmethod
+    def get_help(cls) -> str:
+        return (
+            "Converts terraform input variable declaration files into an config file."
+        )
+
+    @staticmethod
+    def _print_single_variable(name: str, description: str, default: str, type: str):
+        at_start = True
+        line_length = 0
+
+        description_words = description.split()
+        for word in description_words:
+            line_length += 1 + len(word)
+
+            if at_start:
+                print("# ", end="")
+                at_start = False
+
+            print(f" {word}", end="")
+
+            if line_length > MAX_LINE_LENGTH:
+                at_start = True
+                line_length = 0
+                print("")
+        print()
+
+        if type == "string":
+            default = f'"{default}"'
+        elif type == "bool":
+            default = "true" if default else "false"
+
+        print(f"{name} = {default}")
+        print("")
+
+    def execute(self) -> None:
+        # Assume you are in a directory which contains a terraform directory
+        # which contains a variables.tf file.
+        with open("terraform/variables.tf", encoding="utf8") as variables_file:
+            obj = hcl2.load(variables_file)
+            for variable in obj["variable"]:
+                for var_name in variable:
+                    self._print_single_variable(
+                        name=var_name,
+                        description=variable[var_name]["description"],
+                        default=variable[var_name].get("default", None),
+                        type=variable[var_name]["type"],
+                    )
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/scripts/cli/write_terraform_example.py
```diff
@@ -0,0 +1,75 @@
...
+        elif type == "bool":
+            default = "true" if default else "false"
+
+        print(f"{name} = {default}")
+        print("")
+
+    def execute(self) -> None:
+        # Assume you are in a directory which contains a terraform directory
+        # which contains a variables.tf file.
+        with open("terraform/variables.tf", encoding="utf8") as variables_file:
```
BarrettOlson, line 66: nit: have this be an optional argument (gives it more reuse utility for submodules etc.)

hasher-matcher-actioner/hmalib/scripts/cli/write_terraform_example.py
```diff
@@ -0,0 +1,75 @@
...
+    Input file is hardcoded. Output is to stdin.
+    """
+
+    @classmethod
+    def init_argparse(cls, ap: argparse.ArgumentParser) -> None:
+        return super().init_argparse(ap)
+
+    @classmethod
+    def get_name(cls) -> str:
+        return "write-terraform-example"
```
BarrettOlson, line 26: nit/unsure: naming is hard but this feels like it is a little too ambiguous thoughts on `gen-tfvars-exanple` or `print-tfvars-example`

BarrettOlson (approved):
nits inline, not blocking given this is just an added utility users can take or leave. Thanks! please fix CI before merge though

hasher-matcher-actioner/hmalib/scripts/cli/write_terraform_example.py
schatten: Do not see a lot of re-use coming out of this, I'll move it out to a constant for now.

hasher-matcher-actioner/requirements-dev.txt
```diff
@@ -5,4 +5,5 @@ black
 moto
 pandas
 WebTest==2.0.35
-Pillow==8.3.2
\ No newline at end of file
+Pillow==8.3.2
```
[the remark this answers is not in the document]
schatten: I conjectured that you asked because if we didn't have the requirement installed, we'd hit a snag.

So, I put the import for hcl2 within a guard and warned anyone trying to use it.

PR #838, opened 2021-10-13
Reduce MAX_DESCRIPTORS_UPDATED in fetcher, increase lambda timeout to 12 minutes

dmukhg:
Summary
---------
* Allowing the fetcher to run for 12 minutes instead of 5 minutes makes more fetches possible. Was getting constant timeouts with a 50,000 update size
* Also reduced MAX_DESCRIPTORS_UPDATED to 20,000. At 50,000 even at 12 minutes, we were timing out.

Test Plan
---------
* make upload_docker and run lambda on production with a large dataset.

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/fetcher.py
@@ -192,7 +192,6 @@ def lambda_handler(_event, _context):
                 api, limit=MAX_DESCRIPTORS_UPDATED, progress_fn=ProgressLogger()
             )
         except Exception:  # pylint: disable=broad-except
-            # Don't need to call .exception() here because we're just re-raising
             logger.exception(
                 "Encountered exception while getting updates. Will attempt saving.."
             )


hasher-matcher-actioner/hmalib/lambdas/fetcher.py
```diff
@@ -167,33 +172,32 @@ def lambda_handler(event, context):
...
-        delta = indicator_store.next_delta
+            delta = indicator_store.next_delta
 
-        try:
             delta.incremental_sync_from_threatexchange(
                 api, limit=MAX_DESCRIPTORS_UPDATED, progress_fn=ProgressLogger()
             )
-        except Exception as e:
+        except Exception:  # pylint: disable=broad-except
             # Don't need to call .exception() here because we're just re-raising
```
BarrettOlson, line 195: nit/ignorable: it could be helpful to log the exception object as well and just not raise it along with the error message below

BarrettOlson: great! (sanity check: and not trigger a retry?)

PR #839, opened 2021-10-14
[hma] fixes/clean (pdq.te and cloudfront)

BarrettOlson:
Summary
---------

### [`3fb74a05`](https://github.com/facebook/ThreatExchange/pull/839/commits/3fb74a05a319357de14a488e06a4e2e47cfcf631) PHASE_OUT_THIS_INTERNAL_DETAIL e.g. removed `pdq_file_extension` var from everywhere as it is built into python-threatexchange now
- note `hash_count` of `_get_threat_exchange_datasets` is already broken (as it was using this variable) added TODO

### [`8f081388`](https://github.com/facebook/ThreatExchange/pull/839/commits/8f081388eff2db7d6d2f129168e6c684339eeb4d) Changed so that if `include_cloudfront_distribution` the http access of the static s3 files should be disabled (updates bucket ACL: 
- note: this change might not take effect unless the bucket is destroyed due to weirdness in post create ACL changes. 

### [`d66fea6b`](https://github.com/facebook/ThreatExchange/pull/839/commits/d66fea6b7b21ca8514dc8e6954a24c7741bc7c17) fix depends_on of sample data to wait until sns is subscribed as well (bug in index file not being created for samples)

Test Plan
---------

Programatic: 
```
make upload_docker
make dev_test_intance
```

Manual testing:
- Test of UI pages - submit, matches, settings, details
- Confirmed that `sync` button on UI now creates privacy groups with new defaults
- Confirmed that destroy apply now makes an index file of sample data again. 
- Confirmed that attempting to access the s3 website outside of cloudfront url (when enabled) results in 403 response.

### DIFF
--- a/hasher-matcher-actioner/terraform/indexer/main.tf
+++ b/hasher-matcher-actioner/terraform/indexer/main.tf
@@ -134,6 +134,8 @@ resource "null_resource" "provide_sample_pdq_data_holidays" {
   # To get a sensible privacy group value, we reverse engineer the filename split at
   # hmalib.common.s3_adapters.ThreatExchangeS3Adapter._parse_file at line 118
   depends_on = [
+    aws_lambda_permission.indexer,
+    aws_sns_topic_subscription.indexer,
     aws_lambda_function.indexer
   ]
 


hasher-matcher-actioner/terraform/indexer/main.tf
```diff
@@ -134,6 +134,8 @@ resource "null_resource" "provide_sample_pdq_data_holidays" {
   # To get a sensible privacy group value, we reverse engineer the filename split at
   # hmalib.common.s3_adapters.ThreatExchangeS3Adapter._parse_file at line 118
   depends_on = [
+    aws_lambda_permission.indexer,
+    aws_sns_topic_subscription.indexer,
     aws_lambda_function.indexer
```
schatten, line 139: I thought the lambda function would auto-depend on lambda_permission. I've been known to be wrong. :P

PR #843, opened 2021-10-15
[hma] have updates to tables recored via streams

BarrettOlson:
Summary
---------

This takes the work in @schatten started with #842 and adds a few fixes to make it merge ready. It also delineates stream events based on 3 types (ignoring the rest)
- ContentObject records
- HashRecords
- MatchRecords

It determines that kind of record based on SK and then increments  the counts.
  
Test Plan
---------

Submitted to HMA using various methods (matching and not) and made sure the ddb record counts were incremented correctly:
Screen Shot 2021-10-15 at 1 08 51 PM
Screen Shot 2021-10-15 at 1 15 17 PM
Screen Shot 2021-10-15 at 1 16 26 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/ddb_stream_counter.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/ddb_stream_counter.py
@@ -7,8 +7,10 @@ from functools import lru_cache
 from mypy_boto3_dynamodbstreams.type_defs import GetRecordsOutputTypeDef, RecordTypeDef
 import boto3
 
+from hmalib.common.models.content import ContentObject
 from hmalib.common.models.count import AggregateCount
 from hmalib.common.logging import get_logger
+from hmalib.common.models.models_base import DynamoDBItem
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/ddb_stream_counter.py
```diff
@@ -0,0 +1,125 @@
...
+    cls.table_name(): cls
+    for cls in [PipelineTableStreamCounter, BanksTableStreamCounter]
+}
+
+current_stream_counter = ENABLED_STREAM_COUNTERS[SOURCE_TABLE_NAME]
+
+
+def lambda_handler(event: GetRecordsOutputTypeDef, _context):
+    """
+    # How do entity counts work in HMA?
```
mengyangwang, line 99: The comment is so clear!

BarrettOlson:
> Everything looks good to me! Just want to know will we ever run into concurrency issues?

Because we use `update_item` and the calls also have 0 handling UpdateExpression, I think this is handled at least well enough for fast aggregate counting. 
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.AtomicCounters
> You can use the UpdateItem operation to implement an atomic counter—a numeric attribute that is incremented, unconditionally, without interfering with other write requests. (All write requests are applied in the order in which they were received.) 

If we need every greater precision  we can always go back to the tables these streams are based on.

PR #848, opened 2021-10-19
add NonThreatExchangeConfig model

mengyangwang:
Summary
---------

Adding a new model, which will be used to control future NonTE Fetcher

Test Plan
---------

black hmalib
mypy hmalib
make docker
make upload_docker
test in localhost to make sure all configs functions are working fine


BarrettOlson (changes requested):
Naming is too specific imo

hasher-matcher-actioner/hmalib/common/configs/fetcher.py
```diff
@@ -26,3 +26,17 @@ class ThreatExchangeConfig(HMAConfig):
...
         return self.name
+
+
+@dataclass
+class NonThreatExchangeConfig(HMAConfig):
+    """
+    Config for the NonThreatExchange datasets(i.e. StopNCII hashes)
+    """
+
+    fetcher_active: bool
```
BarrettOlson, line 37: looks like the majority of these fields are shared with `ThreatExchangeConfig` is it possible to use a shared base class?

mengyangwang:
> LGTM. You confirmed the fetcher and other setting still work as expected and don't effect underlying storage? (e.g. existing configs wont be affected by the change? Based on the fact we use keyword args everywhere I think it should be fine.)

yeah I verified all existing configs will not be impacted

PR #850, opened 2021-10-20
[hma] add try-catch on hasher to avoid pointless tries

BarrettOlson:
Summary
---------

see title.

Test Plan
---------

Before:
submit to HMA with values that will error (bad url or file format)
see that it is being retried for no reason.

After:
submit to HMA with values that will error (bad url or file format)
see that error is logged once but not repeated and other hashes are sent to matcher.


mengyangwang (approved):
LGTM! Good catch!

PR #852, opened 2021-10-20
Add fetcher interface

mengyangwang:
Summary
---------

add a non-threatexchange api to fetch data from, will bump the version file after this is merged

Test Plan
---------

black hmalib
python -m mypy hmalib
pytest test_non_threatexchange_api.py

 add a new unitest test_non_threatexchange_api.py to test a hash get call from the very beginning.
Screen Shot 2021-10-21 at 9 21 42 PM

Screen Shot 2021-10-21 at 9 22 37 PM

### DIFF
--- a/hasher-matcher-actioner/hmalib/hashexchanges/api/non_threatexchange_api.py
+++ b/hasher-matcher-actioner/hmalib/hashexchanges/api/non_threatexchange_api.py
@@ -26,7 +26,7 @@ class NonThreatExchangeAPI(BaseAPI):
         Returns a paginated list of all hash records from start_timestamp.
         """
 
-        params = {
+        params: t.Dict[str, t.Union[str, int]] = {
             "startTimestamp": start_timestamp,
             "pageSize": page_size,
         }


python-threatexchange/threatexchange/api.py
```diff
@@ -601,3 +601,101 @@ def get_threat_descriptors_from_indicator(
...
+        """
+        Perform an HTTP GET request, and return the JSON response payload.
+        Same timeouts and retry strategy as `_get_session` above.
+        """
+        with self._get_session() as session:
+            response = requests.get(url, params=params or {}, headers=headers or {})
+            response.raise_for_status()
+            return response.json(object_hook=json_obj_hook)
+
+    def _get_session(self):
```
BarrettOlson, line 636: a lot methods look like complete copy of the ones above docstring and all. Can we consolidate?

python-threatexchange/threatexchange/api.py
```diff
@@ -601,3 +601,101 @@ def get_threat_descriptors_from_indicator(
         response = self.get_json_from_url(url)
 
         return response["descriptors"]["data"]
+
+
+class NonThreatExchangeAPI:
+    _BASE_URL = "https://nciiapi.azure-api.net/api"
```
BarrettOlson, line 607: . 
have this be a parameter?

python-threatexchange/threatexchange/api_representations.py
```diff
@@ -45,3 +46,49 @@ def from_graph_api_dict(cls, d: dict) -> "ThreatPrivacyGroup":
...
+        return cls(
+            d.get("source", None),
+            d.get("feedbackValue", None),
+            d.get("tags", []),
+        )
+
+
+@dataclass
+class HashRecord:
+    lastModtimestamp: int
```
BarrettOlson, line 68: nit: can we store the timestamp as datetime within the data model?

python-threatexchange/threatexchange/non_threatexchange_api.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import typing as t
+from .api_representations import HashRecord
+from .api import ThreatExchangeAPI
+
+
+class NonThreatExchangeAPI(ThreatExchangeAPI):
```
schatten, line 8: This feels very clunky. Why is `NonThreatExchangeAPI ` a subclass of `ThreatExchangeAPI`? Feels semantically disconnected.

I understand that this is what you're trying to do:
```
python-threatexchange

ThreatExchangeAPI ----> NonThreatExchangeAPI

hmalib 

Fetcher uses a class of ThreatExchangeAPI
```

But this will very quickly run into issues. Because NonThreatExchangeAPI does not belong in python-threatexchange.

Instead, how about this:

Let's not touch ThreatExchangeAPI at all. Imagine the OSS hash-exchange provided its own python API package and you had to create a middle layer to abstract `python-threatexchange` and the `oss-hash-exchange-api` package.

You'd need to do that in hmalib.

python-threatexchange/threatexchange/non_threatexchange_api.py
```diff
@@ -0,0 +1,41 @@
...
+    ) -> t.List[HashRecord]:
+        """
+        Returns a paginated list of all hash records from start_timestamp.
+        """
+
+        params = {
+            "startTimestamp": self.start_timestamp,
+            "pageSize": self.page_size,
+        }
+        url = self._get_graph_api_url("FetchHashes", params)
```
schatten, line 35: This is really confusing. The OSS API is *not* a graph API.

python-threatexchange/threatexchange/non_threatexchange_api.py
mengyangwang: I was trying to do this in hmalib first, but then I found there are lots of common methods I can use in ThreatExchangeAPI, so that's why I put both them in the same place. But you're right, it's confusing! I will create a common API class in hmalib so that both ThreatExchangeAPI and NonThreatExchangeAPI can extend it.

PR #854, opened 2021-10-26
Support parameterized counts in stream counters

dmukhg:
Summary
---
At this point, only adds support for and test a 'buffer' which receives increments and decrements for both aggregate and parameterized counts. It flushes only on-demand.

Next steps would be to identify which parameterized counts we want to maintain and inc them in the lambda's `PipelineTableStreamCounter.update_increments_for_records(...)`

Test Plan
---
```
$ python -m black hmalib tests
$ python -m mypy hmalib
```

Added a unittest with a mock ddb table for counts.
```
$ python -m py.test
```

Verified by submitting an image that the ddb counts are updated.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/count.py
+++ b/hasher-matcher-actioner/hmalib/common/models/count.py
@@ -15,15 +15,21 @@ class BaseCount:
     """
 
     def get_pkey(self) -> str:
-        """Get partition key for this count."""
+        """
+        Get partition key for this count.
+        """
         raise NotImplementedError
 
     def get_skey(self) -> str:
-        """Get sort key for this count."""
+        """
+        Get sort key for this count.
+        """
         raise NotImplementedError
 
     def get_value(self, table: Table) -> int:
-        """Get current value for the counter."""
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/common/models/count.py
```diff
@@ -79,3 +92,153 @@ def get_pkey(self) -> str:
...
+
+    At this point, only supports a single parameter value.
+
+    Think about it as
+        ParameterizedCount(of="hma.pipeline.matches", by="privacy_group"), value="4567896456789000976"))
+    or  ParameterizedCount(of="hma.pipeline.hashes", by="content_type", value="photo")
+    or  ParameterizedCount(of="hma.pipeline.hashes", by="signal_type", value="pdq")
+    """
+
+    SKEY_PREFIX = "val#4"
```
BarrettOlson, line 110: We missed a random 4 @schatten

schatten: Nooo!!!

PR #857, opened 2021-11-04
Remove the use of sks keyservers

garnesen:
Summary
---------

The key server sks-keyservers.net has been deprecated for some time which means the only valid keyserver in the Dockerfile is pgp.mit.edu. This change adds two other key servers that are verified working in replacement of the sks ones.

When I first tried to get the devcontainer working for HMA, I was getting errors that none of the keyservers were working which lead me down this rabbit hole. An hour later, the mit one was working 🤷 , but the sks links definitely do not work anymore.

From the sks-keyservers.net page:
> This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.

Test Plan
---------

Verified both new key servers work locally during the devcontainer setup:
```
#9 0.550 + gpg --batch --keyserver hkp://keyserver.ubuntu.com --recv-keys 94AE36675C464D64BAFA68DD7434390BDBE9B9C5
#9 0.747 gpg: key 7434390BDBE9B9C5: public key "Colin Ihrig <cjihrig@gmail.com>" imported
#9 0.748 gpg: Total number processed: 1
#9 0.748 gpg:               imported: 1
```
```
#9 1.604 + gpg --batch --keyserver hkp://keys.openpgp.org --recv-keys 94AE36675C464D64BAFA68DD7434390BDBE9B9C5
#9 2.308 gpg: key 7434390BDBE9B9C5: public key "Colin Ihrig <cjihrig@gmail.com>" imported
#9 2.309 gpg: Total number processed: 1
#9 2.309 gpg:               imported: 1
```

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/Dockerfile
+++ b/hasher-matcher-actioner/.devcontainer/Dockerfile
@@ -53,8 +53,8 @@ RUN ARCH= && dpkgArch="$(dpkg --print-architecture)" \
   108F52B48DB57BB0CC439B2997B01419BD92F80A \
   B9E2F5981AA6E0CD28160D9FF13993A75599653C \
   ; do \
-  gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys "$key" || \
-  gpg --batch --keyserver hkp://ipv4.pool.sks-keyservers.net --recv-keys "$key" || \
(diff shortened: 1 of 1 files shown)


BarrettOlson (approved):
Thanks for patching this!

PR #859, opened 2021-11-05
Support bank members without media attached

schatten:
Summary
---
To support the upcoming fetcher overhaul which requires signals without members, we need this capability.

Test
---
```
$ python -m black hmalib
$ python -m mypy hmalib
$ python -m py.test
```


mengyangwang (approved):
LGTM!

PR #860, opened 2021-11-08
Use parameterized counts for the HMADataStore stream counter

garnesen:
Summary
---------

Adds the parameterized count to the HMADataStore stream counter.

Test Plan
---------

Updated my lambda function to use these code changes and testing submitting a photo and video.

Test data added (images/videos were uploaded via the HMA UI, the match was added to dynamo manually):
image

Resulting counts:
image

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/ddb_stream_counter.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/ddb_stream_counter.py
@@ -89,12 +89,22 @@ class PipelineTableStreamCounter(BaseTableStreamCounter):
             # TODO These should maybe have a strong connection to the objects instead of class constants
             # otherwise changes to the object might not correctly update these count checks
             if sk == ContentObject.CONTENT_STATIC_SK:
-                count_buffer.inc_aggregate(AggregateCount.PipelineNames.submits)
+                content_type = record["dynamodb"]["NewImage"]["ContentType"]["S"]
+                count_buffer.inc_parameterized(
+                    AggregateCount.PipelineNames.submits, "content_type", content_type
+                )
             elif sk.startswith(DynamoDBItem.TYPE_PREFIX):
                 # note hashes should always match submits (submits == hashes)
-                count_buffer.inc_aggregate(AggregateCount.PipelineNames.hashes)
+                signal_type = record["dynamodb"]["NewImage"]["SignalType"]["S"]
+                count_buffer.inc_parameterized(
+                    AggregateCount.PipelineNames.hashes, "signal_type", signal_type
+                )
             elif sk.startswith(DynamoDBItem.SIGNAL_KEY_PREFIX):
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/hmalib/lambdas/ddb_stream_counter.py
```diff
@@ -85,16 +85,32 @@ def update_increments_for_records(
 
             pk: str = record["dynamodb"]["Keys"]["PK"]["S"]
             sk: str = record["dynamodb"]["Keys"]["SK"]["S"]
+            # The raw stream values have a different schema than what is returned by boto3 meaning
+            # they can't be deserialized into one of the HMA datamodels using existing functionality.
+            inserted_record = record["dynamodb"]["NewImage"]
```
BarrettOlson, line 90: What happens when the record does not have "NewImage" as a key in the sub-dict?
I worry this might error for count increments that aren't around images. Can you confirm match and hash records are still incremented correctly after this change is deployed?

garnesen: I'll preface by saying this is my first time working with Dynamo DB. My understanding is that since we are only handling "INSERT" events here, there should always be a "NewImage" field that contains the data that was added. It's still probably safer to check though? On that note, should I also be safer when accessing the fields within `inserted_record`?

garnesen: I could add something like
```python
if (content_type := inserted_record.get("ContentType", {}).get("S")) is not None:
    count_buffer.inc_parameterized(
        AggregateCount.PipelineNames.submits, "content_type", content_type
    )
```

BarrettOlson: Good point, I was finally able to trace down the format in the docs.
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_StreamRecord.html

It looks like it is possible for INSERT events not have NewImage if the ViewType is KEYS_ONLY, however given this is something we define in terraform and the later existing checks we have would break anyway. I think it is fine as is.

BarrettOlson: However the checks you are consider for `inserted_record` sound good to me. As it's possible those change could be missed and it would be better if those errors only affected parameterized counts instead of erroring on all records from this innovation. Thanks!

BarrettOlson: However the checks you are consider for inserted_record sound good to me. As it's possible those change could be missed and it would be better if those errors only affected parameterized counts instead of erroring on all records from this innovation. Thanks!

BarrettOlson (approved):
LGTM, thanks!

PR #861, opened 2021-11-09
[hma] add hma_extension entry point for custom impl of ActionPerformer

BarrettOlson:
Summary
---------

Create the initial structure such that custom code written outside of hmalib can be used and run by hmalib lambda's in this first case `hmalib/lambdas/actions/action_performer.py`

More typing, documenting, etc are in the works but want to get input on this approach now.

Test Plan
---------

### Test locally:
```python3
from hmalib.common.extensions import load_extension_impl

fn = load_extension_impl("echo")

fn("Hello World!")
>> Hello World
```

### Test Deploy:
```bash
make upload_docker 
``` 
then
Manually created an ActionPerformer & ActionRule in my <prefix>-HMAConfig table:

ConfigType | ConfigName | action_label | config_subtype | entry_point_name | must_have_labels | must_not_have_labels
-- | -- | -- | -- | -- | -- | --
ActionPerformer | CustomImplAction1 |   | CustomImplActionPerformer | echo |   |   |  
ActionRule | trigger-on-tag-holidays_jpg1_dataset-custom | {"value":{"S":"CustomImplAction1"},"key":{"S":"Action"}} |   |   | [{"M":{"value":{"S":"holidays_jpg1_dataset"},"key":{"S":"Classification"}}}] | [] |  

Submitted the b.jpg photo the triggers match and manually created action.

Made sure the ActionPerformer lambda cloudwatch logs have the result of the echo extension printed along with the action history record for the content_id.

Note: right now manually creating this action will break the UI settings page for actions. 

Updates since publish
---------
- action_performer_iam_role_arn now part of tf outputs
- add some missing returns to `hmalib/scripts/common/utils.py` post requests
- `hmalib/common/configs/actioner.py` was incorrectly typed to handle`MatchMessage` even though action_eval sends its child class `ActionMessage` fixed this.
- Added kwargs field to new action type. 
(repeated test plan after changes)

### DIFF
--- a/hasher-matcher-actioner/hmalib_extensions/main.py
+++ b/hasher-matcher-actioner/hmalib_extensions/main.py
@@ -1,0 +1,30 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+This file holds simple hma_extension implementations. 
+Primary as an example for incorporating others.
+
+To use an extention you must specific a new entry point in 
+setup.py 's entry_points group "hmalib_extensions"
+
+For example the `print_arg` method below has been added to setup.py as
+  "echo = hmalib_extensions.main:print_arg"
+
+Once added hmalib can load these extentions via
+`hmalib.common.extensions.load_extension_impl` 
+e.g. 
+```python3
+from hmalib.common.extensions import load_extension_impl
+
+fn = load_extension_impl("echo")
+
+fn("Hello World!")
+>> Hello World
+
+```
+"""
+
+
+def print_arg(arg):
+    """Example method use could use to test basic hma_extension support"""
+    print(arg)
--- a/hasher-matcher-actioner/setup.py
+++ b/hasher-matcher-actioner/setup.py
@@ -30,5 +30,8 @@ setup(
         "requests>=2.25.1",
     ],
     extras_require=extras_require,
-    entry_points={"console_scripts": ["hmacli = hmalib.scripts.cli.main:main"]},
+    entry_points={
+        "console_scripts": ["hmacli = hmalib.scripts.cli.main:main"],
+        "hmalib_extensions": ["echo = hmalib_extensions.main:print_arg"],
+    },
 )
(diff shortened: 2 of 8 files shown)


hasher-matcher-actioner/setup.py
```diff
@@ -30,5 +30,8 @@
         "requests>=2.25.1",
     ],
     extras_require=extras_require,
-    entry_points={"console_scripts": ["hmacli = hmalib.scripts.cli.main:main"]},
+    entry_points={
+        "console_scripts": ["hmacli = hmalib.scripts.cli.main:main"],
+        "hmalib_extensions": ["echo = hmalib_extensions.main:print_arg"],
```
schatten, line 35: Why is this step necessary? Why not make everything importable via their full path. Eg. instead of doing load_extension("echo"), I could be doing load_extension("hmalib_extensions.main:print_arg") instead.

BarrettOlson: IMO the mappings is helpful in the context of creating the action on the UI or otherwise. It keeps things a little clean in if in the settings UI if you can specify the short and specific with `entry_point_name="echo"` for example

schatten: > IMO the mappings is helpful in the context of creating the action on the UI or otherwise. It keeps things a little clean in if in the settings UI if you can specify the short and specific with `entry_point_name="echo"` for example

I get the brevity aspect, but in this case we had made it necessary for partners to make changes to a file that's checked in to version control. This will potentially make upgrading an issue. Urge you to think of an alternate way to retain the vanity -> actual method mapping that does not touch source control.

hasher-matcher-actioner/hmalib_extensions/main.py
```diff
@@ -0,0 +1,30 @@
...
+fn = load_extension_impl("echo")
+
+fn("Hello World!")
+>> Hello World
+
+```
+"""
+
+
+def print_arg(arg):
```
schatten, line 28: Say I want print_arg to get `content_id`, and the `labels`. How would I configure that?

BarrettOlson:
restructuring. changes to be added to future PR

PR #872, opened 2021-12-02
Run indexer on a schedule instead of on events

schatten:
Banks will not emit update events for every member added/removed. Instead indexer will run on a schedule and:
a) if rebuilding, get all datafiles for TE and all members for banks.
b) if updating, get updated datafiles for TE and updated members for banks.

Note, this PR only reads existing s3 datafiles on a schedule. The banks related things will happen as part of a later PR.

To ensure we don't develop what is unnecessary, we'll start off with the rebuilding arc, do some performance tests to see at what data size rebuilding is ineffective and then work on delta updates.

Test Plan
---
1. Deployed to my instance.
2. Deleted index/ prefix objects in the s3 bucket
3. Manually ran the indexer
4. Verified that the index was re-created

### DIFF
--- a/hasher-matcher-actioner/terraform/indexer/main.tf
+++ b/hasher-matcher-actioner/terraform/indexer/main.tf
@@ -1,16 +1,16 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 # First define the Indexing Schedule
-resource "aws_cloudwatch_event_rule" "indexing_schedule" {
-  name                = "${var.prefix}-HMA_IndexingSchedule"
+resource "aws_cloudwatch_event_rule" "indexing_trigger" {
+  name                = "${var.prefix}-HMA_IndexingTrigger"
   description         = "Rebuild Index on a regular cadence"
   schedule_expression = "rate(${var.indexer_frequency})"
-  role_arn            = aws_iam_role.indexing_schedule.arn
+  role_arn            = aws_iam_role.indexing_trigger.arn
 }
 
-resource "aws_iam_role" "indexing_schedule" {
-  name_prefix        = "${var.prefix}_indexing_schedule"
-  assume_role_policy = data.aws_iam_policy_document.indexing_schedule_assume_role.json
+resource "aws_iam_role" "indexing_trigger" {
+  name_prefix        = "${var.prefix}_indexing_trigger"
+  assume_role_policy = data.aws_iam_policy_document.indexing_trigger_assume_role.json
 
   tags = merge(
     var.additional_tags,
@@ -20,7 +20,7 @@ resource "aws_iam_role" "indexing_schedule" {
   )
 }
 
-data "aws_iam_policy_document" "indexing_schedule_assume_role" {
+data "aws_iam_policy_document" "indexing_trigger_assume_role" {
   statement {
     effect  = "Allow"
     actions = ["sts:AssumeRole"]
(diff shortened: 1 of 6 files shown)


hasher-matcher-actioner/terraform/variables.tf
```diff
@@ -58,6 +58,12 @@ variable "fetch_frequency" {
   default     = "15 minutes"
 }
 
+variable "indexer_frequency" {
```
BarrettOlson, line 61: update tfvars.example please

BarrettOlson:
hmm, These changes seem a tad wasteful until we have incremental updates. As I understand it instead of waiting for specific files to update now we just rebuild everything ever 15 minutes(adjustable)? 

It also adds another layer of latency just under 30min in the worst case (15 for fetch and then 15 index) by default. 
Can we have both options available? I understand not wanting to rebuild for every bank update but for time critical signals from TE I think event based updating is a nice feature to have as an option (and without needs to change tf state for to temporarily increased frequency).

schatten:
> hmm, These changes seem a tad wasteful until we have incremental updates. As I understand it instead of waiting for specific files to update now we just rebuild everything ever 15 minutes(adjustable)?

I'd consider this to be wasteful only if the time (CPU/wall-clock) to rebuild index is significant. By the time we are done with banks fully, even threatexchange hashes will go into banks, so the evented way will not be useful anymore.

> It also adds another layer of latency just under 30min in the worst case (15 for fetch and then 15 index) by default. Can we have both options available? I understand not wanting to rebuild for every bank update but for time critical signals from TE I think event based updating is a nice feature to have as an option (and without needs to change tf state for to temporarily increased frequency).

My plan is to surface the index as a construct in the UI too and allow it to be managed (rebuilt on demand) from there. That will hopefully fill the gap. I'm wary of trying to do delta updates to the index or trying to make index rebuilding more optimized until we get some hard numbers on what the cost (compute / time) is.

That's also planned in subsequent PRs from my workstream.

hasher-matcher-actioner/terraform/indexer/main.tf
```diff
@@ -1,5 +1,72 @@
...
+}
+
+## Create a permission policy from policy document
+resource "aws_iam_policy" "indexing_schedule" {
+  name_prefix = "${var.prefix}_indexing_schedule_role_policy"
+  description = "Permissions for Indexing Schedule"
+  policy      = data.aws_iam_policy_document.indexing_schedule.json
+}
+
+## Attach a permission policy to the fetech trigger role
```
BarrettOlson, line 55: warning copypasta detected

hasher-matcher-actioner/terraform/indexer/main.tf
```diff
@@ -1,5 +1,72 @@
...
+      identifiers = ["events.amazonaws.com"]
+    }
+  }
+}
+
+## Define a policy document to asign to the role
+data "aws_iam_policy_document" "indexing_schedule" {
+  statement {
+    actions   = ["lambda:InvokeFunction"]
+    resources = ["*"]
```
BarrettOlson, line 38: seems a tad dubious to allow the trigger/scheduler to invoke all lambdas, can we put the unified indexer lambda arn here? (I know the issue also exists for the fetcher)

BarrettOlson (approved):
Fair points. 
Accepted, added a few nonblocking comments inline 
Also nit: I think we might miss in the future that `indexing_schedule` and `fetcher_trigger` are basically doing the same thing given how different the names are, might be worth an additional comment at least. Is it not worth having them share a reusable module given the change you have in mind in the next PRs? 

One request before merge though:
(if you haven't already) verify that running the lambda from the aws console (or even cli) works to build just the sample data index as well? (Bonus if you can add a note somewhere on how to do it quickly). Reason: Ideally a fresh deploy won't have to wait 15 minutes by default before a basic match test can be run. 

Thanks!

schatten:
@BarrettOlson Have addressed a bunch of issues you flagged. 

Still in progress <because of weirdness on my local and I'm still working on indexing>:
- verify that sample indexes are in-fact built from 0 state.

PR #873, opened 2021-12-02
[pytx] update pdq_index wrapper and add typing CI for python-threatexchange 

BarrettOlson:
Summary
---------

This PR was too unwieldy so I broke it up into different parts. Leaving for posterity.

mypy before:
> Found 104 errors in 25 files (checked 51 source files)

mypy after:
> Success: no issues found in 51 source files

Commits moved to separate PR:  #875
------
### [`b8156ff1`](https://github.com/facebook/ThreatExchange/pull/873/commits/b8156ff173cc57c6260e2156ef919e968733827d) spelling and have query support custom threshold
- ~~[Small] change that allows propagation of threshold to query in pdq index~~
- instead we want to return the distance

### [`4ede5ea5`](https://github.com/facebook/ThreatExchange/pull/873/commits/4ede5ea588d0294c47b586bb07ebc66d4008d83a) fix typing and bugs in unused parts of pdq_index.py while adding flat_index option
- Start of type fixes (found bug in `dataset.py` `load_index` but it is unused in hma anyway)
- ~~Adds `flat_index: bool = False` option to `PDQIndex` which will build `PDQFlatHashIndex` instead of `PDQMultiHashIndex`~~
  - top commit changed this to use `_get_empty_index` instead and added `PDQFlatIndex`

Commits moved to separate PR: #876
------
### [`d71ea72e`](https://github.com/facebook/ThreatExchange/pull/873/commits/d71ea72e448749a53d9b74ddac7ab330e9c85026) update url md5 to have correct type signature and complete docstring
- [Small] used updated var name to make mypy happy 

### [`85fab77f`](https://github.com/facebook/ThreatExchange/pull/873/commits/85fab77fd6e06344c7651e55a8ee80cdadd53f21) add tlsh support to make INDICATOR_TYPE=HASH_TEXT_TLSH
- [Small-ish] remove fake state hack from `python-threatexchange/threatexchange/signal_type/tlsh_pdf.py` and update `INDICATOR_TYPE` + tests
  - once the sample data exists this should allow for tlsh hash matching on those signals

Commits moved to separate PR:  #877
------
### [`9a46145d`](https://github.com/facebook/ThreatExchange/pull/873/commits/9a46145d8dd9913ecca72cfdfd2492f5a9b57bdf) update dockerfile with steps for testing local pytx changes
- [Small] comment only change in Docker added because how to test local pytx changes for hma without a version bump/push is common enough that documenting how to do so seems worth it.

Commits moved to separate PR:  #878
---------
### [`a9300673`](https://github.com/facebook/ThreatExchange/pull/873/commits/a930067316bf87c7909ca6e25bdc85e6cd2c746f) fix mypy issues throughout pytx
- [Large] mostly simple type fixes. 
- Tricky/dubious changes maybe worth a second look:
  -  `python-threatexchange/threatexchange/cli/match.py` : `parse_input` - generators are tricky
  - `python-threatexchange/threatexchange/cli/dataset_cmd.py` last line - now pass signal_type instance instead
 
### [`61a91350`](https://github.com/facebook/ThreatExchange/pull/873/commits/61a9135063fec286f7b5b992b33fa10d1f76b9c6) fix most typing but need to test
- [Large] fixes typing in rest of `python-threatexchange/threatexchange`
- Tricky/dubious changes maybe worth a second look:
  - added a bunch of `None` checks for `self.checkpoint` in `python-threatexchange/threatexchange/threat_updates.py` that otherwise would have errored.
    - particularly unsure about `return self.checkpoint.stale if self.checkpoint else False` for stale (think maybe True is better but wanted to make the falseyness on None)
   - `self.start_time` doesn't exist at all in `python-threatexchange/threatexchange/cli/tag_fetch.py`. based on `start_time_str` I think `fetch_type.from_timestamp` was the expected value for that `or` expression
   - `delta.tries` in `python-threatexchange/threatexchange/threat_updates.py` : `merge` also isn't a thing (anymore at least)...
   - some float to int casting for time values

### [`0b62029f`](https://github.com/facebook/ThreatExchange/pull/873/commits/0b62029f1c226345374acb254e76f48634a75629) suppress last set of error and add type checking to python-threatexchange
- [Minor] added mypy checks to github actions for `python-threatechange/threatexchange` (had to force push this amended commit a few times to get it to work because I couldn't repro locally: fix was updating setup.py with type-stubs)

Test Plan
---------

### python-threatexchange testing

Played around with various `threatexchange` command and made sure they still work. examples of checks made include:

```
make local install
pytest
# green with expected warnings
export TX_ACCESS_TOKEN="<redacted>"
threatexchange fetch 
    Processed 0 updates:
    Rebuilding match indices...
threatexchange hash photo data/sample-b.jpg
    photo_md5 d35c785545392755e7e4164457657269
    /Users/barrett/.venv/hma/lib/python3.8/site-packages/pdqhash/__init__.py:3: UserWarning: Hash vector order changed between version 0.1.8 and 0.2.0. See https://github.com/faustomorales/pdqhash-python/issues/1 for more details.
      warnings.warn(
    pdq f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22
    pdq_ocr f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22, 
threatexchange match text "bball now?" 
    <ID1> raw_text disputed media_priority_samples
    <ID2> trend_query disputed media_priority_samples
threatexchange -c no_commit.config.json fetch # no_commit.config.json has a nonsample dataset id)
# fetches updates correctly
threatexchange hash --signal-type url_md5 --as-text  url 'https://www.facebook.com/?user=123' 
    url_md5 e359430911fe80c2dd526d3cca21da30
```

### hma testing 

Verified that pytx changes did not effect hma via the following steps.

1) followed new steps in Dockerfile to push changes to my prefix of hma
2) made sure `make dev_test_instance` works
3) deleted threatexchange data and indexes from s3
  3a) made sure the fetcher was still able to get the sample datasets
  3b) made sure the indexer was still able to build the index 
and update 
4) submit manually add make sure a match is still found.

### DIFF
--- a/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
@@ -53,6 +53,7 @@ class PDQHashIndex(ABC):
         queries: t.Sequence[PDQ_HASH_TYPE],
         threshhold: int,
         return_as_ids: bool = False,
+        **kwargs
     ):
         """
         Searches this index for PDQ hashes within the index that are no more than the threshold away from the query hashes by
@@ -92,7 +93,7 @@ class PDQHashIndex(ABC):
             # for custom ids, we understood them initially as uint64 numbers and then coerced them internally to be signed
             # int64s, so we need to reverse this before returning them back to the caller. For non custom ids, this will
             # effectively return the same result
-            output_fn = int64_to_uint64
+            output_fn: t.Callable[[int], t.Any] = int64_to_uint64
         else:
             output_fn = self.hash_at
 
@@ -209,9 +210,15 @@ class PDQMultiHashIndex(PDQHashIndex):
             return faiss.downcast_IndexBinary(self.faiss_index.index)
         return self.faiss_index
 
-    def search(self, queries: t.Sequence[PDQ_HASH_TYPE], threshhold: int, **kwargs):
+    def search(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+        **kwargs
+    ):
         self.mih_index.nflip = threshhold // self.mih_index.nhash
-        return super().search(queries, threshhold, **kwargs)
+        return super().search(queries, threshhold, return_as_ids, **kwargs)
 
     def hash_at(self, idx: int):
         i64_id = uint64_to_int64(idx)
--- a/python-threatexchange/threatexchange/signal_type/pdq_index.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq_index.py
@@ -31,15 +31,15 @@ class PDQIndex(SignalTypeIndex):
     def __len__(self) -> int:
         return len(self.local_id_to_entry)
 
-    def query(self, hash: str) -> t.List[IndexMatch[T]]:
+    def query(
+        self, hash: str, threshhold=PDQ_CONFIDENT_MATCH_THRESHOLD
+    ) -> t.List[IndexMatch[T]]:
         """
         Look up entries against the index, up to the max supported distance.
         """
 
         # query takes a signal hash but index supports banch quries hence [hash]
-        results = self.index.search(
-            [hash], self.PDQ_CONFIDENT_MATCH_THRESHOLD, return_as_ids=True
-        )
+        results = self.index.search([hash], threshhold, return_as_ids=True)
         matches = []
         for result_ids in results:
             for id in result_ids:
--- a/python-threatexchange/threatexchange/signal_type/pdq_index.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq_index.py
@@ -21,20 +21,20 @@ from threatexchange.hashing.pdq_faiss_matcher import (
 class PDQIndex(SignalTypeIndex):
     """
     Wrapper around the pdq faiss index lib
+    using PDQMultiHashIndex
     """
 
     PDQ_CONFIDENT_MATCH_THRESHOLD = 31
     T = IndexT
 
-    def __init__(
-        self, entries: t.Iterable[t.Tuple[str, IndexT]], flat_index: bool = False
-    ) -> None:
+    @classmethod
+    def _get_empty_index(cls) -> PDQHashIndex:
+        return PDQMultiHashIndex()
+
+    def __init__(self, entries: t.Iterable[t.Tuple[str, IndexT]]) -> None:
         super().__init__()
         self.local_id_to_entry: t.OrderedDict = collections.OrderedDict()
-        if flat_index:
-            self.index: PDQHashIndex = PDQFlatHashIndex()
-        else:
-            self.index = PDQMultiHashIndex()
+        self.index: PDQHashIndex = self._get_empty_index()
         self.add(entries=entries)
 
     def __len__(self) -> int:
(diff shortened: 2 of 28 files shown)


python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -22,32 +26,37 @@ class PDQIndex(SignalTypeIndex):
     PDQ_CONFIDENT_MATCH_THRESHOLD = 31
     T = IndexT
 
-    def __init__(self, entries: t.Iterable[t.Tuple[str, T]]) -> None:
+    def __init__(
```
BarrettOlson, line 29: @schatten FYI I've decided an override flag is the wrong way to do this. I am going to change this in the next commit. Thanks!

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -31,15 +31,15 @@ def __init__(self, entries: t.Iterable[t.Tuple[str, T]]) -> None:
     def __len__(self) -> int:
         return len(self.local_id_to_entry)
 
-    def query(self, hash: str) -> t.List[IndexMatch[T]]:
+    def query(
+        self, hash: str, threshhold=PDQ_CONFIDENT_MATCH_THRESHOLD
```
Dcallies, line 44: (blocking): This no longer matches the interface on the base class @ https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/index.py#L89

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -53,6 +53,7 @@ def search(
         queries: t.Sequence[PDQ_HASH_TYPE],
         threshhold: int,
         return_as_ids: bool = False,
+        **kwargs
```
Dcallies, line 56: (blocking?): I generally recommend against this pattern - this means that typos on calling the search method will be ignored, unless you always pass **kwargs into something else. You sometimes see this in older code when they want to force you use a keyword argument, but there is new syntax for that (`foo(a, *, kw_only)`)

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -209,9 +210,15 @@ def mih_index(self):
             return faiss.downcast_IndexBinary(self.faiss_index.index)
         return self.faiss_index
 
-    def search(self, queries: t.Sequence[PDQ_HASH_TYPE], threshhold: int, **kwargs):
+    def search(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+        **kwargs
```
Dcallies, line 218: Same as above - most typers can't follow this, and so suggest being explicit.

python-threatexchange/threatexchange/signal_type/pdq_index.py
Dcallies: If you still want two options a simple enum is what is suggested by (internal facebook link): https://www.internalfb.com/intern/wiki/Facebook-code-smells/boolean-arguments/

Dcallies (commented):
Finished reviewing first 2. I had blocking comments but don't know if you fixed them later in the stack, so not RCing

b8156ff1 spelling and have query support custom threshold
4ede5ea5 fix typing and bugs in unused parts of pdq_index.py while adding flat_index option

BarrettOlson:
Dropping to draft and will split up changes.

BarrettOlson:
@Dcallies  thanks for reviewing the first two commits. Those are the changes I'd like merged sooner anyway so I moved them from this PR and added them here: #875 let me know if I addressed your feedback on avoiding a new boolean arg and avoiding **kwargs. Thanks

PR #875, opened 2021-12-06
[pytx] update faiss_matcher to have option to return distance & add PDQFlatIndex

BarrettOlson:
Summary
---------

To support variable threshold in HMA ideally each match returned can be paired with its distance.
Though since `python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py` has been around for over a year I added a new `search_with_distance` instead of changing the results given form `search` 

I also added a PDQFlatIndex to `python-threatexchange/threatexchange/signal_type/pdq_index.py` so that index type can be used in HMA as well.

Test Plan
---------
fixed a bug in pytest so it now checks the distance for matches for `PDQIndex`

### DIFF
--- a/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
@@ -106,7 +106,6 @@ class PDQHashIndex(ABC):
         queries: t.Sequence[PDQ_HASH_TYPE],
         threshhold: int,
         return_as_ids: bool = False,
-        **kwargs
     ):
         query_vectors = [
             numpy.frombuffer(binascii.unhexlify(q), dtype=numpy.uint8) for q in queries
@@ -239,9 +238,23 @@ class PDQMultiHashIndex(PDQHashIndex):
             return faiss.downcast_IndexBinary(self.faiss_index.index)
         return self.faiss_index
 
-    def search(self, queries: t.Sequence[PDQ_HASH_TYPE], threshhold: int, **kwargs):
+    def search(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+    ):
         self.mih_index.nflip = threshhold // self.mih_index.nhash
-        return super().search(queries, threshhold, **kwargs)
+        return super().search(queries, threshhold, return_as_ids)
+
+    def search_with_distance(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+    ):
+        self.mih_index.nflip = threshhold // self.mih_index.nhash
+        return super().search_with_distance(queries, threshhold, return_as_ids)
 
     def hash_at(self, idx: int):
         i64_id = uint64_to_int64(idx)
--- a/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
@@ -103,7 +103,7 @@ class PDQHashIndex(ABC):
 
     def search_with_distance_in_result(
         self,
-        queries: t.Sequence[PDQ_HASH_TYPE],
+        queries: t.Sequence[str],
         threshhold: int,
     ):
         """
@@ -256,7 +256,7 @@ class PDQMultiHashIndex(PDQHashIndex):
 
     def search_with_distance_in_result(
         self,
-        queries: t.Sequence[PDQ_HASH_TYPE],
+        queries: t.Sequence[str],
         threshhold: int,
     ):
         self.mih_index.nflip = threshhold // self.mih_index.nhash
--- a/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
@@ -107,6 +107,23 @@ class PDQHashIndex(ABC):
         threshhold: int,
         return_as_ids: bool = False,
     ):
+        """
+        Same as search however instead of returning only sequence of matches per query
+        it returns a mapping from query strings to a list of matched hashes (or ids) and distances
+        e.g.
+        result = {
+            "000000000000000000000000000000000000000000000000000000000000FFFF": [
+                ("00000000000000000000000000000000000000000000000000000000FFFFFFFF", 16.0)
+            ]
+        }
+        or if return_as_ids=True
+        result_with_as_ids = {
+            "000000000000000000000000000000000000000000000000000000000000FFFF": [
+                (12345678901, 16.0)
+            ]
+        }
+        """
+
         query_vectors = [
             numpy.frombuffer(binascii.unhexlify(q), dtype=numpy.uint8) for q in queries
         ]
--- a/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
+++ b/python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
@@ -101,25 +101,21 @@ class PDQHashIndex(ABC):
             for i in range(len(query_vectors))
         ]
 
-    def search_with_distance(
+    def search_with_distance_in_result(
         self,
         queries: t.Sequence[PDQ_HASH_TYPE],
         threshhold: int,
-        return_as_ids: bool = False,
     ):
         """
-        Same as search however instead of returning only sequence of matches per query
-        it returns a mapping from query strings to a list of matched hashes (or ids) and distances
+        Search method that return a mapping from query_str =>  (id, hash, distance)
+
+        This implementation is the same as `search` above however instead of returning just the sequence of matches
+        per query it returns a mapping from query strings to a list of matched hashes (or ids) and distances
+
         e.g.
         result = {
             "000000000000000000000000000000000000000000000000000000000000FFFF": [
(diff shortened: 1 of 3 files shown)


python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -101,6 +101,52 @@ def search(
...
         ]
 
+    def search_with_distance(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+    ):
+        """
+        Same as search however instead of returning only sequence of matches per query
```
Dcallies, line 111: (nit, ignorable): Recommended way of doing long docstrings is to have a one-line at the top that is a short summary, and then more details below: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -101,6 +101,52 @@ def search(
             for i in range(len(query_vectors))
         ]
 
+    def search_with_distance(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
```
Dcallies, line 107: Why not call this variable distance, or the method "search_with_threshold" ?

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -101,6 +101,52 @@ def search(
...
+        return_as_ids: bool = False,
```
Dcallies, line 108: (Nit): Hmm, this makes the return of this impossible to type, and as a result is slightly smelly to me. If you wanted to fix it, would suggest having two methods:

1. Return with ids
2. Return without ids (impl is to call with ids and then strip them)

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -101,6 +101,52 @@ def search(
...
+        if return_as_ids:
+            # for custom ids, we understood them initially as uint64 numbers and then coerced them internally to be signed
+            # int64s, so we need to reverse this before returning them back to the caller. For non custom ids, this will
+            # effectively return the same result
+            output_fn: t.Callable[[int], t.Any] = int64_to_uint64
+        else:
+            output_fn = self.hash_at
+
+        result = {}
+        for i in range(len(queries)):
```
Dcallies, line 141: (ignorable, code golf): `for i, query in enumerate(queries):`

There's also stupid fancy things you can do zipping limits and queries together, but I can't tell what the below code is trying to do at a glance to advise you on a way to clean it up

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -101,6 +101,52 @@ def search(
...
+            match_tuples = []
+            matches = [output_fn(idx.item()) for idx in I[limits[i] : limits[i + 1]]]
+            distances = [idx for idx in similarities[limits[i] : limits[i + 1]]]
+            for match, distance in zip(*(matches, distances)):
```
Dcallies, line 145: (blocking?): lol wut? Isn't `fn(*(a, b))` is the same as `fn(a, b)` ?

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -209,9 +255,23 @@ def mih_index(self):
...
+    def search(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+    ):
+        self.mih_index.nflip = threshhold // self.mih_index.nhash
+        return super().search(queries, threshhold, return_as_ids)
+
+    def search_with_distance(
```
Dcallies, line 267: blocking: Wait a second, doesn't this shadow your two definitions on L51 and L104?

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -11,21 +11,33 @@
...
-    Wrapper around the pdq faiss index lib
+    Wrapper around the pdq faiss index lib using PDQMultiHashIndex
     """
 
     PDQ_CONFIDENT_MATCH_THRESHOLD = 31
     T = IndexT
 
+    @classmethod
+    def get_match_threshold(cls):
+        return cls.PDQ_CONFIDENT_MATCH_THRESHOLD
```
Dcallies, line 31: (nit): Unless that value is reused somewhere else, feel free to inline and prevent a class constant from being created. (is there another one inside SignalType as well?)

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -75,3 +86,22 @@ def deserialize(cls, fin: t.BinaryIO) -> "SignalTypeIndex[T]":
         Instanciate an index from a previous call to serialize
         """
         return pickle.loads(fin)
+
+
+class PDQFlatIndex(PDQIndex):
```
Dcallies, line 91: (Unsure) It looks like the only things that change between these two is the match threshold and the class to use for the inside, which are unchanged after creation.

Why not just pass these in, or a bool that selects between the two instead of having two classes? Serialization reasons?

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -209,9 +255,23 @@ def mih_index(self):
...
 
-    def search(self, queries: t.Sequence[PDQ_HASH_TYPE], threshhold: int, **kwargs):
+    def search(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
+        threshhold: int,
+        return_as_ids: bool = False,
+    ):
+        self.mih_index.nflip = threshhold // self.mih_index.nhash
+        return super().search(queries, threshhold, return_as_ids)
```
Dcallies, line 265: What's the superclass in this case? Isn't is ABC/object at this point?

Dcallies (changes requested):
Thanks for splitting, but think that it's now caused trouble in the PR contents :( Feel free to schedule a meeting with me tomorrow to try and pair program it. 

Toplevel comment:
* Think something got borked in the split in pdq_faiss_matcher.py :'(

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
BarrettOlson: with_distance is about the return result. updating to `search_with_distance_in_result`

Yisu123:
你好，我是瞿文政，你的邮件我已经收到，我会尽快回复。

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
BarrettOlson: I agree, this is not part of these changes however I am trying to add what I need in  `threatexchange/hashing/pdq_faiss_matcher.py` without rewritting too much, however I can make this change for `search_with_distance_in_result` methods

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
BarrettOlson: thanks for catching result of refactoring too many times

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
BarrettOlson: it overrides then call the base version. because for `PDQMultiHashIndex` there are so prep work before a query based on threshold

python-threatexchange/threatexchange/signal_type/pdq_index.py
BarrettOlson: Because index types use multi inheritance with a mixin later on and don't access __init__ directly. This way they can keep the `SignalTypeIndex` interface.

(Also in the last PR you strongly recommended against a bool that selects between the two as an arg)

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
BarrettOlson: this was just the removal of **kwargs
I'll remove this change to address later.

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
```diff
@@ -101,6 +101,47 @@ def search(
             for i in range(len(query_vectors))
         ]
 
+    def search_with_distance_in_result(
+        self,
+        queries: t.Sequence[PDQ_HASH_TYPE],
```
Dcallies, line 106: ymmv: This is a new method, you can choose to just pick a winner between string and bytes. I'm not sure, since while bytes is more accurate for PDQ specifically, all of the other signal types just generalize to str (i.e. text input). 

I think in all the calling points you are probably coming from str, but dunno

Dcallies (approved):
Thanks for following up, I guess maybe my browser was not showing the unfold?

python-threatexchange/threatexchange/hashing/pdq_faiss_matcher.py
BarrettOlson: Done

PR #876, opened 2021-12-06
[pytx][minor] clean up in `signal_type/tlsh_pdf.py` and `signal_type/url_md5.py`

BarrettOlson:
Summary
---------

Remove fake state hack from `signal_type/tlsh_pdf.py` and update INDICATOR_TYPE + tests. once the sample data exists this should allow for tlsh hash matching on those signals

and updated var name to make mypy happy in `url_md5.py`
Test Plan
---------

`pytest`


Dcallies (approved):
freeshrugs

PR #877, opened 2021-12-06
[pytx][minor] update dockerfile with steps for testing local pytx changes

BarrettOlson:
Summary
---------

Comment only change in Docker added because how to test local pytx changes for hma without a version bump/push is common enough that documenting how to do so seems worth it.

Test Plan
---------

comment only change


Dcallies (approved):
_y_

Issue #868, opened 2021-11-20
incorrect license information and legal advice


Hello71:
LICENSE and pdq/LICENSE.txt claim that the entire repository is under BSD license, which is not accurate.

pdq/cpp/README.md claims that "`CImg.h` ... license does not allow commercial use", which is not accurate.

Dcallies:
Woah, thanks for the report!

Dcallies:
Alright after some internal discussion, here's what we think we should do:

1. Delete the LICENSE files except the top level
2. Remove the mention on interpretation of the CImg.h license
3. Explicitly call out the exceptions to the entire repository statement in the readme. So for that it will say something like:

```
README
----------

# LIcense
The entire repository is under the BSD license, with a small number of exceptions listed in the file itself. With any luck, this is the complete list of exceptions:
1. pdq/cpp/CImg.h - Only included for demonstration purposes
```

@Hello71 - Do you think that will cover your concerns?

Hello71:
that's not my preference, but at least it's accurate, if very minimal. LIcense is incorrectly capitalized though.

Dcallies:
Unless there's anything else, will close us out!

PR #881, opened 2021-12-09
Update PDQ README for  #868

Dcallies:
Instead point to the top-level license, pending a deletion of the duplicate license files.

Summary
---------

I could have done this as part of the commit that deleted them, but preview is useful. I can probably delete files from the github UI, but am lazy. 

Test Plan
---------

Viewed preview, clicked on link.


BarrettOlson (approved):
why not?
#868 for our ref.

PR #883, opened 2021-12-09
[hma] record match distance in match record and surface on API

BarrettOlson:
Summary
---------

Now that the PDQIndex provides distance for matches why not record it in match record and surface them on the API?
Note: All video_md5 matches recorded at 0 (as expected)

Test Plan
---------
Deployed and took a hash from `sample_data/holidays-jpg1-pdq-hashes.csv` e.g. `431dfd4204ad7b54ec6203bdfd42033fbfa5ec42033dfcc4033bf631dccc031c`

Modified it slightly and ran (after setting auth token):
`hmacli storm  --media "431dfd4204ad7b54ec6203bdfd42033fbfa5ec42033dfcc4033bf631dccc0310" --verbose -c 1 hash`  

Checked results in postman (with auth token in header):
GET `https://{{api_id}}.execute-api.us-east-1.amazonaws.com/api/matches/match/?content_id=<id-of-match-ui>`
```JSON
{
    "match_details": [
        {
            "content_id": "<id-of-match-ui>",
            "content_hash": "431dfd4204ad7b54ec6203bdfd42033fbfa5ec42033dfcc4033bf631dccc0310",
            "signal_id": "<id>",
            "signal_hash": "431dfd4204ad7b54ec6203bdfd42033fbfa5ec42033dfcc4033bf631dccc031c",
            "signal_source": "te",
            "signal_type": "pdq",
            "updated_at": "2021-12-09T20:42:21.489901",
            "match_distance": 2,
            "metadata": ["..."]
        }
    ]
}
```

Checked that match record created before deploying changes return `"match_distance": null,`

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/pipeline.py
+++ b/hasher-matcher-actioner/hmalib/common/models/pipeline.py
@@ -268,6 +268,7 @@ class _MatchRecord(PipelineRecordBase):
     signal_id: str
     signal_source: str
     signal_hash: str
+    match_distance: t.Optional[int] = None
 
 
 @dataclass
@@ -300,6 +301,7 @@ class MatchRecord(PipelineRecordDefaultsBase, _MatchRecord):
                 "GSI1-SK": self.get_dynamodb_content_key(self.content_id),
                 "HashType": self.signal_type.get_name(),
                 "GSI2-PK": self.get_dynamodb_type_key(self.__class__.__name__),
+                "MatchDistance": self.match_distance,
             },
         )
 
@@ -399,6 +401,7 @@ class MatchRecord(PipelineRecordDefaultsBase, _MatchRecord):
                 signal_specific_attributes=cls.deserialize_signal_specific_attributes(
                     item
                 ),
+                match_distance=item.get("MatchDistance"),
             )
             for item in items
         ]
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/common/models/pipeline.py
```diff
@@ -268,6 +268,7 @@ class _MatchRecord(PipelineRecordBase):
     signal_id: str
     signal_source: str
     signal_hash: str
+    match_distance: t.Optional[int] = None
```
schatten, line 271: Ooo! This might help with retroaction when we finally get to it.

schatten: Brownie points for making this backward compatible by using Optional!

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -135,6 +136,9 @@ def get_match_details(table: Table, content_id: str) -> t.List[MatchDetail]:
             signal_source=record.signal_source,
             signal_type=record.signal_type.get_name(),
             updated_at=record.updated_at.isoformat(),
+            match_distance=int(record.match_distance)
```
schatten, line 139: Is this necessary. Does DDB return the match_distance in some other format? Decimal etc?

schatten (approved):
LGTM. A question inline about `int(..)` being strictly necessary.

hasher-matcher-actioner/hmalib/lambdas/api/matches.py
BarrettOlson: yes but only our API middleware seemed to care. if passed along directly it gets turned into Decimal(<val>)

PR #884, opened 2021-12-09
Remove confusing duplicate LICENSE files for #868

Dcallies:
Summary
---------

The whole repository is supposed to be under a single LICENSE. However, by including all these strange variations, we're making that message unclear. Delete everything except for the topline.

Test Plan
---------

I checked to make sure that setup.py in python-threatexchange was not referencing it
```
$ python setup.py check
$ pip install -e .
```


BarrettOlson (approved):
why not? 
#868 for our ref.

PR #885, opened 2021-12-09
[hma] add AdditionalMatchSettingsConfig for pdq threshold to filter matches

BarrettOlson:
Summary
---------

Base for a config that can add additional similarity distance filtering beyond the ceiling set in `python-threatexchange/threatexchange/signal_type/pdq_index.PDQIndex`

Test Plan
---------

Rebased with #883 to set and checked new logs

created and saved config in python
```python3
from hmalib.common.configs.fetcher import AdditionalMatchSettingsConfig
from hmalib.common import config as hmaconfig
import getpass

hmaconfig.HMAConfig.initialize(f"{getpass.getuser()}-HMAConfig")
config = AdditionalMatchSettingsConfig("inria-holidays-test", pdq_match_threshold=20)
hmaconfig.create_config(config)
print(AdditionalMatchSettingsConfig.get("inria-holidays-test"))
```
Simliar to #883 test plan + read logs:
"Match was found at 31.0 distance but exceeds custom threshold of 20 for pg, inria-holidays-test"

### DIFF
--- a/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
+++ b/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
@@ -161,8 +161,8 @@ class Matcher:
                     str(privacy_group)
                 ):
                     if match.distance >= threshold:
-                        logger.warning(  # ToDo debug is probably a better log
-                            "Match was found at %s distance but exceeds custom theshhold of %s for pg, %s",
+                        logger.debug(
+                            "Match was found at %s distance but exceeds custom theshold of %s for pg, %s",
                             match.distance,
                             threshold,
                             privacy_group,
--- a/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
+++ b/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
@@ -24,7 +24,10 @@ from hmalib import metrics
 from hmalib.common.logging import get_logger
 from hmalib.common.mappings import INDEX_MAPPING
 from hmalib.common.messages.match import BankedSignal, MatchMessage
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -114,9 +150,34 @@ def filter_match_results(self, results: t.List[IndexMatch]) -> t.List[IndexMatch
...
+                            "Match was found at %s distance but exceeds custom theshold of %s for pg, %s",
+                            match.distance,
+                            threshold,
+                            privacy_group,
+                        )
+                    else:
+                        filtered_privacy_groups.append(privacy_group)
+                else:
+                    # no custom threshold found
+                    filtered_privacy_groups.append(privacy_group)
```
schatten, line 174: If index was created when privacy_group's threshold was defined at say 40, and the matcher kicked in when the threshold was removed, does this mean the default threshold of 31 will be ignored?

Right now, we rebuild indexes frequently enough, but soon, that might not be the case. If so, we are introducing an implicit dependency. 

Might be totally wrong though. :)

schatten (approved):
Question about an implicit dependency on the index and matcher being used with different threshold values. 

Wouldn't say it is a blocker. Possibility of undefined behavior, though.

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
BarrettOlson: Not sure I understand your questions, if no custom threshold is found we do NOT filter so the default 31 would be respected and a match recorded. these changes don't care about index building; they are built independent of any pg's thresholds

PR #887, opened 2021-12-10
[HMA] Indexer sources hashes from banks as well as te data

schatten:
Summary
---------
### [`f961defb`](https://github.com/facebook/ThreatExchange/pull/887/commits/f961defbe981bc5329ad831b211787e3d172cbd0) [banks] PendingBankMemberSignalsIndex -> BankMemberSignalCursorIndex
Indexer is not the only consumer of bank member signals. So, the Pending index which was sparse worked out fine. However, there are two complications:
a) rebuilding is not going away, there needs to be a way to read all bank member signals. All of them even if there are writes between page reads. So sort_key needs to increment.  
b) other consumers (eg. publish bank to threatexchange) The pending index was meant for a single consumer only and for a one time use. We need something more durable.  

The new BankMemberSignalCursorIndex is durable. We don't remove items from it on successfully processing. Since the SK is monotonically increasing, there is no chance of missing out on keys.  Updates to existing signals (remove) will result in bringing the updated item to the end of the index because updated_at == current_time.  You can theoretically use this structure as a feed.  

### [`a70e42d6`](https://github.com/facebook/ThreatExchange/pull/887/commits/a70e42d6e83d09eaced6378417b34742dfd2d9dd)  Paginate bank_member_signal cursored query
Since we are no longer going to "mark" signals as procssed, we need to "seek" into the index. Which requires querying for pages.  

### [`b4e5c11d`](https://github.com/facebook/ThreatExchange/pull/887/commits/b4e5c11dec92db0505bc9c80415b593f3361310c) Add BaseIndexMetadata and subclasses for banks, threatexchange
Use the index with typed metadata objects so that downstream systems can 'expect' more than a dict.  Also ensures fetched data is converted into typed objects for indexer's convenience.  

### [`aa2f3fcd`](https://github.com/facebook/ThreatExchange/pull/887/commits/aa2f3fcde178a875344c9b25eb368069e5e9a08d) bug/improvement: legacy BankedSignal objects add source classification on init()

Test Plan
---------
Locally, `black`, `mypy` and `py.test`. 

Deployed to AWS.

Added a new member to bank. Waited (polling ddb console manually) for bank member to be hashed.
Ran indexer manually (using lambda console).

**Screenshot showing bank member**
Screen Shot 2021-12-10 at 11 59 25

Submitted an image after configuring an action rule to allow actioning for bank members.

**Screenshot showing the progress page of the submitted image**
Screen Shot 2021-12-10 at 12 00 52

### DIFF
--- a/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
+++ b/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
@@ -9,8 +9,6 @@ import datetime
 import functools
 
 from mypy_boto3_sns.client import SNSClient
-from hmalib.common.models.signal import ThreatExchangeSignalMetadata
-
 from mypy_boto3_dynamodb.service_resource import Table
 from threatexchange.signal_type.pdq import PdqSignal
 from hmalib.common.models.pipeline import MatchRecord
@@ -28,11 +26,22 @@ from hmalib.common.configs.fetcher import (
     ThreatExchangeConfig,
     AdditionalMatchSettingsConfig,
 )
+from hmalib.common.models.bank import BankMember, BanksTable
+from hmalib.common.models.signal import ThreatExchangeSignalMetadata
+from hmalib.indexers.metadata import (
+    BANKS_SOURCE_SHORT_CODE,
+    THREAT_EXCHANGE_SOURCE_SHORT_CODE,
+    BaseIndexMetadata,
+    BankedSignalIndexMetadata,
+)
 
 
 logger = get_logger(__name__)
 
 PG_CONFIG_CACHE_TIME_SECONDS = 300
+
+# This is clearly wrong, but will still work. Someone tell the correct value, please.
+MAX_PDQ_DISTANCE = 1000
 
 
 @functools.lru_cache(maxsize=128)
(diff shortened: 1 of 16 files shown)


hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -183,68 +192,84 @@ def filter_match_results(self, results: t.List[IndexMatch]) -> t.List[IndexMatch
...
-                            "Match was found at %s distance but exceeds custom theshold of %s for pg, %s",
-                            match.distance,
-                            threshold,
-                            privacy_group,
+            match.metadata = [
+                metadata_obj
+                for metadata_obj in match.metadata
+                if metadata_obj.get_source() == BANKS_SOURCE_SHORT_CODE
+                or (
+                    # If metadata obj is from threatexchange (one privacy group
```
schatten, line 202: @BarrettOlson Special attention here. This merges changes you and I made separately.

Your changes: filter out privacy groups that have match settings threshold < the current match distance.

My changes: each metadata_obj only contains a single Indicator or BankMember.

BarrettOlson (approved):
Approving to unblock, 
Some the of s3_adapter stuff still feels a tad wacky, but with it typed we can make it less wacky as needed!

I have a few longer term design thoughts about where we want to have  metadata from banks vs te use the same interfaces instead of them needing to special cased but I think it makes sense to table until the fetcher interface is refined as well.

PR #890, opened 2021-12-10
[hma] add support for PDQ match threshold >31

BarrettOlson:
Summary
---------

1) Updates indexer to build new PDQFlatIndex and store it in s3
2) Updates matcher_base to consider configured thresholds when picking matcher have a choice between  
- PDQIndex (31 max)
- PDQFlatIndex (52 max) 
  - limits based on hardcoded value in python-threatexchange (possible to make dynamic as well) 

What is missing but plan to leave for next PR (if it is alright with reviewer):
- tf variable for enable disable FlatIndex usage
- UI support

Test Plan
---------
(Gave live demo of postman tests)
Tested e2e by submitting hashes of variable distance from given sample set hash and making sure matcher we found or not found accordingly when thresholds were configured for various distances. 
`hmacli storm  --media "ffffffffffffff4f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22" --verbose -c 1 hash`

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/mappings.py
+++ b/hasher-matcher-actioner/hmalib/common/mappings.py
@@ -28,31 +28,10 @@ from hmalib.indexers.s3_indexers import (
     S3BackedInstrumentedIndexMixin,
 )
 
-# Maps from signal type → index to use for that signal type.
+# Maps from signal type → [index] i.e. list of indexes the can be use for that signal type.
 INDEX_MAPPING: t.Dict[
     t.Type[SignalType], t.List[t.Type[S3BackedInstrumentedIndexMixin]]
 ] = {
     PdqSignal: [S3BackedPDQIndex, S3BackedPDQFlatIndex],
     VideoMD5Signal: [S3BackedMD5Index],
 }
-
-
-def get_index_for_signal_type_matching(
-    signal_type: t.Type[SignalType], max_custom_threshold: int = 0
-):
-    indexes = INDEX_MAPPING[signal_type]
-    # disallow empty list
-    assert indexes
-    if len(indexes) == 1:
-        # if we only have one option just return
-        return indexes[0]
-
-    indexes.sort(key=lambda i: i.get_index_max_distance())
-
-    for index in indexes:
-        if max_custom_threshold <= index.get_index_max_distance():
-            return index
-
-    # if we don't have an index that supports max threshold
-    # just return the one if the highest possible max distance
-    return indexes[-1]
--- a/hasher-matcher-actioner/hmalib/common/mappings.py
+++ b/hasher-matcher-actioner/hmalib/common/mappings.py
@@ -24,11 +24,35 @@ from threatexchange.signal_type.signal_base import SignalType
 from hmalib.indexers.s3_indexers import (
     S3BackedMD5Index,
     S3BackedPDQIndex,
+    S3BackedPDQFlatIndex,
     S3BackedInstrumentedIndexMixin,
 )
 
 # Maps from signal type → index to use for that signal type.
-INDEX_MAPPING: t.Dict[t.Type[SignalType], t.Type[S3BackedInstrumentedIndexMixin]] = {
-    PdqSignal: S3BackedPDQIndex,
-    VideoMD5Signal: S3BackedMD5Index,
+INDEX_MAPPING: t.Dict[
+    t.Type[SignalType], t.List[t.Type[S3BackedInstrumentedIndexMixin]]
+] = {
+    PdqSignal: [S3BackedPDQIndex, S3BackedPDQFlatIndex],
+    VideoMD5Signal: [S3BackedMD5Index],
 }
+
+
+def get_index_for_signal_type_matching(
+    signal_type: t.Type[SignalType], max_custom_threshold: int = 0
+):
+    indexes = INDEX_MAPPING[signal_type]
+    # disallow empty list
+    assert indexes
+    if len(indexes) == 1:
+        # if we only have one option just return
+        return indexes[0]
+
+    indexes.sort(key=lambda i: i.get_index_max_distance())
+
+    for index in indexes:
+        if max_custom_threshold <= index.get_index_max_distance():
+            return index
+
+    # if we don't have an index that supports max threshold
+    # just return the one if the highest possible max distance
+    return indexes[-1]
--- a/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
+++ b/hasher-matcher-actioner/hmalib/matchers/matchers_base.py
@@ -22,7 +22,7 @@ from threatexchange.signal_type.signal_base import SignalType
 
 from hmalib import metrics
 from hmalib.common.logging import get_logger
-from hmalib.common.mappings import get_index_for_signal_type_matching
+from hmalib.common.mappings import INDEX_MAPPING
 from hmalib.common.messages.match import BankedSignal, MatchMessage
 from hmalib.common.configs.fetcher import (
     ThreatExchangeConfig,
@@ -58,7 +58,7 @@ def get_privacy_group_matcher_active(privacy_group_id: str) -> bool:
     )
 
 
-@functools.lru_cache(maxsize=128)
+@functools.lru_cache(maxsize=None)
 def _get_all_matcher_active_privacy_groups(cache_buster) -> t.List[str]:
     configs = ThreatExchangeConfig.get_all()
     return list(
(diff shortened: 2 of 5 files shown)


hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -24,11 +24,35 @@
...
+    indexes = INDEX_MAPPING[signal_type]
+    # disallow empty list
+    assert indexes
+    if len(indexes) == 1:
+        # if we only have one option just return
+        return indexes[0]
+
+    indexes.sort(key=lambda i: i.get_index_max_distance())
+
+    for index in indexes:
```
schatten, line 52: nit: index_cls

schatten, line 52: nit: `index_cls`

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -24,11 +24,35 @@
...
-    VideoMD5Signal: S3BackedMD5Index,
+INDEX_MAPPING: t.Dict[
+    t.Type[SignalType], t.List[t.Type[S3BackedInstrumentedIndexMixin]]
+] = {
+    PdqSignal: [S3BackedPDQIndex, S3BackedPDQFlatIndex],
+    VideoMD5Signal: [S3BackedMD5Index],
 }
+
+
+def get_index_for_signal_type_matching(
```
schatten, line 40: Any arguments for or against pushing this to the `get_index()` method on the `Matcher` class in matchers_base.py?

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -24,11 +24,35 @@
...
     S3BackedPDQIndex,
+    S3BackedPDQFlatIndex,
     S3BackedInstrumentedIndexMixin,
 )
 
 # Maps from signal type → index to use for that signal type.
-INDEX_MAPPING: t.Dict[t.Type[SignalType], t.Type[S3BackedInstrumentedIndexMixin]] = {
-    PdqSignal: S3BackedPDQIndex,
-    VideoMD5Signal: S3BackedMD5Index,
+INDEX_MAPPING: t.Dict[
```
schatten, line 32: Can you add a comment here? this is no longer signal_type → index, it is signal_type → [index].

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -58,6 +58,49 @@ def get_privacy_group_matcher_active(privacy_group_id: str) -> bool:
...
+    signal_type: t.Type[SignalType],
+) -> int:
+    if signal_type == PdqSignal:
+        pg_to_pdq_threshold = _get_all_privacy_group_active_matcher_pdq_theshold(
+            time.time() // PG_CONFIG_CACHE_TIME_SECONDS
+        )
+        # no custom threshold set/active
+        if not pg_to_pdq_threshold:
+            return 0
+        return max(pg_to_pdq_threshold.values())
```
schatten, line 99: nit: this way, you end up computing `max` for every match. Could this be moved into the cached function?

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -58,6 +58,49 @@ def get_privacy_group_matcher_active(privacy_group_id: str) -> bool:
...
+
+
+@functools.lru_cache(maxsize=128)
+def _get_all_privacy_group_active_matcher_pdq_theshold(
+    cache_buster,
+) -> t.Dict[str, int]:
+    active_pg_names = _get_all_matcher_active_privacy_groups(cache_buster)
+    if not active_pg_names:
+        return {}
+    return {
```
schatten, line 82: nit: the key of the results dict is never used, only the values. A list would be better unless you plan on doing more filtering later.

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -58,6 +58,49 @@ def get_privacy_group_matcher_active(privacy_group_id: str) -> bool:
     )
 
 
+@functools.lru_cache(maxsize=128)
```
schatten, line 61: In both this and line 75, the maxsize value is going to cause the space used to be many times more than necessary.

Use maxsize when you want to store results for multiple arg tuples. In this case (and L75), we do not want to store results for multiple values of cache_buster.

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -123,7 +166,8 @@ def match(
         index = self.get_index(signal_type)
 
         with metrics.timer(metrics.names.indexer.search_index):
-            match_results: t.List[IndexMatch] = index.query(signal_value)
+            match_results: t.List[IndexMatch] = []
```
schatten, line 169: Not following this change. If you are creating a list and then immediately extending it, why not use the `index.query(...)` result directly?

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -260,13 +304,19 @@ def get_metadata_objects_from_match(
     def get_index(self, signal_type: t.Type[SignalType]) -> SignalTypeIndex:
```
schatten, line 304: Asked elsewhere too, but this seems like THE place where HMA is figuring out dynamically which index to use. mappings.py seems more static. 

Do you see this as a better place to plop the logic?

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -260,13 +304,19 @@ def get_metadata_objects_from_match(
...
         # If cached, return an index instance for the signal_type. If not, build
         # one, cache and return.
-        if not signal_type in self._cached_indexes:
-            index_cls = INDEX_MAPPING[signal_type]
+        max_custom_threshold = (
```
schatten, line 307: super-nit: weird use of paranthesis. Easily confusable as a tuple. Did black do this?

schatten: We end up computing max_custom_threshold and index_cls even if cached. Can avoid.

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -260,13 +304,19 @@ def get_metadata_objects_from_match(
...
+            get_max_threshold_of_active_privacy_groups_for_signal_type(signal_type)
+        )
+        index_cls = get_index_for_signal_type_matching(
+            signal_type, max_custom_threshold
+        )
 
+        if not signal_type in self._cached_indexes or not isinstance(
+            self._cached_indexes[signal_type], index_cls
```
schatten, line 315: AFAIK, self._cached_indexes is only updated by this method itself. Why do the type assertion?

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -260,13 +304,19 @@ def get_metadata_objects_from_match(
...
+        ):
             with metrics.timer(metrics.names.indexer.download_index):
-                self._cached_indexes[signal_type] = index_cls.load(
-                    bucket_name=self.index_bucket_name
-                )
+                index = index_cls.load(bucket_name=self.index_bucket_name)
+                self._cached_indexes[signal_type] = index
```
schatten, line 319: superfluous variable `index`.

schatten (approved):
Overall, looks good. I like that we are building both indexes and using whichever works better at match time. That means users will not need to rebuild indexes when changing thresholds.

Questions that I'd want answered before merging.

1. Can the dynamic index selection logic be moved into `Matcher.get_index`?

Issues (flagged inline):
1. We are doing a bit of computation for every match and that can be avoided.

2. Space utilization when using lru_cache is more than necessary.

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
BarrettOlson: oops refactor error

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
BarrettOlson: sure!

hasher-matcher-actioner/hmalib/common/mappings.py
BarrettOlson: Done!

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
BarrettOlson: re: parenthesis -> black did do this.

We should check because if it changed it is possible the cached index could be the incorrect type. This is pretty niche but here is where using cached index blindly breaks down.

1. No custom threshold or custom threshold < 31 is set.
2. Matcher lambda is invoked and because of current config state PDQIndex is cached for pdq signal type
3. Custom threshold is set for > 31 (matcher that should be used is now PDQFlatIndex)
4. Matcher lambda is invoked again where `PG_CONFIG_CACHE_TIME_SECONDS` has a set value that is pass but `_cached_indexes` is still in the lambda/matchers memory.
5. without this check: `get_index` would incorrectly return `PDQIndex` because it didn’t recompute (or at least check the other cache of) `max_custom_threshold`

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
BarrettOlson: Answered in other still open conversation and added comment to file.

PR #892, opened 2021-12-14
[hma] Add a settings pane for getting index last modified and rebuilding on demand

schatten:
Summary
---------
With banks, indexing runs on a pre-determined frequency. This may not be fast enough for our users when they are demoing something or need to test changes out.

This shows the last_modified times of the current set of indexes on the 'settings' portion of the HMA UI and allows a user to trigger an async rebuild of the index.

Test Plan
---------
mypy, black, py.test on local.

**Screenshot 1: the new UI pane with the rebuild button**
Screen Shot 2021-12-13 at 22 37 58

**Screenshot 2: logs of indexer when the rebuild button was clicked**
Screen Shot 2021-12-13 at 22 38 10

### DIFF
--- a/hasher-matcher-actioner/hmalib/indexers/s3_indexers.py
+++ b/hasher-matcher-actioner/hmalib/indexers/s3_indexers.py
@@ -8,6 +8,7 @@ S3 Backed Index Management. Offers a menu of indexes.
   understand TMK better. :) -- @schatten
 """
 
+from datetime import datetime
 import pickle
 import boto3
 import functools
@@ -35,6 +36,8 @@ class S3BackedInstrumentedIndexMixin:
     Also overrides methods to wrap instrumentation with hmalib.metrics library.
     """
 
+    INDEXES_PREFIX = "index/"
+
     def save(self, bucket_name: str):
         with metrics.timer(metrics.names.indexer.upload_index):
             index_file_bytes = pickle.dumps(self)
@@ -53,6 +56,16 @@ class S3BackedInstrumentedIndexMixin:
                 .read()
             )
             return pickle.loads(index_file_bytes)
+
+    @classmethod
+    def get_oldest_last_modified(cls, bucket_name: str) -> datetime:
+        """
+        Get the update time of the oldest index.
+        """
+        objects = get_s3_client().list_objects_v2(
+            Bucket=bucket_name, Prefix=cls.INDEXES_PREFIX
+        )
+        return max(map(lambda item: item["LastModified"], objects["Contents"]))
 
     @classmethod
     def _get_index_s3_key(cls):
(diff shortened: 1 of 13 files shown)


hasher-matcher-actioner/hmalib/indexers/s3_indexers.py
```diff
@@ -35,6 +36,8 @@ class S3BackedInstrumentedIndexMixin:
     Also overrides methods to wrap instrumentation with hmalib.metrics library.
     """
 
+    INDEXES_PREFIX = "index/"
```
BarrettOlson, line 39: (nonblocking) We haven't been consistent about this being a variable vs just locally defining it. It used to be something defined in terraform...

BarrettOlson: nevermind, looks like this is a step in the right direction regardless

hasher-matcher-actioner/hmalib/indexers/s3_indexers.py
```diff
@@ -54,6 +57,16 @@ def load(cls, bucket_name: str):
...
 
+    @classmethod
+    def get_oldest_last_modified(cls, bucket_name: str) -> datetime:
+        """
+        Get the update time of the oldest index.
+        """
+        objects = get_s3_client().list_objects_v2(
+            Bucket=bucket_name, Prefix=cls.INDEXES_PREFIX
+        )
+        return max(map(lambda item: item["LastModified"], objects["Contents"]))
```
BarrettOlson, line 68: How is "LastModified" stored? depending on the timestamp counting up wouldn't this actually be the most recently modified?

hasher-matcher-actioner/hmalib/lambdas/api/indexes.py
```diff
@@ -0,0 +1,59 @@
...
+def get_indexes_api(
+    indexes_bucket_name: str, indexer_function_name: str
+) -> bottle.Bottle:
+    indexes_api = SubApp()
+
+    @indexes_api.get("/last-modified", apply=[jsoninator])
+    def all_indexes_last_modified() -> IndexesLastModifiedResponse:
+        """
+        Returns the max of last_modified time of all indexes. Read as: when was
+        the latest index rebuilt.
```
BarrettOlson, line 40: hmm, 'latest' or 'oldest' am I miss understanding these as contradictory.

BarrettOlson (approved):
Approving to unblock. Thanks
two asides though
1) can you confirm intended behavior is correct e.g. "oldest" vs "latest"

2) to me this index button is really nice to have but it could just be a button/card on a existing settings or bank page instead of its own dedicated one.

hasher-matcher-actioner/hmalib/indexers/s3_indexers.py
schatten: Yes. This is the most recently modified timestamp.

I am not accounting for a case where one index is built but the other is not.

schatten:
>     1. can you confirm intended behavior is correct e.g. "oldest" vs "latest"

Fixed. It is latest. Not oldest. I started with oldest but quickly realized this would be really hard to get right. So switched everywhere but the method name. This is now addressed.

>     2. to me this index button is really nice to have but it could just be a button/card on a existing settings or bank page instead of its own dedicated one.

I agree. I struggled for a bit, but I could not figure out a place where this would fit better. Maybe a global settings page, but we don't have one. The equivalent settings for actions and action rules have their own page, so decided to go down this path, but am not beholden to it. 

Happy to discuss and move elsewhere once this lands.

PR #896, opened 2021-12-16
Detached bank UI

schatten:
Summary
---------
We have supported bank member signals that do not have media attached, (eg added via a fetcher), in ddb for a while. This PR adds:
a) an API to create such member signals
b) Ensure that the all-bank-members and single-bank-member pages both render correctly and show some information.

### [`06cc6f04`](https://github.com/facebook/ThreatExchange/pull/896/commits/06cc6f04200b524bdf72bfdd4077d27911d58bdd) [Bonus] Extract content-details's ← link into a component; use in ContentDetails & ViewBankMember [tsx only]

Test Plan
---------
black, mypy and py.test on local.

Added a few detached video members using postman.

**Screenshot 1: Single Bank Member page for a detached member signal**
Screen Shot 2021-12-15 at 23 02 44

**Screenshot 2: All Bank Members page with some detached member signals**
Screen Shot 2021-12-15 at 23 02 52

### DIFF
--- a/hasher-matcher-actioner/webapp/src/messages/BankMessages.tsx
+++ b/hasher-matcher-actioner/webapp/src/messages/BankMessages.tsx
@@ -29,6 +29,7 @@ export type BankMember = {
   notes: string;
   created_at: Date;
   updated_at: Date;
+  is_media_unavailable: boolean;
 };
 
 export type BankMemberSignal = {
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/webapp/src/messages/BankMessages.tsx
```diff
@@ -29,6 +29,7 @@ export type BankMember = {
   notes: string;
   created_at: Date;
   updated_at: Date;
+  is_media_unavailable: boolean;
```
BarrettOlson, line 32: nit (and not new to this PR) but why have `is_media_unavailable` it feels like the null status of `storage_key` and `raw_content` should always match its value and if this is the case then having this flag is a way for things to get out of sync for bank members. Given this is already in API and model this is not blocking for this PR.

BarrettOlson (approved):
Approved.

Works for me given existing model for `BankMember` but probably worth a follow up conversation around `is_media_unavailable`  Thanks

PR #902, opened 2021-12-23
[devcontainer] Try and make devcontainer work windows

Dcallies:
Summary
---------

Discover that all the environment variables in example given is for mac. Spend a few hours trying to figure out what is going wrong. Conclude that there's no default that works for both. Force people to edit their own.

Will require updating CONTRIBUTING.MD afterwards.

Test Plan
---------

I was able to boot it on windows, but many errors related to CLRF. Going to attempt to refresh repository in lf mode and try again.

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/.gitignore
+++ b/hasher-matcher-actioner/.devcontainer/.gitignore
@@ -1,0 +1,2 @@
+# You likely have to customize this, so copy it
+# devcontainer.json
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/.devcontainer/.gitignore
```diff
@@ -0,0 +1,2 @@
+# You likely have to customize this, so copy it
+# devcontainer.json
```
Dcallies, line 2: It was unclear to me what the behavior would be if I checked this in the same commit as moving it, so do it in two steps.

schatten: .gitignore is 'observed' at `git add`, so same commit would have worked.

schatten (approved):

PR #906, opened 2021-12-23
[py-tx] Interface for collab configuration

Dcallies:
Summary
---------

This will replace what is currently represented by:
1. txtoken
2. collab config files

And gives us the option to do multi-config fetching and so on.

A separate PR will contain the proposed interface for fetching.

Test Plan
---------

Uhhh... there were no squiggles in VSCode?

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -1,0 +1,54 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import typing as t
+import pathlib
+import os
+
+from threatexchange.collab_config import CollaborationConfig
+
+"""
+Local storage and configuration for the CLI.
+
+The CLI and Hasher-Matcher-Actioner are roughly parallel, but this isn't a 
+scalable service running on AWS. Instead, we have all of our state in
+a file (likely ~/.threatexchange)
+"""
+
+FETCH_STATE_DIR_NAME = "fetched_state"
+COLLABORATION_CONFIG_DIR_NAME = "collaborations"
+INDEX_STATE_DIR_NAME = "index"
+CONFIG_FILENAME = "config.json"
+
+
+class CliState:
+    """
+    A wrapper around stateful information stored for the CLI.
+
+    Everything is just in a single file (usually ~/.threatexchange).
+    """
+
+    def __init__(self, dir: pathlib.Path):
+        assert dir.is_file
+        self._dir = dir
+
+    def get_collab_by_name(self, name: str) -> t.Optional[CollaborationConfig]:
+        """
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -0,0 +1,51 @@
...
+        assert dir.is_file
+        self._dir = dir
+
+    def get_collab_by_name(self) -> t.Optional[CollaborationConfig]:
+        """
+        Gets only a single stored collaboration config by its given name.
+        """
+        return None
+
+    def update_collab(self, collab: CollaborationConfig, delete: bool = False) -> None:
```
schatten, line 41: nit: not a lot more effort to create a dedicated `delete_collab` method. I find the usage of `delete` as an argument confusing.

schatten: Also, question: Can collab names be changed?

If they can't, then all is well.

If collab.name can be updated, though, how would the implementation figure out the collab to be updated?

schatten (approved):
strongly held nit about not using a delete flag in the `update_collab` method.

Probably blocking: `get_collab_by_name`  is missing a `name` argument.

Approving to unblock.

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
Dcallies: Copy pasted - it's used as a hint for some linters. I also see some without it, so can just delete.

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -0,0 +1,51 @@
...
+    A wrapper around stateful information stored for the CLI.
+
+    Everything is just in a single file (usually ~/.threatexchange).
+    """
+
+    def __init__(self, dir: pathlib.Path):
+        assert dir.is_file
+        self._dir = dir
+
+    def get_collab_by_name(self) -> t.Optional[CollaborationConfig]:
```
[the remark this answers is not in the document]
Dcallies: Lol, yes

python-threatexchange/threatexchange/cli/cli_config.py
Dcallies: Yea, plan is for names to be immutable, because I actually need to key them several places in the system (index). The alternative would be generating IDs for collabs, but then I have ID assignment problems, so seemed easier to just make the collab name immutable, but also was going to see what happened when I did some of the later diffs.

Dcallies:
Thanks for review! No strong preference on a delete or not, was just minimizing methods.

Dcallies:
Cannot reproduce lint error locally, and not on file edited.

PR #907, opened 2021-12-23
[py-tx] Fetcher API interface

Dcallies:
Summary
---------

Third time's the charm? This is an attempt to both generalize the fetch interface, and to delete mistakes of the past with the threat_updates API. 

It's basically the threat_updates API again, except with less weird intermediate translations via function as arguments.

# What's in this PR
A simple interface that represents fetch APIs, which should cover all 5 fetch approaches I am tracking:
1. ThreatExchange tagged_objects
2. ThreatExchange threat_updates
3. StopNCII.org 
4. NCMEC Industry Hash DB
5. A hardcoded set of signals

It also contains a simple interface for storage that I think should work for both HMA and py-te, in line with goals of aligning them.

# What the end of this road hopefully contains (3w of coding + debugging, which probably means 6w in real time)
1. py-te CLI handles multiple APIs/collabs
2. py-te state moves to .threatexchange (since its now a lot more complicated)
3. py-te uses Fetcher API => SQL Lite Storage Impl => SignalTypeIndex, which would be very close to HMA
4. ThreatExchange fetch by tags is deleted because I'm too lazy to preserve it
5. Hardcoded sample data signal fetcher source, threat_updates, NCII API, NCMEC API
6. The whole thing is stable enough to fetch ~200k range signals

Test Plan
---------

Pending

### DIFF
--- a/python-threatexchange/threatexchange/fetcher/fetch_api.py
+++ b/python-threatexchange/threatexchange/fetcher/fetch_api.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 """
@@ -12,45 +11,70 @@ import typing as t
 
 from threatexchange.collab_config import CollaborationConfig
 from threatexchange.signal_type.signal_base import SignalType
-from . import fetch_state_base as state
+from threatexchange.fetcher import fetch_state_base as state
 
-
+# TODO t.Generic[TFetchDelta, TFetchedSignalData]
+#      In order to make it easier to track the expected extensions for an API
 class SignalExchangeAPI:
     """
-    An external APIs to get and maybe put signals
+    APIs to get and maybe put signals.
 
     Fetchers ideally can checkpoint their progress, so that they can tail
-    updates from the server.
+    updates.
+
+    While this interface is primarily intended for connecting with
+    externally hosted servers, it might be useful to write adopters for
+    certain formats of local files, which could be valuable for testing.
 
     Is assumed that fetched signals have some metadata attached to them,
     which is unique to that API. Additionally, it a assumed that there
-    might be multiple contributors to signals inside of an API.
+    might be multiple contributors (owners) to signals inside of an API.
 
-    The Fetcher can assume it is stateful, and
+    An instance of this class can retain state (caching connecting, etc)
+    as needed.
+
+    Methods with an implementation are optional, but may rely on
+
+    = On Owner IDs =
+    Some APIs may not have any concept of owner ID (because the owner is the
+    API itself). In that case, it's suggested to use 0 for all IDs. If the you
+    own the API, it may also make sense to have 0 be the return of
+    get_own_owner_id, which is used by matching to tell if a signal be
+    considered "confirmed" or not. If your API does support the concept of
+    owners, make sure to implement resolve_owner and get_own_owner_id
+
     """
 
     @classmethod
     def name(cls) -> str:
         """
-        A string name unique to all SignalExchangeAPIs
+        A simple string name unique to SignalExchangeAPIs in use.
 
-        This will get stored for lookup from a map of supported APIs
-        keyed str: API
+        This shouldn't be changed once comitted, or you may break naive
+        storage solutions (like the one in the CLI) which stores fetched
+        data by (SignalExchangeAPI.name(), collab_name).
         """
         return cls.__name__
 
     def resolve_owner(self, id: int) -> str:
         """
-        Convert an owner ID into a human readable name (if available)
+        Convert an owner ID into a human readable name (if available).
 
-        If empty is returned, its assumed that the collaboration name itself
-        is good enough to describe this content.
+        If empty string is returned, a placeholder will be used instead.
         """
         return ""
 
     def get_own_owner_id(self) -> int:
         """
-        Return the owner ID of this caller. Opinions with that ID are "ours"
+        Return the owner ID of this caller. Opinions with that ID are "ours".
+
+        SignalOpinions returned by fetch() where the owner id is
+        the return of get_own_owner_id() are assumed to be owned by you, which
+        can result in some additional metadata being added to matches, for
+        example in the `match` command of the CLI.
+
+        A default implementation is provided that is assumed to not match any
+        real owner ID.
         """
         return -1
 
@@ -65,7 +89,12 @@ class SignalExchangeAPI:
     def report_seen(
         self, s_type: SignalType, signal: str, metadata: state.FetchedSignalData
     ) -> None:
-        """Report that this signal was observed on your platform"""
+        """
+        Report that you observed this signal.
+
+        This is an optional API, and places that use it should catch
+        the NotImplementError.
+        """
         raise NotImplementedError
 
     def report_opinion(
@@ -89,7 +118,12 @@ class SignalExchangeAPI:
         signal: str,
         metadata: state.FetchedSignalData,
     ) -> None:
-        """Report that a previously seen signal was a true positive"""
+        """
+        Report that a previously seen signal was a true positive.
+
+        This is an optional API, and places that use it should catch
+        the NotImplementError.
+        """
         raise NotImplementedError
 
     def report_false_positive(
@@ -98,5 +132,10 @@ class SignalExchangeAPI:
         signal: str,
         metadata: state.FetchedSignalData,
     ) -> None:
-        """Report that a previously seen signal is a false positive"""
+        """
+        Report that a previously seen signal is a false positive.
+
+        This is an optional API, and places that use it should catch
+        the NotImplementError.
+        """
         raise NotImplementedError
--- a/python-threatexchange/threatexchange/fetcher/fetch_api.py
+++ b/python-threatexchange/threatexchange/fetcher/fetch_api.py
@@ -4,7 +4,7 @@
 """
 The fetcher is the component that talks to external APIs to get and put signals
 
-@see Fetcher
+@see SignalExchangeAPI
 """
 
 
@@ -63,7 +63,7 @@ class SignalExchangeAPI:
         raise NotImplementedError
 
     def report_seen(
-        self, s_type: SignalType, signal: str, metadata: state.FetchMetadata
+        self, s_type: SignalType, signal: str, metadata: state.FetchedSignalData
     ) -> None:
         """Report that this signal was observed on your platform"""
         raise NotImplementedError
--- a/python-threatexchange/threatexchange/fetcher/fetch_api.py
+++ b/python-threatexchange/threatexchange/fetcher/fetch_api.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+The fetcher is the component that talks to external APIs to get and put signals
+
+@see Fetcher
+"""
+
+
+import typing as t
+
+from threatexchange.collab_config import CollaborationConfig
+from threatexchange.signal_type.signal_base import SignalType
+from . import fetch_state_base as state
+
+
+class SignalExchangeAPI:
+    """
+    An external APIs to get and maybe put signals
+
+    Fetchers ideally can checkpoint their progress, so that they can tail
+    updates from the server.
+
+    Is assumed that fetched signals have some metadata attached to them,
+    which is unique to that API. Additionally, it a assumed that there
+    might be multiple contributors to signals inside of an API.
+
+    The Fetcher can assume it is stateful, and
+    """
+
+    @classmethod
+    def name(cls) -> str:
+        """
+        A string name unique to all SignalExchangeAPIs
+
+        This will get stored for lookup from a map of supported APIs
+        keyed str: API
+        """
+        return cls.__name__
+
+    def resolve_owner(self, id: int) -> str:
+        """
+        Convert an owner ID into a human readable name (if available)
+
+        If empty is returned, its assumed that the collaboration name itself
+        is good enough to describe this content.
+        """
+        return ""
+
+    def get_own_owner_id(self) -> int:
+        """
+        Return the owner ID of this caller. Opinions with that ID are "ours"
+        """
+        return -1
+
+    def fetch_once(
+        self, collab: CollaborationConfig, checkpoint: t.Any
+    ) -> state.FetchDelta:
+        """
+        Call out to external resources, pulling down one "batch" of content.
+        """
+        raise NotImplementedError
+
+    def report_seen(
+        self, s_type: SignalType, signal: str, metadata: state.FetchMetadata
+    ) -> None:
+        """Report that this signal was observed on your platform"""
+        raise NotImplementedError
+
+    def report_opinion(
+        self,
+        collab: CollaborationConfig,
+        s_type: t.Type[SignalType],
+        signal: str,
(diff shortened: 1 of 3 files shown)


python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+The fetcher is the component that talks to external APIs to get and put signals
+
+@see Fetcher
```
Dcallies, line 7: Out of date

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
...
+    Is assumed that fetched signals have some metadata attached to them,
+    which is unique to that API. Additionally, it a assumed that there
+    might be multiple contributors to signals inside of an API.
+
+    The Fetcher can assume it is stateful, and
+    """
+
+    @classmethod
+    def name(cls) -> str:
+        """A human-friendly name unique to all fetchers"""
```
Dcallies, line 36: out of date

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
...
+        return cls.__name__
+
+    def resolve_owner(self, id: int) -> str:
+        """Convert an owner ID into an owner record"""
+        return ""
```
Dcallies, line 41: One of my earlier drafts relied on this more, but I think I'll keep it in for now, since I think this is where that name resolution would happen.

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Call out to external resources, pulling down one "batch" of content.
+        """
+        raise NotImplementedError
+
+    def report_seen(
+        self, s_type: SignalType, signal: str, metadata: state.FetchMetadata
+    ) -> None:
+        """
+        Report that this signal is a true positive
```
Dcallies, line 55: fix

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        raise NotImplementedError
+
+    def report_opinion(
+        self, s_type: t.Type[SignalType], signal: str, opinion: state.SignalOpinion
```
Dcallies, line 60: This needs CollabConfig to have the right connection details

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
...
+        raise NotImplementedError
+
+    def report_true_positive(
+        self, s_type: t.Type[SignalType], signal: str, metadata: state.FetchMetadata
+    ) -> None:
+        """Report that a previously seen signal was a true positive"""
+        raise NotImplementedError
+
+    def report_false_positive(
+        self, s_type: t.Type[SignalType], signal: str, metadata: state.FetchMetadata
```
Dcallies, line 74: Note that these don't need CollabConfig because FetchMetadata is unique to collab, and so should have enough context to make a write (TBD when actually trying to implement it though)

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,79 @@
...
+    ) -> None:
+        """
+        Report that this signal is a true positive
```
Dcallies, line 77: update

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,121 @@
...
+    Some APIs may not support all of these, but each of these should influence
+    what action you might take as a result of matching, otherwise it might
+    make more sense as a tag
+    """
+
+    FALSE_POSITIVE = 0  # Signal generates false positives
+    WORTH_INVESTIGATING = 1  # Indirect indicator
+    TRUE_POSITIVE = 2  # Confirmed meets category
+
+    DISPUTED = 100  # When multiple opinions - has both false positive and others
```
Dcallies, line 29: I *think* you can just extend an enum, which makes it simpler to have AggregateOpinion and SingleOpinion

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@dataclass
+class SignalOpinion:
```
Dcallies, line 33: docbloc

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,121 @@
...
+    owner: int
+    category: SignalOpinionCategory
+    category_labels: t.List[str]
+    additional_labels: t.List[str]
+
+
+@dataclass
+class FetchMetadata:
```
Dcallies, line 41: Note that this class is not final, and so HMA might choose to extend this to deal with the specific needs of that storage. This metadata should be unique on a single collab.

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,121 @@
...
+    """
```
Dcallies, line 42: Also add the main details of what it's doing, and not just the weird side details

python-threatexchange/threatexchange/fetcher/fetch_api.py
BarrettOlson: can we have it "raise NotImplementedError" to it does fail silently/non-uniquely

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,121 @@
...
+    Information needed by the SignalExchangeAPI to look up the record
+
+    This powers seen/true positive/false positive reporting
+    """
+
+    collab_name: str  # Corresponds to the collab config that fetched it
+    net_opinion: SignalOpinionCategory
+    opinion_count: int
+    net_category_labels: t.Set[str]
```
Dcallies, line 51: rename all labels tags

BarrettOlson (changes requested):
Chatted offline.
Overall I like were this is headed.
A few things left to do/determine imo
- Making it clear that `FetchMetadata` is a base that must be extended to be usable
- Refining/documenting how `net_opinion` is/should be calculated (and updated)
  - and/or making the list of opinions something that is also populated by default  
- Defining the configuration state of the fetcher that filters the set of labels / tags for each config.

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: I'd imagined this being used only by default for human readable stuff, so I think it's okay to leave this way.

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,123 @@
...
+    Metadata to make decisions on matches and power feedback on the fetch API.
+    """
+
+    collab_name: str  # Corresponds to the collab config that fetched it
+    opinions: t.List[SignalOpinion]
+
+
+class FetchDelta:
+    """
+    Contains the result of a fetch.
```
schatten, line 63: This interface /base class does not contain the actual results. Is that a miss or did you intend it?

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Finish writing the results of previous merges to persistant state.
+
+        This should also persist the checkpoint.
+        """
+        return None
+
+    def get_as_signals(self) -> t.Dict[t.Type[SignalType], t.List[t.Tuple[str, str]]]:
+        """
+        Get as a map of SignalType.name() => (signal, MetataData ID).
```
schatten, line 109: nit: type signature says SignalType, but docstring says SignalType.name() which is str.

schatten: having been the consumer of multiple such tuple style result classes when doing index refactors, I'll be frank, it gets really confusing. 

Thoughts on using [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) here instead? 

They are equally lightweight and self-documenting / self-explanatory.

schatten (approved):
Have left some comments. Approving to unblock because the fetch api looks like a step in the right direction. When implementing concrete classes, more issues might emerge, but LGTM right now.

Somewhat strong questions/nudges:
* Why does FetchDelta not have any 'records'? Is that to be provided by FetchDelta subclasses? 
* Request to sub out t.Tuple[str, str] in favour of a NamedTuple.

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
Dcallies: Hmm, weird, I swear I deleted these once in a previous version already.

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
Dcallies: Intended

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,102 @@
...
+
+@see SignalExchangeAPI
+"""
+
+
+import typing as t
+
+from threatexchange.collab_config import CollaborationConfig
+from threatexchange.signal_type.signal_base import SignalType
+from . import fetch_state_base as state
```
BarrettOlson, line 15: nit/ignorable: stick to absolute even if trivial

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,102 @@
...
+    An external APIs to get and maybe put signals
+
+    Fetchers ideally can checkpoint their progress, so that they can tail
+    updates from the server.
+
+    Is assumed that fetched signals have some metadata attached to them,
+    which is unique to that API. Additionally, it a assumed that there
+    might be multiple contributors to signals inside of an API.
+
+    The Fetcher can assume it is stateful, and
```
BarrettOlson, line 29: nit: and wut?

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,102 @@
...
+        This will get stored for lookup from a map of supported APIs
+        keyed str: API
+        """
+        return cls.__name__
+
+    def resolve_owner(self, id: int) -> str:
+        """
+        Convert an owner ID into a human readable name (if available)
+
+        If empty is returned, its assumed that the collaboration name itself
```
BarrettOlson, line 46: ? is the string returned about a collaboration, owner, or content?

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -0,0 +1,102 @@
...
+        is good enough to describe this content.
+        """
+        return ""
+
+    def get_own_owner_id(self) -> int:
+        """
+        Return the owner ID of this caller. Opinions with that ID are "ours"
+        """
+        return -1
```
BarrettOlson, line 55: nit: why` -1` instead of `raise` or `None`

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,123 @@
...
+class SignalOpinion:
+    """
+    The metadata of a single signal upload.
+
+    Certain APIs won't have any concept of owner, category, or tags,
+    in which case owner=0, category=TRUE_POSITIVE, tags=[] is reasonable
+    default.
+    """
+
+    owner: int
```
BarrettOlson, line 46: nit/ignore: int vs string?

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Note - this currently implies that you are going to load the entire dataset
+        into memory, which once we start getting huge amounts of data, might not make
+        sense.
+        """
+        raise NotImplementedError
+
+    def get_metadata_from_id(self, metadata_id: int) -> FetchedSignalData:
+        """
+        Returns additional metadata from a match
```
BarrettOlson, line 121: nit: ~~match~~ fetched signal

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Merge a subsequent FetchDelta into this one.
+
+        At the implementation's discretion, it may call flush() or the
+        equivalent work.
+        """
+        raise NotImplementedError
+
+    def flush(self) -> None:
+        """
```
BarrettOlson, line 100: nit/ignore/question: does returning None here have a meaning and why does the base class do so instead of throwing?

BarrettOlson (approved):
LGTM, added a few nonblocking nits and questions inline. 

(I think most of my question will surface/be answered once we get to the storage impl)

Dcallies:
Discussed offline, some things are punting forward, some are doing now.

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
Dcallies: Had str, changed to int because I thought it might work , might have to change it back to str depending on implementation.

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: Picking an implementation that will probably do the right thing if unimplemented. Might have to change eventually. Overall goal was to leave the minimum number of methods to get something working. Having a self owner ID isn't strictly needed to fetch a first round of hashes, so provided a default implementation.

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: owner. If the API doesn't have a concept of owners, assume the collaboration name is the owner name. Probably just delete it.

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: AND?

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
Dcallies: Derp, you are right, this should be NotImplemented, since there's expectations on the interface that are not satisfied by a default implementation of `None`

python-threatexchange/threatexchange/fetcher/fetch_state_base.py
Dcallies: Fixed comment, added TODO for namedtuple

PR #912, opened 2022-01-13
[py-te][fetcher] collab config

Dcallies:
Summary
---------

We need to specialize collabconfig in order to give specific instructions to the fetcher, so go ahead and add an interface for that (even though it's duplicated).

Also add "Base" to the name of a few classes that are expected to be overridden to help with that,
Also minor delints in other files.

Test Plan
---------

Lets see if my local linter works!


schatten (approved):
LGTM!

PR #913, opened 2022-01-13
Upgrade Node to v16, v15 is no longer supported

schatten:
## Summary
v15 is no longer supported and it is advised to not use odd number releases in production.
Additionally:
- remove sks keyservers when installing yarn
- fix missing comma in devcontainer example

## Test plan
- rebuild container without cache (installs node, installs dependencies)
- webapp build (npm start) browser opens

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/devcontainer.json.example
+++ b/hasher-matcher-actioner/.devcontainer/devcontainer.json.example
@@ -49,7 +49,7 @@
     "python.analysis.extraPaths": [
       // Unless we do this, pylance is not able to resolve imports
       "~/.local/lib/python3.9/site-packages"
-    ]
+    ],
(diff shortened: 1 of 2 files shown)


Dcallies (approved):
> and it is advised to not use odd number releases in production.

Lol, wut? Is this a node thing?

hasher-matcher-actioner/.devcontainer/devcontainer.json.example
```diff
@@ -49,7 +49,7 @@
     "python.analysis.extraPaths": [
       // Unless we do this, pylance is not able to resolve imports
       "~/.local/lib/python3.9/site-packages"
-    ]
+    ],
```
Dcallies, line 52: Derp. I thought this was just my local copy.

PR #915, opened 2022-01-18
[hma] pin python dependency versions

BarrettOlson:
Summary
---------

Starting point for having a reproducible tagged version of HMA, specifically the API
As well as a draft of a CHANGELOG.md to record updates, breaking changes, and migration instructions 

Test Plan
---------
```
make upload_docker
make dev_test_instance
```


schatten (approved):
LGTM. Do we want to create a v0 Changelog? Something that documents the current capabilities? 

We can start a document internally and then add it as part of a second PR though. Not blocking on this one.

PR #916, opened 2022-01-18
Move Delete button into Edit view for ActionRules tab

obokaiwele:
Summary
---------
Feedback says it's too easy to delete ActionRules in the HMA UI. This moves the delete button from the top level of the ActionRules tab view into the Edit view. You have to start editing an ActionRule before you can delete it,

Test Plan
---------
Before:

After:
AR_002
AR_003

### DIFF
--- a/hasher-matcher-actioner/webapp/src/components/settings/ActionRulesTableRow.tsx
+++ b/hasher-matcher-actioner/webapp/src/components/settings/ActionRulesTableRow.tsx
@@ -136,12 +136,6 @@ export default function ActionRulesTableRow({
             onClick={() => setEditing(true)}>
             <IonIcon icon={pencil} size="large" color="white" />
           </Button>{' '}
-          <Button
-            variant="secondary"
-            className="table-action-button"
-            onClick={() => setShowDeleteActionRuleConfirmation(true)}>
-            <IonIcon icon={trashBin} size="large" className="white" />
-          </Button>
           <Modal
             show={showDeleteActionRuleConfirmation}
             onHide={() => setShowDeleteActionRuleConfirmation(false)}>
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/webapp/src/components/settings/ActionRulesTableRow.tsx
```diff
@@ -190,6 +184,13 @@ export default function ActionRulesTableRow({
...
             <IonIcon icon={checkmark} size="large" color="white" />
           </Button>{' '}
+          <Button
+            variant="secondary"
+            className="mb-2 table-action-button"
+            onClick={() => setShowDeleteActionRuleConfirmation(true)}
+            disabled={!editing}>
+            <IonIcon icon={trashBin} size="large" className="white" />
+          </Button>{' '}
           <Button
```
BarrettOlson, line 194: nit: I think it is slightly more intuitive to have delete (bin) below cancel (X). Thoughts @schatten ?

schatten: Would agree. @obokaiwele can you change the order?

'Save', 'Cancel', some spacing, even an `<hr>` and _then_ 'Delete'.

schatten:
Note, for completeness, I also added a 'before' screenshot from my own instance.

schatten (changes requested):
Thanks for doing this @obokaiwele. As @BarrettOlson flagged, (and I seconded), can you make the order of buttons: 'Save', 'Cancel', some spacing, even an `<hr>` and _then_ 'Delete'.

hasher-matcher-actioner/webapp/src/components/settings/ActionRulesTableRow.tsx
obokaiwele: @BarrettOlson, @schatten I'll reorder the buttons and add the `<hr>`

schatten (approved):
Looks good to me!

PR #918, opened 2022-01-24
Add Troubleshooting Tips

obokaiwele:
Summary
---------

Add two troubleshooting tips to help users through issues with setting up HMA locally.

Test Plan
---------
- CONTRIBUTING.md
Screen Shot 2022-01-24 at 9 49 36 AM

- devcontainer.json:
    - Clone ThreatExchange: `git clone https://github.com/facebook/ThreatExchange.git`
    - Open `hasher-matcher-actioner/` directory in a remote container in Visual Studio Code
    - Choose option in VSCode to open using an existing `Dockerfile`
    - Remote container builds successfully and the directory is opened in the container.

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/devcontainer.json
+++ b/hasher-matcher-actioner/.devcontainer/devcontainer.json
@@ -1,25 +1,20 @@
 // For format details, see https://aka.ms/devcontainer.json.
 {
-  // =================== CTRL+F TODO TO FIND THINGS TO FIX ==============
   "build": {
     "dockerfile": "Dockerfile",
     "args": {
-      // MacOS can try ${env:USER} 
-      // Windows can try ${env:USERNAME}
-      "unixname": "TODO-YOUR-USER-HERE" 
+      // MacOS and Linux use ${localEnv:USER}, Windows uses ${localEnv:USERNAME}
+      "unixname": "${localEnv:USER}${localEnv:USERNAME}"
     }
   },
-  // MacOS can try ${localEnv:USER}
-  // Windows can try ${localEnv:USERNAME}
-  "remoteUser": "TODO-YOUR-USER-HERE",
-    "mounts": [
+  // MacOS and Linux use ${env:USER}, Windows uses ${env:USERNAME}
+  "remoteUser": "${localEnv:USER}${localEnv:USERNAME}",
+  "mounts": [
     "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind",
-    // MacOS can try ${localEnv:HOME}${localEnv:USERPROFILE}
-    // Windows can try ${localEnv:HOMEPATH}
-    "source=TODO-YOUR-HOME-DIR-WITH-AWS/.aws,target=/var/run/aws-config,type=bind,consistency=cached",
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/CONTRIBUTING.md
```diff
@@ -209,3 +209,25 @@ $ hmacli run-api
...
+### Default devcontainer.json
+- Context:
+    - You clone the `ThreatExchange` repo and open the `hasher-matcher-actioner/` directory in a VSCode remote container.
+    - You select VSCode to use a preexisting `Dockerfile`
+    - The VSCode console logs show
+        ```bash
+        [2022-01-24T14:14:49.974Z] /bin/sh: line 26: tar: command not found
+        [2022-01-24T14:14:49.974Z] Exit code 127
+        ```
+- Cause: You did not create a `devcontainer.json` before opening the directory in the remote container. This caused VSCode to create a default `devcontainer.json` which points to the wrong `Dockerfile`.
```
BarrettOlson, line 224: doesn't your change above prevent this now?

hasher-matcher-actioner/.devcontainer/devcontainer.json
```diff
@@ -1,25 +1,20 @@
 // For format details, see https://aka.ms/devcontainer.json.
```
BarrettOlson, line 1: did you mean to update devcontainer.json.example or rename back to devcontainer.json?

BarrettOlson (commented):
Thanks for much for doing this! 
Did you mean to have the file renamed?

hasher-matcher-actioner/CONTRIBUTING.md
obokaiwele: Yes, the change to rename `devcontainer.json` will prevent this particular problem. I left the documentation for this problem partly to explain why the renaming was done, and partly because if future features make `devcontainer.json` not work out of the box, we could be back to using `devcontainer.json.example` which will reopen the possibility of this problem.

hasher-matcher-actioner/.devcontainer/devcontainer.json
obokaiwele: Yes, the renaming was intentional from `devcontainer.json.example` to `devcontainer.json`

hasher-matcher-actioner/CONTRIBUTING.md
BarrettOlson: ahh, I feel like adding a reference to `devcontainer.json.example` while also removing it is a tad questionable. 

We could bring it back but that is when I would add this line. Could we either keep around a copy of `devcontainer.json.example` or instead point them to the new `devcontainer.json`

hasher-matcher-actioner/.devcontainer/devcontainer.json
BarrettOlson: great, I prefer that option it just make the docs a tad confusing

hasher-matcher-actioner/CONTRIBUTING.md
obokaiwele: That makes sense. I've removed the reference to `devcontainer.json.example` since the file has been deleted.

BarrettOlson: Awesome, I'll go ahead and merge. Thanks again for adding this

PR #923, opened 2022-01-28
[hma][ui][bugfix] update create new bank request field to work

BarrettOlson:
Summary
---------

One field was missed from my requested rename. 

Test Plan
---------
Before: cannot create new banks on UI
After: can create new banks on UI


schatten (approved):
Thanks. Glad to know that you've got my back! 👍🏼

PR #924, opened 2022-02-02
BankMember tags are not editable

schatten:
Summary
---------
Editing bank member tags can potentially lead to partners overusing them. We want banks to be the unit of classification rules. Until we discuss this more, disallowing edits to bank member tags when creating and editing them.

Test Plan
---------
Screen Shot 2022-02-01 at 22 18 14
Screen Shot 2022-02-01 at 22 19 11
Screen Shot 2022-02-01 at 22 20 12


BarrettOlson (approved):
LGTM! Thanks
(from the first image I though the red text was on the page)

Dcallies (approved):
Thanks for doing this - I think leaving it visible will let us get some feedback - I think there may be a potential usecase for smaller companies that want to add more context to their reviews.

PR #925, opened 2022-02-02
[hma][api][ui] update match details table to include signal bank details if present

BarrettOlson:
Summary
---------

The content details page definitely had/has some cobwebs on it. This PR does a little clean up and updates the Matches table to show Bank and Dataset membership. 

This resulted in changing one of the fields returned from the `/matches/match/` endpoint
-  `te_signal_details` (instead of the ambiguous `metadata`)  
- and adding the new `banked_signal_details` 
  - which unfortunately at the moment is forced to be backed by an inefficient ddb scan (added warning comment)

Test Plan
---------

Befores:
Screen Shot 2022-01-28 at 12 02 03 PM
Screen Shot 2022-02-02 at 4 13 41 PM

Afters:

Now with Links!
Screen Shot 2022-02-02 at 4 13 54 PM

https://user-images.githubusercontent.com/7664526/152242138-49fd27db-f2c2-41d9-bde6-63394a746771.mov

### DIFF
--- a/hasher-matcher-actioner/webapp/src/pages/ContentDetails.tsx
+++ b/hasher-matcher-actioner/webapp/src/pages/ContentDetails.tsx
@@ -4,7 +4,7 @@
 
 import React, {useState, useEffect} from 'react';
 import {useParams} from 'react-router-dom';
-import {Col, Row, Table, Button} from 'react-bootstrap';
+import {Col, Container, Row, Table} from 'react-bootstrap';
 
 import {
   HashDetails,
@@ -18,7 +18,7 @@ import {formatTimestamp} from '../utils/DateTimeUtils';
 import {getContentTypeForString} from '../utils/constants';
 import ContentMatchTable from '../components/ContentMatchTable';
 import ActionHistoryTable from '../components/ActionHistoryTable';
-import FixedWidthCenterAlignedLayout from './layouts/FixedWidthCenterAlignedLayout';
+import FullWidthLocalScrollingLeftAlignedLayout from './layouts/FullWidthLocalScrollingLeftAlignedLayout';
 import ContentPreview from '../components/ContentPreview';
 import ReturnTo from '../components/ReturnTo';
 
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/webapp/src/pages/ContentDetails.tsx
```diff
@@ -67,73 +67,78 @@ export default function ContentDetailsSummary(): JSX.Element {
...
-              contentId={id}
-              contentType={getContentTypeForString(contentDetails.content_type)}
-              url={img}
-            />
-          ) : null}
-        </Col>
-      </Row>
-      <ContentMatchTable contentKey={id} />
-    </FixedWidthCenterAlignedLayout>
+    <FullWidthLocalScrollingLeftAlignedLayout title="Summary">
```
schatten, line 70: blocking: (kinda): This layout is meant for use where the children implement their own scrolling by using the `overflow` css property. See how [Matches.tsx](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/webapp/src/pages/Matches.tsx#L257) does it. 

In your present impl, if the page becomes longer than a screenful, users will not be able to scroll vertically at all.

schatten: You could create a `FullWidthLeftAlignedLayout` component that copies from `FixedWidthCenterAlignedLayout` but uses container-fluid.

schatten (changes requested):
Thanks for the screenshot. Love the links instead of just duplicating a table! So much more maintainable!

Blocking comment on the scrollability of ContentDetails. Everything else LGTM!

hasher-matcher-actioner/webapp/src/pages/ContentDetails.tsx
BarrettOlson: Done

schatten (approved):

PR #931, opened 2022-02-22
implement pagination for bank members page

crhird:
Summary
---------
In bank members page, each request returns up to 100 photos/videos. With this change whenever a user scrolls to the bottom of this list, a new request will be sent and the additional photos/videos will be added to the list.

Test Plan
---------
Video file too large to add here, so took a couple screenshots showing the steps instead.
Screen Shot 2022-02-23 at 9 27 11 AM
Screen Shot 2022-02-23 at 9 27 15 AM


hasher-matcher-actioner/.devcontainer/post-create
```diff
@@ -14,4 +14,6 @@ python3 -m pip install --upgrade pip
 python3 -m pip install -r requirements-dev.txt
 
 # Install javascript toolchain
-cd webapp && npm install && cd ..
+cd webapp && npm install && npm install react-waypoint --save && yarn add react-waypoint
```
BarrettOlson, line 17: if you updated `package.json` this shouldn't be needed

hasher-matcher-actioner/scripts/create_test_bank.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""
+Replace this with the table to which you want to add data.
+If you don't already have a table, source the script set_tf_outputs_in_local_env.sh
+to create table. Name will be stored in environemnt variable BANKS_TABLE.
+"""
+
+dynamodb = boto3.resource("dynamodb")
+# add the name of your table here
+table_name = ''
```
BarrettOlson, line 22: nit: totally fine with requiring script edits before running for simplicity but can you also make `num_members` a variable as well as `test_bank_name` instead of `"test bank"` and  `1000` respectively.

Also having an assert/conditional that exits and print a warning if you try to use this script without the edits would be appreciated

hasher-matcher-actioner/.devcontainer/post-create
crhird: It doesn't look like package.json is a tracked file in git, so is it somehow created in devcontainer creation and will find this package?

BarrettOlson (approved):
L(Great)TM, thanks!

PR #932, opened 2022-02-23
Action rules UI refresh

schatten:
Summary
---------

Test Plan
---------


BarrettOlson:
> @BarrettOlson getting this into main will make editing difficult if privacy groups were used as conditions. How do we want to handle it?

I have concerns about locking this out on the UI for two reasons
1) The ability to have that as a condition is valuable, if in bank (or te dataset) -> do this   
2) if someone is already using this they could no longer change it.

For 1 there is probably a longer design discussion

As for 2 it could be address with a legacy page/drop down that lets you delete older action-rules  but I am unsure

PR #933, opened 2022-02-24
Add an EFS durable file system for hashing lambda

schatten:
Summary
---
This file system will be used to write content hashes by the hashing lambda so that we can build indexes to use for retroaction.

Test Plan
---
```
$ terraform init
$ terraform apply
```

Added the following snippet to the hashing lambda and deployed.

```python
    with open(f"/mnt/durable-storage/{context.aws_request_id}.file", "w") as f:
        f.write("Hello, World!")

    from os import listdir

    logger.info(listdir("/mnt/durable-storage/"))
```

Hit test on the AWS console a few times, deployed a new version of the lambda and then checked the logs..

Saw the following

```
[INFO]	2022-02-24T03:21:05.008Z	9e057c0d-2027-468a-b797-f261d3964178	['944d7e76-0fb5-4a7b-9104-de86fd90dfc0.file', 'a36b2c1b-1ef4-47d0-aa5c-7f69498c12fd.file', '57e95dc9-efaa-4707-abac-ab1017c7c7ed.file', 'abff2ab6-fd19-47dd-85d8-3bcf2f63c5b7.file', '51191f91-75fc-4f6d-a770-44c46725c054.file', '9e057c0d-2027-468a-b797-f261d3964178.file']
```

See how files accumulate in the file system.

### DIFF
--- a/hasher-matcher-actioner/terraform/main.tf
+++ b/hasher-matcher-actioner/terraform/main.tf
@@ -22,6 +22,8 @@ locals {
   te_data_folder             = module.hashing_data.threat_exchange_data_folder_info.key
   te_api_token_secret_name   = "threatexchange/${var.prefix}_api_tokens"
   hma_api_tokens_secret_name = "hma/${var.prefix}_api_tokens"
+
+  durable_storage_path = "/mnt/durable-storage"
 }
 
 ### Config storage ###
@@ -218,6 +220,12 @@ module "authentication" {
   webapp_and_api_shared_user_pool_client_id = var.webapp_and_api_shared_user_pool_client_id
 }
 
+module "durable_fs" {
+  source          = "./durable-fs"
+  prefix          = var.prefix
+  additional_tags = var.additional_tags
+}
+
 
 /**
  * # Primary S3 Bucket:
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/terraform/main.tf
```diff
@@ -218,6 +220,12 @@ module "authentication" {
   webapp_and_api_shared_user_pool_client_id = var.webapp_and_api_shared_user_pool_client_id
 }
 
+module "durable_fs" {
```
BarrettOlson, line 223: I image this is not free, thoughts on an enable flag for tfvars?

BarrettOlson (approved):
Approving to unblock. Until this feature is more or less supported end to end I would prefer it require an enable/disable configuration in tfvars. Thoughts?

hasher-matcher-actioner/terraform/main.tf
schatten: Setting up an EFS is not going to cost $$. IOPS are billed instead.

So, while I want to make this configurable, I'd rather do that at the individual feature level. 

eg. `tfvars.enable_live_content_clustering` determines whether we write and read. But the EFS exists either way.

I decided against having the EFS optional. It needs to be added to the hashing lambda, api_root and a few other lambdas. Making all of that configurable would have been an exercise in terraform dyanmics that's best avoided. 

LMK if you think otherwise.

PR #935, opened 2022-02-24
[hma][hmalib] bump mypy version and update hmacli install command

BarrettOlson:
Summary
---------

`mypy=0.812` has a dependency that was breaking for the install commands for python3.8+ macOS on arm64. This should hopefully address that. I also updated the make install for hmalib to use `-e` which fixed an issue where my dev venv was not find `hmalib` when I tried running `hmacli` commands 

This should change anything for devcontainer users but I think it is important that install works outside of this as well.
(aside: My personal use case is if I am jumping between hmalib and python-theatexchange (or a third project) being able to use the hmacli to run quick checks or tests outside of the devcontainer  is helpful)

Test Plan
---------

Confirmed that `make dev_install_hmacli` run in both native macOS and linux devcontainer without error.

### DIFF
--- a/hasher-matcher-actioner/setup.py
+++ b/hasher-matcher-actioner/setup.py
@@ -15,6 +15,7 @@ all_extras = set(sum(extras_require.values(), []))
 extras_require["test"] = sorted({"pytest==6.2.1"} | all_extras)
 extras_require["package"] = ["wheel"]
 extras_require["lint"] = ["black==21.12b0"]
+extras_require["type"] = ["types-requests==2.27.1"]
 extras_require["all"] = sorted(set(sum(extras_require.values(), [])))
 
 setup(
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/setup.py
```diff
@@ -15,6 +15,7 @@
 extras_require["test"] = sorted({"pytest==6.2.1"} | all_extras)
 extras_require["package"] = ["wheel"]
 extras_require["lint"] = ["black==21.12b0"]
+extras_require["type"] = ["types-requests==2.27.1"]
```
schatten, line 18: If you are adding requests types, why are we ignoring requests.packages.urllib3 in mypy.ini?

schatten (approved):
Approve to unblock. Have a question inline.

hasher-matcher-actioner/setup.py
BarrettOlson: because the current stub doesn't seem to cover this usage

PR #937, opened 2022-03-02
Support writing typed records to files grouped in disjointed time buckets

Samyakk123:
Summary
---------

The time buckets will be used to cluster pieces of content together upon being hashed by HMA. It will use a time interval provided as the grouping property and group all pieces of content within the given time frame together. 

The benefits to clustering this information together lies in 
- **Retro-actioning** - action media submitted before a hash was added
- **Viewing Representative Media** - While viewing a hash it allows the ability to view media similar to it. For example it allows us to check if a media has been seen in the previous time period and use it as representative media (if missing access to the media)

image

Test Plan
---------

Added a new unittest file test_timebucket.py to ensure working across various time intervals and varying timedeltas
Checked using mypy to resolve any warnings
Uses the freezegun library to run test cases through different time intervals
Run pytest and mypy to see results

```
$ python -m py.test
$ mypy TimeBucketizer.py
```

After performing benchmark tests by testing various speed and memory allocations on the AWS lambda with different buffer sizes, it came to a conclusion that the optimal speed and manageable memory happens with a maximum buffer size of 3200 records. 

image

image

From the chart under, after a size of 3200 the rate of memory usage increases significantly for a trade off of very minimal growth in performance. 
image
image


python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+    def add_record(self, record: T) -> None:
+        """
+        Adds the record to the current bucket.
+        """
+        if len(self.buffer) >= 100 or self.end <= datetime.datetime.now():
+            self._flush()
+        self.buffer.append(record)
+
+    def destroy_remaining(self) -> None:
+        if len(self.buffer):
```
schatten, line 77: This will **return** if self.buffer **has** values. That means lost values. 

Write a test to check for this and then correct it.

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+            return
+        self._flush()
+
+    def get_records(self) -> t.List[T]:
+        """
+        Used for testing. Returns a list of all filepathways found starting from the storage_path
```
schatten, line 83: Out of date docstring.

a) this is not for testing.
b) Should not return list of filepathways.

How is mypy not catching this?

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+            val
+            for sublist in [
+                [os.path.join(i[0], j) for j in i[2]] for i in os.walk(directory_path)
+            ]
+            for val in sublist
+        ]
+
+        return list_of_pathways
+
+    def get_csv_file_count(self, directory_path) -> int:
```
schatten, line 98: This should not be part of the public API.

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Returns all the data in the csv files
+        """
+        with open(os.path.join(directory_path, str(self.id)) + ".csv") as f:
+            data_count = sum(1 for _ in f)
```
schatten, line 103: Wouldn't `len(list(f))` work just fine?

The comprehension feels weird.

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+        return data_count - 1
```
schatten, line 104: Why -1?

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _flush(self) -> None:
+        """
+        Flushes the files currently stored onto the database
```
schatten, line 108: What database?

schatten: Flushes the files? Doesn't this flush _into_ files?

schatten: Into _a_ file?

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+            str(self.start.year),
+            str(self.start.month),
+            str(self.start.day),
+            str(self.start.hour),
+            str(self.start.minute),
+        )
+
+        file_name = str(self.id) + ".csv"
+        directory_path = os.path.join(self.storage_path, self.type, accurate_date)
+        file_pathway = os.path.join(directory_path, file_name)
```
schatten, line 120: path, not pathway..

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if not os.path.isdir(directory_path):
+            os.makedirs(directory_path)
+
+        with open(file_pathway, "a+", newline="") as outfile:
```
schatten, line 125: Explain why you are opening in append mode.

schatten: And why newline is overridden.

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+
+        self.start, self.end = self._calculate_start(self.bucket_width)
+        self.buffer = []
+
+    def get_file_contents(self, file_path) -> List[str]:
+        """
+        Returns all the data stored inside csv file given the file_path
+        """
+        my_file = open(file_path, "r")
+        return list(csv.reader(my_file))
```
schatten, line 137: Remember to Use .from_csv here to convert to CSViable subclasses.

python-threatexchange/threatexchange/signal_type/tests/test_timebucket.py
```diff
@@ -0,0 +1,104 @@
...
+import tempfile
+import shutil
+from threatexchange.timebucketizer import CSViable, TimeBucketizer
+from freezegun import freeze_time
+import os
+from os import walk
+
+
+class TestTimeBuckets(unittest.TestCase):
+    class testA(CSViable):
```
schatten, line 13: A Class (even an inner class) should always start with a capital letter.

python-threatexchange/threatexchange/timebucketizer.py
```diff
@@ -0,0 +1,137 @@
...
+            os.makedirs(directory_path)
+
+        with open(file_pathway, "a+", newline="") as outfile:
+            writer: _csv._writer = csv.writer(outfile)
+            writer.writerows(map(lambda x: x.to_csv(), self.buffer))
+
+        self.start, self.end = self._calculate_start(self.bucket_width)
+        self.buffer = []
+
+    def get_file_contents(self, file_path) -> List[str]:
```
schatten, line 132: This could be a static method. It does not use the instance.

python-threatexchange/threatexchange/signal_type/tests/test_timebucket.py
```diff
@@ -0,0 +1,104 @@
...
+                for _ in range(5):
+                    for _ in range(3):
+                        sample.add_record(self.testA())
+                    frozen_datetime.tick(delta=datetime.timedelta(minutes=1))
+                fileContent = []
+                sample.add_record(self.testA())
+                for i in range(5):
+                    fileContent.extend(
+                        sample.get_file_contents(
+                            os.path.join(
```
schatten, line 73: Ideally, get_file_contents will not be a required method.

And the construction of the file path will be an internal detail of the Timebucketizer class. Imagine if I changed how the path is constrctued in the timebucket class. 

We'll need to change it here too. That means the file path construction is no longer internal.

Try to rework your tests so that they depend on returning records and not files.

All tests will have 3 parts. (setup, query, compare). 

In setup, you create the instance of TimeBucketizer and add records to it, change time values etc.

In query, you retrieve records.

In compare, you compare.

Best if the query method deals with CSViable subclasses rather than records.

Line number 86 (to_csv() is a red flag).

schatten (changes requested):
Sort your imports. `pip install isort` and then `isort <filename>` both your files.

In your test plan, you should use `mypy` too.

Other comments inline.

Overall, this is in better shape than before, but still some way to go..

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+    def from_csv(cls: t.Type[Self], value: t.List[str]) -> Self:
+        raise NotImplementedError
+
+
+T = t.TypeVar("T", bound=CSViable)
+
+
+class TimeBucketizer(t.Generic[T]):
+
+    # TODO Make this generic, so we can have TimeBucket[Tuple[int, int]] or TimeBucket[MoreComplexObject].
```
schatten, line 33: Remove this. We have addressed the TODO.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+        self.start, self.end = self._calculate_bucket_endpoints(datetime.datetime.now(), bucket_width)
+        self.storage_path = storage_path
+        self.id = id
+        self.type = type
+
+        self.buffer_size = buffer_size
+        self.buffer: t.List[T] = []
+
+
+    def _calculate_bucket_endpoints(self, time: datetime.datetime, bucket_width: datetime.timedelta):
```
schatten, line 72: As discussed offline, make this a static method. So that it can be reused by class and instance methods.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import csv
+import datetime
+import os
+import typing as t
+
+MAX_BUFFER_SIZE = 100
```
schatten, line 7: This should be 3200 as we have discussed.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def add_record(self, record: T) -> None:
+        """
+        Adds the record to the current bucket.
+        """
+        if len(self.buffer) >= MAX_BUFFER_SIZE or self.end <= datetime.datetime.now():
+            self._flush()
+        self.buffer.append(record)
+
+    def destroy_remaining(self) -> None:
```
schatten, line 85: This method name is really scary. We are not destroying, are we?

Can we have something a bit more benign like force_flush()?

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Used to trigger a force flush of remaining data stored inside buffer
+        """
+        if not len(self.buffer):
+            return
+        self._flush()
+
+    # Change back to t.List[T] after other method is implemented
+    def get_records(self) -> t.List[str]:
```
schatten, line 94: This "was" only used for testing. Is not anymore. Can you remove this method?

It is supplanted by get_file_contents.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+        # Opening in append mode because it's possible we have to write to the same file multiple times with preexisting data
+        # newline is overriden to prevent spaces in the csv file between sections
+        with open(file_path, "a+", newline="") as outfile:
+            writer = csv.writer(outfile)
+            writer.writerows(map(lambda x: x.to_csv(), self.buffer))
+
+        self.start, self.end = self._calculate_bucket_endpoints(datetime.datetime.now(), self.bucket_width)
+        self.buffer = []
+
+    def get_file_contents(self, since: datetime.datetime, until: datetime.datetime, type: str, storage_path: str, bucket_width: datetime.timedelta, type_class: t.Type[CSViable]):
```
schatten, line 139: This method can be renamed `get_records`. The fact that things are written to a file is an implementation detail.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Divides the day into 24h / bucket_width buckets. When add_record
+        is called, based on current time, appends record to current bucket.
+        Each bucket's data is stored as a serialized CSV file.
+        Note that there may be multiple instances of TimeBucket writing to the
+        same file system so add some uniqueness for this instance.
+
+        Say your storage path is /var/data/threatexchange/timebuckets
+        The file for current datetime (2022/02/08/20:49 UTC) for a bucket_width
+        of 5 minutes should be "/var/data/threatexchange/timebuckets/<type>/2022/02/08/20:45/<unique_id>.csv"
```
schatten, line 46: The path is no longer correct. I think the 20:45 will be split into 20/45.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+        with open(file_path, "a+", newline="") as outfile:
+            writer = csv.writer(outfile)
+            writer.writerows(map(lambda x: x.to_csv(), self.buffer))
+
+        self.start, self.end = self._calculate_bucket_endpoints(datetime.datetime.now(), self.bucket_width)
+        self.buffer = []
+
+    def get_file_contents(self, since: datetime.datetime, until: datetime.datetime, type: str, storage_path: str, bucket_width: datetime.timedelta, type_class: t.Type[CSViable]):
+        """
+        Returns all the data stored inside csv file given the file_path
```
schatten, line 141: Docstring is wrong. No file_path is given. It does not return "all" data. Fix this.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        
+        since_nearest = self._calculate_bucket_endpoints(since, bucket_width)[0]
+        until_nearest = self._calculate_bucket_endpoints(until, bucket_width)[1]
+    
+        content_list = []
+        file_list = [] 
+        while since_nearest <= until_nearest:
+            
+            directory_path = os.path.join(storage_path, type, str(since_nearest.year), str(since_nearest.month), str(since_nearest.day), str(since_nearest.hour), str(since_nearest.minute))
```
schatten, line 151: Can you create a small utility method that creates a path given some parameters. I do not like that the same logic (os.path.join(...many arguments)) is being done in two places.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+            if os.path.isdir(directory_path): 
+                file_list.extend([os.path.join(directory_path, file) for file in os.listdir(directory_path) if os.path.isfile(os.path.join(directory_path, file))])
+            since_nearest += bucket_width
+            
+        for file in file_list:
+            my_file = open(file, "r")
```
schatten, line 157: You must close files that you open. Do this instead:

```python
with open(file, "r") as fp:
  ... do stuff with file...
```

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -0,0 +1,161 @@
...
+
+            content_list.extend(list(map(type_class().from_csv, csv.reader(my_file))))
+        
+        return content_list
```
schatten, line 161: Add new line here. This could be a windows / linux thing.

schatten (approved):
Thanks for doing this @Samyakk123!

As discussed offline, I'm approving this to unblock you.

I have left some comments inline. Lets address them and merge this in!

PR #938, opened 2022-03-08
Memory pressure test of indexes

franklin-phan:
Content
---------
To use the existing FAISS based indexes for Live Content Clustering, we wanted to identify how many content ids could be accommodated in a FAISS index in a lambda index. If users wanted to exceed the limits, they would have to go for a larger lambda instance.

We tested PDQIndex, PDQFlatIndex and TrivialSignalType indexes at increasing lambda memory size values and measured how many hash + content_id entries it could take before AWS killed the lambda for taking up more memory than allocated.

Methodology
---------
A new lambda `eat_memory` was written and terraformed. It takes in the number of entries to generate in the event payload. It then tries to create an index of that size.

Because AWS will kill a lambda that takes more memory than is allocated, we were able to use the killed lambda logs as a signal and identify at what point the lambda memory is exhausted. We tested each index max breakpoint at least 3 times and occasionally the index would still build because of slightly different index build sizes. Each datapoint is subject to a margin of error of 1-2 thousand above what was recorded because of that but that shouldn't affect the slope of each index greatly since millions of Content IDs are required to determine an overall trend.  

Automation
---------

The tests for the `eat_memory` lambda were automated with the lambda-terminal script. The index being tested and how many starting content IDS added to the index can be adjusted in the payload of the script. By adjusting the for loop range amount, you can determine how many content IDs the tests increment by. To change how much memory is being allocated to the lambda, in the AWS Management Console, navigate to the Lambda page, select the eat_memory function, and go to the configuration section and under memory you can allot up to 10240 MBs. After configuring the index type, starting content ID amount, testing range, and allotted lambda memory, you can run the script to obtain test results with
```
python lambda-terminal.py 
```
The readout for if the lambda calls was successful should print out in the terminal. You can read the logs for the `eat_memory` lambda in the Cloud Watch page of the AWS Management Console.

Summary
---------
Below I have provided test results of how many content IDs a certain index can build at specific AWS Lambda memory allocations. The three indexes I tested were the PDQ Index, PDQ Flat Index, and Trivial Signal Type Index. The lambda for these tests is written in eat_memory.py and I automated the tests with my lambda-terminal script. 

Out Of Memory Content ID Count Breakpoints (In Thousands)
| # MB | PDQ | PDQF | TST |
| --- | --- | ---| --- | 
| 128 | 50  | 109  | 174  |
| 256  | 138  | 248 | 369  |
| 512 | 347  | 524 | 783  |
| 1024  | 720  | 1077  | 1657 |
| 2048  | 1550  | 2180 | 3462 |
| 4096 | 3251  | 4388  | 7100  |

PDQ Index Breakpoints
PDQ Flat Index Breakpoints
Trivial Signal Type Index Breakpoints
Index Breakpoint Comparison

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/eat_memory.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/eat_memory.py
@@ -1,0 +1,28 @@
+from hmalib.common.logging import get_logger
+from threatexchange.signal_type.pdq_index import PDQIndex,PDQFlatIndex
+from threatexchange.signal_type.signal_base import TrivialSignalTypeIndex
+import secrets
+
+
+# Make upload_docker
+# terraform -chdir=terraform apply
+
+logger = get_logger(__name__)
+
+items = []
+indexDict = {'PDQIndex':PDQIndex,'PDQFlatIndex':PDQFlatIndex,'TrivialSignalTypeIndex':TrivialSignalTypeIndex}
+def index_increaser(indexType):
+    logger.info(len(items))
+    indexType.build(items)
+    logger.info(f'{indexType} Built')
+
+
+def lambda_handler(event, context):
+    events = list(event.values())
+    i= 1
+    while i < (int(events[1])+1):
+        items.append((secrets.token_hex(32),{"content_id": "8a642764-efab-4d53-a601-523abdcebee3"}))
+        i +=1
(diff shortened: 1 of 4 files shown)


schatten:
This looks good, @franklin-phan. 

As discussed offline, not merging this in. We'll retain the pull request for reference.

PR #944, opened 2022-03-15
[pytx][breaking] ThreatExchange Multifetching Omnibus

Dcallies:
# Summary

Hoo boy.

In the background for the last few months, I've been iterating on what amounts to a complete rewrite of the py-threatexchnage library.

The biggest single change is that it no longer is about Facebook's ThreatExchange libraries, but instead about the exchange of signals in general. In order to support that, all the core interfaces also need to be rejiggered.

## Merge Strategy
I have utterly failed to make this easy to review, and instead tried to sprint to getting into a functional version before we got to review. However, I've taken too long, and now the best way into production is to turn it into a group project. 

Per @schatten and I's discussion offline, we are going to choose to deliberately break trunk. HMA is going to need to be pinned to a version for a while.

## Review Strategy
@schatten is going to be the main reviewer, and he and I will come up with a strategy that thoroughly reviews every file. That may take place after the merge, rather than before, but it will all be read through.

### General Concepts
1. All the key interfaces exist in memory with no opinion on storage
2. Since we know that things will be persisted, but not necessarily how, make stringly-typed lookups for collab, api, signal type, content type
3. A system using the library simple needs to be able to load all APIs, signal types, and content types, and it can do work, and have an interface to load stored collabs and fetch results to have an end-to-end system.
4. There is a "dumb" implementation for everything that would be slow, but is always correct, and allows for attempting fetch+hash+match+writeback even for a previously unknown interface.

### Collaboration Config
The main change is that collaborations can now be for other APIs other than ThreatExchange. However, collaborations might be mixed with other sources for matching.

<TODO david to fill out rest>

# Test Plan

Pending - it is my intent that a lovely collection of unittests are produced. However, I'm not sure it will be in this PR.

Manual test plan:

### Quick e2e validation
```
# Remember to pip install local first

## Interfaces still work
$ threatexchange fetch
$ threatexchange dataset -r  # Manual right now
$ threatexchange match "bball now"
self.execute_print_signal_summary(settings)
```

### Factory Reset
```
$ rm -r ~/.threatexchange
```

### Creating config
```
$ threatexchange config edit threatexchange ...
```


Dcallies:
David's suggested review order:
1. [CollabConfig](https://github.com/facebook/ThreatExchange/pull/944/files#diff-c0fa8e5a9cda3dbc02da7c2880a167d28fa7421fcabb5ae41beb9a4d744bae7b)
2. ContentType
3. SignalType
4. Hash Command
5. Fetcher State
6. Fetcher
7. CliState
8. main.py
9. The rest

python-threatexchange/threatexchange/fb_threatexchange/api.py
```diff
@@ -21,6 +21,47 @@
 from .api_representations import ThreatPrivacyGroup
 
 
+def get_app_token(cli_option: str = None) -> str:
```
schatten, line 24: TODO: Move this to storage / collabConfig storage.

python-threatexchange/threatexchange/fb_threatexchange/api.py
```diff
@@ -21,6 +21,47 @@
...
+    """Get the API key from a variety of fallback sources"""
```
schatten, line 25: doc: Enumerate the fallback sources?

schatten (commented):
Reviewed setup.py and the rename of files from threatexchange/foo → threatexchange/fb_threatexchange/foo

schatten (commented):
Reviewed collab_config.py

python-threatexchange/threatexchange/content_type/pdf.py
```diff
@@ -4,21 +4,47 @@
 """
 Wrapper around the pdf content type.
 """
+
+from io import StringIO
```
schatten, line 8: @schatten Note to self: verify this class is actually able to parse a PDF using a repl session.

schatten (commented):
Reviewed content types. LGTM!

schatten (commented):
Reviewed signals! Phew.

Comments inline.

python-threatexchange/threatexchange/signal_type/trend_query.py
```diff
@@ -57,61 +57,69 @@ class TrendQuerySignal(signal_base.SignalType, signal_base.StrMatcher):
...
-                    set(descriptor.tags),
-                ),
-            )
-        else:
-            old_val[1].merge(descriptor)
-        return True
+    @classmethod
+    def validate_signal_str(cls, signal_str: str) -> str:
+        tq = TrendQuery(
+            json.loads(hash)
```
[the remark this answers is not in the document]
Dcallies: 1. It's a typo from a rename
2. The interface says throw exceptions if you are malformed

python-threatexchange/threatexchange/fb_threatexchange/api.py
Dcallies: I think it has a place in cli main.py

python-threatexchange/threatexchange/fb_threatexchange/api.py
Dcallies: I can, but seems easy to get out of date vs reading the code below

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -5,20 +5,24 @@
...
 import collections
 import typing as t
 import pickle
 
-from threatexchange.signal_type.index import SignalTypeIndex, IndexMatch, T as IndexT
+from threatexchange.signal_type.index import (
+    PickledSignalTypeIndex,
+    SignalTypeIndex,
+    IndexMatch,
+    T as IndexT,
```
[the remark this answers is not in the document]
Dcallies: Not sure - I'm not sure I fully understand how the generics are supposed to work, but this allows me to follow any changes to the variance on Index, but pyre might catch the same thing, I dunno

python-threatexchange/threatexchange/signal_type/pdq_index.py
```diff
@@ -55,36 +59,18 @@ def query(self, hash: str) -> t.List[IndexMatch[IndexT]]:
...
-        """
-        Instantiate an index from a previous call to serialize
-        """
-        return pickle.loads(fin.read())
+    def add(self, signal_str: str, entry: IndexT) -> None:
+        self.add_all(((signal_str, entry),))
+
+    def add_all(self, entries: t.Iterable[t.Tuple[str, IndexT]]) -> None:
+        start = len(self.local_id_to_entry)
+        self.local_id_to_entry.extend(entries)
```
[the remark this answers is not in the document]
Dcallies: It actually isn't! I made the same mistake originally, but the type appears to want to be (str, T), even though that was hidden by incomplete typing in the original.

For review, you can probably generally ignore things that are typing errors as long as the typing as passing.

python-threatexchange/threatexchange/signal_type/pdq_ocr.py
```diff
@@ -56,43 +63,43 @@ def hash_from_file(cls, file: pathlib.Path) -> str:
...
+        if not all((pdq_hash_1, ocr_text_1, pdq_hash_2, ocr_text_2)):
+            return signal_base.HashComparisonResult.no_match_result(255)
+
+        dist = simple_distance(pdq_hash_1, pdq_hash_2)
+        match = False
+        if dist <= cls.PDQ_PLUS_OCR_CONFIDENT_MATCH_THRESHOLD:
+            # Check for text match
+            text1 = common.normalize_string(ocr_text_1)
+            text2 = common.normalize_string(ocr_text_2)
+            match = _levenshtein_text_match(
```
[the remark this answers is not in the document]
Dcallies: Sounds like a reasonable idea, let me see if I can swing it.

python-threatexchange/threatexchange/content_type/pdf.py
Dcallies: It may not - there's a unittest I broke that needs to be restored - I added a card in the project to come back around and fix it.

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -133,3 +142,12 @@ def serialize(self, fout: t.BinaryIO) -> None:
...
         """Instanciate an index from a previous call to serialize"""
         raise NotImplementedError
+
+
+class PickledSignalTypeIndex(SignalTypeIndex[T]):
+    def serialize(self, fout: t.BinaryIO) -> None:
+        fout.write(pickle.dumps(self))
+
+    @classmethod
+    def deserialize(cls, fin: t.BinaryIO) -> "SignalTypeIndex[T]":
```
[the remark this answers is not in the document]
Dcallies: Looks like the answer might be hereabouts: https://peps.python.org/pep-0673/

python-threatexchange/threatexchange/fetcher/collab_config.py
```diff
@@ -23,5 +22,43 @@ class CollaborationConfigBase:
     """
 
     name: str
-    enabled: bool  # Whether to fetch from this or not
-    fetcher_name: str  # Fetch_api.SignalExchangeAPI.name()
+    api: str  # Often a const for implementations
+    enabled: bool  # Whether to fetch/sync
```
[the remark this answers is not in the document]
Dcallies: Discussed offline, I originally attempted to do this, but then ran into an issue where I wanted to gate matches on the CLI by this as well, but HMA definitely wouldn't 

Resolution: Keep as generic "enabled", with the only promise being fetch/sync, and note some implementations may also gate match

python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -5,59 +5,34 @@
...
-    labels: t.Set[str]
-    primary_descriptor_id: int
-    # TODO: Someday, also include the following:
-    # contested: bool
-    # plurality_opinion: t.Tuple(bool, t.Set[str])
-    # highest_action_severity: str
-    # opinions_by_owner: t.Dict[int, MatchOpinion]
+    @classmethod
+    def from_bool(cls, matches: bool) -> "HashComparisonResult":
+        return cls(matches, int(matches))
```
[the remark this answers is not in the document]
Dcallies: Whoops, reversed it.

I think expecting the concept of distance to be around is useful, but it's unclear to me if making it nullable helps. I'd rather punt for now to see how this develops in the post-expansions world.

python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
```diff
@@ -4,29 +4,54 @@
...
+    def test_signal_types_have_content(self):
+        for s in self.SIGNAL_TYPES_TO_TEST:
+            assert s.get_content_types(), "{s!r} has no content types"
 
     def test_str_hashers_have_impl(self):
-        signal_types = [s() for s in meta.get_all_signal_types()]
-        str_hashers = [s for s in signal_types if isinstance(s, StrHasher)]
-        for str_hasher in str_hashers:
-            str_hasher.hash_from_str("test string")
+        text_hashers = [
```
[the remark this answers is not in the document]
Dcallies: The dumb answer:
* It was already here and I just refactored it

The marginally smarter answer:
* I don't think checking that content types are non-empty is typeable, but could be wrong
* ABC is an option for checking that an implementation exists, but just checking that the implementation is callable can shake out a couple of spiders.
* I'm a fan of smoke testing as opposed to no testing, and since we're low, I'd prefer to keep it in for now

python-threatexchange/threatexchange/cli/cli_state.py
```diff
@@ -0,0 +1,198 @@
...
+    def __init__(self, indice_dir: pathlib.Path) -> None:
+        self.dir = indice_dir
+
+    def get_available(self) -> t.List[str]:
+        return [
+            str(f)[-len(self.FILE_EXTENSION)]
+            for f in self.dir.glob(f"*{self.FILE_EXTENSION}")
+        ]
+
+    def clear(
```
schatten, line 59: docstring, please.

python-threatexchange/threatexchange/cli/cli_state.py
```diff
@@ -0,0 +1,198 @@
...
+class CliIndexStore:
+
+    FILE_EXTENSION = ".index"
+
+    def __init__(self, indice_dir: pathlib.Path) -> None:
+        self.dir = indice_dir
+
+    def get_available(self) -> t.List[str]:
+        return [
+            str(f)[-len(self.FILE_EXTENSION)]
```
schatten, line 55: Probably missing a colon here.

str(f)[-len(self.FILE_EXTENSION] will return a single character.

python-threatexchange/threatexchange/cli/cli_state.py
```diff
@@ -0,0 +1,198 @@
...
+            for f in self.dir.glob(f"*{self.FILE_EXTENSION}")
+        ]
+
+    def clear(
+        self, only_types: t.Optional[t.Iterable[t.Type[SignalType]]] = None
```
schatten, line 60: nit: for collections, default to an empty list works just as good as t.Optional[t.Iterable[...]]. 

You'll not need to add guard clauses.

python-threatexchange/threatexchange/cli/cli_state.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def collab_file(self, collab_name: str) -> pathlib.Path:
+        return self.dir / f"{collab_name}.state.json"
+
+    def clear(self, collab: CollaborationConfigBase) -> None:
+        file = self.collab_file(collab.name)
+        if file.is_file():
+            logging.info("Removing %s", file)
+            file.unlink(missing_ok=True)
+        if file.parent.is_dir():
```
schatten, line 113: When can file.parent.is_dir() be False?

schatten (commented):
Reviewed cli.json, cli.cli_config, cli.cli_state and a few of their dependencies.

python-threatexchange/threatexchange/fetcher/fetch_state.py
```diff
@@ -46,11 +66,78 @@ class SignalOpinion:
...
+        new: t.Union["AggregateSignalOpinionCategory", SignalOpinionCategory],
+    ) -> "AggregateSignalOpinionCategory":
+        """
+        Combine signal opinions into an aggregate opinion.
+
+        In general, take the highest confidence/severity of true positives,
+        unless you have both a true + false positive, in which case the result
+        is disputed.
+        """
+        # For some reason, mypy can't deal with new being retyped
```
schatten, line 111: We've used t.cast for this in a lot of hmalib.

python-threatexchange/threatexchange/fetcher/fetch_state.py
```diff
@@ -61,10 +148,27 @@ class FetchedSignalDataBase:
...
+        return [SignalOpinion.get_trivial()]
+
+    @classmethod
+    def merge_metadata(
+        cls, _older: "FetchedSignalMetadata", newer: "FetchedSignalMetadata"
+    ) -> "FetchedSignalMetadata":
+        """
+        The merge strategy when streaming updates.
+        """
+        return newer
```
schatten, line 161: Only if newer.owner == older.ower, right? 

Otherwise, we should not try to merge them?

Or prefer our own owner_id?

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -109,33 +146,56 @@ def report_opinion(
...
+            ),
+        )
 
     def report_false_positive(
         self,
+        collab: CollaborationConfigBase,
         s_type: t.Type[SignalType],
         signal: str,
-        metadata: state.FetchedSignalDataBase,
+        _metadata: state.FetchedSignalMetadata,
```
schatten, line 184: Probably not use _ for an interface method?

python-threatexchange/threatexchange/fetcher/fetch_state.py
schatten: I see that this is done as part of SimpleFetchedSignalMetadata, so feel free to ignore.

python-threatexchange/threatexchange/cli/cli_state.py
Dcallies: Happened in testing where parent folder was missing

python-threatexchange/threatexchange/cli/cli_state.py
Dcallies: This is a deliberate choice to avoid confusing with "I choose to restrict to nothing" versus "I choose not to restrict". I believe the change you are suggesting will cause an actual behavior change due to the suggestion.

python-threatexchange/threatexchange/fetcher/simple/state.py
```diff
@@ -0,0 +1,183 @@
...
+        updates = newer.get_as_update_dict()
+        if not updates:
+            return
+        newer_by_type: t.DefaultDict[
+            str, t.List[t.Tuple[str, t.Optional[fetch_state.FetchedSignalMetadata]]]
+        ] = defaultdict(list)
+        for (stype, signal_str), record in updates.items():
+            newer_by_type[stype].append((signal_str, record))
+
+        for n_type, n_updates in newer_by_type.items():
```
schatten, line 87: Could definitely use a test case. Add as a card to project if not already done.

python-threatexchange/threatexchange/cli/cli_state.py
Dcallies: Whomp whomp

schatten (commented):
Fetcher state reviewed. LGTM. Questions inline, but like the direction this is headed in.

schatten (commented):
Reviewed all fetcher_apis except fb_threatexchange_api.py. That's next.

python-threatexchange/threatexchange/fetcher/fetch_state.py
Dcallies: TIL

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -14,37 +8,251 @@
...
+
+    @property
+    def index_dir(self) -> pathlib.Path:
+        return self._dir / "index/"
+
+    @property
+    def config_file(self) -> pathlib.Path:
+        return self._dir / "config.json"
+
+    def path_for_config(
```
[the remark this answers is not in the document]
Dcallies: A lot of these issues are because this class inherits "CollabStoreBase" (which I tempted to delete), and also is pulling double duty as the general storage. 

I have a card for splitting up state access into a series of helpers where this will be easier to pry apart. I'd like to just TODO for now

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -14,37 +8,251 @@
...
+    def update_persistent_config(self, config: CLiConfig):
+        cli_json.dataclass_dump_file(self.config_file, config)
+
+    def dir_for_fetched_state(
+        self,
+        api: t.Type[SignalExchangeAPI],
+    ) -> pathlib.Path:
+        return self.fetched_state_dir / f"{api.get_name()}/"
+
+    def get_names_without_loading(self) -> t.List[str]:
```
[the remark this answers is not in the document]
Dcallies: Would be solved with a config split as above

python-threatexchange/threatexchange/fetcher/simple/state.py
Dcallies: Done

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: It generates a lint error because it's unused otherwise :/ Maybe that has gone away now that I have children...

python-threatexchange/threatexchange/fetcher/fetch_state.py
Dcallies: Nope, it really means completely replace without checking. However, the docstring is wrong

python-threatexchange/threatexchange/meta.py
```diff
@@ -0,0 +1,102 @@
...
+            for content_type in signal_type.get_content_types():
+                self.signal_type_by_content[content_type].append(signal_type)
+
+    def get_fetcher_classes(self):
+        return [f[0] for f in self.fetcher_data_by_name.values()]
+
+    def get_supported_signal_types_for_content(
+        self, content: t.Type[ContentType]
+    ) -> t.List[t.Type[SignalType]]:
+        return list(self.signal_type_by_content.get(content, ()))
```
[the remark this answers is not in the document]
Dcallies: It's not to defend against wrong types, it's to defend against changing the contents of the list, which causes a failure mode that is very hard to debug. Storing it as a tuple instead might be preferred.

```
stypes = x.get_supported_signal_types_for_content()
stypes.append(other)  # Whoops
```

python-threatexchange/threatexchange/meta.py
```diff
@@ -0,0 +1,102 @@
...
+        self.content_by_name = {c.get_name(): c for c in content_types}
+        self.signal_type_by_name = {s.get_name(): s for s in signal_types}
+        self.signal_type_by_content: t.Dict[
+            t.Type[ContentType], t.List[t.Type[SignalType]]
+        ] = {c: [] for c in content_types}
+        for signal_type in signal_types:
+            for content_type in signal_type.get_content_types():
+                self.signal_type_by_content[content_type].append(signal_type)
+
+    def get_fetcher_classes(self):
```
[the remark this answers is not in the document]
Dcallies: Copypasta failure

python-threatexchange/threatexchange/meta.py
```diff
@@ -0,0 +1,102 @@
...
+        _validate_content_and_signal(content_types, signal_types)
+
+        self.content_by_name = {c.get_name(): c for c in content_types}
+        self.signal_type_by_name = {s.get_name(): s for s in signal_types}
+        self.signal_type_by_content: t.Dict[
+            t.Type[ContentType], t.List[t.Type[SignalType]]
+        ] = {c: [] for c in content_types}
+        for signal_type in signal_types:
+            for content_type in signal_type.get_content_types():
+                self.signal_type_by_content[content_type].append(signal_type)
```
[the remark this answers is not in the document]
Dcallies: I have been burned many time by default dict, and avoid it unless I can keep all the uses of it in sight (in this case, this variable may be queried externally, so I don't want to accidentally create new content types)

python-threatexchange/threatexchange/meta.py
```diff
@@ -0,0 +1,102 @@
...
+from threatexchange.fetcher.fetch_api import SignalExchangeAPI
+from threatexchange.fetcher.fetch_state import FetchedStateStoreBase
+from threatexchange.fetcher.collab_config import CollaborationConfigStoreBase
+
+
+class SignalTypeMapping:
+    def __init__(
+        self,
+        content_types: t.List[t.Type[ContentType]],
+        signal_types: t.List[t.Type[SignalType]],
```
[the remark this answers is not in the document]
Dcallies: I can think of theoretical ones, but those might not be useful. The usecase I thinking of is from inside contexts where we would delibately only expose a subset of functionality to the library. I.e. inside the PDQ special-cased lamda method, the only signal type that exists is PDQ.

python-threatexchange/threatexchange/meta.py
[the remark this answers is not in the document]
Dcallies: You can argue that the the variables themselves are public, but it's much easier to spot intent with this:

  x.signal_type_by_content[x].append(y)

python-threatexchange/threatexchange/meta.py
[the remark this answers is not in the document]
Dcallies: Another note - this is a moot point right now because I don't think there are any types with multiple types right now.

Imagine that we made PDQ work on videos, and it looked for violating frames against the same index - I might enable PDQ for photos, but not want to process videos (for time).

python-threatexchange/threatexchange/cli/json.py
```diff
@@ -0,0 +1,53 @@
...
+import pathlib
+import dacite
+import dataclasses
+from enum import Enum
+import typing as t
+
+T = t.TypeVar("T")
+
+
+def dataclass_dump_file(path: pathlib.Path, obj) -> None:
```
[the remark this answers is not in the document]
Dcallies: Kept it for symmetry with json itself, can refactor cli_config

python-threatexchange/threatexchange/cli/json.py
```diff
@@ -0,0 +1,53 @@
...
+    with path.open("w") as fp:
+        dataclass_dump(fp, obj)
+
+
+def dataclass_dump(fp: t.IO[str], obj) -> None:
+    json_dict = dataclasses.asdict(obj)
+    # Sanity check - we want to make sure it will also come out the other end
+    # And this will wrong-type error if it can't
+    dataclass_load_dict(json_dict, obj.__class__)
```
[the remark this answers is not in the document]
Dcallies: I'm worried about it failing randomly for benign type failiures, but I assume that means there's an incorrect assignment not being caught by typechecking somewhere

python-threatexchange/threatexchange/cli/json.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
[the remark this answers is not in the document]
Dcallies: It stores dataclasses in json format? (the original name was cli_json, but then the path is cli/cli_json)

I am accepting name suggestions - you give me problems but not solutions!

python-threatexchange/threatexchange/fetcher/apis/fb_threatexchange_api.py
```diff
@@ -0,0 +1,287 @@
...
+
+    descriptor_id: t.Optional[int]
+
+
+@dataclass
+class FBThreatExchangeIndicatorRecord(state.FetchedSignalMetadata):
+
+    opinions: t.List[FBThreatExchangeOpinion]
+
+    def get_as_opinions(  # type: ignore  # Why can't mypy tell this is a subclass?
```
schatten, line 74: mypy can't work with this because list is an invariant type in python.

so, you can't do:

```python
b: t.List[SubClass] = []
a: t.List[SuperClass] = []

a = b # mypy implodes!
```


Two ways to handle this...

Either, use `Sequence` which is a covariant type.

Or, in the super class, redefine the return type as a typevar. 

```python
SignalOpinionOrSubclass = TypeVar('SignalOpinionOrSubclass', bound=SignalOpinion)

class FetchedSignalMetadata:
  def get_as_opinions(...) → t.List[SignalOpinionOrSubclass]:
    pass
```

schatten (commented):
Reviewed fb_threatexchange. comment inline.

python-threatexchange/threatexchange/cli/dataset_cmd.py
```diff
@@ -1,184 +1,256 @@
...
+        self.only_tags = set(only_tags)
+        self.signals_only = signals_only
+        self.limit = limit
+
+    def execute(self, settings: CLISettings) -> None:
+        # Maybe consider subcommands?
+        if self.clear_indices:
+            self.execute_clear_indices(settings)
+        elif self.rebuild_indices:
+            self.execute_generate_indices(settings)
```
schatten, line 143: clear before generating?

schatten (approved):
Phew! That was a lot. 

If CI passes, I am fine with this being merged in. There is definitely a lot of work ahead, but I guess that is best done in the form of small PRs on main!

Let's do this!

python-threatexchange/threatexchange/cli/dataset_cmd.py
Dcallies: I'll clear them inline so that you don't end up with no indices if it errors

python-threatexchange/threatexchange/fetcher/apis/fb_threatexchange_api.py
Dcallies: Yea, I eventually learned the reasons (in relation to Dict, which has the same relationship with Mapping).

Dcallies:
Hold on to your butts, we're now continuing in https://github.com/facebook/ThreatExchange/projects/3

PR #947, opened 2022-03-16
[hma][hotfix] comment out vpc setting for hasher lambda 

BarrettOlson:
Summary
---------

Adding a VPC to the hasher lambda seem to prevent it from adding to an SQS queue (for matching) or writing to ddb (for HashRecord)
Because VPC is only needed for efs which is not part of the HMA pipeline yet. I think the best hotfix is just to comment out these setting to unblock the hashing lambda and follow up with the needed changes / network setting to support the filesystem that won't break things. 

Test Plan
---------
```
make upload_docker
...
...
make dev_test_instance
...
```


schatten (approved):
Thanks!

PR #948, opened 2022-03-16
Docs workflow fixes

schatten:
Summary
---------
Got tired of having to do hotfixes in 946, 941 and 940.

Test Plan
---------
Can only test after this lands on main, since workflows are pulled from main. Has to be a better way, but would take a more motivated contributor to figure that out.


BarrettOlson (approved):
whynot

PR #950, opened 2022-03-21
added LCC Indexer class and Hash Record class

franklin-phan:
Summary
---------

This first commit of the LCC Indexer class has pseudo code for the get_recent_index method  and build_index_from_last_24h method. It also has a HashRecord class that contains data for content_hash and content_id.

### DIFF
--- a/hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
+++ b/hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
@@ -1,0 +1,65 @@
+from threatexchange.signal_type.pdq_index import PDQIndex
+from hmalib.common.logging import get_logger
+from datetime import timedelta, datetime
+import time
+from threatexchange.signal_type.timebucketizer import TimeBucketizer, CSViable
+import pickle
+starttime = time.time()
+logger = get_logger(__name__)
+import typing as t
+from dataclasses import dataclass
+
+# @staticmethod
+# def get_file_contents2(
+#         since: datetime.datetime,
+        # until: datetime.datetime,
+        # type: str,
+        # storage_path: str,
+        # bucket_width: datetime.timedelta,
+        # type_class: t.Type[CSViable],
+# ):
+
+Self = t.TypeVar("Self", bound="CSViable")
+
+@dataclass(eq=True)
+class HashRecord(CSViable):
+    """
+    Example class used for testing purposes.
+    """
+    content_hash:str
+    content_id:str
+
+    def to_csv(self) -> t.List[t.Union[str, int]]:
+        return [self.content_hash, self.content_id]
+
+    @classmethod
+    def from_csv(cls: t.Type[Self], value: t.List[str]) -> Self:
+        return HashRecord(value[0],value[1])
+class LCCIndexer:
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
```diff
@@ -0,0 +1,65 @@
+from threatexchange.signal_type.pdq_index import PDQIndex
```
schatten, line 1: Add Facebook copyright notice. Copy from any other file.

Use isort for your imports.

```sh
$ sudo apt install isort 
```

Followed by `isort filename`.

hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
```diff
@@ -0,0 +1,65 @@
...
+    content_hash:str
+    content_id:str
+
+    def to_csv(self) -> t.List[t.Union[str, int]]:
+        return [self.content_hash, self.content_id]
+
+    @classmethod
+    def from_csv(cls: t.Type[Self], value: t.List[str]) -> Self:
+        return HashRecord(value[0],value[1])
+class LCCIndexer:
```
schatten, line 38: nit: format. Use black on each file before submitting. @Samyakk123 can help you with it.

hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
```diff
@@ -0,0 +1,65 @@
...
+        # type: str,
+        # storage_path: str,
+        # bucket_width: datetime.timedelta,
+        # type_class: t.Type[CSViable],
+# ):
+
+Self = t.TypeVar("Self", bound="CSViable")
+
+@dataclass(eq=True)
+class HashRecord(CSViable):
```
schatten, line 25: This PR should only add HashRecord. I want to review LCCIndexer separately. That might be a longer discussion. 

Re-raise this PR (or beat it into shape).

hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
schatten: Typical module names in python are all lower case. Since this is already in `hmalib.indexers`, maybe lcc.py would have sufficed?

hasher-matcher-actioner/hmalib/indexers/LCCIndexer.py
schatten: Also, move the HashRecord class to a more 'common' place. Both indexers and hashers will use it. So putting it in `hmalib.indexers` is sending the wrong message.

eg. [here](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/hmalib/common/models/pipeline.py)

schatten (changes requested):
Thanks for this, @franklin-phan.

Overall comments:
1/ One PR for adding hash-record and another one for the LCC Indexer.
2/ Find a better place to put the hash-record class. See inline-comments.

schatten:
Abandoning per discussion with @franklin-phan

PR #958, opened 2022-03-25
Cli e2e testing

Dcallies:
Summary
---------

This diff adds a backstop for testing, which is simple e2e tests that pretend to be executing the CLI. Each test:
1. Is treated like a fresh install of threatexchange. You can change this by just overriding the setUp() method and move it to a setUpBeforeClass type deal.
2. Has helpers for triggering CLI runs, or looking for specific output.

Test Plan
---------

New unittests!

### DIFF
--- a/python-threatexchange/setup.py
+++ b/python-threatexchange/setup.py
@@ -57,7 +57,7 @@ setup(
     ],
     keywords="facebook threatexchange",
     url="https://www.github.com/facebook/ThreatExchange",
-    packages=find_packages(exclude=["tests*"]),
+    packages=find_packages(exclude=["tests*"]),  # TODO also exclude subdir tests
     install_requires=[
         "python-Levenshtein",
         "requests>=2.26.0",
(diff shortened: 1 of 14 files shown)


python-threatexchange/setup.py
```diff
@@ -57,7 +57,7 @@
     ],
     keywords="facebook threatexchange",
     url="https://www.github.com/facebook/ThreatExchange",
-    packages=find_packages(exclude=["tests*"]),
+    packages=find_packages(exclude=["tests*"]),  # TODO also exclude subdir tests
```
schatten, line 60: Do we have evidence that this does not exclude test packages? 

Why would subdir tests add new packages. New modules I understand, but packages?

schatten (approved):
Looks good!

sahanapeters (approved):
👍

python-threatexchange/setup.py
Dcallies: Maybe I misunderstood an error message I've seen, but I thought threatexchange/foo/tests stuff was being vacuumed into the package

Issue #211, opened 2019-10-17
TMK Clusterize on thousands of .tmk files


jlekas:
I was attempting to use the tmk-clusterize tool on a set of around 50,000 videos and the program crashed before it was finished running. Is there a limit to the number of hash files that the clusterize tool can operate on?

johnkerl:
We need to incorporate FAISS into the clusterizer.

That said, the program might have run out of RAM as it's a single-threaded all-in-one demo -- do you have more context on why it crashed, and/or what you saw on the terminal around the time of the crash?

jlekas:
The program was left running for 1 - 2 days and at a certain point I switched to the open terminal and had stopped running although I unfortunately do not remember the output. Do you think it would it be worthwhile attempting to fork the project and writing a multi threaded or multi processed version of the file to help it run on larger data sets, or would it be better to wait for FAISS to be incorporated into the clusterizer?

johnkerl:
@jlekas both. :)

I really apologize for the delay. My team is actively working on unblocking a couple key flows in ThreatExchange ...

jlekas:
Oh, no worries thank you so much for your timely responses and help in solving this problem. I will look into writing a multi threaded and multi processed version of the file to see if it works well for my larger data set.

Dcallies:
@dxdc  - How are things going after your changes - is it performant enough to consider this task closed out?

dxdc:
@Dcallies I've tested up to 25k tmk files, works great. There may be some better ways to optimize the parallelization, but it would be for smaller scale improvements at this point I would imagine.

PR #314, opened 2020-10-30
Add memory information to faiss index benchmarks

bodnarbm:
Summary
=========

To help understand index selection choice on memory requirements,
adding some benchmark output for how approximately how much storage
is needed for the serialized form for the indexes.

Test Plan
========

Ran the benchmark locally and noted the below output

```
(threatexchange) ➜  pytx3 git:(add-memory-benchmark-for-faiss-indexes) python benchmarks/benchmark_pdq_faiss_matchers.py                   
Benchmark: PDQ Faiss Matcher Comparison

Options:
         faiss_threads :  1
         dataset_size :  250000
         num_queries :  1000
         thresholds :  [0, 15, 31, 47]
         seed :  None

using random seed of  1604086833819350000
use --seed  1604086833819350000  to rerun with same random values

Building Stats:
        PDQFlatHashIndex: time to build (s):  0.3096351623535156
        PDQFlatHashIndex: approximate size: 7,812KB
        PDQMultiHashIndex: time to build (s):  2.423464298248291
        PDQMultiHashIndex: approximate size: 20,858KB

Benchmarks for threshold:  0
        PDQFlatHashIndex - Total Time to search  (s):  0.7060251235961914
        PDQMultiHashIndex - Total Time to search  (s):  0.029154062271118164
        PDQFlatHashIndex - Precent of targets found:  100.0
        PDQMultiHashIndex - Precent of targets found:  100.0

Benchmarks for threshold:  15
        PDQFlatHashIndex - Total Time to search  (s):  0.6720812320709229
        PDQMultiHashIndex - Total Time to search  (s):  0.022841930389404297
        PDQFlatHashIndex - Precent of targets found:  100.0
        PDQMultiHashIndex - Precent of targets found:  100.0

Benchmarks for threshold:  31
        PDQFlatHashIndex - Total Time to search  (s):  0.6760938167572021
        PDQMultiHashIndex - Total Time to search  (s):  0.35967183113098145
        PDQFlatHashIndex - Precent of targets found:  100.0
        PDQMultiHashIndex - Precent of targets found:  100.0

Benchmarks for threshold:  47
        PDQFlatHashIndex - Total Time to search  (s):  0.6572327613830566
        PDQMultiHashIndex - Total Time to search  (s):  2.8741109371185303
        PDQFlatHashIndex - Precent of targets found:  100.0
        PDQMultiHashIndex - Precent of targets found:  100.0
```

Issue #197, opened 2019-08-27
TMK integration with FAISS


bbhopesh:
README.md says that integration has been successfully prototyped and information will be posted soon. Can you please give more information about integration with FAISS? We are using TMK in one of our applications and we need an efficient way of querying.

bbhopesh:
Hey @johnkerl, now that I can compile `tmk-query-with-faiss`, I am running into some issues while running it. It fails with FaissException. It seems to be some error because of way faiss works. Below is the error snapshot. Can you please take a look?

![image](https://user-images.githubusercontent.com/8807496/64060487-0ca77980-cb82-11e9-8148-bbc758c1fa95.png)

johnkerl:
@bbhopesh I checked with the FAISS maintainers & it turns out we should transparently fall back to linear search for small-sized datasets. Using FAISS when the sizes are big enough to be useful. This would take a wrapper method.

bbhopesh:
If it's not lot of work, can you please write that wrapper function? I don't know details of FAISS to implement it correctly.

johnkerl:
No worries, we can do so! :)

johnkerl:
I don't know FAISS well either but it would be the current guts of `tmk-query.cpp` whenever the current guts of `tmk-query-with-faiss.cpp` fails due to sizing issues.

kenkyusha:
has anybody made reasonable guide how this should be set up from scratch?
I did clone this repo and ran into the problem saying *faiss* is missing.
After cloning faiss ran into the issue of not having OpenMP running properly.
Luckily this was quick fix just adapting the g++ in the makefile.inc to the one which works with fopenMP, the faiss compiles and installes (lots of warning in the log).
And now I am trying to build the TMK again and get following:
*using the g++ working with fopenmp*
ld: unknown option: -rpath=/usr/local/lib
collect2: error: ld returned 1 exit status
make: *** [tmk-query-with-faiss] Error 1

*using the default one*:
ld: unknown option: -rpath=/usr/local/lib
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [tmk-query-with-faiss] Error 1

any hints would be helpful :-)

johnkerl:
@kenkyusha thanks, this definitely needs a write-up. :^/

My team is having a heavy conference month & I hope we can get to this very very soon.

kenkyusha:
> @kenkyusha thanks, this definitely needs a write-up. :^/
> 
[...]

PR #959, opened 2022-03-26
Parallel, multicore processing for TMK

dxdc:
Summary
---------

After carefully reviewing the code in `tmk-query-with-faiss.cpp`, I don't think that the current implementation is working as expected.

- FAISS search is limited to the k-nearest neighbors, which is hardcoded to 5. This is very limited for larger datasets.
- TMK comparison is using `computeCosSim`, and I'm not sure the similarity evaluation is similar (or, needs to be similar?) by FAISS
- Inputs for level 1 and level 2 scores are accepted, but not considered or used at all

However, there is a desire to increase performance (e.g., https://github.com/facebook/ThreatExchange/issues/211).

I took a stab at implementing parallel processing with OMP. I left the code base the same, and the OMP library as optional at compile time, for now so this could be merged without disrupting users that may not have OMP installed or if different flags are required on other OS's.

Sample results (3438 TMK files), `c1` = 0.7, `c2` = 0.7:
- `tmk-query-with-faiss` @ 0.000143 sec ## note: an unfair comparison, because c1/c2 are not considered
> note that, total runtime for query-with-faiss vs. query-parallel is around 80%, that is just for the query step
- `tmk-query-parallel` @ 0.323135 sec
- `tmk-query` @ 4.336303 sec

I haven't yet addressed `tmk-find-cluster-reps.cpp`, but the implementation would be nearly identical.

Test Plan
---------

- OMP must be installed. On a Mac, this can be done with `brew install libomp`.
- Run `make parallel` to build the multi-cpu variants. I also updated the faiss compiler directives to work properly on Mac.


dxdc:
FYI, I have been getting some segfaults with tmk-clusterize-parallel above 3000 tmk files or so. In some earlier iterations, I tried parallelizing only the inner loop and that also gave excellent results (maybe 80-90% as good), but not as good as parallelizing both.

Tweaking the stack size limits seems to resolve it. Completes in 22 seconds for the below query with 3438 tmk files.

```sh
# see:  https://stackoverflow.com/questions/13264274/why-segmentation-fault-is-happening-in-this-openmp-code
ulimit -v 1048576
KMP_STACKSIZE=512m ./tmk-clusterize-parallel --c1 0.9 --c2 0.4 --min 2 ./tmkfiles/*.tmk
```

dxdc:
Have now also addressed these points:

- FAISS search is limited to the k-nearest neighbors, which is hardcoded to 5. This is very limited for larger datasets.
- Inputs for level 1 and level 2 scores are accepted, but not considered or used at all
- Default fallback to tmk-query if number of points are below the FAISS recommended threshold (see notes in #197). Simple implementation for now.

Regarding:
> TMK comparison is using computeCosSim, and I'm not sure the similarity evaluation is similar (or, needs to be similar?) by FAISS

Would be helpful to get some help on this for CosSim type searches since I'm far from an expert at understanding how FAISS works. I've tried the following, based on some guidance I've found (https://github.com/facebookresearch/faiss/issues/61), which *does* seem to give better slightly better results.

- Use `faiss::IndexFlatIP` instead of `faiss::IndexFlatL2`
- It seems we're supposed to L2-normalize the vectors. I found `faiss::fvec_renorm_L2`, which looks like it could be the correct method? I'm not entirely sure about the parameters. E.g., maybe something like `faiss::fvec_renorm_L2(vector_dimension, num_database_vectors, database.data())`

Still, the current FAISS implementation appears to be trying to make the level 1 search faster, but that is actually the fastest part of the search. The level 2 search is a lot slower, more accurate, and I think is [happening here](https://github.com/facebook/ThreatExchange/blob/f9aff6dd0c90e6c47ffe4151bced4de1676d84f6/tmk/cpp/algo/tmkfv.cpp#L566-L672).

Dcallies:
Hey Daniel! Thanks a bunch for digging into this! I haven't finished digesting the earlier message you sent me, but as far as I got was concluding that `faiss::fvec_renorm_L2` was the correct method according to the note on cosign distance in the FAISS documentation at https://github.com/facebookresearch/faiss/wiki/MetricType-and-distances#how-can-i-index-vectors-for-cosine-similarity.

Sadly there has been turnover on the project since the original authors have written this module, I think this is going to be make or break by the test plan. I'll ask around internally (@BarrettOlson might be the best reviewer, but we'll all be making it up as it goes), but what would be convincing to me as a test plan is:
1. Show the expected distance with the brute force with X videos (c1 and c2 dists if they matter)
2. A couple of queries with FAISS, show that the correct videos are returned.

Your performance numbers are super useful! This isn't apples-to-apples because it was on PDQ and L1 distances, but here's what a previous author did for benchmarking: #314. You've already put in a lot of work, so as long as we're reasonably certain it's correct, I'm personally willing to merge as-is and will try and get you a fast turnaround.

tmk/cpp/bin/tmk-query-with-faiss.cpp
```diff
@@ -168,16 +166,29 @@ int main(int argc, char** argv) {
...
+  if (verbose) {
+    printf("VECTOR_DIMENSION     %d\n", (int)vector_dimension);
+    printf("NUM_DATABASE_VECTORS %d\n", (int)num_database_vectors);
+    printf("NUM_CENTROIDS        %d\n", (int)num_centroids);
+  }
 
   // The coarse quantizer should not be deallocated before the index.
   // 4 = number of bytes per code (vector_dimension must be a multiple of this)
   // 8 = number of bits per sub-code (almost always 8)
   faiss::IndexIVFPQ faiss_index(&coarse_quantizer, vector_dimension, num_centroids, 4, 8);
```
Dcallies, line 178: I'm confused about why this isn't inner product, since that's what is needed to query cosin distance. Do you know?

Ref: https://github.com/facebookresearch/faiss/wiki/Faiss-indexes

dxdc: @Dcallies I found the same references as you, and had the exact same set of thoughts. To query cosine similarity, it seems like we're supposed to use IP and L2-normalized vectors. I just pushed the changes I mentioned at the end of https://github.com/facebook/ThreatExchange/pull/959#issuecomment-1079741989: [a5863bd](https://github.com/facebook/ThreatExchange/pull/959/commits/a5863bddadd91b5c1308ed0e02a482cadd6517e7). Hopefully, it's more clear.

I do see an improved result set, but I have no idea if I've called `fvec_renorm_L2` correctly. I don't see any errors though.

Dcallies:
Whoops, no need to run the PDQ indice tests, just was showing how they did the benchmarking. You can see the variables they listed on the top for faiss inputs. 

The thresholds for PDQ would instead be c1 distance for this.

dxdc:
Ah, OK!

Modified strategy:

- `haystack.txt` now to include arbitrary collection of ~4100 TMK files I have on my machine + all TMK files in `./regtest`
  - additional ~4100 files were only added so as to have sufficient data without aborting with an error.
- `needles.txt` set to `./regtest/chair-19-sd-bar.tmk` (one of the README test cases)
- tmk-query-with-faiss thresholds set to c1=0 c2=1, to return maximal hits

NOTE: `tmk-query-with-faiss` doesn't return all the same movies, but that was true initially. It's only returning the "best" case of a 100% match. As for the comparison of a5863bd, it seems slightly less helpful with this test case:

```
Query  0 ./regtest/chair-19-sd-bar.tmk:
  distance 0.062066 L1 1.000000 L2 1.000000 metadata ./regtest/chair-19-sd-bar.tmk ## original
  distance 0.090200 L1 1.000000 L2 1.000000 metadata ./regtest/chair-19-sd-bar.tmk ## with changes from [a5863bd]
```

Here are the full results. Note that the parallel and non-parallel variants have the same results, but in a different order due to the nature of how it's being run. We don't care about that because ultimately we can just sort the results at the end.

```
Performance comparison for: 4140 TMK files

With FAISS: 'tmk-query-with-faiss'

VECTOR_DIMENSION     256
NUM_DATABASE_VECTORS 4140
NUM_CENTROIDS        257
MINIMUM POINTS NOT MET FOR FAISS: 4140 FOUND vs 10023 EXPECTED
Start training the quantizer:
Training level-1 quantizer
Training level-1 quantizer on 4140 vectors in 256D
WARNING clustering 4140 points to 257 centroids: please provide at least 10023 training points
Training IVF residual
computing residuals
training 4x256 product quantizer on 4140 vectors in 256D
Training PQ slice 0/4
WARNING clustering 4140 points to 256 centroids: please provide at least 9984 training points
Clustering 4140 points in 64D to 256 clusters, redo 1 times, 25 iterations
  Preprocessing in 0.00 s
  Iteration 24 (0.05 s, search 0.05 s): objective=329.363 imbalance=2.207 nsplit=0       
Training PQ slice 1/4
WARNING clustering 4140 points to 256 centroids: please provide at least 9984 training points
Clustering 4140 points in 64D to 256 clusters, redo 1 times, 25 iterations
  Preprocessing in 0.00 s
  Iteration 24 (0.05 s, search 0.05 s): objective=77.461 imbalance=2.638 nsplit=0        
Training PQ slice 2/4
WARNING clustering 4140 points to 256 centroids: please provide at least 9984 training points
Clustering 4140 points in 64D to 256 clusters, redo 1 times, 25 iterations
  Preprocessing in 0.00 s
  Iteration 24 (0.05 s, search 0.04 s): objective=31.4667 imbalance=2.855 nsplit=0       
Training PQ slice 3/4
WARNING clustering 4140 points to 256 centroids: please provide at least 9984 training points
Clustering 4140 points in 64D to 256 clusters, redo 1 times, 25 iterations
  Preprocessing in 0.00 s
  Iteration 24 (0.05 s, search 0.04 s): objective=18.6312 imbalance=2.662 nsplit=0       
IndexIVFPQ::precompute_table: precomputed tables not needed for inner product quantizers
End training the quantizer.
 add_core times: 0.001 17.034 0.455 
imbalance factor: 3.68157
LOAD SECONDS   = 1.420
NEEDLES COUNT  = 1
HAYSTACK COUNT = 4140

CHECK SECONDS = 0.001
Searching for the 15 nearest neighbors

QUERY SECONDS = 0.000193
MEAN QUERY SECONDS = 0.000193
Query results (vector IDs, then distances):
(Note that the nearest neighbor is not always at distance 0 due to quantization errors.)
Query  0 ./regtest/chair-19-sd-bar.tmk:
  distance 0.090200 L1 1.000000 L2 1.000000 metadata ./regtest/chair-19-sd-bar.tmk


 -- Finished: 2.107 seconds

Without FAISS: 'tmk-query-parallel'
LOAD SECONDS   = 1.140
NEEDLES COUNT  = 1
HAYSTACK COUNT = 4140

CHECK SECONDS = 0.002

QUERY FOR ./regtest/chair-19-sd-bar.tmk
  0.727866 0.733801 ./regtest/chair-orig-22-fhd-no-bar.tmk
  1.000000 1.000000 ./regtest/chair-19-sd-bar.tmk
  0.952330 0.961902 ./regtest/chair-orig-22-sd-bar.tmk
  0.895393 0.901900 ./regtest/chair-22-with-small-logo-bar.tmk
  0.727833 0.733890 ./regtest/chair-orig-22-hd-no-bar.tmk
  0.953348 0.962158 ./regtest/chair-22-sd-grey-bar.tmk
  0.989083 0.991138 ./regtest/chair-20-sd-bar.tmk
  0.952936 0.962041 ./regtest/chair-22-sd-sepia-bar.tmk

QUERY SECONDS = 0.010351
MEAN QUERY SECONDS = 0.010351

 -- Finished: 1.793 seconds

Without FAISS: 'tmk-query'
LOAD SECONDS   = 1.147
NEEDLES COUNT  = 1
HAYSTACK COUNT = 4140

CHECK SECONDS = 0.002

QUERY FOR ./regtest/chair-19-sd-bar.tmk
  1.000000 1.000000 ./regtest/chair-19-sd-bar.tmk
  0.989083 0.991138 ./regtest/chair-20-sd-bar.tmk
  0.953348 0.962158 ./regtest/chair-22-sd-grey-bar.tmk
  0.952936 0.962041 ./regtest/chair-22-sd-sepia-bar.tmk
  0.895393 0.901900 ./regtest/chair-22-with-small-logo-bar.tmk
  0.727866 0.733801 ./regtest/chair-orig-22-fhd-no-bar.tmk
  0.727833 0.733890 ./regtest/chair-orig-22-hd-no-bar.tmk
  0.952330 0.961902 ./regtest/chair-orig-22-sd-bar.tmk

QUERY SECONDS = 0.059822
MEAN QUERY SECONDS = 0.059822

 -- Finished: 1.760 seconds
 ```

File: tmk/cpp/bin/tmk-clusterize.cpp [hunk 2]
```diff
@@ -224,45 +231,75 @@ void snowballClusterize(
...
+
+  if (verbose) {
+    printf("\n");
+    printf("CALCULATING THE ADJACENCY MATRIX\n");
+  }
+
+  #pragma omp parallel
+  #pragma omp single nowait
+  {
+    for (const auto& it1 : metadataToFeatures) {
```
Dcallies, line 252: Dumb question - why not parallel for?

https://stackoverflow.com/questions/1448318/omp-parallel-vs-omp-parallel-for

tmk/cpp/bin/tmk-two-level-score.cpp
```diff
@@ -142,39 +150,66 @@ int main(int argc, char** argv) {
...
+                printf(
+                    "%.6f %.6f %s %s\n", s1, s2, metadata1.c_str(), metadata2.c_str());
+              }
+            }
+          } // end omp task it2
+        } // end for loop it2
+      } // end omp task it1
+    } // end for loop it1
+  } // end parallel
+  #pragma omp taskwait
```
Dcallies, line 205: (Dumb question): If you do single without nowait, does that just generate this wait here?

Dcallies (commented):
Some quick questions/comments

tmk/cpp/bin/tmk-clusterize.cpp
dxdc: Using `parallel for` was actually my first attempt, but it gave a bunch of errors that I couldn't understand. I think it doesn't like the use of map in parallel for. There is a way to use the collapse syntax with parallel for, e.g. https://stackoverflow.com/a/28483812/4626770 that seems very applicable if that could be sorted out.

tmk/cpp/bin/tmk-two-level-score.cpp
dxdc: I wish I had a deeper understanding of this library to answer your question. I'm not familiar enough with it, and I'm confident there are some improvements that could be made! I tried a few different methods until I stumbled on this one that worked.

https://stackoverflow.com/a/13810539/4626770

tmk/cpp/Makefile
```diff
@@ -42,6 +42,12 @@ LIBOBJ= \
 ./raster/rasterwriters.o \
 ./raster/timeresamplers.o
 
+EXE_PARALLEL= \
+ tmk-query-parallel \
+ tmk-clusterize-parallel \
+ tmk-two-level-score-parallel \
+ tmk-query-with-faiss  # Pending https://github.com/facebook/ThreatExchange/issues/197
```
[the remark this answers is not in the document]
dxdc: a6499bc

tmk/cpp/bin/tmk-clusterize.cpp
```diff
@@ -224,45 +231,75 @@ void snowballClusterize(
...
+      #pragma omp task firstprivate(it1)
+      {
+        const std::string& filename1 = it1.first;
+        const std::shared_ptr<TMKFeatureVectors> pfv1 = it1.second;
+        // printf("... %s\n", filename1.c_str());
```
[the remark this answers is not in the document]
dxdc: a6499bc (note: I am not the author of this, but removed)

tmk/cpp/bin/tmk-two-level-score.cpp
Dcallies: My level of expertise:

image

tmk/cpp/bin/tmk-query-with-faiss.cpp
```diff
@@ -161,23 +164,36 @@ int main(int argc, char** argv) {
...
   faiss_index.verbose = verbose;
 
+  // check for min data size for optimal FAISS performance
+  int min_points = faiss_index.cp.min_points_per_centroid * num_centroids;
+  if (num_database_vectors < min_points) {
+    if (verbose) {
+      printf("MINIMUM POINTS NOT MET FOR FAISS: %d FOUND vs %d EXPECTED\n", (int)num_database_vectors, min_points);
+    }
+    if (!force) {
+      return tmkQuery(argc, argv);
```
Dcallies, line 193: (blocking): This seems spooky to me - tmkQuery might have a different set of arugments and parsing function, and the error messages seem like they'll get confusing, since they each have their own usage() function.

Additionally, this file is called "query with faiss" and so I'd expect to it query with faiss or just error out, without fallback behavior.

Dcallies (changes requested):
Been in conversation with @BarrettOlson and we both are in agreement we probably don't have enough knowledge to do an expert review.

For now, we're going to go with, if:
1. It seems to work on Daniel's machine
2. The test plan is clear/easy to reproduce/looks right from our best guess

Then we'll merge it in, and hopefully it's a marginal improvement on your usecase.

One thing that is currently making this harder to review is that the PR is starting to get up there in size. It would be faster to review chunks of it (i.e. adding parallel to tmk-clusterize) than as part of a bigger PR. It can be a pain to split up PRs if you aren't used to it, so I'm open to keep iterating on this one as long as it doesn't get any bigger.

There was some functionality around trying to make tmk-query re-usable for different executables that I was unsure about merging, and a #ifdef gated main function which weirded me out. I've noted these with "blocking"

tmk/cpp/bin/tmk-query-with-faiss.cpp
```diff
@@ -161,23 +164,36 @@ int main(int argc, char** argv) {
   size_t num_database_vectors = haystackMetadataToFeatures.size();
 
   // Make the index object and train it
-  faiss::IndexFlatL2 coarse_quantizer(vector_dimension);
+  faiss::IndexFlatIP coarse_quantizer(vector_dimension);
```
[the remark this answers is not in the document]
dxdc: added note and made other changes. hopefully more straightforward now!

tmk/cpp/bin/tmk-query-with-faiss.cpp
```diff
@@ -161,23 +180,38 @@ int main(int argc, char** argv) {
...
 
+  // check for min data size for optimal FAISS performance
+  int min_points = faiss_index.cp.min_points_per_centroid * num_centroids;
+  if (num_database_vectors < min_points) {
+    if (verbose) {
+      printf("MINIMUM POINTS NOT MET FOR FAISS: %d FOUND vs %d EXPECTED\n", (int)num_database_vectors, min_points);
+    }
+    if (!force) {
+      printf(stderr, "Number of points is below minimum size for faiss. Run with --force flag to ignore this error.\n");
+      exit(2);
```
Dcallies, line 211: 👍

Dcallies (approved):
Thanks for those changes! Let's merge this in!

Issue #189, opened 2019-08-09
[Error] no matching function for call to'getline (ch) Ar**, size_t*, FILE*)'


Yisu123:
Excuse me, when I implemented PDQ by CPP , I found that in pdq-photo-hasher.cpp there seemed to be no getline namespace (line 125). When I used "std:: getline (& filename, & linelen, stdin)" to solve this problem, I met [Error] no matching function for call to'getline (ch) Ar**, size_t*, FILE*)',Do you know any way to solve this problem?Thank you.

johnkerl:
What is your compiler version? E.g. `g++ --version`.

johnkerl:
FYI, this should be `getline` from `stdio.h` not `std::getline` (which involves `iostream`).

johnkerl:
Can you please share the compiler output you saw from "there seemed to be no getline namespace (line 125)", before changing from `getline` to `std::getline`?

Yisu123:
Thank you very much for your reply. I've solved this problem, probably because I compiled it on Windows and lacked some necessary libraries. When I compiled it in linux, I didn't find this error. so I think it may be a problem of my computer environment. I'm sorry to disturb you.

johnkerl:
No worries -- thank you! :)

PR #960, opened 2022-03-26
WIP: Compilation changes for Windows

dxdc:
Summary
---------

Had similar issues as described in https://github.com/facebook/ThreatExchange/issues/189 so made several edits which allow compilation to succeed on Windows.

I also added some flags to reduce `ffmpeg`'s noisy output (if verbose was not enabled).

There are still changes needed to the Makefile, but I haven't attempted merging those yet. In brief:

Instead of:
`CC=g++ -O2 -std=c++14 -I../..`

Use:
`CC_WIN=g++ -O2 -std=gnu++14 -I../..`

for:
- ./hashing/filehasher.o
- tmkdump
- tmk-clusterize
- tmk-query
- tmk-two-level-score

You will also need to install `make`, e.g., via [choco](https://community.chocolatey.org/packages/make) and [MinGW](https://osdn.net/projects/mingw/). Note that, MinGW needs to be added to the system path in order to compile and use the executables.

The tests will not run after `make` without more changes, but they can be run manually. Haven't extensively tested, but output appears identical.

Still need:
- [ ] README
- [ ] Makefile changes (including testing)

Test Plan
---------

### DIFF
--- a/pdq/cpp/hashing/pdqhashing.cpp
+++ b/pdq/cpp/hashing/pdqhashing.cpp
@@ -13,7 +13,6 @@
 
 #if defined(_WIN32)
   #define _USE_MATH_DEFINES
-  #include <iostream>
 #endif
 
 #include <cassert>
--- a/pdq/cpp/hashing/pdqhashing.cpp
+++ b/pdq/cpp/hashing/pdqhashing.cpp
@@ -11,6 +11,11 @@
 // https://github.com/facebookexternal/ThreatExchange-PDQ/blob/main/pdqhash-2017-10-09.pdf
 // ================================================================
 
+#if defined(_WIN32)
+  #define _USE_MATH_DEFINES
+  #include <iostream>
+#endif
+
 #include <cassert>
 #include <chrono>
 #include <cmath>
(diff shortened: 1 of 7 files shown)


pdq/cpp/hashing/pdqhashing.cpp
```diff
@@ -11,6 +11,11 @@
 // https://github.com/facebookexternal/ThreatExchange-PDQ/blob/main/pdqhash-2017-10-09.pdf
 // ================================================================
 
+#if defined(_WIN32)
+  #define _USE_MATH_DEFINES
+  #include <iostream>
```
Dcallies, line 16: (Dumb question, it's been a long time since cpp): I don't see cin/cout or friends anywhere, did you include this for functionality or to help with debugging? (Same question on other imports)

tmk/cpp/hashing/filehasher.cpp
```diff
@@ -9,6 +9,17 @@
 #include <stdio.h>
 #include <memory>
 
+#include <iostream>
```
Dcallies, line 12: Is this used?

Dcallies (commented):
This all looks reasonable to me, and my next ask was for README updates, which you said you've already planned.

@BarrettOlson - Do you want to take a second look since I think you've touched c++ land more recently than I?

dxdc:
Thanks @Dcallies :

- Removed unused imports (good catch, forgot to remove those)
- Update Makefile to support basic OS detection (and simplify README)
- Add some notes to README

PR #961, opened 2022-03-28
[HMA] Add dumb API for batch detached signal additions

schatten:
Summary
---
Add a bank_operation that supports batch signal additions to a bank. With threatexchange → banks being worked on, I have a hunch this will be useful. No optimizations yet, but get the API up.

- An NGO partner had asked me for this feature
- I felt the need to use this when creatng a bank with a large number of objects.

Because CI started failing, I added a second [commit](https://github.com/facebook/ThreatExchange/pull/961/commits/1884fec1c625249343daf848a917720dc2698a81) upgrading black and click to this PR. If reviewer wants, can be moved to its own PR. 

Test Plan
---
Used this to successfully create a bank with million + records. Usual python suspects (mypy, black, py.test)

### DIFF
--- a/hasher-matcher-actioner/hmalib/banks/bank_operations.py
+++ b/hasher-matcher-actioner/hmalib/banks/bank_operations.py
@@ -115,3 +115,40 @@ def add_detached_bank_member_signal(
         signal_type=signal_type,
         signal_value=signal_value,
     )
+
+
+"""
+Represents a single detached signal being added to a bank. Container
+ContentType, SignalType and the signal value.
+"""
+
+
+class Signal(t.NamedTuple):
+    content_type: t.Type[ContentType]
+    signal_type: t.Type[SignalType]
+    signal_value: str
+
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/banks/bank_operations.py
```diff
@@ -115,3 +115,40 @@ def add_detached_bank_member_signal(
...
+    signal_value: str
+
+
+def add_detached_bank_member_signal_batch(
+    banks_table: BanksTable,
+    bank_id: str,
+    signals: t.Iterable[Signal],
+) -> t.Iterable[BankMemberSignal]:
+    """
+    Dump a large number of detached signals into a bank. Check
```
BarrettOlson, line 138: nit: ~~large~~ could be any number yeah?

BarrettOlson (approved):
LGTM.

PR #962, opened 2022-03-28
During index building, measure time to fetch data from banks.

schatten:
Summary
---
The indexing process has multiple stages. One of the newer ones is getting member signals from banks.

So far, it was not getting measured. Now that we are building 1M+ banks, this becomes a significant part of the indexing duration.

Add a metric name and measure it during index building.

Test Plan
---
Deployed to my personal hma instance.

Attached screenshot of cloudwatch metrics. Notice `get_bank_data` in the graph.
Screen Shot 2022-03-28 at 15 32 29

During the course of 1 day, as I added more members to the bank, the index building time increased specifically because the `get_bank_data` time increased.

### DIFF
--- a/hasher-matcher-actioner/hmalib/banks/bank_operations.py
+++ b/hasher-matcher-actioner/hmalib/banks/bank_operations.py
@@ -135,7 +135,7 @@ def add_detached_bank_member_signal_batch(
     signals: t.Iterable[Signal],
 ) -> t.Iterable[BankMemberSignal]:
     """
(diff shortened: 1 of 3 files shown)


schatten:
Putting in draft until black / click [fixes](https://github.com/facebook/ThreatExchange/pull/961/commits/1884fec1c625249343daf848a917720dc2698a81) from #961 land and are rebased against.

BarrettOlson (approved):
LGTM

PR #969, opened 2022-04-01
Parallel processing - improved

dxdc:
Summary
---------

This is an improvement to https://github.com/facebook/ThreatExchange/pull/959.

#### Bug fix for faiss

- Minor bug fix: 0bed18fd86869b349689bb30aa6b1c375ff95822

#### Simplified/optimized use of omp parallel

- Eliminates the OOM faults from the existing version, and is generally a lot simpler. It seems to be very slightly faster as well.

#### Unordered maps

- Switch to unordered maps. Since we are running this in parallel, we can achieve faster lookups, and there is no advantage to use ordered maps since we are sorting the final output anyway.

#### Optimized for loops

- Rather than iterating through the entire loop twice, and checking whether or not it's a degenerate pair within the inner loop, this optimization allows for just a single run through to produce all of the unique pairs.

```cpp

    for (unsigned int i = 0; i < filenames.size() - 1; i++) {
      for (unsigned int j = i + 1; j < filenames.size(); j++) {
        // comparison here
      }
    }
```

- The iteration through the TMK arrays was modified to accommodate a more facile parallelization. Essentially, we build a list of filenames like this first. This also allows us to prefill the adjacencyMatrix diagonal and avoid an unneeded extra computation.

```cpp
  std::vector<std::string> filenames;
  filenames.reserve(metadataToFeatures.size());

  for (const auto &s : metadataToFeatures) {
    // prefill adjacencyMatrix diagonal
    adjacencyMatrix[s.first].insert(s.first);
    filenames.push_back(s.first);
  }
  ```

#### Documentation

- Updated README to add notes about how to build the parallel versions

Test Plan
---------

### DIFF
--- a/tmk/README.md
+++ b/tmk/README.md
@@ -77,12 +77,30 @@ Notes:
 ## Windows users
 
 Some basic modifications have been made to enable support for Windows. To compile, you will also need to install `make`, e.g., via [choco](https://community.chocolatey.org/packages/make) and [MinGW](https://osdn.net/projects/mingw/).
-  - MinGW needs to be added to the system path in order to compile and use the executables
+
+- MinGW needs to be added to the system path in order to compile and use the executables
 
 Known issues:
-  - The tests will not run after `make` without more changes, but they can be run manually
-  - Avoid BOM or CRLF line endings (e.g., if `haystack.txt` files are supplied)
-  - Use of absolute file paths may need more work in some cases. Try relative paths, or copying executables (like `ffmpeg.exe`) to your local directory.
+
+- The tests will not run after `make` without more changes, but they can be run manually
+- Avoid BOM or CRLF line endings (e.g., if `haystack.txt` files are supplied)
+- Use of absolute file paths may need more work in some cases. Try relative paths, or copying executables (like `ffmpeg.exe`) to your local directory.
+
+## Performance and parallelization
+
+Parallelized variants of some of the binaries can be compiled using `make parallel`.
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by Dcallies — APPROVED** [review]
Hmm, this is definitely an argument for more unit testing in the longer term. I'm worried about our team's continued knowledge gap and the ability to verify the outputs.

File: tmk/README.md [hunk 1]
```diff
@@ -77,12 +77,30 @@ Notes:
...
 Some basic modifications have been made to enable support for Windows. To compile, you will also need to install `make`, e.g., via [choco](https://community.chocolatey.org/packages/make) and [MinGW](https://osdn.net/projects/mingw/).
-  - MinGW needs to be added to the system path in order to compile and use the executables
+
+- MinGW needs to be added to the system path in order to compile and use the executables
+
+Known issues:
+
+- The tests will not run after `make` without more changes, but they can be run manually
+- Avoid BOM or CRLF line endings (e.g., if `haystack.txt` files are supplied)
+- Use of absolute file paths may need more work in some cases. Try relative paths, or copying executables (like `ffmpeg.exe`) to your local directory.
```
Dcallies, line 87: Can you create issues for these?

dxdc: Sure. I combined them into a single Windows-based issue: https://github.com/facebook/ThreatExchange/issues/971

dxdc:
> Hmm, this is definitely an argument for more unit testing in the longer term. I'm worried about our team's continued knowledge gap and the ability to verify the outputs.

Yeah. I have been manually verifying the outputs myself against the standard set of TMK files.

PR #970, opened 2022-04-01
Automate local copy of py-tx when a directory found

schatten:
Summary
---
While we have had a readme entry and some instructions on how to do this. Automating this is a QoL improvement that I have longed for.

The new script exposes three functions. Swithing to a local copy, switching back to a pypi copy and verifying that we are on the pypi copy.

The test plan requires you to know that in the pypi version, a module called `threatexchange.content_type.meta` exits. While in the current main version, it does not.

Test Plan
---
```sh
$ ./scripts/set_threatexchange_source local
<local-threatexchange directory created>
$ python -c "from threatexchange.content_type import meta"
  File "<string>", line 1, in <module>
ImportError: cannot import name 'meta' from 'threatexchange.content_type' (/workspaces/ThreatExchange/hasher-matcher-actioner/local-threatexchange/threatexchange/content_type/__init__.py)
$

$ ./scripts/set_threatexchange_source pypi
<local-threatexchange directory deleted>
$ python -c "from threatexchange.content_type import meta"
$ <No error>
```

Test Plan for Docker
---
```sh
$ ./scripts/set_threatexchange_source pypi 
$ make upload_docker
$ docker run hma-lambda-dev <does-not-matter>
$ docker exec -it vibrant_mclean /bin/bash 
bash-4.2# python -c "from threatexchange.content_type import meta"
bash-4.2#
```

```sh
$ ./scripts/set_threatexchange_source local
$ make upload_docker
$ docker run hma-lambda-dev <does-not-matter>
$ docker exec -it hopeful_satoshi /bin/bash    
bash-4.2# python -c "from threatexchange.content_type import meta"
  File "<string>", line 1, in <module>
ImportError: cannot import name 'meta' from 'threatexchange.content_type' (/var/task/threatexchange/content_type/__init__.py)
bash-4.2#
```


hasher-matcher-actioner/scripts/set_threatexchange_source
```diff
@@ -0,0 +1,58 @@
...
+        fi
+    done
+
+    exit 0
+elif [ $1 == "local" ]; then
+    sed -e 's/"threatexchange/#\"threatexchange/' -i setup.py # remove from setup.py
+    pip uninstall -y threatexchange                           # remove from current site-packages
+
+    # Copy and install from source
+    cp -r ../python-threatexchange/ ./local-threatexchange
```
BarrettOlson, line 47: Ideally we could live edit `/python-threatexchange/` without having to run this command again.
Would symlinks work/allow for this?

hasher-matcher-actioner/hmalib/tests/test_threatexchange_from_sitepackages.py
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import unittest
+import site
+import os
+import pathlib
+
+
+@unittest.skipIf(
+    "IGNORE_PY_TX_LOCAL" in os.environ,
```
Dcallies, line 10: Hmm, unittest skips make me nervous, and gating behavior by environment variables seems like an especially messy way in the kong term.

Dcallies (commented):
Super duper dumb question - is it possible to build this into the docker image as its own layer or composition? Rather than making the docker file dirty and thus in danger of being checked in dirty, that it's an ephemeral choice at deployment time?

I'm very spooked about sed-ing files in the directory and having random changes in danger of being checked in. The test you've added seemed like an indirect measure if a change has occurred. 

Rather than a solution that takes a test to verify this class of errors has been made, is there a solution that makes mistakes impossible?

hasher-matcher-actioner/scripts/set_threatexchange_source
schatten: Ideally, yes. I do not think Docker context allows following symlinks though.

https://stackoverflow.com/questions/31881904/docker-follow-symlink-outside-context

hasher-matcher-actioner/hmalib/tests/test_threatexchange_from_sitepackages.py
schatten: Ouch. This file should not exist. Have replaced with a shell script test.

schatten: Will fix.

schatten:
> Super duper dumb question - is it possible to build this into the docker image as its own layer or composition? Rather than making the docker file dirty and thus in danger of being checked in dirty, that it's an ephemeral choice at deployment time?

Far from a dumb question. 🥇 

Each `RUN` line in a dockerfile actually ends up creating a new layer, but that does not answer your question.

It may be possible to do a [conditional RUN](https://stackoverflow.com/questions/43654656/dockerfile-if-else-condition-with-external-arguments) in dockerfile. I'll investigate.

> I'm very spooked about sed-ing files in the directory and having random changes in danger of being checked in. The test you've added seemed like an indirect measure if a change has occurred.

That's fair. I'll plan changes to this diff.

> Rather than a solution that takes a test to verify this class of errors has been made, is there a solution that makes mistakes impossible?

Agreed. I'll spend some more time figuring out a clean / non-dirty way of doing this.

hasher-matcher-actioner/scripts/set_threatexchange_source
BarrettOlson: thanks for checking

hasher-matcher-actioner/scripts/set_threatexchange_source
```diff
@@ -0,0 +1,44 @@
...
+    rm -r $LOCAL_LIB_LOCATION 2>/dev/null
+    pip install . # reinstall from pypi
+    exit 0
+elif [ $1 == "local" ]; then
+    pip uninstall -y threatexchange # remove from current site-packages
+
+    # Copy and install from source. Copying is necessary. A link would not work.
+    # Dockerbuilds can't follow symlinks outside their context directory.
+    rm -r $LOCAL_LIB_LOCATION 2>/dev/null
+    cp -r ../python-threatexchange/ $LOCAL_LIB_LOCATION
```
Dcallies, line 36: Symlink doesn't work on windows, but is it an option?

hasher-matcher-actioner/scripts/set_threatexchange_source
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
Dcallies, line 1: ignorable: I went through a big bash period in my youth, and eventually regretted writing so much scripting, and wished I had written more python instead to do the same things. Bash has all sorts of exciting difficult-to-debug failure modes (if you don't `set -e` at the beginning, a single line may fail to execute, but it still executes the rest of the script).

BarrettOlson (approved):

Dcallies (approved):
I want to feel included!

hasher-matcher-actioner/Dockerfile
```diff
@@ -8,16 +8,13 @@ RUN yum install -y gcc gcc-c++
 COPY setup.py settings.p[y] ./
 RUN pip install -e .
 
+# Copy local-threatexchange if it exists and install threatexchange from it in
+# editable mode. This will override a copy of threatexchange fetched from pypi.
+COPY local-threatexchang[e] ./
```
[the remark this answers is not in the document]
schatten: That's correct. Picking a pattern @BarrettOlson started in line 8.

hasher-matcher-actioner/scripts/set_threatexchange_source
schatten: It is "not" an option. Symlinks outside of docker context do not work.

When you do a `docker build`, you'll notice docker sends the entire context (the working directory) to the docker daemon. Symlinks are not traversed.

hasher-matcher-actioner/scripts/set_threatexchange_source
schatten: Nice catch. Adding now.

PR #972, opened 2022-04-02
[pytx] Incremental progress fixing up fb_threat_exchange

Dcallies:
Summary
---------

This is the result of a few hours of work trying to fix up fb_threatexchange.

Features:
1. HasFbThreatExchangeIndicatorType allows signal types to record their fb type
2. (Maybe dangerous/should be removed) Will fall back to trying the signal name otherwise

Misc fixes:
1. Fix cli storage not using all correct translations by switching to the helper
4. Fix errors during serialization leaving partial files by writing to a tmpfile first
5. Fix display from fetch always showing "since the beginning of time"

Test Plan
---------

```
$ threatexchange config api fb_threat_exchange -I 303636684709969

$ threatexchange fetch
[fb_threat_exchange] HMA Test Collaboration (Superset) - Up to date, at 2021-12-19T04:09:52
Rebuilding match indices...
Building index for raw_text with 1 signals...
Index for raw_text ready
Building index for pdq with 138 signals...
Index for pdq ready
Building index for url with 3 signals...
Index for url ready
```

Sadly, the data doesn't actually match, but that's a future diff.

### DIFF
--- a/python-threatexchange/threatexchange/fetcher/apis/fb_threatexchange_signal.py
+++ b/python-threatexchange/threatexchange/fetcher/apis/fb_threatexchange_signal.py
@@ -1,0 +1,33 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Helper to allow you to hint at the ThreatExchange type string on SignalTypes
+"""
+
+import typing as t
+
+
+class HasFbThreatExchangeIndicatorType:
+    """
+    A mixin to hint to fb_threatexchange_api how to handle this SignalType
+
+    For example, PDQSignalType, aka "pdq" in ThreatExchange is "HASH_PDQ"
+
+    We could hardcode all of these into the fetch() method, but that doesn't
+    play nicely with expansions. Our solution:
+
+    ```
+    class PDQSignalType(SignalType, HasFbThreatExchangeIndicatorType):
+        INDICATOR_TYPE = "HASH_PDQ"
+        ...
+    ```
+    """
+
+    INDICATOR_TYPE: t.ClassVar[t.Union[str, t.Tuple[str, ...]]] = ()
+
+    @classmethod
+    def facebook_threatexchange_indicator_applies(cls, indicator_type: str) -> bool:
+        types = cls.INDICATOR_TYPE
+        if isinstance(cls.INDICATOR_TYPE, str):
+            types = (cls.INDICATOR_TYPE,)
+        return indicator_type in types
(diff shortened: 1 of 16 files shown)


python-threatexchange/threatexchange/fetcher/apis/fb_threatexchange_signal.py
```diff
@@ -0,0 +1,33 @@
...
+    play nicely with expansions. Our solution:
+
+    ```
+    class PDQSignalType(SignalType, HasFbThreatExchangeIndicatorType):
+        INDICATOR_TYPE = "HASH_PDQ"
+        ...
+    ```
+    """
+
+    INDICATOR_TYPE: t.ClassVar[t.Union[str, t.Tuple[str, ...]]] = ()
```
schatten, line 26: Can you explain when INDICATOR_TYPE could be a tuple?

schatten (approved):
If the 'supported_signal_types' were to change, would I need to clear the entire downloaded dataset and start from scratch?

Assume the following timeline
t1 : A partner is using python-threatexchange to pull signals via threat_updates
t2: New signal_type is made available on the threatexchange graph-api
t3: Other partners start using the new signal_type
t4: The partner installs the extension that enables the new signal_type for them.

Would the first call to threat_updates with the new signal_type pull all descriptors of that type?

python-threatexchange/threatexchange/fetcher/apis/fb_threatexchange_signal.py
Dcallies: URI, RAW_URI => URLSignalType

Dcallies:
> If the 'supported_signal_types' were to change, would I need to clear the entire downloaded dataset and start from scratch?

The intent is for the answer to be no if dropping types, maybe yes if adding type (but simpler to be no). A more dangerous change is if signal types are renamed. 

> Assume the following timeline
> t1 : A partner is using python-threatexchange to pull signals via threat_updates
> t2: New signal_type is made available on the threatexchange graph-api
> t3: Other partners start using the new signal_type
> t4: The partner installs the extension that enables the new signal_type for them.
> 
> Would the first call to threat_updates with the new signal_type pull all descriptors of that type?

I can think of implementations where the answer could be either, which comes down to whether to store everything added to the collaboration or not, regardless of whether we know how to use it. We have TMK as example of a signal type where there is big benefits to excluding, so I'm tending towards the answer being "no" by default. However, we also have cases where it can be useful to download data, even if you don't plan on using it (signals that could be used manually if not by threatexchange/HMA).

tl;dr: Haven't decided yet.

PR #973, opened 2022-04-03
[pytx] fix subtle bug with match

Dcallies:
Summary
---------

Lambdas can capture variables in unexpected ways. In this case, the "index" variable was being overridden each loop, being left with the last value. A trick you can use to force it to capture the right value is to set it as a default value to an argument, which is evaluated at declaration time (whereas captured variables are evaluated at runtime).

Also minor renames, of the index accessor, which I plan to follow up with others as well:
```
settings.index  # Indices
settings.collabs  # collab configs
settings.fetchers  # apis
...
```

Test Plan
---------

I had a collaboration that had URLs, but not URL MD5s, and the md5s were evaluated second, which would be None. This generated a None attribute error without the fix.

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -167,7 +167,7 @@ class CLISettings:
         self._state = cli_state
         self._sample_message_printed = False
         self._config: t.Optional[CLiConfig] = None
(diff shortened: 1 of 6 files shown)


schatten (commented):
Neat!

schatten (approved):

PR #977, opened 2022-04-05
[pytx] Fix text matching, add a stupid framework for testing signal types

Dcallies:
Summary
---------

I noticed this bug. Instead of fixing it like a normal person, I also wasted several hours making you a testing framework of questionable usefulness. Please enjoy responsibly, if at all. 

Test Plan
---------

mypy and py.test

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/tests/signal_type_test_helper.py
+++ b/python-threatexchange/threatexchange/signal_type/tests/signal_type_test_helper.py
@@ -1,0 +1,121 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import pytest
+import typing as t
+
+from threatexchange.signal_type.signal_base import (
+    MatchesStr,
+    TextHasher,
+    SignalType,
+)
+
+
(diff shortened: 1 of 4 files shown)


python-threatexchange/threatexchange/signal_type/tests/signal_type_test_helper.py
```diff
@@ -0,0 +1,115 @@
...
+    SignalType,
+)
+
+
+class SignalTypeHashTest:
+    """
+    Helpers to help you quickly test common functionality
+
+    Most functions are designed to be used to be used in a provider fasion
+    to rapidly make simple tests.
```
BarrettOlson, line 18: nit: add a little more context on how to use and/or point read to `test_raw_text.py` as an example.

BarrettOlson (approved):
yay tests

python-threatexchange/threatexchange/signal_type/tests/signal_type_test_helper.py
Dcallies: I have something stacked on top of this, will add it there.

PR #981, opened 2022-04-06
[hma][lcc] Add scheduled custodian to reduce number of files per timebucket in durable EFS

Samyakk123:
Summary
---------

image

Created new custodian lambda with scheduled events

Test Plan
---------

Added a new unittest file test_timebucket.py to ensure working across various time intervals and varying timedeltas
Checked using mypy to resolve any warnings
Uses the freezegun library to run test cases through different time intervals
Run pytest and mypy to see results

```
$ python -m py.test
$ mypy TimeBucketizer.py
```

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
+++ b/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
@@ -1,13 +1,14 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
-from dataclasses import dataclass
 import datetime
 import os
 import random
 import string
 import tempfile
 import unittest
+from dataclasses import dataclass
 
 from freezegun import freeze_time
+
 from hmalib.common.models.pipeline import HashRecord
 from hmalib.common.timebucketizer import CSViable, TimeBucketizer
 
--- a/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
+++ b/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
@@ -1,10 +1,14 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 from dataclasses import dataclass
 import datetime
+import os
+import random
+import string
 import tempfile
 import unittest
 
 from freezegun import freeze_time
+from hmalib.common.models.pipeline import HashRecord
 from hmalib.common.timebucketizer import CSViable, TimeBucketizer
 
 
@@ -21,7 +25,7 @@ class SampleCSViableClass(CSViable):
     def to_csv(self):
         return [self.a, self.b]
 
-    def from_csv(self, value):
+    def from_csv(self):
         return SampleCSViableClass()
 
 
@@ -152,3 +156,39 @@ class TestTimeBuckets(unittest.TestCase):
                 [],
                 "Destroy method should not have executed as the buffer is empty",
             )
+
+    @freeze_time("2012-08-13 14:04:00")
+    def test_squash_content(self):
+        with tempfile.TemporaryDirectory() as td:
+            with freeze_time(datetime.datetime.now()) as frozen_datetime:
+
+                VALUE_1 = 5
+                VALUE_2 = 10
+                VALUE_3 = 3
+                for i in range(VALUE_1):
+                    for i in range(VALUE_2):
+                        sample = TimeBucketizer(
+                            datetime.timedelta(minutes=1), td, "hasher", str(i)
+                        )
+                        for i in range(VALUE_3):
+                            content = "".join(
+                                random.choice(string.ascii_lowercase) for _ in range(10)
+                            )
+                            sample.add_record(HashRecord(content, str(i)))
+                        sample.force_flush()
+                    frozen_datetime.tick(delta=datetime.timedelta(minutes=1))
+
+                TimeBucketizer.squash_content(
+                    datetime.datetime.now(), "hasher", td, datetime.timedelta(minutes=1)
+                )
+
+                records = TimeBucketizer.get_records(
+                    datetime.datetime.now() - datetime.timedelta(minutes=7),
+                    datetime.datetime.now(),
+                    "hasher",
+                    td,
+                    datetime.timedelta(minutes=1),
+                    HashRecord,
+                )
+
+                self.assertEqual(len(records), VALUE_1 * VALUE_2 * VALUE_3)
--- a/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
+++ b/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
@@ -163,7 +163,6 @@ class TestTimeBuckets(unittest.TestCase):
     def test_squash_content(self):
         with tempfile.TemporaryDirectory() as td:
             with freeze_time(datetime.datetime.now()) as frozen_datetime:
-
                 VALUE_1 = 5
                 VALUE_2 = 10
                 VALUE_3 = 3
@@ -186,12 +185,13 @@ class TestTimeBuckets(unittest.TestCase):
                 frozen_datetime.tick(delta=datetime.timedelta(minutes=1))
                 frozen_datetime.tick(delta=datetime.timedelta(minutes=1))
                 frozen_datetime.tick(delta=datetime.timedelta(minutes=1))
+
                 TimeBucketizer.squash_content(
                     "hasher",
                     td,
                     datetime.timedelta(minutes=1),
-                    datetime.datetime.now() - datetime.timedelta(minutes=2),
                     datetime.datetime.now() - datetime.timedelta(days=1),
+                    datetime.datetime.now() - datetime.timedelta(minutes=2),
                 )
 
                 records = TimeBucketizer.get_records(
@@ -202,6 +202,21 @@ class TestTimeBuckets(unittest.TestCase):
                     datetime.timedelta(minutes=1),
                     HashRecord,
                 )
-                print(records)
+
+                now = datetime.datetime(2012, 8, 13, 14, 4, 0)
+                file_count = 0
+                for i in range(VALUE_1):
+                    directory_path = os.path.join(
+                        td,
+                        "hasher",
+                        str(now.year),
+                        str(now.month),
+                        str(now.day),
+                        str(now.hour),
+                        str((now + datetime.timedelta(minutes=i)).minute),
+                    )
+                    file_count += len(os.listdir(directory_path))
+
                 self.assertEqual(len(records), VALUE_1 * VALUE_2 * VALUE_3)
                 self.assertCountEqual(records, expected_records)
+                self.assertEqual(file_count, VALUE_1)
--- a/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
+++ b/hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
@@ -2,6 +2,7 @@
 import datetime
 import os
 import random
+import typing as t
 import string
 import tempfile
 import unittest
@@ -26,7 +27,8 @@ class SampleCSViableClass(CSViable):
     def to_csv(self):
         return [self.a, self.b]
 
-    def from_csv(self):
+    @classmethod
+    def from_csv(cls, value: t.List[str]):
         return SampleCSViableClass()
 
 
@@ -166,6 +168,7 @@ class TestTimeBuckets(unittest.TestCase):
                 VALUE_1 = 5
                 VALUE_2 = 10
                 VALUE_3 = 3
+                expected_records = []
                 for i in range(VALUE_1):
                     for i in range(VALUE_2):
                         sample = TimeBucketizer(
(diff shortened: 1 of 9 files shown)


hasher-matcher-actioner/hmalib/common/tests/test_timebucket.py
```diff
@@ -152,3 +156,39 @@ def test_destroy_empty_buffer(self):
...
+                records = TimeBucketizer.get_records(
+                    datetime.datetime.now() - datetime.timedelta(minutes=7),
+                    datetime.datetime.now(),
+                    "hasher",
+                    td,
+                    datetime.timedelta(minutes=1),
+                    HashRecord,
+                )
+
+                self.assertEqual(len(records), VALUE_1 * VALUE_2 * VALUE_3)
```
schatten, line 194: We had discussed offline that you can test for equality of the records and not just their counts.

Try `self.assertEqual(records, expected_records)` where expected_records is a list you append to along `sample.add_record()`.

To allow for the lists to be out of order, you can use the [`TestCase.assertCountEqual`](https://docs.python.org/3.2/library/unittest.html#unittest.TestCase.assertCountEqual) function.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +178,53 @@ def get_records(
...
+
+    @staticmethod
+    def squash_content(
+        location: datetime.datetime,
+        type: str,
+        storage_path: str,
+        bucket_width: datetime.timedelta,
+        date_to: datetime.timedelta = datetime.timedelta(days=1),
+    ):
+
```
schatten, line 193: nit: Doc string, please. What does this method do, and why.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +178,53 @@ def get_records(
...
+        if bucket_width > date_to:
+            return
```
schatten, line 195: Good catch!

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +178,53 @@ def get_records(
...
-                content_list.extend(
-                    list(map(type_class().from_csv, csv.reader(my_file)))
-                )
+                content_list.extend(list(map(type_class.from_csv, csv.reader(my_file))))
 
         return content_list
+
+    @staticmethod
+    def squash_content(
+        location: datetime.datetime,
```
schatten, line 187: What is location? Do not remember discussing this. Can you add an explanation to the docstring?

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +178,53 @@ def get_records(
...
+                    os.path.join(directory_path, "squash" + str(uuid.uuid1())) + ".csv",
+                    "w",
+                ) as new_file:
+                    for file in file_list:
+                        with open(file, "r") as reader:
+                            new_file.write(reader.read())
+
+                        os.remove(file)
+
+            until_nearest -= bucket_width
```
schatten, line 230: I would suggest you break down the problem into two parts: a) Identify the buckets to squash b) actually squash.

A problem with the current approach is if the loop gets terminated prematurely because of an error or a timeout. It is impossible to know what buckets could not be squashed. 

eg. time increases as we go down: 
1. squashed_in_an_earlier_run
2. squashed_in_an_earlier_run 
3. squashed_in_an_earlier_run,
4. errored 
5. errored
6. squashed_in_current_run 
7. squashed_in_current_run
8. too_new_to_squash.

Instead, run your 'identify loop', build a list of all buckets to process, and then process them chronologically. So, even in case of an error, you end up with.

1. squashed_in_an_earlier_run
2. squashed_in_an_earlier_run 
3. squashed_in_an_earlier_run,
4. squashed_in_current_run 
5. squashed_in_current_run
6. errored 
7. errored
8. too_new_to_squash.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
schatten: My memory (and I may be making stuff up here) was that we keep going back in time until we find a squashed bucket. We wanted to go back at most 1 day so that we don't keep looping infinitely.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +178,53 @@ def get_records(
...
+
+                with open(
+                    os.path.join(directory_path, "squash" + str(uuid.uuid1())) + ".csv",
+                    "w",
+                ) as new_file:
+                    for file in file_list:
+                        with open(file, "r") as reader:
+                            new_file.write(reader.read())
+
+                        os.remove(file)
```
schatten, line 228: Good candidate for an internal private style method.

There are two parts to this algorithm (identify, execute). If you pull out "execute" into a different method, in the future, we could define other "identify" algorithms while relying on the robust "execute" method you wrote.

hasher-matcher-actioner/hmalib/lambdas/custodian.py
```diff
@@ -0,0 +1,21 @@
...
+BUCKET_WIDTH = datetime.timedelta(minutes=10)
+
+
+def lambda_handler(event, context):
+    """
+    Squash the records of multiple timebucketizers into a single file
+    """
+
+    TimeBucketizer.squash_content(
+        datetime.datetime.now(), "hasher", "/tmp/makethisdirectory/", BUCKET_WIDTH
```
schatten, line 20: Make the path argument ("/tmp/..." as an environment variable.)

eg. at line 12: `DURABLE_FS_PATH = os.getenv("DURABLE_FS_PATH", None)`

Only call squash_content if it is defined. If None, log and move on.

schatten: Extract "hasher" to a constant too and make it a list we'd iterate over. There may be other types in the future.

hasher-matcher-actioner/terraform/custodian/main.tf
```diff
@@ -0,0 +1,129 @@
...
+resource "aws_lambda_function" "custodian" {
+  function_name = "${var.prefix}_custodian"
+  package_type  = "Image"
+  role          = aws_iam_role.custodian.arn
+  image_uri     = var.lambda_docker_info.uri
+  image_config {
+    command = ["hmalib.lambdas.custodian.lambda_handler"]
+  }
+
+  # Timeout is kept less than the fetch frequency. Right now, fetch frequency is
```
schatten, line 25: Copy pasted comment. Delete or update.

hasher-matcher-actioner/terraform/custodian/main.tf
```diff
@@ -0,0 +1,129 @@
...
+  timeout = 60 * 12
+
+  memory_size = 128
+}
+
+resource "aws_cloudwatch_event_target" "custodian" {
+  arn  = aws_lambda_function.custodian.arn
+  rule = aws_cloudwatch_event_rule.recurring_fetch.name
+}
+resource "aws_cloudwatch_event_rule" "recurring_fetch" {
```
schatten, line 38: Search for all mentions of "fetch" and update them.

hasher-matcher-actioner/terraform/custodian/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+
+variable "additional_tags" {
+  description = "Additional resource tags"
+  type        = map(string)
+}
+
+variable "custodian_frequency" {
+  description = "How long to wait between calls to ThreatExchange. Must be an AWS Rate Expression. See here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
```
schatten, line 20: Update docs.

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -226,6 +226,16 @@ module "durable_fs" {
   additional_tags = var.additional_tags
 }
 
+module "custodian" {
```
schatten, line 229: This is a very specific custodian. Let's use "lcc_custodian" while in terraform/main.tf.

hasher-matcher-actioner/terraform/main.tf
```diff
@@ -226,6 +226,16 @@ module "durable_fs" {
...
+  source              = "./custodian"
+  prefix              = var.prefix
+  additional_tags     = merge(var.additional_tags, local.common_tags)
+  custodian_frequency = var.custodian_frequency
```
schatten, line 233: Same. Use lcc_custodian.

hasher-matcher-actioner/terraform/variables.tf
```diff
@@ -65,6 +65,13 @@ variable "indexer_frequency" {
   default     = "15 minutes"
 }
 
+variable "custodian_frequency" {
```
schatten, line 68: See main.tf. Use Same. Use lcc_custodian...

hasher-matcher-actioner/hmalib/common/timebucketizer.py
Samyakk123: Oh I thought our plan was to take in a location value, and we squash all records from that point till 1 day before / last squash. I thought this way if they wanted it for now it'd be easy to just pass datetime.now()

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +176,55 @@ def get_records(
...
-                content_list.extend(
-                    list(map(type_class().from_csv, csv.reader(my_file)))
-                )
+                content_list.extend(list(map(type_class.from_csv, csv.reader(my_file))))
 
         return content_list
+
+    @staticmethod
+    def squash_directory(directory_path):
+
```
schatten: Add docstring.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +176,55 @@ def get_records(
...
+                    new_file.write(reader.read())
+                os.remove(file)
+
+    @staticmethod
+    def squash_content(
+        type: str,
+        storage_path: str,
+        bucket_width: datetime.timedelta,
+        max_bucket_start_time: datetime.datetime,
+        # Add logging
```
schatten: duh!

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +176,61 @@ def get_records(
...
+        with open(
+            os.path.join(directory_path, "squash_" + str(uuid.uuid1())) + ".csv",
+            "w",
+        ) as new_file:
+            for file in file_list:
+                logger.info("getting content of: %s", file)
+                with open(file, "r") as reader:
+                    new_file.write(reader.read())
+                logger.info("removing file: %s", file)
+                os.remove(file)
```
schatten, line 203: Removing the source file before its content has been written, flushed and the destination file closed can be problematic.

Can you write all the content to the destination and then delete all the files?

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +176,61 @@ def get_records(
...
+        logger.info("successfully squashed: %s", directory_path)
+
+    @staticmethod
+    def squash_content(
+        type: str,
+        storage_path: str,
+        bucket_width: datetime.timedelta,
+        max_bucket_start_time: datetime.datetime,
```
schatten, line 211: nit: I do not expect to pass a max value BEFORE a min value. Natural ordering suggests min and then max.

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +176,61 @@ def get_records(
...
+        until_nearest = TimeBucketizer._calculate_bucket_endpoints(
+            max_bucket_start_time, bucket_width
+        )[1]
+        since_nearest = TimeBucketizer._calculate_bucket_endpoints(
+            min_bucket_start_time, bucket_width
+        )[0]
+
+        if until_nearest > datetime.datetime.now() - bucket_width:
+            logger.info(
+                "Start time indicates nearest bucket is not ready for squashing"
```
schatten, line 223: max_start_time indicates ....

hasher-matcher-actioner/hmalib/common/timebucketizer.py
```diff
@@ -174,8 +176,61 @@ def get_records(
...
+            )
+            return
+
+        while until_nearest <= since_nearest:
```
schatten, line 227: This is weird. Why is until < since at all? 

since is always < until, right?

schatten: I'm worried we are not even entering this loop because until > since. 

This means that your test (which is passing) is not capturing all the behavior. The test is counting and comparing records, but that is not enough.

Internally, whether the number of files reduced at all is not being checked for. Your unit test should do that.

schatten (changes requested):
I'm reasonably sure the `while` iteration in squash_content is wrong.  

It is kind of guaranteed that until > since.

Which also tells me that the unit test is inadequate. Have a look at comments inline.

schatten (approved):
LGTM.

schatten (approved):

PR #987, opened 2022-04-08
[vPDQ] mv to a toplevel project

Dcallies:
Summary
---------

I think this is different enough from PDQ (a photo algorithm), and is analogous to TMK (which also uses PDQ as it's frame algorithm). I think it should be a toplevel project.

Test Plan
---------

Hope and prayer


BarrettOlson (approved):

Dcallies:
Discussed offline - @BarrettOlson is considering moving all the similarity detection projects back under /hashing, but for now since the others are still toplevel, will leave as toplevel.

PR #984, opened 2022-04-07
Add text support for hashing for python threatexchange

sahanapeters:
Summary
---------
Previously `text` was an invalid choice for the hash command, both for inline and for file hashing.

#### Before
```
$ threatexchange hash text -I "sdfsdhkfhsdj"
usage: threatexchange hash [-h] [--signal-type {pdq,video_md5,url_md5}] [--inline] {
     photo,video,url} content [content ...]
threatexchange hash: error: argument content_type: invalid choice: 'text' (choose from 'photo', 'video', 'url')
```

```
threatexchange hash text file.txt                                 
usage: threatexchange hash [-h] [--signal-type {pdq,video_md5,url_md5}] [--inline]
                           {photo,video,url} content [content ...]
threatexchange hash: error: argument content_type: invalid choice: 'text' (choose from 'photo', 'video', 'url')
```

#### After

There is a new SignalType `text_md5`. I am not sure why it didn't exist before. Was it deleted?
Anyway, adding it here in this PR. The only thing is I just copied the hashing for the file from what we do on md5 😢 .
 Not the hashing expert so if anyone has other thoughts, feel free to comment.

```
threatexchange hash text -I "helllooooooooo" 
text_md5 eb6238a4a35130945111fd77bd3bf988
```

```
threatexchange hash text file.txt  
text_md5 6074299e0bea6a7842a2148e23d914f4
```

##### Note:

Not having an existing file will now throw a FileNotFoundError

```
threatexchange hash text blahblah.txt        speters/fix_cmd_validation  ✱
Traceback (most recent call last):
  File "/Users/sahanapeters/Code/ThreatExchange/python-threatexchange/env/bin/threatexchange", line 11, in <module>
    load_entry_point('threatexchange', 'console_scripts', 'threatexchange')()
  File "/Users/sahanapeters/Code/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 218, in main
    execute_command(settings, namespace)
  File "/Users/sahanapeters/Code/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 101, in execute_command
    command.execute(settings)
  File "/Users/sahanapeters/Code/ThreatExchange/python-threatexchange/threatexchange/cli/hash_cmd.py", line 112, in execute
    hash_str = hash_fn(signal_type, inp)
  File "/Users/sahanapeters/Code/ThreatExchange/python-threatexchange/threatexchange/cli/hash_cmd.py", line 105, in <lambda>
[...]

PR #989, opened 2022-04-09
Fix validation on threatexchange hash inputs

sahanapeters:
Summary
---------

Opening this PR in favor of closing https://github.com/facebook/ThreatExchange/pull/984 for cleanliness.

What this PR does

1. Adds an error handling for file not found exception for `hash_cmd.py`
2. Makes TextHasher extend FileHasher. 
 Since `threatexchange hash text file.txt` or `threatexchange hash url file.txt` would not return hashes.
3. ~~Raise a ValueError for text_tlsh when the input text is too short.~~

##### For future reference

How to add `tlsh` extension for Threatexchange

```
$ threatexchange config extensions add threatexchange.extensions.text_tlsh
```

Test Plan
---------
`make test`

### DIFF
--- a/python-threatexchange/threatexchange/cli/hash_cmd.py
+++ b/python-threatexchange/threatexchange/cli/hash_cmd.py
@@ -106,8 +106,12 @@ class HashCommand(command_base.Command):
             if isinstance(inp, str):
                 hash_fn = lambda s, t: s.hash_from_str(t)
                 signal_types = str_hashers
-
             for signal_type in signal_types:
-                hash_str = hash_fn(signal_type, inp)
-                if hash_str:
-                    print(signal_type.get_name(), hash_str)
+                try:
+                    hash_str = hash_fn(signal_type, inp)
+                    if hash_str:
+                        print(signal_type.get_name(), hash_str)
+                except FileNotFoundError:
+                    self.stderr(
+                        f"The file {inp} doesn't exist or the file path is incorrect"
+                    )
--- a/python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
+++ b/python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
@@ -33,18 +33,21 @@ class ThreatExchangeCLIE2eTest(unittest.TestCase):
         self.addCleanup(lambda: shutil.rmtree(str(state_dir)))
         self._state_dir = state_dir
 
-    def cli_call(self, *given_args: str) -> str:
+    def cli_call_helper(self, std_op: str, *given_args: str) -> str:
         args = list(self.COMMON_CALL_ARGS)
         args.extend(given_args)
         print("Calling: $ threatexchange", " ".join(args), file=sys.stderr)
         try:
-            with patch("sys.stdout", new=StringIO()) as fake_out:
+            with patch(std_op, new=StringIO()) as fake_out:
                 main(args, state_dir=self._state_dir)
             return fake_out.getvalue()
         except SystemExit as se:
             if se.code == 0:  # ERROR: Everything is fine
                 return fake_out.getvalue()
             raise E2ETestSystemExit(se.code)
+
+    def cli_call(self, *given_args: str) -> str:
+        return self.cli_call_helper("sys.stdout", *given_args)
 
     def assert_cli_output(
         self, args: t.Iterable[str], expected_output: t.Union[str, t.Dict[int, str]]
@@ -70,3 +73,9 @@ class ThreatExchangeCLIE2eTest(unittest.TestCase):
         self.assertEqual(exception.returncode, 2)
         if msg_regex is not None:
             self.assertRegex(str(exception), msg_regex)
+
+    def assert_cli_error_output(
+        self, given_args: t.Iterable[str], expected_err: t.Union[str, t.Dict[int, str]]
+    ) -> None:
+        output = self.cli_call_helper("sys.stderr", *given_args)
+        self.assertEqual(expected_err, output.strip())
--- a/python-threatexchange/threatexchange/signal_type/signal_base.py
+++ b/python-threatexchange/threatexchange/signal_type/signal_base.py
@@ -95,7 +95,18 @@ class SignalType:
         return []
 
 
-class TextHasher:
+class FileHasher:
+    """
+    This class can hash files.
+    """
+
+    @classmethod
+    def hash_from_file(cls, file: pathlib.Path) -> str:
+        """Get a string representation of the hash from a file"""
+        raise NotImplementedError
+
+
+class TextHasher(FileHasher):
     """
     This class can turn text into intermediary representations (hashes)
     """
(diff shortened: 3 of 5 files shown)


python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -95,7 +95,21 @@ def get_examples() -> t.List[str]:
...
+    to prefer file hashing to reading the file in as a Str.
+    """
+
+    @classmethod
+    def hash_from_file(cls, file: pathlib.Path) -> str:
+        """Get a string representation of the hash from a file"""
+        raise NotImplementedError
+
+
+class TextHasher(FileHasher):
```
Dcallies, line 112: Hmm, thought this was already the case. 

Ah derp, it just implements the function without actually extending it. Yay python.

python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -95,7 +95,21 @@ def get_examples() -> t.List[str]:
         return []
 
 
-class TextHasher:
+class FileHasher:
+    """
+    This class can hash files.
+
+    If also inheriting from StrHasher, put this second in the inheritance
+    to prefer file hashing to reading the file in as a Str.
```
Dcallies, line 103: (Ignorable): I think this is referring to an old implementation - I think it's safe to remove.

python-threatexchange/threatexchange/extensions/text_tlsh/text_tlsh.py
```diff
@@ -42,7 +42,7 @@ def validate_signal_str(cls, signal_str: str) -> str:
     def hash_from_str(cls, text: str) -> str:
         hash_str = str(tlsh.hash(text.encode()))
         if hash_str == "TNULL":  # Likely too short
-            return ""
+            raise ValueError("text is too short")
```
Dcallies, line 45: (blocking): Hmm, I'm not sure about this behavior change. The contract (which is undocumented in the docstring...) previously was that if your hash doesn't apply for the input (too short, etc), you just return empty string. We check for empty string in all the places where hash is called for this reason. Otherwise we'd need to catch exceptions anyways. If behavior is exceptional, then exceptions are a good choice. However, if your API will regularly need to deal with it, Exceptions are actually much slower to process than having a truthy return, or making the return optional. 

Otherwise, imagine if you have 3 applicable SignalTypes for a piece for a content, and one of them needs a minimum length.

We'd expect this:

```
$ threatexchange hash -I 'example text'
type_a blah
type_b blah

$ 
```
tlsh which needs 100 bytes doesn't get printed

Instead what we'd get is:

```
$ threatexchange hash -I 'example text'
type_a blah
type_b blah
<Big stacktrace>
ValueError
```

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -101,13 +101,17 @@ def execute(self, settings: CLISettings) -> None:
...
+                    hash_fn = lambda s, t: s.hash_from_str(t)
+                    signal_types = str_hashers
+                for signal_type in signal_types:
+                    hash_str = hash_fn(signal_type, inp)
+                    if hash_str:
+                        print(signal_type.get_name(), hash_str)
+            except FileNotFoundError:
+                self.stderr(
+                    f"The file {inp} doesn't exist or the file path is incorrect"
+                )
```
Dcallies, line 117: (blocking): You want to make this try: catch narrower to the statements you expect to throw this exception. In this case it's L111

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -101,13 +101,17 @@ def execute(self, settings: CLISettings) -> None:
...
-                    print(signal_type.get_name(), hash_str)
+            try:
+                hash_fn = lambda s, t: s.hash_from_file(t)
+                signal_types: t.List[t.Any] = file_hashers
+                if isinstance(inp, str):
+                    hash_fn = lambda s, t: s.hash_from_str(t)
+                    signal_types = str_hashers
+                for signal_type in signal_types:
+                    hash_str = hash_fn(signal_type, inp)
+                    if hash_str:
```
Dcallies, line 112: Here's a demonstration of the old contract on empty string in action.

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -70,3 +70,13 @@ def assert_cli_usage_error(
         self.assertEqual(exception.returncode, 2)
         if msg_regex is not None:
             self.assertRegex(str(exception), msg_regex)
+
+    def assert_cli_error_output(
+        self, given_args: t.Iterable[str], expected_err: t.Union[str, t.Dict[int, str]]
+    ) -> None:
+        args = list(self.COMMON_CALL_ARGS)
+        args.extend(given_args)
+        print("Calling: $ threatexchange", " ".join(args))
```
Dcallies, line 79: (blocking): This makes a maintenance challenge, since now the args extension behavior needs to updated in multiple places if we change it. Instead use `self.cli_call()` which will handle it for you.

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -70,3 +70,13 @@ def assert_cli_usage_error(
...
+        with patch("sys.stderr", new=StringIO()) as fake_out:
+            main(args, state_dir=self._state_dir)
```
Dcallies, line 81: ... instead, just do self,cli_call here

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -70,3 +70,13 @@ def assert_cli_usage_error(
         self.assertEqual(exception.returncode, 2)
         if msg_regex is not None:
             self.assertRegex(str(exception), msg_regex)
+
+    def assert_cli_error_output(
+        self, given_args: t.Iterable[str], expected_err: t.Union[str, t.Dict[int, str]]
+    ) -> None:
```
Dcallies, line 76: ignorable: Consider making cli_usage_error a subset of this behavior - searching by error regex is slightly more flexible than searching by exact string.

Dcallies (changes requested):
Nice job! Thanks for adding tests! 

If you are planning on replacing the contents of one PR with another, especially if you are planning on just re-implementing it, it's also fine to use `git push --force` to just rewrite the history, although you are right in that it can create a lot of dangling comments.

Can you tell me more about the behavior change to throwing exceptions if the SignalType hash doesn't apply?

python-threatexchange/threatexchange/extensions/text_tlsh/text_tlsh.py
sahanapeters: I see what you're saying with the large stacktrace for API usage - that makes sense. :) 
However, I still don't think its actually good user behavior to just return an empty string because we haven't documented this anywhere. It seems buggy rather than a known and understood behavior.
Do you think it makes sense if we caught the ValueError, like we're doing for FileNotFound?

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
sahanapeters: Ahhh good idea. 
What threw me off was the hard-coded "sys.stdout" for my use case and I was averse to changing `self.cli_call()` since it is called in multiple places. My solution was to just create another method 😛 
DRYing this bit with a helper method .

BarrettOlson (approved):
LGTM, thanks!

PR #990, opened 2022-04-10
[pytx] StopNCII API

Dcallies:
Summary
---------

Create a simple API wrapper for StopNCII.org. I haven't gotten the API keys yet, so instead enjoy this test with the json copied from the example.

Test Plan
---------

Added unittest


python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+            records.append(
+                {
+                    "lastModtimestamp": now,
+                    "signalType": str(signal_type),
+                    "hashValue": hashValue,
+                    "CSPFeedbacks": [fb_dict],
+                }
+            )
+        payload = {"count": len(feedbacks), "hashRecords": records}
+        self._post("SubmitHashes", json=payload)
```
Dcallies, line 251: Whoops, wrong API, redo with the correct one (though can't test it yet)

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+        Same timeouts and retry strategy as `_get_session` above.
+        """
+
+        url = "/".join((self.BASE_URL, self.API_VERSION, endpoint))
+        with self._get_session() as session:
+            response = session.get(url, json=json)
+            response.raise_for_status()
+            return response.json()
+
+    def _post(self, endpoint: str, *, json=None, params=None) -> t.Any:
```
schatten, line 156: same: `_post_json_or_fail`?

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+            params["nextPageToken"] = next_page
+        json_val = self._get("FetchHashes", **params)
+        logging.debug("StopNCII FetchHashes returns: %s", json_val)
+        return dacite.from_dict(
+            data_class=FetchHashesResponse,
+            data=json_val,
+            config=dacite.Config(cast=[enum.Enum, set]),
+        )
+
+    def fetch_hashes_iter(
```
schatten, line 191: Question: With hashexchange APIs, there must be a translator layer between fb_threatexchange → fetcher and stopncii → fetcher. The layer will have to translate for example -- StopNCIIHashRecord to BankMemberSignal arguments.

Where do you imagine that will happen?

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+        """Submit feedback on multiple hashes"""
+        now = int(time.time())
+        # Now for the fun hacky part, convert to expected format
+        records = []
+        for (signal_type, hashValue), feedback in feedbacks.items():
+            fb_dict = asdict(feedback)
+            fb_dict["tags"] = list(fb_dict["tags"])  # Convert from set
+            fb_dict["feedbackValue"] = str(
+                fb_dict["feedbackValue"]
+            )  # Convert from enum
```
schatten, line 240: Can't this be done in an instance method on the `StopNCIICSPFeedback` class?

If the behavior were to change, tracking this usage down might be easy to miss.

schatten (approved):
Looks good. Am happy merging this as is. If you want to wait before testing against an actual test API instance, that works too. 

Have a few questions and suggestions inline. Can you address them?

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+    Unknown = "Unknown"
+    Received = "Received"  # Initial state
+    Active = "Active"  # Unclear what this represents
+    Withdrawn = "Withdrawn"  # Should be deleted by client
+    Deleted = "Deleted"  # Should be deleted by client
+
+
+@enum.unique
+class StopNCIICSPFeedbackValue(enum.Enum):
+    """"""
```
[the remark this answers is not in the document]
Dcallies: Missed one!

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+
+        If using without a context manager, ensure you end up calling close() on
+        the returned value.
+        """
+        session = requests.Session()
+        session.headers.update(self._auth_headers)
+        session.mount(
+            self.BASE_URL,
+            adapter=TimeoutHTTPAdapter(
+                timeout=60,
```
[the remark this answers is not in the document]
Dcallies: I prefer keeping inline when there is only one use.

python-threatexchange/threatexchange/stopncii/api.py
Dcallies: I prefer the shorter names, and think they would not be surprising to a future person. Should add docstring though

python-threatexchange/threatexchange/stopncii/api.py
Dcallies: Not able to parse your question - the answer is in the existing interfaces. You must implement the interfaces for fetching to work. You can put the translation layer anywhere you want as long as long as the interface returns the right value. Here's the exit of fetch/sync so it must be before/during here https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/fetcher/fetch_state.py#L267-L269

python-threatexchange/threatexchange/fb_threatexchange/api.py
```diff
@@ -128,7 +128,7 @@ def get_json_from_url(self, url, params=None, *, json_obj_hook: t.Callable = Non
         Same timeouts and retry strategy as `_get_session` above.
         """
         with self._get_session() as session:
-            response = requests.get(url, params=params or {})
+            response = requests.get(url, params=params or {})  # !!! Typo? sessions.get?
```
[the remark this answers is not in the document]
Dcallies: @schatten Why I think there's a typo: Where is the session object used?

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+                max_retries=Retry(
+                    total=4,
+                    status_forcelist=[429, 500, 502, 503, 504],
+                    allowed_methods=["HEAD", "GET", "OPTIONS"],
```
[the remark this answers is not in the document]
Dcallies: Hmm, on second though, retrying post requests I'm not sure about. Timeout yes, but not retry. Just going to leave for now.

python-threatexchange/threatexchange/stopncii/api.py
```diff
@@ -0,0 +1,251 @@
...
+                    backoff_factor=0.2,  # ~1.5 seconds of retries
+                ),
+            ),
+        )
+        return session
+
+    def _get(self, endpoint: str, **json) -> t.Any:
```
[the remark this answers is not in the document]
Dcallies: I prefer the shorter names, and find _get less ambiguously maps to GET than the suggested, which might be interpreted as "Get JSON"

PR #992, opened 2022-04-14
[breaking][hma] 1/n: Use signal_type_mapping instead of static signal / content types

### LINKED ISSUE #995: Get HMA to build with unreleased refactored threatexchange
- add a script to make switching between local an pypi hosted threatexchange versions
- get black , mypy and py.test to pass with the local version of threatexchange

schatten:
Summary: Fixes #995 
---------
1. Switch to using local threatexchange for Github workflows
2. Replace `threatexchange.api` with `threatexchange.fb_threatexchange.api`
3. Add a new class (`HMASignalTypeMapping` extends `SignalTypeMapping`), this pulls signal types from dynamodb, with fetcher and collab store coming soon.
4. Get an instance of HMASignalTypeMapping in major entrypoints and pass it down the dependency chain to model deserializers
5. Add a tiny migrations framework that uses terraform for state-maintenance and hmalib.scripts.cli.main to run the migrations.

Test Plan
---------
black, mypy and py.test pass.

NOTE: This DOES NOT make hma work with new pytx. That's still a WIP. Other work will build on top.
NOTE: I plan on landing this PR and breaking trunk.


hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
+from pyclbr import Function
```
schatten, line 3: Remove this.

hasher-matcher-actioner/hmalib/common/models/content.py
```diff
@@ -2,6 +2,7 @@
 
 import datetime
 from enum import Enum
+from multiprocessing.sharedctypes import Value
```
schatten, line 5: Remove this.

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,206 @@
...
+            )
+            create_config(signal_config)
+            signal_types.append(_type)
+
+        return (content_types, signal_types)
+
+    def get_signal_type(self, name: str) -> t.Optional[t.Type[SignalType]]:
+        return self.signal_type_by_name.get(name, None)
+
+    def get_content_type(self, name: str) -> t.Optional[t.Type[ContentType]]:
```
Dcallies, line 145: (ignorable): This method only used privately

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,206 @@
...
+        return self.content_by_name.get(name, None)
+
+    def get_signal_type_enforce(self, name: str) -> t.Type[SignalType]:
```
Dcallies, line 148: (alt, ignorable): Another convention I've seen that is less verbose is just sticking `x` at the end. 
* `get` -> nullable
* `getx` -> non-nullable

hasher-matcher-actioner/scripts/set_threatexchange_source
```diff
@@ -34,7 +34,7 @@ elif [ $1 == "local" ]; then
 
     # Copy and install from source. Copying is necessary. A link would not work.
     # Dockerbuilds can't follow symlinks outside their context directory.
-    rm -r $LOCAL_LIB_LOCATION 2>/dev/null
+    rm -r $LOCAL_LIB_LOCATION 2>/dev/null || true
```
Dcallies, line 37: Is there an rm flag you can use or gate this instead with an if? Otherwise you permission fail your removal this may behave unexpectedly.

Also feeds into my opinion of "use python for scripts if script is longer than a few lines"

hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
```diff
@@ -63,9 +64,24 @@ def from_dict(cls, d: t.Dict) -> "DictParseable":
...
+    """
+
+    @classmethod
+    def from_dict(
+        cls, d: t.Dict, signal_type_mapping: HMASignalTypeMapping
+    ) -> "DictParseableWithSignalTypeMapping":
+        raise NotImplementedError
+
+
 def jsoninator(
```
Dcallies, line 80: Lol wut is this name?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -53,6 +54,13 @@
 INDEXER_FUNCTION_NAME = os.environ["INDEXER_FUNCTION_NAME"]
 WRITEBACK_QUEUE_URL = os.environ["WRITEBACKS_QUEUE_URL"]
 
+# Initialize hmaconfig at module level. Mounted SubApps need not initialize
+# their own HMAConfigs.
+HMAConfig.initialize(HMA_CONFIG_TABLE)
+
+functionality_mapping = get_pytx_functionality_mapping()
```
Dcallies, line 61: (dumb question): This might be a significant amount of latency since it's making dynamodb calls. Are we already doing this kind of work at the outer level?

hasher-matcher-actioner/hmalib/common/tests/test_mappings.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
Dcallies, line 1: yaytests

hasher-matcher-actioner/hmalib/common/models/tests/test_pipeline_models.py
```diff
@@ -98,7 +101,7 @@ def get_example_md5_hash_record():
         # Use this somewhere
         return models.PipelineHashRecord(
             TestPDQModels.TEST_CONTENT_ID,
-            PhotoMD5Signal,
```
Dcallies, line 101: rest in pepperonis

Dcallies (commented):
Thanks for doing this! 

I think I need to meet with you to ask a series of dumb questions about HMA - it's been too long since I worked on it.

My biggest concern is that HMASignalTypeMapping is slightly more complicated than needs to be to set up. The SignalTypeMapping class is not expected by any interface, and was a focusing function for me when designing the CLI (since I knew I'd need all the capabilities). That means that you aren't constrained to it's interface if you don't like the shape of it. I think it's totally fine for HMASignalTypeMapping to not inherit from anything.

I have no idea if this is helpful to you, but here's some changes I was considering making the CLI to make it work more intuitively for commands. This isn't to say you should do the same thing, and underneath the hood, I was planning on maintaining something like signal mapping to make it all work, but it shows how you can expose a narrower band to the code to try and separate concerns

```
settings: CliConfig  # The god object

settings.collab  # Helper interface to collabs
settings.collab.get(name)
settings.collab.get_enabled()

settings.content  # Helper to content types
settings.content.get(name)
settings.content.get_all()

settings.signal  # Helper to signal types
settings.signal.get(name)
settings.signal.get_all()
settings.signal.get_for_content(content)

settings.index  # Helper to indices
settings.index.get(signal)
settings.index.get_all()

settings.fetcher  # APIs
settings.fetcher.get(name)
settings.fetcher.get_all()

settings.fetcher.state # API state
settings.fetcher.state.get_records(name)
settings.fetcher.state.get_records_by_signal(signal)
```

I also planned on putting the write commands in there as well
```
settings.config.update()
settings.index.clear_all()
```
but also potentially hiding them slightly
```
settings.write.config.update()
```

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
schatten: Since this is done before the lambda handler is called, the latency is ammortized over the lifetime of a lambda instance. During such a lifetime, it probably handles thousands (if not tens of thousands) of requests. 

That, and DynamoDB is quite performant. Individual gets are p99 10ms. I don't think scans for a single partition will be much larger. Config table is expected to be tiny.

hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
schatten: Oh, first time you are coming across this. :P 

JSON + terminator. I'm really horrible with portmanteaus.

schatten: Try speaking this name out loud. It gets funnier.

hasher-matcher-actioner/hmalib/common/mappings.py
schatten: I thought the same, but "getx python" search results always point to graphics packages. 

If you feel strongly, this is just a search-and-replace. Can easily do. LMK.

hasher-matcher-actioner/hmalib/common/mappings.py
schatten: Intentional. It is part of the public API for `HMASignalTypeMapping`.

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,206 @@
...
+
+def get_pytx_functionality_mapping() -> HMAFunctionalityMapping:
+    """
+    Call from HMA entrypoints. Ensure HMAConfig.initialize() has already been
+    called.
+    """
+    fetchers: t.List[SignalExchangeAPI] = []
+
+    threatexchange_api_token = AWSSecrets().te_api_token()
+    if not threatexchange_api_token in (None, ""):
```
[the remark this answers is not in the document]
schatten: Lol. Me trying to talk to cpython in English. :P Will fix.

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,206 @@
...
+            default_content_types, default_signal_types = self._init_configs()
+            super().__init__(default_content_types, default_signal_types)
+            return
+
+        enabled_content_types = list(
+            map(
+                lambda c: import_class(c.content_type_class),
+                filter(lambda ct: ct.enabled, all_content_types),
+            )
+        )
```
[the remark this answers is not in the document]
schatten: Yours is more readable. Copying!

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,206 @@
...
+    Enhancments:
+
+    1. State is pulled in from dynamodb when __init__ is first called.
+    2. Dedicated methods for getting signal and content_type by name. Instead of
+       depending on SignalTypeMapping's internal state.
+    """
+
+    def __init__(self):
+        all_content_types = ToggleableContentTypeConfig.get_all()
+        all_signal_types = ToggleableSignalTypeConfig.get_all()
```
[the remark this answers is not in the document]
schatten: Yeah. I see that value in that now. I've had to invent a test class to get it to work without doing "work" in the init. Let me try to address this.

.github/workflows/hma-ci.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements-dev.txt
+          bash ./scripts/set_threatexchange_source local
```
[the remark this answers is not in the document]
schatten: No. Only until we release new version of python-threatexchange into pypi.

schatten:
> settings.content  # Helper to content types
> settings.content.get(name)
> settings.content.get_all()

I thought signal_type_mapping and siblings were the narrowest part of the stack. In that they would be used to communicate between the various components. Should I consider FunctionalityMapping and its member's classes part of the internal API of threatexchange, and not use them directly?

Even if you were to re-design SignalTypeMapping (or anything), I can mypy my way out of it. But I see the value in separating out signals / contents and everything into smaller more focussed collections.

hasher-matcher-actioner/hmalib/common/mappings.py
Dcallies: I don't - in my personal reviewing convention:

blocking: I feel strongly about this. Don't land without explicitly confirming it's okay
nit: I wouldn't write it this way, and I think it's sub-optimal, or it's a suboptimal but still functionally correct
ignorable: If you didn't change it, my feelings are not hurt :P 
alt: Here's another way to write this, but your way is still correct. Sometimes I'm not sure if it's better either. Implicitly ignorable
code golf: Here's another way to write this, but I did it mostly to see if this would work. Implicitly ignorable

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,284 @@
...
+
+    @classmethod
+    def get_from_config_or_default(cls) -> "HMASignalTypeMapping":
+        """
+        Pull configs from HMAConfigs if available. If not, use defaults.
+        """
+        all_content_types = ToggleableContentTypeConfig.get_all()
+        all_signal_types = ToggleableSignalTypeConfig.get_all()
+
+        if len(all_content_types) + len(all_signal_types) == 0:
```
Dcallies, line 112: alt, ignorable: `if not (all_content_types or all_signal_types):`

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,284 @@
...
+            # We have never written content or signal types to database, so use
+            # default values instead.
+            (
+                all_content_types,
+                all_signal_types,
+            ) = cls._get_default_configs()
```
Dcallies, line 118: blocking: This means as soon as they add the first st/ct config, they may suddenly have an unexpected change in configuration because it erases the others.

I'd suggest instead this behavior:

```
all_content_types, all_signal_types = cls._get_default_configs()
by_class = {ct.content_type_class: ct for ct in all_content_types}
by_class.update(ct.content_type_class: ct for ct in ToggleableContentTypeConfig.get_all()}
all_content_types = list(by_class.values())
# Same for signal
```

Basically, implicitly, these configs exist until they are inserted into the database, in which case the database version exists.

This creates unusual behavior when you delete these records unfortunately, where disabled => deleted turns into enabled. To fix would require us to prevent deletion from the UI for these, and instead turn into disable (which I think makes sense, but want your input).

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,284 @@
...
+
+        signal_type_configs = []
+        for _type in DEFAULT_SIGNAL_AND_CONTENT_TYPES.signal_type_by_name.values():
+            signal_config = ToggleableSignalTypeConfig(
+                name=ToggleableSignalTypeConfig.get_name_from_type(_type),
+                signal_type_class=full_class_name(_type),
+            )
+            signal_type_configs.append(signal_config)
+
+        return (content_type_configs, signal_type_configs)
```
Dcallies, line 158: ignorable: Parenthesis not required here

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,284 @@
...
+
+    def write_as_configs(self):
+        """
+        Write current state to dynamodb. This will create or update HMAConfigs
+        so that subsequent calls to get_from_config_or_default() return current
+        state.
+
+        Note: this will also disable signal and content types that are not part
+        of current state.
+        """
```
Dcallies, line 168: Is the most common behavior that we write everything, or we'd write only single changes? I'm imagine a UI with a series of checkboxes

hasher-matcher-actioner/hmalib/scripts/cli/run_api.py
```diff
@@ -60,7 +60,9 @@ def execute(self, tf_outputs: t.Dict) -> None:
         for k in fn_env_vars:
             os.environ[k] = fn_env_vars[k]
 
-        from hmalib.lambdas.api.api_root import app
+        from hmalib.lambdas.api.api_root import bottle_init_once
```
Dcallies, line 63: nit: can this be at the toplevel? Code smell to have inline imports

If not, can you add comment?

hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
```diff
@@ -63,9 +64,24 @@ def from_dict(cls, d: t.Dict) -> "DictParseable":
         raise NotImplementedError
 
 
+class DictParseableWithSignalTypeMapping:
+    """
+    A special case of dict parseables where the dynamic signal type config is
+    required.
```
Dcallies, line 70: During rewriting the CLI I constantly struggled on whether the list of signal types should be stored statically or not (SignalType.get_all(), which would error if not set in init). The number of places where you want to treat it as a statically known thing is high.

hasher-matcher-actioner/hmalib/common/tests/test_mappings.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_bootstrapped_values_work(self):
+        with self.fresh_dynamodb():
+            HMAConfig.initialize(self.__class__.table.table_name)
+            first_response = HMASignalTypeMapping.get_from_config_or_default()
+            second_response = HMASignalTypeMapping.get_from_config_or_default()
+
+            # Check that the first and second responses are the same.
+            self.assertEqual(
+                first_response.content_by_name, second_response.content_by_name
+            )
```
Dcallies, line 29: Are you using pytest in HMA? I recently spent a lot of time reading through the pytest docs, and as a result understand it a lot better now. Basically unnitest.assertEqual and friends strip all the pytest helper information out of your tests, and unittest.TestCase doesn't actually get you anything that pytest can't do similarly/better (thought it may also just be functional vs object-oriented taste differences).

hasher-matcher-actioner/hmalib/common/tests/test_mappings.py
```diff
@@ -0,0 +1,62 @@
...
+
+from hmalib.common.config import HMAConfig
+from hmalib.common.models.tests.ddb_test_common import HMAConfigTestBase
+from hmalib.common.mappings import (
+    HMASignalTypeMapping,
+)
+
+
+class SignalTypeMappingInitTestCase(HMAConfigTestBase, unittest.TestCase):
+    def test_bootstrapped_values_work(self):
```
Dcallies, line 20: blocking: Can you add a test case where someone starts with the boostrapped configs, and then overrides PDQ with enabled (writing to DB), and then overrides PDQ to disabled (writing to DB) and make sure the state of the whole system is expected the whole way?

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -136,75 +210,27 @@ def process_s3_event(event: dict) -> None:
...
-    get_indexes_api(
-        indexes_bucket_name=INDEXES_BUCKET_NAME,
-        indexer_function_name=INDEXER_FUNCTION_NAME,
-    ),
-)
 
-if __name__ == "__main__":
-    app.run()
+if __name__ == "__main":
+    bottle_init_once()[0].run()
```
Dcallies, line 236: <3

ignorable: for clarity, it might be nicer to split this
```
app, _handler = bottle_init_once()
app.run()
```

hasher-matcher-actioner/hmalib/lambdas/api/api_root.py
```diff
@@ -53,68 +52,143 @@
 INDEXER_FUNCTION_NAME = os.environ["INDEXER_FUNCTION_NAME"]
 WRITEBACK_QUEUE_URL = os.environ["WRITEBACKS_QUEUE_URL"]
 
+
+@lru_cache(maxsize=1)
+def bottle_init_once() -> t.Tuple[
+    bottle.AppStack, t.Callable[[t.Dict[str, t.Any], t.Any], t.Dict[str, t.Any]]
+]:
```
Dcallies, line 59: Thanks for doing this

Dcallies (changes requested):
Okay, phwew, did make it through the entire PR, so future reviews should be faster.

Had a couple of things I thought were blocking, but if those are solved, can review fast and get you an accept.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -353,6 +353,18 @@ def update_config(config: HMAConfig) -> "HMAConfig":
...
+    """
+    Tries to update config, if found. Else creates. Will do two ddb calls.
+    """
+    try:
+        update_config(config)
+    except:
+        create_config(config)
+    finally:
+        return config
+
```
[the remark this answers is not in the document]
schatten: Thanks for the thorough expression! TIL about the flaw with finally returns.

hasher-matcher-actioner/hmalib/common/mappings.py
schatten: I believe we discussed this to death on call. The default set of records need to be handled, and is not handled at the moment. 

Solution we agreed on is close to (exactly?) what you suggested. I will put it on top of this stack.

hasher-matcher-actioner/hmalib/common/mappings.py
schatten: I am not entirely sure how this would look in reality. I can't even remember why I wrote this method as it currently exists. You can't modify HMASignalTypeMapping any way. Let me rethink this bit.

hasher-matcher-actioner/hmalib/scripts/cli/run_api.py
schatten: I'll add comment. In this case, it can't because of a second code smell which is that the lambdas read environment levels at module scope.

hasher-matcher-actioner/hmalib/common/tests/test_mappings.py
schatten: I see. Time to RTFM! :)

hasher-matcher-actioner/hmalib/lambdas/api/middleware.py
schatten: Even though it is painful in the short run to get signal_type_mapping over to _everything_ under the sun, I believe it lets us truly achieve your vision of "if it works in pytx, it will work in hma". 

To that end, I don't mind the requirement to have signal_type_mapping _everywhere_.

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -1,37 +1,206 @@
...
+
+@dataclass
+class ToggleableContentTypeConfig(HMAConfig):
+    content_type_class: str
+
+    # Allow soft disables.
+    enabled: bool = True
+
+
+class HMASignalTypeMapping(SignalTypeMapping):
```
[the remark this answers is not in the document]
schatten: It is news to me that no interface will take SignalTypeMapping as an argument. I thought surely FunctionalityMapping and children would be the weapon of choice when encapsulating functionality. I don't mind it either way.

To reduce unintended dependencies, I'll take up your suggestion of making a 'contains' check rather than an 'is a` check.

Dcallies (changes requested):
No changes since last review

hasher-matcher-actioner/hmalib/common/tests/test_mappings.py
schatten: No longer necessary because we are doing migrations (adding configs) a different way. Check out [this commit](https://github.com/facebook/ThreatExchange/pull/992/commits/d2e0f584a8fad5f9094b0db3423fdfb899884b30) for details.

hasher-matcher-actioner/hmalib/common/config.py
```diff
@@ -353,6 +353,18 @@ def update_config(config: HMAConfig) -> "HMAConfig":
     return config
 
 
+def create_or_update_config(config: HMAConfig) -> "HMAConfig":
+    """
+    Tries to update config, if found. Else creates. Will do two ddb calls.
```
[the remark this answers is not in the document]
schatten: Have removed this altogether. We are not doing create or update at all anymore.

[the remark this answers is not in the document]
schatten: I could not find a native upsert. It is possible to achieve the same outcome as upsert, but would need to know the attribute names in advance. Would require some wizardry in generating a SET statement.

hasher-matcher-actioner/hmalib/common/mappings.py
```diff
@@ -90,33 +83,31 @@ def get_name_from_type(content_type: t.Type[ContentType]) -> str:
         return f"content_type:{full_class_name(content_type)}"
 
 
-class HMASignalTypeMapping(SignalTypeMapping):
+class HMASignalTypeMapping:
```
Dcallies, line 86: (y)

Dcallies (approved):
shipit

hasher-matcher-actioner/hmalib/scripts/cli/migrate.py
```diff
@@ -0,0 +1,73 @@
...
+
+    eg. "$ python -m hmalib.scripts.cli.main migrate
+    2022_04_02_default_content_signal_type_configs"
+
+    Will execute the do_migrate function in the
+    "hmalib.scripts.migrations.2022_04_02_default_content_signal_type_configs"
+    module.
+
+    Right now, there is no state management done within the command.. Doing a
+    migration twice will cause it to happen twice. State-management is left to
```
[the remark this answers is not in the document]
schatten: Run only once is handled by terraform. We can add some bells and whistles to make it _impossible_ to run twice. 

Right now, the only way to have a migration run twice is if:
- some one used `python -m hmacli....` directly. 
- tainted the resource on terraform.

Both, in my opinion, require strong intent on part of the end-user.

PR #994, opened 2022-04-14
[pytx] StopNCIISignalExchangeAPI

Dcallies:
Summary
---------

Now with a Fetcher implementation! Still needs a place to store the keys (same as TX).

Also contains these random, hard to review changes:
1. fetch_iter makes it simpler to store state between fetch attempts, which is actually the most common way to fetch
2. Attempt to massage types more towards a sane typed future (If you have the superclass, assume covariant supertypes, if you have the exact subtype, returns the exact types).

Test Plan
---------

Has a unittest, haven't tried fetching hot yet.


python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -78,6 +81,8 @@ def get_config_class(cls) -> t.Type[CollaborationConfigBase]:
         """Returns the dataclass used to store records for this API"""
         return CollaborationConfigBase
 
+    # TODO - this doens't work for StopNCII which sends the owner as a string
+    #        maybe this should take the full metadata?
```
BarrettOlson, line 85: why not have owner id as an str in the first place, does the break the other API's impl?

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -199,3 +204,56 @@ def report_false_positive(
...
+    def fetch_iter(
+        self,
+        supported_signal_types: t.List[t.Type[SignalType]],
+        collab: CollaborationConfigBase,
+        # None if fetching for the first time,
+        # otherwise the previous FetchDelta returned
+        checkpoint: t.Optional[state.TFetchCheckpoint],
+    ) -> t.Iterator[state.FetchDelta[state.TFetchCheckpoint]]:
+        """
+        An alternative fetch_once implementation simplifies tracking state.
```
BarrettOlson, line 252: stale docstring?

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -199,3 +204,56 @@ def report_false_positive(
...
+
+
+class SignalExchangeAPIWithIterFetch(
+    t.Generic[
+        state.TFetchCheckpoint
+    ],  # TODO move to SignalExchangeAPI in a future diff
+    SignalExchangeAPI,
+):
+    """
+    An alternative fetch_once implementation simplifies tracking state
```
BarrettOlson, line 216: docstring meant for `def fetch_once` below?

BarrettOlson (approved):
Mostly LGTM!
More than happy with you merging as is, all changes look like steps in the right direction :) (and doesn't change the TX API besides some typing)

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: I think I'm trending in that direction, but I haven't decided yet if this leads to weird results on ThreatExchange, where we currently aren't resolving the owner ids. I'm planning on tackling this in a followup, and I'll create a note in https://github.com/facebook/ThreatExchange/projects/3

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: This appears to be the right docstring - can you elaborate?

python-threatexchange/threatexchange/fetcher/fetch_api.py
Dcallies: No, the alternative fetch_once implementation is referring to fetch_iter

PR #996, opened 2022-04-16
[pytx] threatexchange hash to give error on invalid file

Dcallies:
Summary
---------
I wasn't looking close enough and missed this in the original PR, apologies!

Return codes can inform you if a program executed as expected. Return code 0 = success, 2 = user error (invalid arguments, etc)
```
# Before
$ threatexchange hash noexist
$ echo $?  # Display return code
0

# After
$ threatexchange hash noexist
$ echo $?  # Display return code
2
```

Test Plan
---------

Updated test

PR #997, opened 2022-04-16
[pytx][stacked] threatexchange match input update

Dcallies:
Summary
---------

This is similar to #984, except on the match command instead. I think I wrote it in a way that should be quick to port later.

Stacked on #996

Test Plan
---------

updated unittests

### DIFF
--- a/python-threatexchange/threatexchange/cli/helpers.py
+++ b/python-threatexchange/threatexchange/cli/helpers.py
@@ -7,6 +7,8 @@ import sys
 import tempfile
 import shutil
 
+from threatexchange.cli.exceptions import CommandError
+
 """
 Common helpers for the CLI
 
@@ -54,6 +56,6 @@ class FlexFilesInputAction(argparse.Action):
                 break
             path = pathlib.Path(filename)
             if not path.is_file():
-                raise ValueError(f"No such file {path}")
+                raise CommandError(f"No such file {path}", 2)
             ret.append(path)
         setattr(namespace, self.dest, ret)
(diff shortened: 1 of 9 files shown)


python-threatexchange/threatexchange/cli/helpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+    By default assumes filename
+    If `--` is passed in, the remainder is interpreted as text
+    If empty, reads from stidin.
+
+    Use with nargs=argparse.REMAINDER for best results, which gives you:
+
+    $ echo "Some text" > file.txt; cmd file.txt
+    $ echo "Some text" | cmd
+    $ cmd -- Some text
```
sahanapeters, line 33: 🥳

python-threatexchange/threatexchange/cli/tests/sample_data_e2e_test.py
```diff
@@ -13,7 +13,7 @@ class SampleDataE2ETest(ThreatExchangeCLIE2eTest):
     def test_direct_to_match(self):
         """The classic first use of the CLI"""
         self.assert_cli_output(
-            ("match", "text", "-I", self.MATCHES_ONE),
+            ("match", "text", "--", self.MATCHES_ONE),
```
sahanapeters, line 16: Do we not want to keep "-I" for inline?

sahanapeters (approved):
LGTM.

python-threatexchange/threatexchange/cli/tests/sample_data_e2e_test.py
Dcallies: The functionality is still there, just renamed as `--`!

PR #998, opened 2022-04-18
[py-tx] NCMEC API. GET status and entries

Dcallies:
Summary
---------

An implementation for the NCMEC hash API from reading https://report.cybertip.org/hashsharing/v2

Test Plan
---------

I pasted some of the examples from the documentation into the test, similar to what I did for StopNCII.

### DIFF
--- a/python-threatexchange/threatexchange/ncmec/hash_api.py
+++ b/python-threatexchange/threatexchange/ncmec/hash_api.py
@@ -45,12 +45,14 @@ class _XMLWrapper:
     def __init__(self, e: ET.Element) -> None:
         self.element = e
 
-    def __getitem__(self, key) -> "_XMLWrapper":
+    def __getitem__(self, key: str) -> "_XMLWrapper":
         """Asserts child with name exists"""
-        child = self.element.find(key)
-        if child is None:
+        child = self.element.findall(key)
+        if not child:
             raise IndexError
-        return _XMLWrapper(child)
+        if len(child) != 1:
+            raise ValueError(f"More than one child named {key}")
+        return _XMLWrapper(child[0])
 
     def maybe(self, key: str, *rest: str) -> "_XMLWrapper":
         """Gets a child node or a placeholder for chaining"""
@@ -81,7 +83,7 @@ class _XMLWrapper:
 
     def __iter__(self) -> t.Iterator["_XMLWrapper"]:
         """Iterate over children"""
-        for child in list(self.element):
+        for child in self.element:
             yield _XMLWrapper(child)
 
     def __str__(self) -> str:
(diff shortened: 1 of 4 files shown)


python-threatexchange/threatexchange/ncmec/hash_api.py
```diff
@@ -0,0 +1,326 @@
...
+    def __len__(self) -> int:
+        return len(self.element)
+
+    def __bool__(self) -> bool:
+        """True if not a non-existant element from .maybe()"""
+        return self.element is not _DEFAULT_ELE
+
+    def __iter__(self) -> t.Iterator["_XMLWrapper"]:
+        """Iterate over children"""
+        for child in list(self.element):
```
BarrettOlson, line 84: dumb question: What part of the `ET.Element` interface makes this work?

BarrettOlson, line 84: dumb question: What part of the `ET.Element` interface makes this work?

python-threatexchange/threatexchange/ncmec/hash_api.py
```diff
@@ -0,0 +1,326 @@
...
+
+    To use the raw methods, just use .element
+    """
+
+    def __init__(self, e: ET.Element) -> None:
+        self.element = e
+
+    def __getitem__(self, key) -> "_XMLWrapper":
+        """Asserts child with name exists"""
+        child = self.element.find(key)
```
BarrettOlson, line 50: possible dumb question to help my understanding: why `find` not `findall`

BarrettOlson, line 50: possible dumb question to help my understanding: why `find` not `findall`

python-threatexchange/threatexchange/ncmec/hash_api.py
BarrettOlson: answer: because python
import anti~~pattern~~gravity

BarrettOlson: answer: because python
import anti~~pattern~~gravity

python-threatexchange/threatexchange/ncmec/hash_api.py
BarrettOlson: synced offline: should be the same because we expect well formed xml we can treat like json (warning if we are wrong this fails silently) so might be worth a todo/fix

BarrettOlson (approved):
from what understand of the code it LGTM. That plus the tests and docstrings make it so I'm comfortable with a merge as is. 

Don't have experience with XML + python so might be worth a follow up if @schatten has design thoughts post merge

python-threatexchange/threatexchange/ncmec/hash_api.py
Dcallies: Converted to a 'splody version

PR #1000, opened 2022-04-20
Dashboard support for LCC and Metrics for gathering data

Samyakk123:
Summary
---------

image

Add new dashboard with 2 metrics following from the LCC API

### DIFF
--- a/hasher-matcher-actioner/hmalib/indexers/lcc.py
+++ b/hasher-matcher-actioner/hmalib/indexers/lcc.py
@@ -34,7 +34,7 @@ class LCCIndexer:
     @classmethod
     def build_index_from_last_24h(cls, signal_type, storage_path, bucket_width) -> None:
         """Create an index"""
-        with metrics.timer(metrics.names.lcc.build_index):
+        with metrics.timer(metrics.names.lcc.get_data):
             d = timedelta(days=1)
 
             # Make 3 different metric.timers
@@ -48,12 +48,12 @@ class LCCIndexer:
                 HashRecord,
             )
 
-        with metrics.timer(metrics.names.lcc.build_index_get_data):
+        with metrics.timer(metrics.names.lcc.in_memory_processing):
             record_list = []
             for record in past_day_content:
                 record_list.append((record.content_hash, record.content_id))
 
-        with metrics.timer(metrics.names.lcc.build_index_pdq_build):
+        with metrics.timer(metrics.names.lcc.build_index):
             return PDQIndex.build(record_list)
 
     @classmethod
--- a/hasher-matcher-actioner/hmalib/indexers/lcc.py
+++ b/hasher-matcher-actioner/hmalib/indexers/lcc.py
@@ -11,6 +11,7 @@ from datetime import datetime, timedelta
 
 from threatexchange.signal_type.index import SignalTypeIndex
 from threatexchange.signal_type.pdq_index import PDQIndex
+from hmalib import metrics
 
 from hmalib.common.logging import get_logger
 from hmalib.common.models.pipeline import HashRecord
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/indexers/tests/test_lcc_indexer.py
```diff
@@ -26,8 +28,13 @@ def test(self):
             )
             tbi.force_flush()
             test_class = LCCIndexer()
-            test_index = test_class.build_index_from_last_24h(
-                tbi.type, tbi.storage_path, tbi.bucket_width
-            )
+            with metrics.timer(metrics.names.lcc.build_index):
```
schatten, line 31: blocking: The metrics timer should be used inside the LCCIndexer class.

In what we have here, only the test ends up writing metrics.

schatten, line 31: I don't think you should be checking in changes to this file as part of the PR. We only wanted to generate some data so that your dashboard was not empty.

hasher-matcher-actioner/terraform/lcc_custodian/main.tf
```diff
@@ -113,6 +113,53 @@ data "aws_iam_policy_document" "custodian" {
 
 }
 
+
+locals {
+
+  # all_lambda_names = flatten([[for lambda in var.pipeline_lambdas : lambda[1]], var.other_lambdas])
+  title_glance = jsonencode({
```
schatten, line 120: This is not a title, can you think of a better name for this variable?

hasher-matcher-actioner/terraform/lcc_custodian/main.tf
```diff
@@ -113,6 +113,53 @@ data "aws_iam_policy_document" "custodian" {
...
+  first_item = jsonencode({
+    width = 12,
+    type  = "metric",
+    properties = {
+      metrics = [["ThreatExchange/HMA", "lcc.build_index-duration"]]
+      view    = "timeSeries",
+      stacked = false,
+      region  = "${data.aws_region.current.name}"
+      period  = 300
+      stat    = "Average"
```
schatten, line 143: Average stats for duration does not add a lot of value, use ` stat = "p90"` instead.

hasher-matcher-actioner/terraform/lcc_custodian/main.tf
```diff
@@ -113,6 +113,53 @@ data "aws_iam_policy_document" "custodian" {
...
+  # all_lambda_names = flatten([[for lambda in var.pipeline_lambdas : lambda[1]], var.other_lambdas])
+  title_glance = jsonencode({
+    height = 6,
+    width  = 6,
+    type   = "metric",
+    properties = {
+      view    = "timeSeries",
+      stacked = false,
+      metrics = [["ThreatExchange/HMA", "lcc.build_index-count"]]
+      # metric = [for lambda in local.all_lambda_names : ["ThreatExchange/HMA", "lcc.build_index-count"]]
```
schatten, line 128: Remove comment.

hasher-matcher-actioner/terraform/lcc_custodian/main.tf
```diff
@@ -113,6 +113,53 @@ data "aws_iam_policy_document" "custodian" {
...
+      region = "${data.aws_region.current.name}"
```
schatten, line 129: Dunno if there is a default 'stat' transform applied. For counts, you'd want to use 'sum'.

schatten (changes requested):
Overall, looking good.

Blocking: move the `with metrics.timer()` block to inside lcc_indexer.py.
Use correct stat transforms 'p90' for duration and 'sum' for count.

Can explain why offline.

hasher-matcher-actioner/hmalib/indexers/lcc.py
```diff
@@ -33,22 +34,27 @@ def get_recent_index(cls, storage_path, signal_type) -> PDQIndex:
     @classmethod
     def build_index_from_last_24h(cls, signal_type, storage_path, bucket_width) -> None:
         """Create an index"""
-        d = timedelta(days=1)
+        with metrics.timer(metrics.names.lcc.build_index):
```
schatten, line 37: This step is not building the index. It is merely fetching the data.

I'd say there are three stages:
- get_data (where we are getting data from timebucketizer)
- in_memory_processing (process the data L53-55
- build_index

schatten (approved):
The approach looks good to me.

Change the names of the stages as I have recommended (don't forget to update terraform) and we are good to go!

schatten (changes requested):
Left some inline comments. I believe (may be wrong) that the `in_memory_processing` has not been added to the dashboard.

Also, there's a dangling print statement in metrics/__init__.py

schatten (approved):
Fixed the remaining issues after @Samyakk123 granted me write permissions on his repo.

PR #1007, opened 2022-04-27
Make hasher work with pytx refactored

schatten:
Summary
---------
Use signal-type-mapping (rely on dynamic config drivent mappings).

Open hma-config access to hashing lambda.

Test Plan
---------
Ran locally. Running remote is harder because submissions don't work with refactor yet.
```sh
$ python -m hmalib.scripts.cli.main run-lambda --lambda-name hasher
Loading faiss with AVX2 support.
Could not load library with AVX2 support due to:
ModuleNotFoundError("No module named 'faiss.swigfaiss_avx2'")
Loading faiss.
Successfully loaded faiss.
Performance measurement requested. Supplying appmetrics instrumentation.
Unable to load THREAT_EXCHANGE_API_TOKEN_SECRET_NAME from env
/home/dipanjanm/.local/lib/python3.8/site-packages/pdqhash/__init__.py:3: UserWarning: Hash vector order changed between version 0.1.8 and 0.2.0. See https://github.com/faustomorales/pdqhash-python/issues/1 for more details.
  warnings.warn(
```

No errors, verified that the hash_record got created in dynamodb. Visited [hma content page](http://fb-dipanjanm-hma-webapp.s3-website-us-east-1.amazonaws.com/matches/006b9c44-0071-4653-abc8-3a1ed53a4f29). Screenshot attached.
Screen Shot 2022-04-27 at 15 39 00

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/hashing.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/hashing.py
@@ -1,7 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 import functools
-from hmalib.common.messages.bank import BankSubmissionMessage
 import boto3
 import os
 import datetime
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/lambdas/hashing.py
```diff
@@ -48,15 +53,23 @@ def get_sqs_client() -> SQSClient:
...
-)
+@functools.lru_cache(maxsize=None)
+def _get_signal_type_mapping() -> HMASignalTypeMapping:
+    """
+    Cache-get signalTypeMapping. Call only if HMAConfig has been initialized.
+    """
+    return get_pytx_functionality_mapping().signal_and_content
+
+
+@functools.lru_cache(maxsize=None)
```
Dcallies, line 67: Is this the default?

Dcallies (approved):
Assuming just 1:1 mappings.

hasher-matcher-actioner/hmalib/lambdas/hashing.py
schatten: No, maxsize defaults to 128. Using the sentinel value None disables the lru counting so saves a tiny bit of resources.

PR #1009, opened 2022-05-10
[hma][config] allow config set attributes to be empty

schatten:
Summary
---
DynamoDB [does not support](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.SetTypes) empty sets. However, some config attributes (like `only_tags`, `except_tags` for APIs) might be empty sets. This change adds a sentinel value to an empty set and removes that on retrieval for integer and string set types.

Test Plan
---
A unit test was added.

While it is possible to verify this behavior (empty set not allowed) on boto, moto does not raise the exception.  [Filed an issue](https://github.com/spulec/moto/issues/5116) for them and might fix it in a PR later.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/aws_dataclass.py
+++ b/hasher-matcher-actioner/hmalib/common/aws_dataclass.py
@@ -33,6 +33,11 @@ import typing as t
 
 T = t.TypeVar("T")
 
+# dynamodb does not allow sets to be empty. Use this as the only member when
+# writing an empty set. When retrieving a set, remove this if it is the only
+# value.
+EMPTY_SET_SENTINEL_VALUE_STR = "::EMPTY_SET_SENTINEL::"
+
 
 class AWSSerializationFailure(ValueError):
     pass
@@ -81,8 +86,14 @@ def py_to_aws(py_field: t.Any, in_type: t.Optional[t.Type[T]] = None) -> T:
     if in_type == bool:  # BOOL
         return py_field  # type: ignore # mypy/issues/10003
     if in_type == t.Set[str]:  # SS
+        if not py_field or len(py_field) == 0:
+            # Handle empty sets.
+            return {EMPTY_SET_SENTINEL_VALUE_STR}  # type:ignore
         return py_field  # type: ignore # mypy/issues/10003
     if in_type == t.Set[int]:  # SN
+        if not py_field or len(py_field) == 0:
+            # Handle empty sets.
+            return {EMPTY_SET_SENTINEL_VALUE_STR}  # type:ignore
         return {i for i in py_field}  # type: ignore # mypy/issues/10003
     if in_type == t.Set[float]:  # SN
         # WARNING WARNING
@@ -150,9 +161,9 @@ def aws_to_py(in_type: t.Type[T], aws_field: t.Any) -> T:
     if in_type is bool:  # BOOL
         return aws_field  # type: ignore # mypy/issues/10003
     if in_type is t.Set[str]:  # SS
-        return aws_field  # type: ignore # mypy/issues/10003
+        return {value for value in aws_field if value != EMPTY_SET_SENTINEL_VALUE_STR}  # type: ignore # mypy/issues/10003
     if in_type is t.Set[int]:  # SN
-        return {int(s) for s in aws_field}  # type: ignore # mypy/issues/10003
+        return {int(value) for value in aws_field if value != EMPTY_SET_SENTINEL_VALUE_STR}  # type: ignore # mypy/issues/10003
     if in_type is t.Set[float]:  # SN
         return {float(s) for s in aws_field}  # type: ignore # mypy/issues/10003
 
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -33,6 +33,11 @@ class Item(HasAWSSerialization):
 
 T = t.TypeVar("T")
 
+# dynamodb does not allow sets to be empty. Use this as the only member when
+# writing an empty set. When retrieving a set, remove this if it is the only
```
BarrettOlson, line 37: nit: current impl seem to always remove this value this if it is in the set, regardless.

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -33,6 +33,11 @@ class Item(HasAWSSerialization):
...
+# value.
+EMPTY_SET_SENTINEL_VALUE_STR = "::EMPTY_SET_SENTINEL::"
+
 
 class AWSSerializationFailure(ValueError):
```
Dcallies, line 42: (Ignorable): After writing this lovely class, I discovered a module that seems to do it much better - `dacite`. I wish I could go back in tine and switch to it. All the the CLI storage is currently powered by `dacite` instead of handrolling something similar.

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -81,8 +86,14 @@ def py_to_aws(py_field: t.Any, in_type: t.Optional[t.Type[T]] = None) -> T:
     if in_type == bool:  # BOOL
         return py_field  # type: ignore # mypy/issues/10003
     if in_type == t.Set[str]:  # SS
+        if not py_field or len(py_field) == 0:
+            # Handle empty sets.
+            return {EMPTY_SET_SENTINEL_VALUE_STR}  # type:ignore
         return py_field  # type: ignore # mypy/issues/10003
     if in_type == t.Set[int]:  # SN
+        if not py_field or len(py_field) == 0:
```
Dcallies, line 94: What are the cases where the first statement can be false be the second statement is true? Don't we expected `py_field` to be a collection object?

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
```diff
@@ -150,9 +161,9 @@ def aws_to_py(in_type: t.Type[T], aws_field: t.Any) -> T:
     if in_type is bool:  # BOOL
         return aws_field  # type: ignore # mypy/issues/10003
     if in_type is t.Set[str]:  # SS
-        return aws_field  # type: ignore # mypy/issues/10003
+        return {value for value in aws_field if value != EMPTY_SET_SENTINEL_VALUE_STR}  # type: ignore # mypy/issues/10003
```
Dcallies, line 164: (ignorable): This compare might actually be costly for huge lists. You could directly check against the empty set case up front.

```
if origin is set and aws_field == {EMPTY_SET_SENTINEL_VALUE_STR}:
  return set()
```

Dcallies (approved):
This is a tiny bit jank, and I'm confused about this strange restriction in DynamoDB itself.

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
schatten: You'd still need to do some handrolling of dynamodb specific things. True some things might have been simpler.

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
schatten: Valid point. Sets are hashed too. So should be 0(1) check in most cases. https://stackoverflow.com/questions/12390298/time-complexity-of-checking-if-two-set-are-equal-in-python

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
schatten: You're correct. The second one should suffice.

hasher-matcher-actioner/hmalib/common/aws_dataclass.py
schatten: David flagged an easier check that performs better. That will also take care of this issue. Thanks!

schatten:
> This is a tiny bit jank, and I'm confused about this strange restriction in DynamoDB itself.
> 
> Blocking Q: Does this work for sets of ints and floats?

Chose to not implement for floats because I could not imagine a use-case where we'd use sets of floats and they could be empty. Do any come to mind?

> Is an alternative not writing the value at all and if you can't find it, assuming its empty set?

That might have been simpler. 🤦🏼 

I'll put this on one hold. Thanks!

PR #1013, opened 2022-05-21
[pytx] Apply generics to SignalExchangeAPI

Dcallies:
Summary
---------

A tiny bit of cleanup while I decide what to do with the Delta/Update problem that implementing the NCMEC API has given me.

Test Plan
---------

```
ThreatExchange/python-threatexchange$ python -m mypy threatexchange
```
(Though I always get confused about which version we are using in production)


BarrettOlson (approved):
Love clean up and typing additions!

PR #1014, opened 2022-05-25
[py-tx][backwards incompatible] Split Update Format and Index Format

Dcallies:
Summary
---------
The flow of information through the system is:

API.fetch() => Delta => Store => Index => Metadata

NCMEC's API is like the tag-based ThreatExchange API, and it's not stored as hash => record, it's actually stored record => [hash, hash, hash] which is incompatible with many assumptions made along the way.

After lots of mulling, I think the solution is to:
1. Separate considerations about storing the record that the fetch() API produces
2. Converting the format from the fetch() update store to the signal => metadata

As a part of that, I need to mess with a lot of the interfaces above. There are more changes that I want to make, and I tried to checkpoint with this diff.

The final version I want to end up with is:
1. Storing what the update format is from fetch should move from Delta => SignalExchangeAPI
2. Delta should become a trivial object, with no need to override it

As a side effect, the storage format for the CLI changed from JSON to pickle. It's possible I can actually revert this in a few more PRs, but right now Delta is the only object that knows the correct format.

Test Plan
---------

Added a test

Also went through:
1. Fetch from threatexchange
3. Match some content
4. Run the dataset command to make sure it isn't broken (yet).

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_state.py
+++ b/python-threatexchange/threatexchange/cli/cli_state.py
@@ -11,16 +11,13 @@ There are a few categories of state that this wraps:
 """
 
 import pickle
-import json
 import pathlib
 import typing as t
-import dataclasses
 import logging
 
 from threatexchange.signal_type.index import SignalTypeIndex
 from threatexchange.signal_type.signal_base import SignalType
 from threatexchange.cli.exceptions import CommandError
-from threatexchange.cli import dataclass_json
 from threatexchange.fetcher.collab_config import CollaborationConfigBase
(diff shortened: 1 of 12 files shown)


Dcallies:
The error doesn't reproduce in 3.8 in devcontainer, let me see if I can get more information.

Dcallies:
Looks like it's fixed in 3.7 https://github.com/python/typing/issues/511

BarrettOlson (approved):
Approve to unblock. I'd rather merge this as a checkpoint than an even bigger PR later. Thanks!

PR #1015, opened 2022-05-25
[pytx] No more 3.6 support

Dcallies:
Summary
---------

3.6 is no longer supported. Let's no longer support in on py-tx

Test Plan
---------

How do I test this?


BarrettOlson (changes requested):
Oops, we probably should also update this to 3.8: https://github.com/facebook/ThreatExchange/blob/main/.github/workflows/python-threatexchange-release.yaml#L23

BarrettOlson (approved):

PR #1018, opened 2022-06-01
[pdq][cpp] add clang-format and CI

BarrettOlson:
Summary
---------

based on clang-format usages in other fb repos see
 https://github.com/facebook/folly/blob/main/folly/.clang-format
and 
https://github.com/facebook/CacheLib/blob/main/.github/workflows/clang-format-check.yml

Test Plan
---------

Ongoing


BarrettOlson:
> Based on the description, I expect no changes to the AST. Checked a handful of files and it seems to just be whitespace adjustments.

Thanks for the quick review! Full disclosure because clang-format reordered some imports it is _possible_ to see behavior changes but they are unlikely and I didn't see any in testing.

PR #1022, opened 2022-06-07
[pytx] De-Interface FetchDelta, impl details moved to API

Dcallies:
Summary
---------

As a final effort to simplify the SignalExchange interfaces, this moves the details on merging updates into the SignalExchangeAPI itself.

This also has some dangling requests from #1017 that I moved here to avoid one pass there.

Test Plan
---------

py.test && mypy


python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -88,7 +91,49 @@ def get_config_class(cls) -> t.Type[TCollabConfig]:
...
+        fetched: state.TUpdateRecord,
+    ) -> t.Dict[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]]:
+        """
+        Convert the record from the API format to the format needed for indexing.
+
+        This is the fallback method of creating state when there isn't a
+        specialized storage for the fetch type.
+
+        """
+        raise NotImplementedError
```
Dcallies, line 134: This is basically the key part of the PR

python-threatexchange/threatexchange/fetcher/fetch_state.py
```diff
@@ -50,8 +75,11 @@ class SignalOpinionCategory(IntEnum):
     make more sense as a tag
     """
 
+    # TODO: Rename to DOES_NOT_FIT_CATEORY
     FALSE_POSITIVE = 0  # Signal generates false positives
+    # TODO: Rename to something else
     WORTH_INVESTIGATING = 1  # Indirect indicator
+    # TODO: Rename to FITS_CATEGORY
     TRUE_POSITIVE = 2  # Confirmed meets category
```
Dcallies, line 83: This leaked from another set of commits, but worth keeping in I think

BarrettOlson (approved):
Approving to unblock. updates seems like steps in the right direction.

python-threatexchange/threatexchange/fetcher/apis/tests/test_stopncii.py
```diff
@@ -25,27 +21,27 @@ def fetcher(api: StopNCIIAPI):
     return signal_exchange
 
 
-def test_fetch(fetcher: SignalExchangeAPI, monkeypatch: pytest.MonkeyPatch):
+def test_fetch(fetcher: StopNCIISignalExchangeAPI, monkeypatch: pytest.MonkeyPatch):
```
schatten, line 24: non-blocking. We've used [freezegun](https://github.com/spulec/freezegun) for moving through time in HMA. If the testing gets any more complex, freezegun's API is much more readable and does not rely on magic numbers.

python-threatexchange/threatexchange/fetcher/apis/tests/test_stopncii.py
```diff
@@ -25,27 +21,27 @@ def fetcher(api: StopNCIIAPI):
...
     monkeypatch.setattr("time.time", lambda: 10**8)
     collab = CollaborationConfigWithDefaults("Test")
     it = fetcher.fetch_iter([], collab, None)
-    delta = t.cast(SimpleFetchDelta, next(it))
+    delta = next(it)
 
-    assert delta.record_count() == 2
+    assert len(delta.updates) == 2
```
schatten, line 30: Do you make actual API calls here? Can this test start failing because of the underlying database receiving updates?

schatten (approved):
Looks good. Comments inline. The only open question I'd like answered before you merge is this:

- If for large datasets, we MUST move the merge impl into the FetchStateStore implementation, why should the API interface support merging at all?
  - Why not make that a responsibility of the store itself?

Also, a strongly held nit about not editing function values if you are also returning said value.

Dcallies:
> If for large datasets, we MUST move the merge impl into the FetchStateStore implementation, why should the API interface support merging at all? 

Because I posit that almost no database has this problem currently. Hashes are generally small. Providing in-memory as a fallback means that a fallback always exists, which is enough to prototype and test, which is the main usecase I'm targeting with the `naive_` functions. If you can prove the value, you can justify the effort of building a store.

> Why not make that a responsibility of the store itself?
My goal was to reduce the total number of interfaces that need to prototype to 1 - SignalAPI. If I did this, it would increase to 2, and also whatever storage you choose might not work in all cases. 

---

I am generally encouraging all interfaces to be built off dataclass, and did at least prototype structured storage that would be needed. There's also a potential solution that I debated, which is enforcing that all updates must be keyed - if that's the case, it always allow partitioning by the key, and so always has an approach to incremental storage. I chose not to take that step. 

However, that step might be worth taking because it does have a translation for key just being a constant, which then puts the burden on the API for not making individual values larger than what is likely to be available to memory.

python-threatexchange/threatexchange/fetcher/apis/tests/test_stopncii.py
Dcallies: I might attempt to convince you in the other direction - I've found the py.test fixture functionality pretty usable, and just has a short barrier to understanding it.

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -225,7 +270,77 @@ def report_false_positive(
...
+        for k, v in new.items():
+            if v is None:
+                old.pop(k, None)
+            else:
+                old[k] = v
+        return old
+
+
+class SignalExchangeAPIWithSimpleUpdates(
+    SignalExchangeAPIWithKeyedUpdates[
```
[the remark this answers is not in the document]
Dcallies: I considered adding them as mixins, but I've confused myself in the past by ordering methods incorrectly when using mixins. Adding the diamond explicitly provides a consistent resolution order.

python-threatexchange/threatexchange/fetcher/fetch_api.py
```diff
@@ -225,7 +270,77 @@ def report_false_positive(
...
+    An API that supports sequential updates with a common key (id, hash, etc).
+
+    This provides a simple approach for merging data.
+    A value of None for a value indicates that the record was deleted.
+    """
+
+    @classmethod
+    def naive_fetch_merge(
+        cls, old: t.Optional[t.Dict[K, t.Optional[V]]], new: t.Dict[K, t.Optional[V]]
+    ) -> t.Dict[K, t.Optional[V]]:
```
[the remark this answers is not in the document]
Dcallies: It's explained you can do so in the interface comment.

PR #1023, opened 2022-06-14
[pytx] Fetcher Implementation for NCMEC, Many Supporting Changes

Dcallies:
Summary
---------

This one got away from me a little bit:
1. SignalExchangeAPI implementation for NCMEC
2. CLI configuration updates to store credentials
3. Random bug fixes where I found them

Test Plan
---------

#### Collab Edit
```
$ threatexchange config collab edit ncmec --help
usage: threatexchange config collab edit ncmec [-h] [--create] [--enable [{0,1}] | --disable] [--only-signal-types [NAME [NAME ...]]] [--not-signal-types [NAME [NAME ...]]] [--only-owners [ID [ID ...]]] [--not-owners [ID [ID ...]]] [--only-tags [TAG [TAG ...]]] [--not-tags [TAG [TAG ...]]] [--json] [--environment [Industry,NGO,Exploitative]] collab_name

positional arguments:
  collab_name           the name of the collab

optional arguments:
  --create, -C          indicate you intend to create a config
  --enable [{0,1}]      enable the config (default on create)
  --disable             disable the config
  --only-signal-types [NAME [NAME ...]], -s [NAME [NAME ...]]
                        limit to these signal types
  --not-signal-types [NAME [NAME ...]], -S [NAME [NAME ...]]
                        dont use these signal types
  --only-owners [ID [ID ...]], -o [ID [ID ...]]
                        only use signals from these owner ids
  --not-owners [ID [ID ...]], -O [ID [ID ...]]
                        dont use signals from these owner ids
  --only-tags [TAG [TAG ...]], -t [TAG [TAG ...]]
                        use only signals with one of these tags
  --not-tags [TAG [TAG ...]], -T [TAG [TAG ...]]
                        don't use signals with one of these tags
  --json, -J            instead, interpret the argument as JSON and use that to edit the config
  --environment [Industry,NGO,Exploitative]
                        [auto generated from config class]
```

#### Configure NCMEC
```
$ threatexchange config api ncmec --help
usage: threatexchange config api ncmec [-h] [--credentials STR STR]

Configure NCMEC hash api integration

optional arguments:
  --credentials STR STR
                        set the username and password to access the NCMEC API
```

### E2E Test
```
$ threatexchange extensions add pytx_photodna  # Non-public extension
$ threatexchange config api ncmec --credentials <secret>
$ threatexchange config collab edit ncmec -C 'Test NCMEC' --environment NGO
$ threatexchange fetch --limit 5000
$ threatexchange match -HS photodna photo -- 0.0.0... # The zero hash
$ threatexchange match -HS photodna photo -- <secret>  # A hash from the database
photodna - (Test NCMEC) TRUE_POSITIVE B2
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -51,7 +51,7 @@ class CLiConfig:
     """A place to store misc configuration for the CLI"""
 
     fb_threatexchange_api_token: t.Optional[str] = None
-    ncmec_access: t.Optional[t.Tuple[str, str]] = None
+    ncmec_credentials: t.Optional[t.Tuple[str, str]] = None
     stop_ncii_keys: t.Optional[StopNCIIKeys] = None
     extensions: t.Set[str] = field(default_factory=set)
     # Every item needs a default for backwards compatibility
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -17,6 +17,7 @@ import logging
 
 from dacite import WrongTypeError
 
+from threatexchange.stopncii import api as stopncii_api
 from threatexchange.fetcher import collab_config
 from threatexchange.fetcher.fetch_api import SignalExchangeAPI
 from threatexchange.content_type import content_base
@@ -32,11 +33,28 @@ CONFIG_FILENAME = "config.json"
 
 
 @dataclass
+class StopNCIIKeys:
+    subscription_key: str
+    fetch_function_key: str
+    # Someday, additional keys
+
+    @property
+    def keys_are_valid(self):
+        return all(
+            stopncii_api.is_valid_key(k)
+            for k in (self.subscription_key, self.fetch_function_key)
+        )
+
+
+@dataclass
 class CLiConfig:
     """A place to store misc configuration for the CLI"""
 
     fb_threatexchange_api_token: t.Optional[str] = None
+    ncmec_access: t.Optional[t.Tuple[str, str]] = None
+    stop_ncii_keys: t.Optional[StopNCIIKeys] = None
     extensions: t.Set[str] = field(default_factory=set)
+    # Every item needs a default for backwards compatibility
 
 
 class CliState(collab_config.CollaborationConfigStoreBase):
--- a/python-threatexchange/threatexchange/cli/config_cmd.py
+++ b/python-threatexchange/threatexchange/cli/config_cmd.py
@@ -7,9 +7,9 @@ Config command to setup the CLI and settings.
 
 import argparse
 from dataclasses import is_dataclass, Field, fields, MISSING
+from enum import Enum
 import itertools
 import json
-import importlib
 import logging
 import os
 import typing as t
@@ -35,6 +35,7 @@ from threatexchange.fetcher.apis.fb_threatexchange_api import (
     FBThreatExchangeSignalExchangeAPI,
 )
 from threatexchange.fetcher.fetch_api import SignalExchangeAPI
+from threatexchange.fetcher.apis.ncmec_api import NCMECSignalExchangeAPI
 from threatexchange.fetcher.apis.static_sample import StaticSampleSignalExchangeAPI
 from threatexchange.signal_type.signal_base import SignalType
 
@@ -191,10 +192,20 @@ class _UpdateCollabCommand(command_base.Command):
         if hasattr(field.type, "__args__"):
             target_type = field.type.__args__[0]
 
+        argparse_type = target_type
+        metavar = target_type.__name__
+
+        if issubclass(target_type, Enum):
+            argparse_type = common.argparse_choices_pre_type(
+                [m.name for m in target_type],
+                lambda s: target_type[s],
+            )
+            metavar = f"[{','.join(m.name for m in target_type)}]"
+
         ap.add_argument(
             f"--{field.name.replace('_', '-')}",
-            type=target_type,
-            metavar=target_type.__name__,
+            type=argparse_type,
+            metavar=metavar,
             help="[auto generated from config class]",
         )
 
@@ -248,6 +259,8 @@ class _UpdateCollabCommand(command_base.Command):
 
         for field in fields(self._API_CLS.get_config_class()):
             if not field.init:
+                if field.name == "api":
+                    self.edit_kwargs.pop("api")
                 continue
             if field.name in self._IGNORE_FIELDS:
                 continue
@@ -524,7 +537,7 @@ class ConfigContentCommand(command_base.Command):
 
 
 class ConfigThreatExchangeAPICommand(command_base.Command):
-    """Configure apis"""
+    """Configure Facebook ThreatExchange integration"""
 
     @classmethod
     def get_name(cls) -> str:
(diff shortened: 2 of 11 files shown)


python-threatexchange/threatexchange/cli/config_cmd.py
```diff
@@ -620,10 +633,39 @@ def execute_config(self, settings: CLISettings) -> None:
...
+    def __init__(
+        self,
+        credentials: t.List[str],
+    ) -> None:
+        self.credentials = (credentials[0], credentials[1]) if credentials else None
+
+    def execute(self, settings: CLISettings) -> None:
+        if self.credentials is not None:
+            config = settings.get_persistent_config()
+            config.ncmec_credentials = self.credentials
```
Dcallies, line 661: Unfortunately, it's hard to directly validate on the command without changing the argparse format (nargs=2 treats each arg separately by default), but at least we could validate here

python-threatexchange/threatexchange/cli/main.py
```diff
@@ -141,8 +143,82 @@ def _get_fb_tx_app_token(config: CLiConfig) -> t.Optional[str]:
 def _get_stopncii_tokens(
     config: CLiConfig,
 ) -> t.Tuple[t.Optional[str], t.Optional[str]]:
-    """Get the API keys for StopNCII from the config"""
-    return None, None  # TODO
+    """
+    Get the API key from a variety of fallback sources
+
+    Examples might be environment, files, etc
+    """
```
Dcallies, line 150: As I was writing this, I wondered whether I shouldn't add a place for this on the signal exchange API itself, since otherwise I keep adding these on individual classes. Getting the default from environment/file seems simple enough to make default behavior in the constructor

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
```diff
@@ -0,0 +1,232 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+SignalExchangeAPI impl for the NCMEC hash exchange
+
+"""
```
Dcallies, line 6: Clean this up

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
```diff
@@ -0,0 +1,232 @@
...
+          (any unique key, a combining function) from the concerns needed for Index
+          to work (group by signal type => Metadata).
+    """
+
+    entry: api.NCMECEntryUpdate
+
+    def get_as_opinions(self) -> t.List[state.SignalOpinion]:
+        raise NotImplementedError(
+            "Placeholder while I figure out the correct model for NCMEC"
+        )
```
Dcallies, line 83: I fixed this with the previous PR, remove

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
```diff
@@ -0,0 +1,232 @@
...
+    ret: t.Dict[t.Tuple[api.NCMECEntryType, str], t.Type[SignalType]] = {}
+    if VideoMD5Signal in signal_types:
+        ret[api.NCMECEntryType.video, "md5"] = VideoMD5Signal
+    if PdqSignal in signal_types:
+        ret[api.NCMECEntryType.image, "pdq"] = PdqSignal
+    for st in signal_types:
+        if st.get_name() == "photodna":
+            ret[api.NCMECEntryType.image, "pdna"] = st
+            break
+    return ret
```
Dcallies, line 117: I need a better way to handle this long term. You end up with a problem where you might have one of the signal type or API. 

This API might be slightly different since it's included in the base threatexchange library, and using string reflection here is okay, but I dunno

python-threatexchange/threatexchange/cli/main.py
```diff
@@ -141,8 +143,82 @@ def _get_fb_tx_app_token(config: CLiConfig) -> t.Optional[str]:
...
+            file=sys.stderr,
+        )
+        # Don't throw because we don't want to block commands that fix this
+        return None, None  # We probably don't expect to fall back here
+    return None, None
+
+
+def _get_ncmec_credentials(config: CLiConfig) -> t.Tuple[str, str]:
+    """Get user+pass from NCMEC from the config"""
+    environment_var = "TX_STOPNCII_KEYS"
```
schatten, line 189: Shouldn't this be more NCMEC specific?

python-threatexchange/threatexchange/cli/main.py
schatten: Why have fallbacks at all? What prompted this feature?

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @classmethod
+    def naive_fetch_merge(
+        cls,
+        old: t.Optional[NCMECUpdate],
+        new: NCMECUpdate,
+    ) -> NCMECUpdate:
+        old = old or {}
+        old.update(new)
+        return old
```
schatten, line 201: Lol! `Naivest_fetch_merge`!

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
```diff
@@ -0,0 +1,232 @@
...
+    def naive_convert_to_signal_type(
+        cls,
+        signal_types: t.Sequence[t.Type[SignalType]],
+        fetched: NCMECUpdate,
+    ) -> t.Dict[t.Type[SignalType], t.Dict[str, NCMECSignalMetadata]]:
+        mapping: t.Mapping[
+            t.Tuple[api.NCMECEntryType, str], t.Type[SignalType]
+        ] = _get_conversion(signal_types)
+        ret: t.Dict[t.Type[SignalType], t.Dict[str, NCMECSignalMetadata]] = {}
+        for entry in fetched.values():
```
schatten, line 213: How does NCMEC (and this API) handle deletes? I don't see any handling of 'entry == None'.

schatten (changes requested):
blocking:

1/ L189 in `cli/main.py` uses the wrong environment variable best I can tell.
2/ how are we handling deletes? 

open-question:
- I'm not sure the fallbacks for credentials are necessary. Esp. because the state (checkpoint) is stored persistently. Allowing change of credential might affect consistency.

This may not affect NCMEC, but any other privacy-group oriented APIs might get affected. eg.

t1: use fetch with credentials which have access. checkpoint moved forward.
t2: Use fetch with credentials which do not have access. no updates, but checkpoint moved forward (based on API. eg GraphQL cursors)
t3. Use fetch again with credentials from t1. 

Unless we store credentials along with the state, I would not want to open up the possibility of this happening. Even though this is unlikely, if it were to happen eg. test-creds vs prod-crds, it is really hard to debug.

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
Dcallies: search for `.deleted`

python-threatexchange/threatexchange/cli/main.py
Dcallies: Copy paste failure

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -31,12 +32,29 @@
...
+            for k in (self.subscription_key, self.fetch_function_key)
+        )
+
+
 @dataclass
 class CLiConfig:
     """A place to store misc configuration for the CLI"""
 
     fb_threatexchange_api_token: t.Optional[str] = None
+    ncmec_credentials: t.Optional[t.Tuple[str, str]] = None
```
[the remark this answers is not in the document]
Dcallies: NCMEC ones were simpler. StopNCII has a complex function key model

Dcallies:
Thanks for review!

Answers inline for blockers

open-question:

> I'm not sure the fallbacks for credentials are necessary. 

The main motivation for this is currently storing them on the CLI requires typing them out in plaintext in a way that inserts them into history, which is generally bad practice. Both the fallback options provide an alternative that is secure, since I didn't do the right things to secure the config file itself (make it writable only by user, etc). 

I just duplicated what was already been done for TX without really thinking about it in the short term while I considered alternatives. 
1. (Plays nicely with extensions) Allow SignalExchange APIs to look for their own credentials during __init__
2. Actually do the right thing to store credentials securely on CLI (secure file, only accept from stdin, etc)
3. Do what TX does and allow putting credentials into collabconfig (doesn't make sense for NCMEC which uses a single credential, but an alternative would be to accept a list of environments instead of one)

Re: your suggestion to store credentials in the checkpoint - worried this conflates fetched state with configuration, especially since the CLI encourages you to clear storage whenever it runs into an issue currently :/

python-threatexchange/threatexchange/fetcher/apis/ncmec_api.py
```diff
@@ -0,0 +1,232 @@
...
+            ret[api.NCMECEntryType.image, "pdna"] = st
+            break
+    return ret
+
+
+NCMECUpdate = t.Dict[str, api.NCMECEntryUpdate]
+
+
+class NCMECSignalExchangeAPI(
+    fetch_api.SignalExchangeAPI[
```
Dcallies, line 124: Should be changed to the keyed version

python-threatexchange/threatexchange/cli/main.py
Dcallies: Offline discussion: Good CLI practice reasons - using the config to create a password will cause it to be stored visibly and in plaintext in bash history, which is an antipattern. Providing alternatives gives security-conscious users an alternative.

schatten (approved):

PR #1024, opened 2022-06-17
Add CI action for tmk/cpp make regtest

BarrettOlson:
Summary
---------

pdq/cpp has a gh action but tmk/cpp was missing one. This adds it.

There is a side benefit that it now you mistakenly commit your local FFMPEG path the CI will fail (something that I missed and committed a few PRs ago by mistake which is also fixed here)

Test Plan
---------

Created this PR made sure both the new and existing gh action pass.

### DIFF
--- a/tmk/cpp/Makefile
+++ b/tmk/cpp/Makefile
@@ -1,7 +1,7 @@
 # ----------------------------------------------------------------
 CXX=g++ # please customize according to your system --- Note: Apple clang is not supported 
 CC=$(CXX) -O2 -std=c++14 -I../.. #
-FFMPEG=/Users/barrett/homebrew/bin/ffmpeg # please customize according to your installation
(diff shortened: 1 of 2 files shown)


tmk/cpp/Makefile
```diff
@@ -1,7 +1,7 @@
 # ----------------------------------------------------------------
 CXX=g++ # please customize according to your system --- Note: Apple clang is not supported 
 CC=$(CXX) -O2 -std=c++14 -I../.. #
-FFMPEG=/Users/barrett/homebrew/bin/ffmpeg # please customize according to your installation
+FFMPEG=/usr/bin/ffmpeg # please customize according to your installation (but do not commit)
```
schatten, line 4: non-blocking: Should we make this an environment variable?

BarrettOlson: yes, good follow up, I don't know why it wasn't one from the start.

PR #1025, opened 2022-06-19
[pytx][cli] Allow printing out signals, csv files

Dcallies:
Summary
---------

This PR adds the ability to print out data to screen from the dataset command. This is designed to close a functionality gap with the old CLI that stored data in simple csv files that could be passed around.

Test Plan
---------

Added test

(Note - below `tx` is used as an alias for `threatexchange`

### Halp
```
$ threatexchange dataset --help
usage: threatexchange dataset [-h] [--rebuild-indices | --clear-indices | --signal-summary | --print-signals] [--only-signals SIGNAL_TYPE [SIGNAL_TYPE ...] | --only-content CONTENT_TYPE
                              [CONTENT_TYPE ...]] [--only-collabs NAME [NAME ...]] [--only-tags STR] [--signals-only] [--csv] [--print-zeroes]

    Introspect fetched data.

    Can print out contents in simple formats
    (ideal for sending to another system), or regenerate
    index files (ideal if distributing indices for some reason)

    Example commands:

    ```
    # Show total size of known signals
    $ threatexchange dataset

    # Show the size of one collaboration
    $ threatexchange dataset -c 'Sample Data'

    # Generate a simple CSV file for one collaboration for ease of distribution
    $ threatexchange dataset -c 'Sample Data' -P --csv > collab.csv
    ```
    

optional arguments:
  -h, --help            show this help message and exit
  --rebuild-indices, -r
                        rebuild indices from fetched data
  --clear-indices, -X   clear all indices
  --signal-summary      print summary in terms of signals (default action)
  --print-signals, -P   print signals to screen
  --only-signals SIGNAL_TYPE [SIGNAL_TYPE ...], -s SIGNAL_TYPE [SIGNAL_TYPE ...]
                        only consider signals of this type
  --only-content CONTENT_TYPE [CONTENT_TYPE ...], -C CONTENT_TYPE [CONTENT_TYPE ...]
                        only process signals for these content types
  --only-collabs NAME [NAME ...], -c NAME [NAME ...]
                        [-S|-P] only count items with this tag
  --only-tags STR, -t STR
                        [-S|-P] only count items with these tags
  --signals-only, -S    [-P] only print signals
  --csv                 [-P] store in csv format
  --print-zeroes, -z    when printing counts, print 0 values
```

### Some examples with sample data

```
$ tx dataset
pdq: 138
raw_text: 3
video_md5: 2
trend_query: 1
url: 1
url_md5: 1

$ tx dataset -P | tail -n5
[...]

PR #1026, opened 2022-06-19
Missing toplevel command

Dcallies:
Summary
---------

Calling with so subcommands is not an error, but generates an assertion failure shortly afterwards. Fix that I guess

Test Plan
---------

Unittest

```
$ threatexchange
<see help>
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/main.py
+++ b/python-threatexchange/threatexchange/cli/main.py
@@ -73,7 +73,9 @@ def get_argparse(settings: CLISettings) -> argparse.ArgumentParser:
     )
     # is_config = Safe mode to try and let you fix bad settings from CLI
     ap.set_defaults(is_config=False)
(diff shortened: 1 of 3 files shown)


python-threatexchange/threatexchange/cli/tests/dataset_cmd_test.py
```diff
@@ -0,0 +1,65 @@
...
+    def test(self):
+        """
+        Test on the output of the dataset command, using the sample signals
+
+        The sample signals will probably change a few times during the course
+        of development, which will unexpectedly break this test, so apologies
+        future developers.
+        """
+        self.assert_cli_output(("dataset",), "")  # No datas yet
+        self.cli_call("fetch")
```
schatten, line 14: Love this surface for tests!

schatten (approved):
Looks good. Thanks!

BarrettOlson (commented):
FYI: Looks like this has changes from #1025 as well

PR #1027, opened 2022-06-19
[pytx] README.md rewrite for 1.0

### LINKED ISSUE #985: Update threatexchange library README for new usage

Dcallies:
Big-ol rewrite. I didn't plug it into a text editor, so probably a ton of grammars and speeling errors

Mostly fixes #985

### DIFF
--- a/python-threatexchange/README.md
+++ b/python-threatexchange/README.md
@@ -1,23 +1,15 @@
 # python-threatexchange
 
-A Python Library for downloading and using datasets stored in ThreatExchange.
+A Python Library to simplify the exchange and use of trust & safety information, especially media hash exchanges. Also comes with a CLI.
+
+python-threatexchange is designed to be extensible, and comes with a simple model of adding new functionality.
+
+To get similar functionality in a deployable service, check out hasher-matcher-actioner.
 
 ![PyPI - Python Version](https://img.shields.io/pypi/pyversions/threatexchange) [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black) 
 
-## Upcoming Changes - Please Pardon Our Dust
-ThreatExchange is in the middle of a big refactor starting in #944 which will make it less about Facebook's ThreatExchange, and more about signal and hash exchange in general. There may be some breakages in the short term, and many of the concepts below will need to be updated. 
-
-
-## Overview
-Fetching signals from ThreatExchange "correctly" is sadly not as straightforward as it should be.
-
-Additionally, once you have the data, the application of those signals can be difficult.
-
-This library provides a reference implementation for fetching datasets, applying many common formats, and uploading your own signals to the dataset.
-
-These reference implementations are not meant to be the best, most efficient, or production ready versions, but to instead be simple enough that they could be easily pasted to other languages.
-
-The bundled cli tool of the same name is meant to demonstrate the library, as well as potentially provide a first draft implementation for prototype or evaluation.
+## Please Pardon Our Dust
+ThreatExchange is in the middle of a big refactor starting in #944, and isn't fully stable yet. Track progress at https://github.com/facebook/ThreatExchange/projects/3
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


python-threatexchange/README.md
```diff
@@ -1,23 +1,15 @@
...
-
-## Upcoming Changes - Please Pardon Our Dust
-ThreatExchange is in the middle of a big refactor starting in #944 which will make it less about Facebook's ThreatExchange, and more about signal and hash exchange in general. There may be some breakages in the short term, and many of the concepts below will need to be updated. 
-
-
-## Overview
-Fetching signals from ThreatExchange "correctly" is sadly not as straightforward as it should be.
-
-Additionally, once you have the data, the application of those signals can be difficult.
+python-threatexchange is designed to be extensible, and comes with a simple model of adding new functionality.
```
BarrettOlson, line 5: ignorable: editing nit: ~~and comes~~

python-threatexchange/README.md
```diff
@@ -27,113 +19,108 @@ If you don't have `pip`, learn how to install it [here](https://pip.pypa.io/en/s
...
 
-**Photo**
-* MD5
+## Philisophy of the Library
+Trust and safety is not owned by Meta. However, this library is maintained by a small team at Meta with a limited range of experience, and so we'll probably prioritize the usecases we are most familiar with. We believe that accessibility is a barrier for many platforms and so will put as much as we can in the open. We also understand that it may not make sense to use only publicly visible approaches, and welcome platform-specific modifications and derivatives. However, we also accept pull requests! If you think functionality is generally applicable, or you have a bug bothering you, we accept pull requests! If you are thinking a larger change may be needed (such as adding an entirely new subcommand to the CLI), we definitely appreciate if you reach out to talk through a feature before submitting it!
 
-**Video**
-* MD5
+## General Expectation for Compatibility and Versioning
+1. Major versions (1.X.X => 2.0.0) are not guaranteed to backwards compatible. However, post 1.0.0, tooling to try and migrate state forward will be available. Please reach out if you need help migrating forward.
```
BarrettOlson, line 33: this is speaking to a real or abstract versions, if abstract we should add "e.g." to the parentheses to make it clear we are not talking about specific pytx versions.
Also nit: "will" -> "should" to cover our bases (thinking of the case where a bug is discovered that makes such tooling impossible)

python-threatexchange/README.md
```diff
@@ -27,113 +19,108 @@ If you don't have `pip`, learn how to install it [here](https://pip.pypa.io/en/s
...
-3425830734108278 raw_text media_priority_samples
-3261912580534814 trend_query media_priority_samples
-
-$ threatexchange label descriptor 3425830734108278 false_positive
-```
+A serialization of data that can be used as an input to detect/match content is called a "signal", and this library enforces that every signal be representable has a python str class.
 
-## Key Concepts
-This library extends some of the concepts from the [ThreatExchange API](https://developers.facebook.com/docs/threat-exchange/reference/apis/) to facilitate content and hash matching. ThreatExchange has a very generic data model, of which the usage has mostly been defined by convention. This library allows for encoding common conventions.
+SignalType enforces that you provide "naive" or brute force versions of the techniques that can be used for correctness testing. By default, python-threatexchange will use a simple linear scan against these brute force methods to However, if there are more efficient methods for scanning large datasets, the SignalTypeIndex interface provides a place to store a more complex scaled technique.
```
BarrettOlson, line 46: ~~to However,~~

python-threatexchange/README.md
Dcallies: We talked about this offline, 

I believe that we "will" have tooling to "try" to migrate. We cannot provide no solution. Barrett counters that this should be a team decision on if we will uphold this, which I think narrows to just @BarrettOlson and @schatten.

Are you two in favor to keeping breaking changes to major version numbers, and to provide a route for backwards compatibility for state between major versions?

BarrettOlson (commented):
LGTM but definitely want @schatten opinion as well

BarrettOlson (approved):

PR #1029, opened 2022-06-20
[pytx] convert CLI test to py.test style

Dcallies:
Summary
---------

py.test is super cool, and one of the things that makes it cool is fixtures, which are a great alternative to unittest classes. The big value add is that all the various .assert methods have been greatly improved upon via just assert methods. 

However, if you still use unittest objects, you actually don't get the benefits from the improved assertions, so you usually want to swap over.

This PR makes the unitest class more py.test-y, and swaps over one command to demonstrate.

Test Plan
---------

py.test


BarrettOlson (approved):
LGTM

PR #1035, opened 2022-06-28
[VPDQ]  Add VPDQ python bidning  and FAISS Flat Index

tigerchensh:
Summary
---------

Implement python binding for VPDQ. Now, CPP functionalities can be used in python. 
Add Faiss FlatIndex matching for VPDQ as well as the quality filter.
Faiss FlatIndex also returns matching percentages

Test Plan
---------
Build python binding:
python3 setup.py build_py build_ext --inplace

Test Faiss Matching by unit test:
python3 -m unittest test_vpdq_faiss.py -v

### DIFF
--- a/vpdq/python/test_vpdq_faiss.py
+++ b/vpdq/python/test_vpdq_faiss.py
@@ -2,7 +2,6 @@
 
 import unittest
 import typing as t
-import functools
 import vpdq_matcher
 import vpdq_util
 import os
--- a/vpdq/python/test_vpdq_faiss.py
+++ b/vpdq/python/test_vpdq_faiss.py
@@ -1,0 +1,75 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import unittest
+import typing as t
+import functools
+import vpdq_matcher
+import vpdq_util
+import os
+
+SAMPLE = "/sample-hashes"
+
+
+class TestVPDQFaiss(unittest.TestCase):
+    def setUp(self):
+        d = os.path.dirname(os.getcwd())
+        sample_folder = d + SAMPLE
+        self.index = vpdq_matcher.VPDQFlatHashIndex()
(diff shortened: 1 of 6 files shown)


vpdq/python/test_vpdq_faiss.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import unittest
```
BarrettOlson, line 3: Yay tests!

Instead of the unittest framework @Dcallies is recommending we use pytest (and switch over the existing unittest)

Here an example PR he merged: https://github.com/facebook/ThreatExchange/pull/1029

vpdq/python/vpdq_matcher.py
```diff
@@ -0,0 +1,266 @@
+import vpdq
+import typing as t
+import faiss  # type: ignore
+import binascii
+import numpy  # type: ignore
+
+BITS_IN_VPDQ = 256
+
+
+def dedupe(hashes):
```
BarrettOlson, line 10: can you add typing and return typing to these methods?

vpdq/python/vpdq_util.py
```diff
@@ -0,0 +1,88 @@
+import vpdq
+import numpy as np
+
+
+def vector_to_hex(hash_vector):
```
BarrettOlson, line 5: This is also too much of copy of `pdq_hasher.py`
maybe we should move more of this into python-threatexchange so these kind of utils can be reused.

BarrettOlson (changes requested):
Happy with the progress. Excited to see this working.

However, I think this copies too much from the existing code in `python-threatexchange` as written. 

This is probably means we are due for a longer design conversation but for now could you break this PR up into the core python binding we know we want to land in `vpdq` and the remaining features that have overlap?

Thanks!

PR #1036, opened 2022-06-29
[VPDQ] Add VPDQ python binding

tigerchensh:
Summary
---------

Implement python binding for VPDQ. Now, CPP-VPDQ hashing can be used in python.
Add a regression test for python binding that match the computed hashes with expected hashes.

Test Plan
---------

Build python binding:
python3 setup.py build_py build_ext --inplace

Test Binding Hashing by pytest:
Replace FFMPEG with your own installation path in test_vpdq_hash.py
pytest test_vpdq_hash.py


vpdq/python/test_vpdq_hash.py
```diff
@@ -0,0 +1,45 @@
...
+FFMPEG = "/usr/bin/ffmpeg"  # please customize according to your installation (but do not commit)
+SECOND_PER_HASH = 1
+DISTANCE_TOLERANCE = 10
+QUALITY_TOLERANCE = 50
+
+
+class TestVPDQHash:
+    def test_VPDQ_hash(self):
+        d = os.path.dirname(os.path.dirname(os.getcwd()))
+        hash_folder = d + SAMPLE_HASH_FOLDER
```
BarrettOlson, line 18: nit: instead of just + for string concat can you use? 
https://docs.python.org/3/library/pathlib.html#pathlib.Path

vpdq/python/vpdq.pyx
```diff
@@ -0,0 +1,131 @@
...
+                width: int,
+                height: int):
+    """Compute vpdq hash
+
+    Args:
+        input_video_filename (str): Input video file path
+        ffmpeg_path (str): ffmpeg path
+        verbose (bool): If verbose, will print detailed information
+        seconds_per_hash (int): The frequence(per second) a hash is generated from the video
+        width (int): Width to downsample the video to before hashing frames.. If it is 0, will use the original width of the video to hash
```
BarrettOlson, line 103: nit: naming -> `downsample_width`, `downsample_height` instead

vpdq/python/vpdq.pyx
```diff
@@ -0,0 +1,131 @@
...
+        vpdq_hash,
+        ffmpeg_path.encode('utf-8'),
+        verbose,
+        seconds_per_hash,
+        width,
+        height,
+        "vpdqPY"
+        )
+
+    hashs= [vpdq_feature(hash.quality,
```
BarrettOlson, line 128: nit nit: hashes

vpdq/python/vpdq.pyx
```diff
@@ -0,0 +1,131 @@
...
+    Returns:
+        int: The hamming distance
+    """
+    return hammingDistance(hash1, hash2)
+
+def computeHash(input_video_filename: str,
+                ffmpeg_path: str,
+                verbose: bool,
+                seconds_per_hash: int,
+                width: int,
```
BarrettOlson, line 94: can width & height use default args = 0?

BarrettOlson (commented):
I have a few clean up nits but overall this looks good.

Unsure: to avoid confusion given the pure python impl and `pdq/python` do you think we should rename this folder to `python_binding`

vpdq/python/test_vpdq_hash.py
```diff
@@ -0,0 +1,50 @@
...
+                )
+                name = os.path.splitext(file)[0]
+                self.test_hashes[name] = ret
+
+        for k in self.sample_hashes.keys():
+            print("Comparing hash for video: " + k)
+            hash1 = self.test_hashes[k]
+            hash2 = self.sample_hashes[k]
+            assert len(hash1) == len(hash2)
+            for h1, h2 in zip(hash1, hash2):
```
Dcallies, line 47: blocking: This is a bit too loose for me. We are relying on the sorted(list) of these files to exactly line up without checking if they actually line up.

Instead, can we establish a convention that we always expect a corresponding named file from the sample hashes, and assert it is there as part of the test?

I.e. instead of two for loops above, only have one for loop that detects one of the files, and then looks up the other by name. We then end up with a dict of name => (sample, test)

Dcallies (changes requested):
So close to working python package!

Initial readthough. Worried your test has some ways to fail silently that makes it look like it's passing.

vpdq/python/vpdq.pyx
```diff
@@ -0,0 +1,131 @@
...
+    cdef vector[vpdqFeature] vpdq_hash;
+    if downsample_width == 0:
+        vid = cv2.VideoCapture(input_video_filename)
+        downsample_width = vid.get(cv2.CAP_PROP_FRAME_WIDTH)
+
+    if downsample_height == 0:
+        vid = cv2.VideoCapture(input_video_filename)
+        downsample_height = vid.get(cv2.CAP_PROP_FRAME_HEIGHT)
+
+    rt = hashVideoFile(
```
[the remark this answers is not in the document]
tigerchensh: I am adding check to see if the rt is True, otherwise will raise an exception in python along with error messages from CPP implementation.

vpdq/python/vpdq.pyx
```diff
@@ -0,0 +1,131 @@
...
+        seconds_per_hash,
+        downsample_width,
+        downsample_height,
+        "vpdqPY"
+        )
+
+    hashes= [vpdq_feature(hash.quality,
+                         hash.frameNumber,
+                         hash.pdqHash)  for hash in vpdq_hash]
+    return hashes
```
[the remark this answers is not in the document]
tigerchensh: No, the black does not work on pyx (cython file).

vpdq/python/test_vpdq_hash.py
tigerchensh: The two loops are checking the hashes inside each VPDQ file, look up stage is actually:
```
for k in self.sample_hashes.keys():
            print("Comparing hash for video: " + k)
            hash1 = self.test_hashes[k]
            hash2 = self.sample_hashes[k]
            
```
I will add test to assert them not be none.

vpdq/python/test_vpdq_hash.py
```diff
@@ -0,0 +1,50 @@
...
+SECOND_PER_HASH = 1
+DISTANCE_TOLERANCE = 10
+QUALITY_TOLERANCE = 50
+
+
+class TestVPDQHash:
+    def test_VPDQ_hash(self):
+        d = os.path.dirname(os.path.dirname(os.getcwd()))
+        hash_folder = pathlib.Path(d + SAMPLE_HASH_FOLDER)
+        video_folder = pathlib.Path(d + SAMPLE_VIDEOS)
```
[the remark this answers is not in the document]
tigerchensh: Will use Path lib for all future python IO

tigerchensh:
Thanks for the review, I have made those changes and manually reformat pyx file.

vpdq/python/vpdq.pyx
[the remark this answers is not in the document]
Dcallies: Hmm, I found https://github.com/psf/black/issues/359 but it's still WIP

After you've checked it in, I'll try pasting it into a new .py file, formatting that, and pasting it back in

vpdq/python/test_vpdq_hash.py
```diff
@@ -1,48 +1,67 @@
...
+    def test_compare_hashes(self):
+        """This regression test is creating hashes from sample videos and compare them with the provided hashes line by line.
+        Two VPDQ features are considered the same if each line of the hashes are within DISTANCE_TOLERANCE.
+        For hashes that have a quality lower than QUALITY_TOLERANCE, the test will skip them for comoparing.
+        """
+        d = Path.cwd().parent.parent
+        hash_folder = d / SAMPLE_HASH_FOLDER
+        video_folder = d / SAMPLE_VIDEOS
         self.test_hashes = {}
         self.sample_hashes = {}
```
Dcallies, line 43: nit: Why make these instance members?
```suggestion
        test_hashes = {}
        sample_hashes = {}
```

vpdq/setup.cfg
```diff
@@ -0,0 +1,4 @@
+[metadata]
```
Dcallies, line 1: What's the advantage of splitting the configuration between setup.cfg and setup.py vs just putting it all in setup.py, or all in setup.cfg and using the dummy setup()?

vpdq/setup.py
```diff
@@ -25,7 +25,7 @@ def run(self):
 
 EXTENSIONS = [
     Extension(
-        'python.vpdq',
+        'vpdq',
```
Dcallies, line 28: Can you run `black` on this file?

Dcallies (commented):
Discussed offline, we're good to go, with some changes to the setup() stuff (merge to just setup.py)!

Dcallies (approved):
shipit

PR #1039, opened 2022-06-30
[pytx] Rename state file to .pickle

Dcallies:
Summary
---------

It's not .json, so let's rename it.

This is a backwards incompatible change! It will make a new version of the CLI not able to find the old files, and will essentially reset fetch, but will not require intervention (such as manually clearing the directory). 

Test Plan
---------

py.test


BarrettOlson (approved):
LGTM

PR #1040, opened 2022-06-30
[VPDQ] Add standalone function for vpdq-python import

tigerchensh:
Summary
---------

Add a standalone function for hashToHex in PDQ CPP implementation. So, VPDQ python binding can use this function instead of NumPy vector translation. 

Test Plan
---------
Won't affect pdq CPP. Test correctness it in python-binding PR.


BarrettOlson (approved):
LGTM

PR #1044, opened 2022-07-03
[pytx] Hide mypy errors that only happen to me apparently

Dcallies:
Summary
---------

I get these in my local checkout:
```
threatexchange/fetcher/apis/file_api.py:98: error: "BinaryIO" has no attribute "read1"
threatexchange/fetcher/apis/file_api.py:100: error: Incompatible types in assignment (expression has type "TextIO", variable has type "BinaryIO")
threatexchange/fetcher/apis/file_api.py:102: error: Argument 1 to "write" of "IO" has incompatible type "str"; expected "bytes"
threatexchange/cli/tests/e2e_test_helper.py:65: error: Need type annotation for "ex"
threatexchange/cli/tests/e2e_test_helper.py:65: error: Argument 1 to "raises" has incompatible type "Tuple[Type[CommandError], Type[E2ETestSystemExit]]"; expected "Union[Type[<nothing>], Tuple[Type[<nothing>], ...]]"
Found 5 errors in 2 files (checked 90 source files)
```

However, on server they are fine. This only happened on my new devcontainer checkout, and after spending about 15 minutes comparing the workflow to what I'm doing locally, decided it's not worth the effort and just ignored it

Test Plan
---------

`python -m mypy threatexchange` and it's clean

### DIFF
--- a/python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
+++ b/python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
@@ -62,7 +62,7 @@ class ThreatExchangeCLIE2eHelper:
     def assert_cli_usage_error(
         self, args: t.Iterable[str], msg_regex: str = None
     ) -> None:
-        with pytest.raises((CommandError, E2ETestSystemExit), match=msg_regex) as ex:
+        with pytest.raises((CommandError, E2ETestSystemExit), match=msg_regex) as ex:  # type: ignore
             self.cli_call(*args)
         exception = t.cast(t.Union[CommandError, E2ETestSystemExit], ex.value)
         assert exception.returncode == 2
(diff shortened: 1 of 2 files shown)


python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -62,7 +62,7 @@ def assert_cli_output(
     def assert_cli_usage_error(
         self, args: t.Iterable[str], msg_regex: str = None
```
BarrettOlson, line 63: nit/ignorable: should `msg_regex` be `t.Optional[str]`

BarrettOlson (approved):
Yeah, as both of these errors look like bugs (and don't show up for me)

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
Dcallies: Seems like it, it used to be accepted by previous versions. Will try and grab in a followup

PR #1045, opened 2022-07-03
[pytx][cli] label --help rewrite

Dcallies:
Summary
---------

Continuing onward, rewrite the help for the label command. This is extra challenging because this command basically doesn't work yet! 

Also some half-hearted, un-tested cleanups of the command itself.

I actually considered hiding this command instead, and saving it for a future version. What do you think?

Test Plan
---------

```
$ threatexchange label --help
usage: threatexchange label [-h] [--labels CSV | --seen | --false-positive | --true-positive] [--only-signals ONLY_SIGNALS [ONLY_SIGNALS ...] | --as-hash SIGNAL_TYPE] collab content_type ...

    [WIP] Label signals and content for sharing.

    Warning! This command is still under construction, and is not yet stable.
    Please open an issue at https://github.com/facebook/ThreatExchange/issues
    if you need the development team to prioritize stabilitizing it.

    There are three main types of labeling:

    1. Seen
       Marking that you've observed a match, which can help others prioritize
       review, or track cross-platform spread.
    2. True Positive / False Positive
       After you've confirmed the results of a match, contributing what that
       result is can help others priotize signals with more precision.
    3. Upload
       If you have your own curated signals, sharing them with others can
       help them find matches, and give them opportunities to label your
       signals.

    Examples:
    ```
    $ threatexchange label "Sample Collab" text -l example,foo -- Some text I'm labeling
    ```
    

positional arguments:
  collab                The name of the collaboration
  content_type          the type of what you are labeling
  files                 list of files or -- to interpret remainder as a string

optional arguments:
  --labels CSV, -l CSV  labels to apply to item
  --seen                mark you've seen this item
  --false-positive      mark that this doesn't belong in the collaboration
  --true-positive       mark that this does belong in the collaboration
  --only-signals ONLY_SIGNALS [ONLY_SIGNALS ...], -S ONLY_SIGNALS [ONLY_SIGNALS ...]
                        limit to this signal type
  --as-hash SIGNAL_TYPE, -H SIGNAL_TYPE
                        interpret input as a hash of this type
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/label_cmd.py
+++ b/python-threatexchange/threatexchange/cli/label_cmd.py
@@ -1,8 +1,10 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 from argparse import ArgumentParser, ArgumentTypeError
-import logging
+import argparse
+import pathlib
 import typing as t
+from threatexchange.cli.helpers import FlexFilesInputAction
 from threatexchange.fetcher.fetch_state import SignalOpinion, SignalOpinionCategory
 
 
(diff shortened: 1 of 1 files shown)


sahanapeters (approved):
I don't have a strong opinion as in I won't argue passionately on whether to hide or keep it for future version.
I think it's fine to keep it in since it's a warning that it is not stable.

PR #1046, opened 2022-07-03
[pytx][cli] match --help rewrite

Dcallies:
Summary
---------

Continuing with rewriting all the helps, now do match.

Test Plan
---------

```
$ threatexchange match --help
usage: threatexchange match [-h] [--only-signal ONLY_SIGNAL] [--hashes] [--show-false-positives] [--hide-disputed] content_type ...

    Match content to fetched signals

    Runs the given content through applicable SignalTypes, and compare it
    with previously fetched signals stored in the index files. Only
    SignalTypes supported by the CLI and your extensions can be matched,
    even if you can fetch them. Any matches will be printed to screen.

    # Input
    You can pass in data to the command in a few different ways:
    ```
    # As an input file that contains one signal
    $ threatexchange match photo my_photo.jpg

    # As stdin
    $ echo This is my cool text | threatexchange match text -

    # Inline
    $ threatexchange match text -- This is my cool text
    ```

    # Output
    The output of this command is in the following format:

    <signal type> - (<Collab Name>) <opinion category> <label1>,<label2>,...

    The category is key to understanding what you might want to do with a match.
    Here's an explanation of the categories. Opinion categories:
    * TRUE_POSITIVE:
      All contributors of this signal believe that matching content should fit
      the collaboration
    * WORTH_INVESTIGATING:
      This content needs manual investigation to confirm or fanout to find the
      content that fits the collaboration
    * DISPUTED:
      Some members have said that this signal can be used to find content
      that fits the collaboration, but others have said it matches content
      that does not belong in the collaboration.
    * FALSE_POSITIVE:
      Members have said content that matches does not belong in the
      collaboration, or that this is information content that should not
      trigger normal side-effects (i.e. human review).
    

positional arguments:
  content_type          what kind of content to match
  files                 list of files or -- to interpret remainder as a string

optional arguments:
  --only-signal ONLY_SIGNAL, -S ONLY_SIGNAL
                        limit to this signal type
  --hashes, -H          force input to be interpreted as signals for the given signal type
  --show-false-positives
                        show matches even if you've marked them false_positive
  --hide-disputed       hide matches if someone has disputed them
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/match_cmd.py
+++ b/python-threatexchange/threatexchange/cli/match_cmd.py
@@ -8,7 +8,6 @@ Match command for parsing simple data sources against the dataset.
 import argparse
 import logging
 import pathlib
-import sys
 import typing as t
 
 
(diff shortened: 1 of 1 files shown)


sahanapeters (approved):
👍

PR #1047, opened 2022-07-03
[pytx][cli] Dash for stdin for `hash`, `match` and friends

Dcallies:
Summary
---------

Folks found the empty argument behavior confusing, and on reflection, I find it confusing too, Instead use `-`. Also force it to be considered required. 

Test Plan
---------

Test on cli, also add regression test.


sahanapeters (approved):
👍

PR #1048, opened 2022-07-05
[VPDQ] Make ffmpeg path optional

tigerchensh:
Summary
---------

Invofe ffmpeg by using $PATH 'ffmpeg' in CPP implementation instead of binary's address. So, the ffmpeg's path is optional if it is installed properly.

Test Plan
---------

Run through the regression test without ffmpeg path argument.
python3 regtest.py -d ../ThreatExchange/vpdq/output-hashes -i ../ThreatExchange/tmk/sample-videos

### DIFF
--- a/vpdq/cpp/bin/vpdq-hash-video.cpp
+++ b/vpdq/cpp/bin/vpdq-hash-video.cpp
@@ -9,11 +9,11 @@ static void usage(char* argv0, int rc) {
   FILE* fp = (rc == 0) ? stdout : stderr;
   fprintf(fp, "Usage: %s [options]\n", argv0);
   fprintf(fp, "Required:\n");
-  fprintf(fp, "-f|--ffmpeg-path ...\n");
   fprintf(fp, "-i|--input-video-file-name ...\n");
   fprintf(fp, "-o|--output-hash-file-name ...\n");
   fprintf(fp, "-r|--seconds-per-hash ...:Must be a positive integer\n");
   fprintf(fp, "Options:\n");
+  fprintf(fp, "-f|--ffmpeg-path: Specific ffmpeg's address you want to use\n");
   fprintf(fp, "-v|--verbose: Show all hash matching information\n");
   fprintf(
       fp,
@@ -71,7 +71,7 @@ string stripExtension(const string& path, const string& delimiter) {
 int main(int argc, char** argv) {
   int argi = 1;
   bool verbose = false;
-  string ffmpegPath = "";
+  string ffmpegPath = "ffmpeg";
   string inputVideoFileName = "";
   string outputHashFileName = "";
   string outputDirectory = "";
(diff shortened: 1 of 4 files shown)


vpdq/cpp/bin/vpdq-hash-video.cpp
```diff
@@ -9,11 +9,11 @@ static void usage(char* argv0, int rc) {
   FILE* fp = (rc == 0) ? stdout : stderr;
   fprintf(fp, "Usage: %s [options]\n", argv0);
   fprintf(fp, "Required:\n");
-  fprintf(fp, "-f|--ffmpeg-path ...\n");
   fprintf(fp, "-i|--input-video-file-name ...\n");
   fprintf(fp, "-o|--output-hash-file-name ...\n");
   fprintf(fp, "-r|--seconds-per-hash ...:Must be a positive integer\n");
   fprintf(fp, "Options:\n");
+  fprintf(fp, "-f|--ffmpeg-path: Specific ffmpeg's address you want to use\n");
```
BarrettOlson, line 16: nit: wording "path to ffmpeg" instead of "ffmpeg's address"

BarrettOlson (changes requested):
Overall looks good. The one blocking concern I have is around error handling.

Can you add a call to `regtest.py` to the CI?

Also I think it is worth checking the error message when `ffmpeg` cannot be found on the path. (you should be able to check this by hiding/overriding the path before running the test locally).

If we error isn't clear/helpful we should add a check for catching this at the start and print out the issue clearly.

tigerchensh:
The error message will be "sh: ffmpeg: command not found" for no ffmpeg installed.

vpdq/cpp/regtest.py
```diff
@@ -8,7 +8,13 @@ def get_argparse() -> argparse.ArgumentParser:
     ap = argparse.ArgumentParser(
         description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
     )
-    ap.add_argument("-f", "--ffmpegPath", metavar="FFMPEG_PATH", help="ffmpeg path")
+    ap.add_argument(
+        "-f",
+        "--ffmpegPath",
+        metavar="FFMPEG_PATH",
```
Dcallies, line 14: (nit, ignorable): this is very close to the default, might be fine to ignore it.

vpdq/cpp/regtest.py
```diff
@@ -8,7 +8,13 @@ def get_argparse() -> argparse.ArgumentParser:
     ap = argparse.ArgumentParser(
         description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
```
Dcallies, line 9: Any reason to use RawDescriptionHelpFormatter? This is usually if you are doing custom things with indenting

tigerchensh: Research the formatter_class a little bit, decide to remove it.

PR #1049, opened 2022-07-07
[pytx][cli] Handle demo dataset empty, settings helpers

Dcallies:
Summary
---------

If you call dataset as your first command, instead of doing nothing, fetch first.

Test Plan
---------

```
$ rm -r ~/.threatexchange
$ tx fetch --clear
Looks like you haven't set up a collaboration config, so using the sample one against sample data
Clearing fetched state
dcallies@eecd010a7c71:/workspaces/devcontainer-ThreatExchange/python-threatexchange$ tx dataset
You haven't fetched any state, so we'll call `fetch` for you!
Looks like you haven't set up a collaboration config, so using the sample one against sample data
[sample] Sample Signals - Up to date
Rebuilding match indices...
Building index for trend_query with 1 signals...
Index for trend_query ready
Building index for video_md5 with 2 signals...
Index for video_md5 ready
Building index for raw_text with 3 signals...
Index for raw_text ready
Building index for pdq with 138 signals...
Index for pdq ready
Building index for url_md5 with 1 signals...
Index for url_md5 ready
Building index for url with 1 signals...
Index for url ready
pdq: 138
raw_text: 3
video_md5: 2
trend_query: 1
url: 1
url_md5: 1

$ threatexchange dataset
Looks like you haven't set up a collaboration config, so using the sample one against sample data
pdq: 138
raw_text: 3
video_md5: 2
trend_query: 1
url: 1
url_md5: 1
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -172,6 +172,48 @@ class CliState(collab_config.CollaborationConfigStoreBase):
         self.path_for_collab_config(collab).unlink(missing_ok=True)
 
 
+@dataclass
+class _SignalExchangeAccessor:
+    """Convenience wrapper for operations on the SignalExchangeAPI"""
+
+    _parent: "CLISettings"
+
+    def get_all(self) -> t.ValuesView[SignalExchangeAPI]:
+        return self._parent._mapping.fetcher.fetchers_by_name.values()
+
+    def get_for_collab(
+        self, collab: collab_config.CollaborationConfigBase
+    ) -> SignalExchangeAPI:
+        return self._parent._mapping.fetcher.fetchers_by_name[collab.api]
+
+    def __iter__(self) -> t.Iterator[SignalExchangeAPI]:
+        yield from self.get_all()
+
+
+@dataclass
+class _FetchStoreAccessor:
+    """Convenience wrapper for operations on the state"""
+
+    _parent: "CLISettings"
+
+    def empty(self) -> bool:
+        """Return the collabs with stored state"""
+        collabs = self._parent.get_all_collabs()
+        return not any(
+            collab for collab in collabs if self.get_for_collab(collab).exists(collab)
+        )
+
+    def get_for_api(self, api: t.Type[SignalExchangeAPI]) -> CliSimpleState:
+        return CliSimpleState(api, self._parent._state.dir_for_fetched_state(api))
+
+    def get_for_collab(
+        self, collab: collab_config.CollaborationConfigBase
+    ) -> CliSimpleState:
+        return self.get_for_api(
+            self._parent._mapping.fetcher.fetchers_by_name[collab.api].__class__
+        )
+
+
 class CLISettings:
     """
     A God object for all miscellanious persisted state to make the CLI work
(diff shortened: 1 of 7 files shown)


python-threatexchange/threatexchange/cli/label_cmd.py
```diff
@@ -157,7 +157,7 @@ def execute(self, settings: CLISettings) -> None:
         self.action(settings)
 
     def execute_upload(self, settings: CLISettings) -> None:
-        api = settings.get_api_for_collab(self.collab)
+        api = settings.apis.get_for_collab(self.collab)
         # signal_types = self.only_signals or settings.get_signal_types_for_content(
```
BarrettOlson, line 161: (ignorable) nit: add ToDo on why this comment is still here or remove?

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -214,33 +258,13 @@ def get_signal_types_for_content(
...
-        )
-
     @property
     def in_demo_mode(self) -> bool:
         """Has no live collabs"""
         return not self._state.get_all_collabs()
 
     def get_all_collabs(
-        self, *, default_to_sample: bool = False
+        self, *, default_to_sample: bool = True
```
BarrettOlson, line 267: (nonblocking) : I found tracing through the calls of where `default_to_sample` is used a tad confusing. 
I think it would be clearer if you either 1) add a docstring for the methods? or 2) make the argument even more verbose something like`default_to_using_only_sample_if_demo` 

unsure: one thing that might make things clearer is if you switch the method to default to false. e.g. the expected usage of this method is production not demo and where the demo state calls occur use the flag.

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -172,6 +172,48 @@ def delete_collab(self, collab: collab_config.CollaborationConfigBase) -> None:
         self.path_for_collab_config(collab).unlink(missing_ok=True)
 
 
+@dataclass
+class _SignalExchangeAccessor:
+    """Convenience wrapper for operations on the SignalExchangeAPI"""
+
+    _parent: "CLISettings"
```
BarrettOlson, line 179: question unrelated to review: is there a name for this pattern python? can you do something like this without needed the quotes and is mypy smart enough to figure out this is a type/class when used like this?

BarrettOlson (approved):
LGTM! 

Thought beyond the scope of this PR: We do print "Looks like you haven't set up a collaboration config, so using the sample one against sample data" quite a lot, is it worth either adjusting when this is printed or at least give a direction on how to "set up" and/or to suppress the warning in the warning itself?

python-threatexchange/threatexchange/cli/cli_config.py
Dcallies: This is a forward reference: https://peps.python.org/pep-0484/#forward-references

python-threatexchange/threatexchange/cli/label_cmd.py
Dcallies: The comment is in the class header - the whole command is still under construction

python-threatexchange/threatexchange/cli/cli_config.py
Dcallies: Not sure more verbose is more clear. When I changed the flag I considered renaming the variable. I plan on continuing a previous PR's switching from settings.get_y() => settings.y.get() for better grouping, and can make a stab at renaming then. 

Re: setting back to default to false - The more common expected behavior is that you want to use the sample setting, there's probably only one case where you don't (list which configs you've configured). Nasty hard-to-detect bugs will be created if you forget to set the flag to true, so it seems better to default to the more safe/expected behavior.

PR #1050, opened 2022-07-07
[pytx][cli] Consistent behavior for config list

Dcallies:
Summary
---------

Some lists were --list, some were 'list'. Now they are all ?list. Also noticed that we don't have an equivalent for signals, but mebbe one day.

Bonus changes:
* Now does fully qualified name (heads up @schatten)

Test Plan
---------
new unittest


BarrettOlson (approved):
LGTM

PR #1052, opened 2022-07-07
[pytx][lib] All Fetches are now keyed

Dcallies:
Summary
---------

As discussed offline with @schatten, we now enforce that all updates can be partitioned by key, which allows for naive solutions that don't need to merge the dataset in memory, since you can always partition by key

Test Plan
---------

py.test

Also did one small ThreatExchange config

```
$ tx config api fb_threat_exchange -L
$ tx config api fb_threat_exchange -I 1144372816354585
$ tx fetch
$ tx dataset -P
tx match url -- https://www.instagram.com/dcallies
url - (<omitted>) WORTH_INVESTIGATING media_priority_test
```


schatten (approved):
This makes sense!

PR #1053, opened 2022-07-07
[VPDQ] Add time stamp in result

tigerchensh:
Summary
---------

Add timeStamp in VPDQFeature and changes the IO.
Calculate the timeStamp outside the hashing loop because video header may not contain frame counts. So, will first place a 0 for timeStamp and then calculate the timeStamp based on the frameNumber

Test Plan
---------

Run through VPDQ regression test

### DIFF
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -29,7 +29,7 @@ bool hashVideoFile(
     const int secondsPerHash,
     const int width,
     const int height,
-    const double duration_in_sec,
+    const double durationInSec,
     const char* argv0) {
   stringstream ss;
 
@@ -104,7 +104,7 @@ bool hashVideoFile(
           (int)fread_rc);
     }
   }
-  double secPerFrame = (double)duration_in_sec / fno;
+  double secPerFrame = (double)durationInSec / fno;
   for (auto& frameFeature : pdqHashes) {
     frameFeature.timeStamp = frameFeature.frameNumber * secPerFrame;
   }
--- a/vpdq/cpp/hashing/filehasher.h
+++ b/vpdq/cpp/hashing/filehasher.h
@@ -21,7 +21,7 @@ namespace hashing {
  * @param secondsPerHash The time period of picking frames in vpdq
  * @param width Specified width of the input video
  * @param height Specified height of the input video
- * @param duration Duration of the input video in seconds
+ * @param durationInSec Duration of the input video in seconds
  * @param downsampleFrameDimension The down-scale dimensions for vpdq
  * @param programName The name of executable program which invokes the function
  *
@@ -36,7 +36,7 @@ bool hashVideoFile(
     const int secondsPerHash,
     const int width,
     const int height,
-    const double duration_in_sec,
+    const double durationInSec,
     const char* programName);
 } // namespace hashing
 } // namespace vpdq
--- a/vpdq/cpp/io/vpdqio.cpp
+++ b/vpdq/cpp/io/vpdqio.cpp
@@ -160,7 +160,7 @@ bool readVideoResolution(
 
 bool readVideoDuration(
     const string& inputVideoFileName,
-    double& duration_in_sec,
+    double& durationInSec,
     const char* programName) {
   AVFormatContext* pFormatCtx = avformat_alloc_context();
   int rc =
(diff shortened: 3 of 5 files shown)


vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -102,6 +104,15 @@ bool hashVideoFile(
           (int)fread_rc);
     }
   }
+  double secPerFrame = (double)duration / fno;
+  for (size_t i = 0; i < pdqHashes.size(); i++) {
+    vpdqFeature cur = pdqHashes[i];
+    pdqHashes[i] = (vpdqFeature){
```
BarrettOlson, line 110: hmm, weird that you have to completely create a new object here.

vpdq/cpp/io/vpdqio.cpp
```diff
@@ -149,6 +153,23 @@ bool readVideoResolution(
...
+      avformat_open_input(&pFormatCtx, inputVideoFileName.c_str(), NULL, NULL);
+  if (rc != 0) {
+    fprintf(
+        stderr,
+        "%s: could not open video \"%s\".\n",
+        programName,
+        inputVideoFileName.c_str());
+    return false;
+  }
+  duration = pFormatCtx->duration / 1000000;
```
BarrettOlson, line 170: can you have 1000000 be a const value that describes the unit conversion it is doing?

vpdq/cpp/io/vpdqio.cpp
```diff
@@ -24,7 +25,7 @@ using namespace facebook::pdq::io;
 namespace facebook {
 namespace vpdq {
 namespace io {
-
+int PRECESION = 3;
```
BarrettOlson, line 28: nit: spelling and variable description e.g.
`const int TIMESTAMP_OUTPUT_PRECISION = 3`

vpdq/cpp/io/vpdqio.cpp
```diff
@@ -50,7 +51,7 @@ bool loadHashesFromFileOrDie(
       tmp.push_back(substr);
     }
 
-    if (tmp.size() != 3) {
+    if (tmp.size() != 4) {
```
BarrettOlson, line 54: nit: (missed this in an old code review) something more descriptive than tmp, like `row` or `frame_values` might be better/clear

BarrettOlson (changes requested):
Nice!

blocking: I think we have a little too much truncation by using and `int` for duration as the output for `readVideoDuration`

vpdq/cpp/hashing/filehasher.cpp
BarrettOlson: example does: `pdqHashes[i].timeStamp = pdqHashes[i].frameNumber * secPerFrame;` not work for you?

tigerchensh: It works. The previous issue is caused by using 
```
for (auto t:pdqHashes):
     t.timeStamp = t.frameNumber * secPerFrame;
```
It did not change the value.

BarrettOlson: even simpler:
```  
for (auto& frame_feature : pdqHashes) {
    frame_feature.timeStamp = frame_feature.frameNumber * secPerFrame;
  }
```

vpdq/cpp/hashing/filehasher.h
```diff
@@ -19,6 +19,9 @@ namespace hashing {
  * @param ffmpegPath Ffmpeg's path
  * @param verbose If produce detailed output for diagnostic purposes
  * @param secondsPerHash The time period of picking frames in vpdq
+ * @param width Specified width of the input video
+ * @param height Specified height of the input video
+ * @param duration Duration of the input video in seconds
```
BarrettOlson, line 24: nit: update to changed name

vpdq/cpp/hashing/filehasher.h
```diff
@@ -33,6 +36,7 @@ bool hashVideoFile(
     const int secondsPerHash,
     const int width,
     const int height,
+    const double duration_in_sec,
```
BarrettOlson, line 39: nit: format inconsistency with rest of arguments `lowerCamelCase` vs `lower_snake_case`

BarrettOlson (approved):
Cool cool cool,
few minor style nits but otherwise LGTM

PR #1056, opened 2022-07-08
[vpdq] Empty init file confuses git

Dcallies:
Summary
---------

This empty file apparently creates an un-checkoutable file that always looks modified to my index, and no amounts of resets, unstages, or checkouts seems to fix it.

Making it not a 0 byte file seems to work though.

Test Plan
---------

No longer shows as a missing file


tigerchensh (approved):
LGTM! Will take care of that in the future.

PR #1057, opened 2022-07-08
[VPDQ] Add timeStamp and make ffmpeg optional in Python-binding

tigerchensh:
Summary
---------

Adjust VPDQ python binding to match the CPP implementation.
Add timeStamp in IO and vpdqFeature.
Make ffmpeg and seconds_per_hash have default values.

Test Plan
---------
Install vpdq locally by
```
python3 setup.py bdist_wheel
cd dist
pip3 install vpdq-0.1.0-cp39-cp39-macosx_12_0_arm64.whl
```
and run test
```
pytest test_vpdq_hash.py
```


vpdq/README.md
```diff
@@ -50,6 +50,7 @@ There are four inputs to the comparison algorithm, which determines if two video
 The query video’s frame PDQ hashes Q
 The comparison video’s frame PDQ hashes C
 The PDQ match distance D (example: 31)
+The PDQ quality filter tolerance F (example: 80), if either hash is below this quality level then they will not be compared
```
Dcallies, line 53: blocking: I know I am a broken record on this point, but why are we suggestion a different value than what we are using in production? We should use the same value as we are going to use in the library, which I believe should be 50.

vpdq/python/tests/test_util.py
```diff
@@ -33,5 +33,5 @@ def output_hash_to_file(output_hash_filename, hashes):
 
     with open(output_hash_filename, "w") as file:
         for cur in hashes:
-            file.write(f"{cur.frame_number},{cur.quality},{hash_to_hex(cur.hash['w'])}")
+            file.write(f"{cur.frame_number},{cur.quality},{hash_to_hex(cur.hash['w'])},{cur.time_stamp}")
```
Dcallies, line 36: These floats might be super long. How many digits are useful? Surely no more than 3 (milliseconds)?

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -36,7 +34,7 @@ def test_compare_hashes(self):
         Two VPDQ features are considered the same if each line of the hashes are within DISTANCE_TOLERANCE.
         For hashes that have a quality lower than QUALITY_TOLERANCE, the test will skip them for comoparing.
         """
-        d = Path.cwd().parent.parent
+        d = Path.cwd().parent.parent.parent
```
Dcallies, line 37: ???

blocking: I think your root problem is that you shouldn't be using cwd() here, you should be using the directory that the file is in. There should be a magic python constant for this (maybe __MODULE__ or something like that?)

vpdq/README.md
```diff
@@ -12,16 +12,16 @@ Both TMK and vPDQ are backed by PDQ, and so inherit both PDQ’s strengths and w
...
-| ------------- | ------------- |
-| 1  | face000...  |
-| 2  | face000...  |
-| 3  | face011...  |
-| 4  | face111...  |
-| ... | ...  |
-| 29999 | 88784444 |
-| 30000 | 88884444 |
+We can annotate these hashes with their frame number, quality(0-100 which measures gradients/features,from least featureful to most featureful) and time stamp(sec). So for a 5 minute video at 1 frame/sec, we might have:
+| Frame | Quality | PDQ Hash | Time Stamp |
```
Dcallies, line 16: (nit): I usually see timestamp is one word: https://www.computerhope.com/jargon/t/timestam.htm#write

(unsure) You may want to clarify somewhere that this is seconds time, but I'm not sure if timestamp is well understood in this space.

vpdq/README.md
```diff
@@ -12,16 +12,16 @@ Both TMK and vPDQ are backed by PDQ, and so inherit both PDQ’s strengths and w
...
+| ------------- | ------------- | ------------- | ------------- |
+| 1  | 100 | face000...  | 0.000  |
+| 2  | 99  | face000...  | 0.033  |
+| 3  | 94  | face011...  | 0.067  |
+| 4  | 97  | face111...  | 1.000  |
+| ... | ...  |...  |...  |
+| 29999 | 89  | 88784444 | 989.933 |
+| 30000 | 92  | 88884444 | 989.967 |
```
Dcallies, line 24: blocking: These two numbers are not valid PDQ hashes, Add an ellipses (...) to show there is more

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -7,8 +7,6 @@
 
 SAMPLE_HASH_FOLDER = Path("vpdq/sample-hashes")
 SAMPLE_VIDEOS = Path("tmk/sample-videos")
-FFMPEG = "/usr/bin/ffmpeg"  # please customize according to your installation (but do not commit)
-SECOND_PER_HASH = 1
 DISTANCE_TOLERANCE = 10
 QUALITY_TOLERANCE = 50
```
Dcallies, line 11: Here is our quality tolerance being 50

Dcallies (changes requested):
Some blocking requests inline

vpdq/README.md
tigerchensh: Yes, it should be 50 instead of 80. Thanks for pointing it out.

vpdq/python/tests/test_util.py
tigerchensh: Forget to do this in python, I thought it was done in CPP. Just found out it was not until the IO.

vpdq/python/tests/test_vpdq_hash.py
tigerchensh: Change to Path(__file__).parents[3]

vpdq/README.md
tigerchensh: It's from the original readme, thanks for pointing that out. Haven't noticed that.

vpdq/README.md
```diff
@@ -12,16 +12,16 @@ Both TMK and vPDQ are backed by PDQ, and so inherit both PDQ’s strengths and w
 ## Producing a Hash
 The algorithm for producing the “hash” is simple: given a video, convert it into a sequence of frame images at some interval (for example, 1 frame/second). For each frame image, use the PDQ hashing algorithm on each.
 
-We can annotate these hashes with their frame number, quality(0-100 which measures gradients/features,from least featureful to most featureful) and time stamp(sec). So for a 5 minute video at 1 frame/sec, we might have:
-| Frame | Quality | PDQ Hash | Time Stamp |
+We can annotate these hashes with their frame number, quality(0-100 which measures gradients/features,from least featureful to most featureful) and timestamp(sec). So for a 5 minute video at 1 frame/sec, we might have:
```
Dcallies, line 15: works for me!

vpdq/README.md
Dcallies: Probably my fault then :P

Dcallies (approved):
shipit

PR #1068, opened 2022-07-11
[VPDQ] Use framesPerSecond to calculate timestamp

tigerchensh:
Summary
---------

Replace duration with FPS to calculate the timestamp.
Fix the wrong implementation of the secondsPerHash. The previous implementation is actually framesPerHash.

Test Plan
---------

Run through CPP regtest, python regtest and check timestamp with the internal VPDQ.


vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -63,7 +63,12 @@ bool hashVideoFile(
   int numRGBTriples = height * width;
   int fno = 0;
   unique_ptr<uint8_t[]> rawFrameBuffer(new uint8_t[numRGBTriples * 3]);
-
+  int frameMod = secondsPerHash * framesPerSec;
```
BarrettOlson, line 66: nit: Add comment that you are doing an intentional  floor operation. I think it is kind of easy to miss given the `double` type declaration  for these two vars is a a fair bit higher up in the file.

BarrettOlson (changes requested):
Like the change.
Blocking: We need to check the typing in .pyx to confirm which is correct out of float/double

Question: you mentioned this change should have better performance, do you have a way to measure that?

vpdq/python/vpdq.pyx
```diff
@@ -42,10 +42,10 @@ cdef extern from "vpdq/cpp/hashing/filehasher.h" namespace "facebook::vpdq::hash
         vector[vpdqFeature]& pdqHashes,
         string ffmpeg_path,
         bool verbose,
-        int seconds_per_hash,
+        float seconds_per_hash,
```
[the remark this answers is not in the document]
tigerchensh: The cython will do an automatic transformation from python float to CPP double. Since there is no inbuild python double, I am just using float in python and let cython do the conversion.
image

tigerchensh:
Thanks for the review. To meseaure the performance improvement, this one should be subtle. We can measure the time of hashing the example videos and compare two approaches.

vpdq/cpp/io/vpdqio.cpp
```diff
@@ -176,6 +176,51 @@ bool readVideoDuration(
   durationInSec = (double)pFormatCtx->duration / MILLISEC_IN_SEC;
```
BarrettOlson, line 176: nit: add note that `readVideoDuration` is not used atm (or remove it, but I think it could be useful eventually, say if we wanted to use it later to optionally print a set of debugging information)

vpdq/cpp/io/vpdqio.cpp
```diff
@@ -176,6 +176,51 @@ bool readVideoDuration(
...
+  }
+  if (videoStream == -1) {
+    fprintf(
+        stderr,
+        "%s: could not find video stream \"%s\".\n",
+        programName,
+        inputVideoFileName.c_str());
+    return false;
+  }
+  AVRational fr = pFormatCtx->streams[videoStream]->avg_frame_rate;
```
BarrettOlson, line 220: blocking: apart from these two lines this looks this is an exact copy of `readVideoResolution` imp. It seems like you could write a method that gets both at the same time. Is there a reason not to do that? 

(If you have to keep the utility of having methods for each for debugging you can have the existing methods call the new one and throw out the un-need result)

BarrettOlson (commented):
Question about copied code/redundancy between `readVideoFPS` and `readVideoResolution`

vpdq/cpp/io/vpdqio.cpp
tigerchensh: Thanks for the suggestion. I was trying to separate the functionality, but I would combine them together.

BarrettOlson (approved):
LGTM!

PR #1074, opened 2022-07-19
[PYTX-VPDQ] Implement VPDQ faiss index

tigerchensh:
Summary
---------

Implement VPDQ FlatHashIndex which can return the percentage match in the indexMatch as well as matched features. When adding to the VPDQ Index, it requires metadata which includes the video_id: [str or int].

Test Plan
---------
A simple test wich runs through query_raw_result and query (returns match percentage).
Will add more tests in separate PR.
python3 -m pytest test_vpdq_faiss.py


python-threatexchange/mypy.ini
```diff
@@ -22,4 +22,7 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 [mypy-vpdq.*]
+ignore_missing_imports = True
+
+[mypy-faiss.*]
```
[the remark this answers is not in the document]
tigerchensh:
```
import faiss  # type: ignore
```
The import faiss in pdq_faiss_matcher.py is using type: ignore.

python-threatexchange/threatexchange/extensions/video_vpdq/vpdq_faiss_matcher.py
```diff
@@ -0,0 +1,183 @@
...
+            queries : The VPDQ features to against the index
+            quality_tolerance : The quality tolerance of matching two frames.
+            If either frames is below this quality level then they will not be queired and added to result
+            distance_tolerance : Threshold value to use for this search. The hamming distance between the result hashes and the related query will
+            be no more than the threshold value. i.e., hamming_dist(q_i,r_i_j) <= threshold.
+
+        Returns:
+        sequence of matches per query
+            For each query provided in queries, the returned sequence will contain a sequence of matches within the index
+            that were within threshold hamming distance of that query. These matches will be (id, video_id, frame_number,
```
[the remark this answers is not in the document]
tigerchensh: The id is the faiss_id when adding the signle vpdq feature. The video id represents a video which is a list of vpdq features

tigerchensh:
Thanks for the suggestion on the design! I have removed the video_id and made dedupe and quality filter before adding to the index. The vpdq_faiss.py now is more flexible with changing to different faiss index types.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(
+        self,
+        entries: t.Iterable[t.Tuple[str, IndexT]] = (),
+        quality_threshold: int = VPDQ_CONFIDENT_QUALITY_THRESHOLD,
+    ) -> None:
+        super().__init__()
+        self.index: VPDQHashIndex = self._get_empty_index()
+        self.entries: t.List = []
+        self.idx_to_vpdq: t.List[t.Tuple[int, vpdq.VpdqFeature]] = []
```
Dcallies, line 46: This name surprised me a little bit - my best guess for alternatives:
1. faiss_id_to_entry_id
2. index_idx_to_entry_idx
(or some variation)

Also, I'm not 100% you actually need the original VpdqFeature anywhere that you would need this unless you plan on verifying with the brute force. You might be doubling your storage cost without too much benefit, but I don't know if you are relying on this for testing.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,137 @@
...
+        for hash in results:
+            match_tuples = []
+            for match in results[hash]:
+                # query_str =>  (matched_idx, distance)
+                entry_id, vpdq_match = self.idx_to_vpdq[match[0]]
+                match_tuples.append([vpdq_match, match[1], self.entries[entry_id]])
+            matches[hash] = match_tuples
+        return matches
+
+    def query_with_match_percentage_in_result(
```
Dcallies, line 97: Doesn't `query` return a VPDQMatchResult too with percentage in the result? Is there's something else that distinguishes these functions?

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -48,6 +48,19 @@ def __init__(self, distance: int, metadata: T) -> None:
...
+
+    def __init__(
+        self, query_match_percent: float, compared_match_percent: float, metadata: T
+    ) -> None:
+        # Dist is meaning less here, because VPDQ match does not return a single dist
+        IndexMatch.__init__(self, -1, metadata)
+        self.query_match_percent = query_match_percent
+        self.compared_match_percent = compared_match_percent
+
+
```
Dcallies, line 63: blocking: This is a detail of vPDQ, which is not installed by default by the library, and so sohuld live in the vPDQ directory

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -48,6 +48,19 @@ def __init__(self, distance: int, metadata: T) -> None:
...
 
+class VPDQIndexMatch(IndexMatch):
+    query_match_percent: float
+    compared_match_percent: float
+
+    def __init__(
+        self, query_match_percent: float, compared_match_percent: float, metadata: T
+    ) -> None:
+        # Dist is meaning less here, because VPDQ match does not return a single dist
+        IndexMatch.__init__(self, -1, metadata)
```
Dcallies, line 59: blocking: You should never call the superclass like this, instead always use super().

The reason for this is that in multiple inheritence, or if someone refactors in a middle layer, this will break the expected behavior of initialization, and possibly be very hard to discover

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -48,6 +48,19 @@ def __init__(self, distance: int, metadata: T) -> None:
         self.metadata = metadata
 
 
+class VPDQIndexMatch(IndexMatch):
+    query_match_percent: float
+    compared_match_percent: float
```
Dcallies, line 53: (not sure): You might be able to leave these out, and the typechecker can figure it out from reading the init, but I don't know

Dcallies (changes requested):
Hey Xiangnan, thanks for the continued work. This PR is basically the biggest chunk of functionality add, and maybe the pivotal PR of the integration, so I'm giving it a high level of scruitany. If you want to try and break it up or land some stuff incrementally so it doesn't feel like we are going back and forth on this one forever, let me know and I can help with that.

Toplevel thoughts:
* Don't use the 'bare' version of generics like t.List, t.Dict, etc. These implicitly turn into t.List[t.Any]. If you really mean Any, use that specifically, but usually you know the type. Each t.Any or #type: ignore weakens the benefits of typechecking. There's a lint rule we can add to warn on this.
* Consider helping a reader figure out which methods are an internal detail, and which ones are part of the public interface. Prefixing methods with `_` like `_private_method()` can help a reader understand that this is an implementation detail that they should not rely on externally.

Originally I wrote you something long here about the need for deduping hashes from the index first. However, it looks like faiss still works without it, but it will become a performance problem in the future: https://github.com/facebookresearch/faiss/wiki/FAQ#searching-duplicate-vectors-is-slow

Up to you if you want to try and fix it now (like I did in the image library I linked you) or just make a # todo for it

tigerchensh:
Thanks for the reviews and suggestions on the design. I have removed query_raw_result and moved query_percentage into query, since we only need percentages from VPDQIndexMatch.  I have implemented supports for duplicated hashes in VPDQIndex. Write tests for adding duplicating hashes and serialization. Will add more tests in next PR.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -0,0 +1,71 @@
...
+        prepare_vpdq_feature,
+        VPDQ_QUALITY_THRESHOLD,
+    )
+    from threatexchange.extensions.vpdq.video_vpdq import VideoVPDQSignal
+
+    from threatexchange.signal_type.index import (
+        VPDQIndexMatch,
+    )
+
+ROOTDIR = Path(__file__).parents[5]
```
Dcallies, line 27: Is this still used?

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.mark.skipif(_DISABLED, reason="vpdq not installed")
+def compare_match_result(res1, res2) -> bool:
+    return (
+        res1.compared_match_percent == res2.compared_match_percent
+        and res1.query_match_percent == res2.query_match_percent
+        and res1.metadata == res2.metadata
+        and res1.distance == res2.distance
+    )
```
Dcallies, line 71: (ignorable): You can also implement `__eq__` on the object, which will let you do `==` directly, you can look up how to do that (there some tricks)

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -0,0 +1,71 @@
...
+
+    from threatexchange.signal_type.index import (
+        VPDQIndexMatch,
+    )
+
+ROOTDIR = Path(__file__).parents[5]
+example_meta_data = {"name": "example_video"}
+
+
+@pytest.mark.skipif(_DISABLED, reason="vpdq not installed")
```
Dcallies, line 31: Check this out on how to skip all the tests at once: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module

python-threatexchange/threatexchange/extensions/vpdq/vpdq_faiss.py
```diff
@@ -0,0 +1,86 @@
...
+            For each query provided in queries, the returned sequence will contain a sequence of matches within the index
+            that were within threshold hamming distance of that query. These matches will be (idx, distance).
+            The inner sequences may be empty in the case of no hashes within the index.
+            The same VPDQ feature may also appear in more than one inner sequence if it matches multiple query hashes.
+            For example the hash "000000000000000000000000000000000000000000000000000000000000ffff" would match both
+            "00000000000000000000000000000000000000000000000000000000fffffff" and
+            "0000000000000000000000000000000000000000000000000000000000000000" for a threshold of 16. Thus it would appear in
+            the entry for both the hashes if they were both in the queries list.
+
+            eg.
```
Dcallies, line 56: nit: e.g.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_faiss.py
```diff
@@ -0,0 +1,86 @@
...
+            distances = [idx for idx in similarities[limits[i] : limits[i + 1]]]
+            result[query.hex] = list(zip(matches, distances))
+        return result
+
+    def __getstate__(self):
+        data = faiss.serialize_index_binary(self.faiss_index)
+        return data
+
+    def __setstate__(self, data):
+        self.faiss_index = faiss.deserialize_index_binary(data)
```
Dcallies, line 86: 👍

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Implementation of SignalTypeIndex abstraction for PDQ by wrapping
```
Dcallies, line 4: vPDQ

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+hashing.pdq_faiss_matcher.
```
Dcallies, line 5: nit: update

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class VPDQIndex(PickledSignalTypeIndex[IndexT]):
+    """
+    Wrapper around the vpdq faiss index lib using VPDQHashIndex
+    """
+
+    def __init__(
+        self,
+        quality_threshold: int = VPDQ_QUALITY_THRESHOLD,
```
Dcallies, line 31: 👍

(ignorable): You may want to consider forcing this to be a kwarg by insert `*,` after self

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+            if idx is None:
+                idx = len(self._unique_vpdqHex_to_idx)
+                self._unique_vpdqHex_to_idx[f.hex] = idx
+                self._index_idx_to_vpdqHex_and_entry.append((f.hex, list()))
+                unique_features.append(f)
+            self._index_idx_to_vpdqHex_and_entry[idx][1].append(entry_id)
+        if unique_features:
+            self.index.add_single_video(unique_features)
+
+    def query(self, query_hash: str) -> t.List[IndexMatch[IndexT]]:
```
Dcallies, line 59: Does the typer let you use VPDQIndexMatch here?

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+                vpdq_match, entry_list = self._index_idx_to_vpdqHex_and_entry[match[0]]
+                for entry_id in entry_list:
+                    if entry_id not in query_matched:
+                        query_matched[entry_id] = set()
+                    query_matched[entry_id].add(hash)
+
+                    if entry_id not in index_matched:
+                        index_matched[entry_id] = set()
+                    index_matched[entry_id].add(vpdq_match)
+                # Dist(-1) is meaningless here, because VPDQ match does not return a single dist
```
Dcallies, line 88: Misplaced comment

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+        for entry_id in query_matched.keys():
+            matches.append(
+                VPDQIndexMatch(
+                    -1,
```
Dcallies, line 92: Suggest using the match percentage (int, 0-100) as the placeholder here, instead of -1, which is a bit misleading.

I agree with you that the types are mismatched here. I would like to redo this to make the distance type an internal detail

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+                    len(query_matched[entry_id]) * 100 / len(features),
+                    len(index_matched[entry_id])
+                    * 100
```
Dcallies, line 95: Super dumb question - should we consider using 0-1.0 instead of 0-100?

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+                    / len(self._entry_idx_to_features_and_entires[entry_id][0]),
```
Dcallies, line 96: This is slightly hard to read, might be worth splitting out the variables here so it's clear what each one means.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
```diff
@@ -11,6 +11,10 @@
 HASH = "hash"
 TIMESTAMP = "timestamp"
 VPDQ_TIMESTAMP_PRECISION = 3
+BITS_IN_VPDQ = 256
+VPDQ_VIDEOID_TYPE = t.Union[str, int]
```
Dcallies, line 15: Is this still used?

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
```diff
@@ -11,6 +11,10 @@
 HASH = "hash"
 TIMESTAMP = "timestamp"
 VPDQ_TIMESTAMP_PRECISION = 3
+BITS_IN_VPDQ = 256
```
Dcallies, line 14: (ignorable): I think you could use BITS_IN_PDQ since this is part of the threatexchange library

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -48,6 +48,19 @@ def __init__(self, distance: int, metadata: T) -> None:
         self.metadata = metadata
 
 
+class VPDQIndexMatch(IndexMatch):
```
Dcallies, line 51: You have to pass through T here to get it type 
```suggestion
class VPDQIndexMatch(IndexMatch[T]):
```

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -0,0 +1,71 @@
...
+        VPDQ_QUALITY_THRESHOLD,
+    )
+    from threatexchange.extensions.vpdq.video_vpdq import VideoVPDQSignal
+
+    from threatexchange.signal_type.index import (
+        VPDQIndexMatch,
+    )
+
+ROOTDIR = Path(__file__).parents[5]
+example_meta_data = {"name": "example_video"}
```
Dcallies, line 28: Traditionally these are in ALL_CAPS

You can also type this as t.Mapping[str, str]  to enforce that it can't accidentally be changed.

You could also add `hash` here as well

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.mark.skipif(_DISABLED, reason="vpdq not installed")
+def test_duplicate_hashes():
+    hash = VideoVPDQSignal.get_examples()[0]
+    index = VPDQIndex.build([[hash, "video1"]])
+    index.add(hash, "video2")
+    res = index.query(hash)
+    # A complete match to itself
+    assert compare_match_result(res[0], VPDQIndexMatch(-1, 100, 100, "video1"))
+    assert compare_match_result(res[1], VPDQIndexMatch(-1, 100, 100, "video2"))
```
Dcallies, line 61: All your tests are all asserting that they are perfect matches, which means that many failure modes might be missed. I am not yet convinced that your implementation is correct as a result. Would like to see some variation here in a followup PR.

i.e.
```
pdq_dist(A, A) = 0
pdq_dist(A, B) = 31
pdq_dist(A, C-F) > 31

v1 = [A]
v2 = [A, B]
v3 = [B]
V4 = [B, C]
V5 = [B, C, D]

for x in range(1, 6):
  for y in range(x, 6):
    vpdq_dist(v{x}, v{y})
    vpdq_dist(V{y}, V{x})

```

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+        self,
+        quality_threshold: int = VPDQ_QUALITY_THRESHOLD,
+    ) -> None:
+        super().__init__()
+        self.index: VPDQHashIndex = VPDQHashIndex()
+        self._entry_idx_to_features_and_entires: t.List[
+            t.Tuple[t.List[vpdq.VpdqFeature], IndexT]
+        ] = []
+        self._index_idx_to_vpdqHex_and_entry: t.List[t.Tuple[int, t.List[int]]] = []
+        self._unique_vpdqHex_to_idx: t.Dict[str, int] = {}
```
Dcallies, line 39: Which idx? We have index_idx and entry_idx (assume entry)

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -0,0 +1,100 @@
...
+        self.quality_threshold = quality_threshold
+
+    def add(self, signal_str: str, entry: IndexT) -> None:
+        entry_id = len(self._entry_idx_to_features_and_entires)
+        features = prepare_vpdq_feature(signal_str, self.quality_threshold)
+        self._entry_idx_to_features_and_entires.append((features, entry))
+        # Use hex to represent the feature because it saves the space
```
Dcallies, line 46: There are even more compact serialization, such as bytes.

Dcallies (commented):
Didn't finish, flushing

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
tigerchensh: I don't think there is any use case for comparing two IndexMatch results except for testing. So, I think __eq__ is overkill.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
tigerchensh: It's actually  _unique_vpdqHex_to_index_idx, because each unique hex hash should only map to one index_idx in faiss.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
tigerchensh: No, mypy doesn't think return type VPDQIndexMatch matches IndexMatch[Any]

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
tigerchensh: Thanks for pointing out!

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
tigerchensh: Not anymore, removed

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
tigerchensh: What is the benefit of using 0-1.0 instead of 0-100? I feel like 0-100 has more readability and it can still have the precision as a float.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
tigerchensh: I think I will leave it as it is for easier checking if the feature(hex) already exists in _unique_vpdqHex_to_idx.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
Dcallies: When you are multiplying things together like durations you don't have to re-normalize it to a fraction again.

```
matching_sec = video_duration * match_pct
```
vs
```
matching_sec = video_duration * match_pct/100
```

The only time when 0-100 is preferred is for humans to look at., and that is what `__str__` is generally used for.

Dcallies (approved):
I have no more blocking comments, so accepting.

However, please sync with me before proceeding to the next block of work - I think we should invest significantly in testing in verification, rather than new features. Can talk to you about what I mean offline.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
tigerchensh: Yes, it is worth doing. But I will create a separate PR for making all the matching percentage to be 0-1.0

tigerchensh:
Thanks for the reviews and suggestions. Will move percetange change (0-100 to 0-1.0) and more through testing in separate prs.

PR #1077, opened 2022-07-19
[pytx][hash_cmd] FileHasher instead BytesHasher in hash cli

BarrettOlson:
Summary
---------

If a signal type class extends `signal_base.FileHasher` the hash_cmd wouldn't recognize it because it was only looking for subclasses of `BytesHasher` or `TextHasher (FileHasher is a parent of both)

both have an "is a" relationship with FileHasher so anything that inherits for BytesHasher or TextHasher should still by supported by the change.

Test Plan
---------

`threatexchange hash --signal-type video_md5 video <video-file>` still works

and if you install something like `video_vpdq` (docs WIP) it will now also work

`threatexchange hash --signal-type video_vpdq video <video-file>`

### DIFF
--- a/python-threatexchange/threatexchange/cli/hash_cmd.py
+++ b/python-threatexchange/threatexchange/cli/hash_cmd.py
@@ -13,7 +13,6 @@ from threatexchange.cli.cli_config import CLISettings
 from threatexchange.signal_type.signal_base import (
     FileHasher,
     SignalType,
-    TextHasher,
 )
 from threatexchange.cli import command_base
 from threatexchange.cli.helpers import FlexFilesInputAction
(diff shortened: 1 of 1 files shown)


sahanapeters (approved):
LGTM 👍

tigerchensh:
I think it you can fix by using hash_from_file from both TextHasher and ByteHasher while not checking any of their types.
```
hashers = [s for s in all_signal_types if issubclass(s, FileHasher)]
for file in self.files:
    for hasher in hashers:
        hash_str = s_hasher.hash_from_file(file)
        _print_hash(s_hasher, hash_str)
```
Because with each subclass, their hash_from_file is override and it utilizes the benefit of having the inheritance from FileHasher.

tigerchensh (changes requested):
I think it you can fix by using hash_from_file from both TextHasher and ByteHasher while not checking any of their types.
```
hashers = [s for s in all_signal_types if issubclass(s, FileHasher)]
for file in self.files:
    for hasher in hashers:
        hash_str = hasher.hash_from_file(file)
        _print_hash(hasher, hash_str)
```
Because within each subclass, their hash_from_file is override and it utilizes the benefit of having the inheritance from FileHasher.

BarrettOlson:
Addressed in #1078

PR #1079, opened 2022-07-20
[pytx][ncmec] esp ID filter

Dcallies:
Summary
---------

Requested by a partner using NCMEC. This was originally "only_owner_ids" which died when I removed a lot of the default common arguments for collab config. Now we bring it back in its original form for NCMEC.

Provides a way to only accept entries from certain partners. The remainder are filtered as thought they do not exist.

## Unsolved Problems
Changing this value invalidates your old data, since you may have failed to previously fetch something. I can think of a few solutions:
1. Something that detects that you changed your config, and invalidates your old state
2. Move the collab_config into the signature of Naive convert. This seems okay, but will create a burden on future implementations like HMA, and also encourages over-fetching
3. Something else

Test Plan
---------

Pending, may send to partner to test.

--help
```
$ tx config collab edit ncmec --help
usage: threatexchange config collab edit ncmec [-h] [--create] [--enable [{0,1}] | --disable] --environment [Industry,NGO,Exploitative,test_Industry,test_NGO,test_Exploitative]
                                               [--only-esp-ids int[,int[,...]]] [--json]
                                               collab_name

positional arguments:
  collab_name           the name of the collab

optional arguments:
  --create, -C          indicate you intend to create a config
  --json, -J            instead, interpret the argument as JSON and use that to edit the config

  common to all collabs

  --enable [{0,1}]      enable the config (default on create)
  --disable             disable the config

  specific to ncmec

  --environment [Industry,NGO,Exploitative,test_Industry,test_NGO,test_Exploitative]
                        which database to connect to
  --only-esp-ids int[,int[,...]]
                        Only take entries from these eletronic service provider (ESP) ids```

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
@@ -60,6 +60,12 @@ class NCMECCollabConfig(
     _NCMECCollabConfigRequiredFields,
 ):
     api: str = field(init=False, default=_API_NAME)
+    only_esp_ids: t.Set[int] = field(
+        default_factory=set,
+        metadata={
+            "help": "Only take entries from these eletronic service provider (ESP) ids"
+        },
+    )
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py [hunk 1]
```diff
@@ -60,6 +60,12 @@ class NCMECCollabConfig(
     _NCMECCollabConfigRequiredFields,
 ):
     api: str = field(init=False, default=_API_NAME)
+    only_esp_ids: t.Set[int] = field(
+        default_factory=set,
+        metadata={
+            "help": "Only take entries from these eletronic service provider (ESP) ids"
```
Dcallies, line 66: speels

schatten: Can you add some help text on how to use for multiple ESP ids? It is not immediately apparent to me whether I should do 

```
--only-esp-ids <esp_1>,<esp_2>

OR

--only-esp-ids <esp_1> --only-esp-ids <esp_2>
```

schatten (approved):
Non-blocking request for help text when using multiple esp-ids.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
Dcallies: I can add the default hint to mention it's CSV (though I could use the argparse features to allow multiple flags as you put, I think csv is more clear/consistent with other commands)

PR #1083, opened 2022-07-21
Update README.md

Dcallies:
Show how to use local files to store simple hashes

Summary
---------

Add some notes on using local file

Test Plan
---------

Ran the commands locally


BarrettOlson (approved):
LGTM

PR #1084, opened 2022-07-21
[pytx] Fix NCMEC esp-filter, add too many more features

Dcallies:
Summary
---------

It started as a fix for the fact that set types weren't being parsed correctly. It completely got away from me.

Test Plan
---------

Added unittest, also run
```
$ tx config extensions add threatexchange.cli.tests.fake_extension
$ tx config collab edit fake fake --an-int 1 --a-str a --an-enum OPTION_A --a-list 1,2,3 --a-set 3,2 -C
$ tx config collab print fake
{
  "an_int": 1,
  "a_str": "a",
  "a_list": [
    "1",
    "2",
    "3"
  ],
  "a_set": [
    2,
    3
  ],
  "an_enum": "a",
  "name": "fake",
  "api": "fake",
  "enabled": true,
  "optional": null
}
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/config_cmd.py
+++ b/python-threatexchange/threatexchange/cli/config_cmd.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 """
@@ -22,7 +21,7 @@ from threatexchange.extensions.manifest import ThreatExchangeExtensionManifest
 from threatexchange import meta as tx_meta
 from threatexchange import common
 from threatexchange.cli.cli_config import CLISettings
-from threatexchange.cli import command_base
+from threatexchange.cli import command_base, dataclass_json
 from threatexchange.cli.exceptions import CommandError
 from threatexchange.exchanges.impl.fb_threatexchange_api import (
     FBThreatExchangeCollabConfig,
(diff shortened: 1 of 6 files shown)


python-threatexchange/threatexchange/cli/config_cmd.py
```diff
@@ -158,7 +199,7 @@ def _add_argument(cls, ap: argparse._ArgumentGroup, field: Field) -> None:
             if isinstance(origin, type) and issubclass(
                 origin, collections.abc.Collection
             ):
-                argparse_type = lambda s: origin(p.strip() for p in s.split(","))  # type: ignore  # mypy not smart enough for origin == type
+                argparse_type = lambda s: origin(arg_type(p.strip()) for p in s.split(","))  # type: ignore  # mypy not smart enough for origin == type
```
BarrettOlson, line 202: am I correct that this is the meaningful  fix/change?

Dcallies: This is the only change needed to fix the core, then I added everything else for testing.

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -23,12 +23,23 @@ def __init__(self, code: int) -> None:
 
 class ThreatExchangeCLIE2eHelper:
 
+    # A keystroke saver, prefix these args to the beginning
+    # of every call.
+    # You can bypass by making the first arg "tx" or
+    # "threatexchange".
```
BarrettOlson, line 29: thank you, I know this would have really confused me eventually

BarrettOlson (approved):
LGTM, fix it, fix it, fix it

PR #1086, opened 2022-07-21
[pytx] Begin making mypy more strict

Dcallies:
Summary
---------

mypy has a lot of flags that can enforce stricter typing. You only get out of typing what you put in.

Turning these all on causes a lot of errors, and I'm not even sure how to fix all of them. Instead mark what our intentions are, and then slowly turn it up.

We have a few strategies to deal with the backlog: try and enable everything, then blocklist all the old files. Turn one at a time on and fix as we go.

Test Plan
---------

If we enable all these flags right now:
```
Found 287 errors in 53 files (checked 102 source files)
```

mypy it passes

### DIFF
--- a/python-threatexchange/mypy.ini
+++ b/python-threatexchange/mypy.ini
@@ -1,4 +1,12 @@
 [mypy]
+strict_optional = True
+; disallow_any_generics = True
+; disallow_untyped_calls = True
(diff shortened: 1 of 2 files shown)


BarrettOlson (approved):
Nice! something to goal on :)

PR #1088, opened 2022-07-22
[PYTX]Add requirement.txt in video_vpdq

tigerchensh:
Summary
---------

Add requirement.txt in video_vpdq for auto-install vpdq when installing pytx.
Clean up the unused imports in vpdq setup.
Add install ffmpeg in github workflow for vpdq pytests.
Maybe worth doing in the future: Move extension pytests into a separate github workflow.

Test Plan
---------
Create a new venv and make local_install.
Before:
image
After:
image

Also the github workflow.


.github/workflows/python-threatexchange-ci.yaml
```diff
@@ -49,6 +49,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
```
BarrettOlson, line 51: @schatten you mentioned that containers for these flows can be reused. Does this happy auto-magically or should we move where this install command is run?
(same usage is found for tmk-cpp-ci)

schatten: Not quite auto-magical in all cases. Works for python, but ffmpeg is a system package. You can try using [cache-action](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action) to store / restore the directory that ends up getting ffmpeg. 

But $PATH updates and dpkg info will become more challenging. 

Let's start with no-caching, see how bad this is. 

Alternatively, we can use [containers for the actions](https://stackoverflow.com/questions/59269850/caching-apt-packages-in-github-actions-workflow/60920684#60920684).

schatten: @tigerchensh can you `time` the installations so we get some idea of how bad this is?

Dcallies:
No concerns from me, leaving to others

.github/workflows/python-threatexchange-ci.yaml
BarrettOlson: What do you think? @schatten, that's a bit long but we could leave it as a follow up issue

schatten: 30s is good! I'd be okay with leaving as is.

BarrettOlson (approved):

PR #1089, opened 2022-07-22
[pytx][exchange] update datetime string 

BarrettOlson:
Summary
---------

On my local (MacOS M1) I was seeing py.test failures:
> FAILED threatexchange/exchanges/clients/ncmec/tests/test_hash_api.py::test_mocked_get_hashes - AssertionError: assert 1508872800 == 1508858400
> FAILED threatexchange/exchanges/impl/tests/test_ncmec.py::test_fetch - AssertionError: assert 1508872800 == 1508858400

This was a difference of 4 hours which led me to look into the offset in datetime offsetss

Looks like at least locally, `strptime()` was interpreting the timestamp as "aware" and converting GMT to EST before calculating   the int values.

From the docs: 
https://docs.python.org/3.10/library/datetime.html#technical-detail
> Changed in version 3.7: When the %z directive is provided to the strptime() method, the UTC offsets can have a colon as a separator between hours, minutes and seconds. For example, '+01:00:00' will be parsed as an offset of one hour. In addition, providing 'Z' is identical to '+00:00'.

Test Plan
---------

py.test now passed locally, I need to see what the CI does to know if this still works for Linux.

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
@@ -23,7 +23,7 @@ from requests.packages.urllib3.util.retry import Retry
 from threatexchange.exchanges.clients.fb_threatexchange.api import TimeoutHTTPAdapter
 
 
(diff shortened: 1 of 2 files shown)


Dcallies (approved):
Nice find!

PR #1090, opened 2022-07-22
[pytx][pdq] Removed unneeded local import checks for Pillow faiss-cpu libs

BarrettOlson:
Summary
---------

`Pillow` & `faiss-cpu` are now install_requires not part of extra_require so these checks are now superfluous 

Test Plan
---------

py.test


Dcallies (approved):
Thanks for doing this!

PR #1093, opened 2022-07-22
[pytx] Dynamically Generate valid index classes based on SignalType

Dcallies:
Summary
---------

This is the most disgusting code I think I've written in this project, and I'm very worried I skipped a much simpler solution along the way.

Using dynamic class generation, abuse of pickle, and more, make it possible for a default linear search to exist for each type.

This might not be worth having classes do what they were doing before, which is to create the trivial linear searches.

Test Plan
---------

mypy && py.test

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_state.py
+++ b/python-threatexchange/threatexchange/cli/cli_state.py
@@ -16,7 +16,10 @@ import typing as t
 import logging
 
 from threatexchange.signal_type.index import SignalTypeIndex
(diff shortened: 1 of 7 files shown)


BarrettOlson (changes requested):
This is some cool python magic but I think maintaining and documenting it clearly will also require the use of magic.

Happy to be convinced otherwise, but more in favor of having signal_type be required to specify its index_type instead.

Dcallies:
I feel dumber for having attempted this

PR #1096, opened 2022-07-22
[pytx] Remove owner_id from base opinion

Dcallies:
Summary
---------

The owner field of this object was always problematic, and many APIs don't have a concept of owner. I originally moved to make ownership via an interface function that defaulted to false, but then realized even "ownerless" APIs might actually make a distinction between whether it was a trusted or untrusted source. Thus, I decided to add it as an attribute in the base class.

This creates a backwards incompatible change state change. We don't yet have a fully formed set of tests or detection for these changes, and I don't think the provided unittest is a great solution, but it's better than nothing.

Test Plan
---------

mypy & test

Test to show pickle compatibility of data.


python-threatexchange/threatexchange/signal_type/url.py
```diff
@@ -40,3 +40,13 @@ def matches_str(
     @staticmethod
     def get_examples() -> t.List[str]:
         return ["https://developers.facebook.com/docs/threat-exchange/reference/apis/"]
+
+    @classmethod
+    def normalize_fb_threatexchange_indicator(
```
BarrettOlson, line 45: Did this sneak in from a different soon to be PR?

Dcallies: Stacked, should have noted

python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
```diff
@@ -86,8 +88,16 @@ class FBThreatExchangeOpinion(state.SignalOpinion):
...
 
+    owner_app_id: int
+    tags: t.Set[str]
     descriptor_id: t.Optional[int]
 
+    def is_mine(self) -> bool:
+        return False  # TODO
+
+    def get_tags(self) -> t.Collection[str]:
+        return []
```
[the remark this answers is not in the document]
Dcallies: Thanks, at one point removed tags from the base, then decided to add it back in, should delete this

Dcallies:
I think I've decided to make this backwards compatible for the CLI, which can be done by creating a special implementation of __getstate__ that looks for the old property names

python-threatexchange/threatexchange/exchanges/tests/test_state.py
```diff
@@ -414,12 +472,14 @@ def key(n):
         },
         # Hash 1 again
         {
-            md5(1): SimpleFetchedSignalMetadata(
+            md5(1): FakeSignalMetadata(
                 [
-                    SignalOpinion(
-                        1, SignalOpinionCategory.INVESTIGATION_SEED, {t2, t3}
+                    SignalOpinionWithOwner(
+                        1, SignalOpinionCategory.INVESTIGATION_SEED, {t2, t, 3}
```
schatten, line 478: This changed significantly. Went from t2, t3 to t2, t, 3. 

Is it a typo?

schatten (approved):
Approve the change.

I am nervous about storing state in pickle. Do we have a note somewhere in the README specifying this dependency? And that users should try to stick to the same stable version of python (or atleast never downgrade).

python-threatexchange/threatexchange/exchanges/tests/test_state.py
Dcallies: Hmm, this looks like a vscode cursor bug, but confused about why tests wouldn't fail

Dcallies:
> I am nervous about storing state in pickle. Do we have a note somewhere in the README specifying this dependency? And that users should try to stick to the same stable version of python (or atleast never downgrade).

The versioning guarantee currently states that the exact format stored by the CLI is an implementation detail, just that it will be compatible. Since we are pickling now, I think it's correct to only try and maintain compatibility for pickle - otherwise it will be impossible to try and maintain compatibility for solutions we don't know about.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -47,14 +47,6 @@ def get_progress_timestamp(self) -> t.Optional[int]:
...
         return cls(response.max_timestamp)
 
-    def __setstate__(self, d: t.Dict[str, t.Any]) -> None:
-        """Implemented for pickle version compatibility."""
-        # 0.99.0 => 1.0.0:
-        ### field 'max_timestamp' renamed to 'get_entries_max_ts'
-        if "max_timestamp" in d:
-            d["get_entries_max_ts"] = d.pop("max_timestamp")
-        self.__dict__ = d
-
```
Dcallies, line 57: Nooo, why did this get removed!

PR #1099, opened 2022-07-24
[UnbreakingFetch][1/n]: Allow adding / toggling SignalExchangeAPIs

schatten:
Summary
---
Hashing has been made to work with pytx-1.0. Fetchers are next, because without them, matchers can't function.

For fetchers to work, they need to get all collabs. To get all collabs, we need to get all supported signal exchanges.

This PR adds toggle-able signal exchanges.

You can add a signal-exchange using their fully-qualified class name. By default, add FBThreatExchange and LocalFile.

Test Plan
---
Added unit tests.

Ran migration using `$ terraform -chdir=terraform apply`.

Then, in console, verified that both are 'enabled'.

```
$ python
>>> [
   t.to_concrete_class()
   for t in ToggleableSignalExchangeAPIConfig.get_all()
   if t.enabled
]
[
  <class 'threatexchange.exchanges.impl.fb_threatexchange_api.FBThreatExchangeSignalExchangeAPI'>,
  <class 'threatexchange.exchanges.impl.file_api.LocalFileSignalExchangeAPI'>
]
```

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/configs/tx_apis.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/tx_apis.py
@@ -1,0 +1,94 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+DynamoDB backed SignalExchangeAPI objects.
+"""
+
+from dataclasses import dataclass
+import typing as t
+from enum import Enum
+
+from botocore.exceptions import ClientError
+from jinja2 import ModuleLoader
+from threatexchange.exchanges.signal_exchange_api import SignalExchangeAPI
+
+from hmalib.common.config import HMAConfig, create_config, update_config
+from hmalib.common.logging import get_logger
+from hmalib.common.mappings import full_class_name, import_class
+
+logger = get_logger(__name__)
+
+
+@dataclass
+class ToggleableSignalExchangeAPIConfig(HMAConfig):
+    signal_exchange_api_class: str
+
+    # Allow soft disables
+    enabled: bool = True
+
+    @staticmethod
+    def get_name_from_type(signal_exchange_api: t.Type[SignalExchangeAPI]) -> str:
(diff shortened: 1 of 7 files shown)


hasher-matcher-actioner/hmalib/common/configs/tx_apis.py
```diff
@@ -0,0 +1,100 @@
...
+    signal_exchange_api_class: str
+
+    # Allow soft disables
+    enabled: bool = True
+
+    @staticmethod
+    def get_name_from_type(signal_exchange_api: t.Type[SignalExchangeAPI]) -> str:
+        return f"signal_exchange_api:{full_class_name(signal_exchange_api)}"
+
+    def to_concrete_class(self):
```
Dcallies, line 33: (nit): base class could be the return type here

hasher-matcher-actioner/hmalib/common/configs/tx_apis.py
```diff
@@ -0,0 +1,100 @@
...
+@dataclass
+class ToggleableSignalExchangeAPIConfig(HMAConfig):
+    signal_exchange_api_class: str
+
+    # Allow soft disables
+    enabled: bool = True
+
+    @staticmethod
+    def get_name_from_type(signal_exchange_api: t.Type[SignalExchangeAPI]) -> str:
+        return f"signal_exchange_api:{full_class_name(signal_exchange_api)}"
```
Dcallies, line 31: (ignorable): I went through all that trouble to guarantee you human readable names, and you throw it away! :P

Dcallies (approved):
I have a non-blocking TODO to do the same thing on the CLI (the reason why `tx config signal` and `tx config content` exists).

hasher-matcher-actioner/hmalib/common/configs/tx_apis.py
schatten: This is only for config storage. Not shown to humans.

PR #1100, opened 2022-07-25
Support Enums in aws_dataclass     

schatten:
Summary
---
For HMAConfig to represent NCMECCollabConfig, it needs to support Enums. Add that support and a test guarding it.

There is also a quality of life improvement as part of a second commit, I recommend reviewing commit by commit. 

There is also a third commit which applies name changes done as part of #1095. If reviewer would like, I'd be happy to break things up into 3 diff PRs, but decided against because the change is simple enough.

P.S. I'm strongly considering dropping our clearly lagging implementation in favour of dacite, but that would throw a wrench into the current project.

### [`42b2e07`](https://github.com/facebook/ThreatExchange/pull/1100/commits/42b2e07c849a6ce6fde7fe5146c68a977448c956) Support Enums in aws_dataclass
For HMAConfig to represent NCMECCollabConfig, it needs to support Enums. Add that support and a test guarding it.  P.S. I'm strongly considering dropping our clearly lagging implementation in favour of dacite, but that would throw a wrench into the current project. 
### [`346afce2d`](https://github.com/facebook/ThreatExchange/pull/1100/commits/346afce2d7e313543cd16065e13aad2e5722f8b8) fix: Apply name changes from #1095

Test Plan
---
Added unit test.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/aws_dataclass.py
+++ b/hasher-matcher-actioner/hmalib/common/aws_dataclass.py
@@ -26,7 +26,9 @@ How to use:
 """
 
 from decimal import Decimal
+import inspect
 from dataclasses import dataclass, field, fields, is_dataclass
+from enum import Enum
 
 import json
 import typing as t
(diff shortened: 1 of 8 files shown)


Dcallies (approved):
I like this PR, it has:
1. Clear explanation in summary
2. Clean split of commits
3. Good backing by unittests

PR #1101, opened 2022-07-25
[pytx] Add pdq utils and tests

tigerchensh:
Summary
---------

Implement zero pdqHash, random pdqHash, similar pdqHash and validate_signal_str in pdq signal. These functions are created for future VPDQ benchmark. Create corresponding tests.

Test Plan
---------

unittest

### DIFF
--- a/python-threatexchange/tests/hashing/utils.py
+++ b/python-threatexchange/tests/hashing/utils.py
@@ -1,0 +1,24 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+import random
+from threatexchange.hashing.pdq_utils import *
+
+def get_zero_hash() -> str:
+    """Return a pdq hash hex str that is zero for every byte"""
+    return "0" * PDQ_HEX_STR_LEN
+
+
+def get_random_hash() -> str:
+    """Return a random pdq hash hex str"""
+    return f"{random.randrange(2**256):0{PDQ_HEX_STR_LEN}x}"
+
+
+def get_similar_hash(pdq_hex: str, dist: int) -> str:
+    """Return a pdq hash hex str with dist hamming distance away from pdq_hex"""
+    if dist > BITS_IN_PDQ or dist < 0:
+        raise ValueError("Invalid distance value")
+    order = random.sample(range(BITS_IN_PDQ), k=dist)
+    bin_list = list(hex_to_binary_str(pdq_hex))
+    for i in order:
+        bin_list[i] = str(int(bin_list[i]) ^ 1)
+    bin_str = "".join(bin_list)
+    return binary_str_to_hex(bin_str)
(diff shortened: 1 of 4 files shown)


schatten (changes requested):
The utility of the added methods is unclear from the title or description of the PR.

Beyond creating test data, I do not see real world use-cases for the added method. I'm not implying that we don't merge this in. Simply that the added methods be added as test helpers.

You can do that adding a new module in `python-threatexchange/tests/`. As long as the module does not have unit-tests, py.test will ignore it.

PS. When you are raising a pull-requests, it is helpful to talk about the reasons why the changes are necessary. In this example, I can see that you have added "pdq utils and tests" by looking at the "Files Changed" tab. What is unclear is why you've added these. 

Example of good PR summaries: #1062 #1090, #1016

tigerchensh:
Thanks for the review. Sorry about the incomplete summary. 
The initiative of this pr is from https://github.com/facebook/ThreatExchange/pull/1097 where I tried to create similar hashes and combine them as a test video (List of VPDQ features). Then, I can generate edge cases situations when matching videos. However, this leads to two separate parts, one is utils in pdq and another one is utils in vpdq. That's why I opened this PR. Sorry for the loss of the conversation.
I will make the changes to move them into the test folder.

python-threatexchange/tests/hashing/utils.py
```diff
@@ -0,0 +1,25 @@
...
+def get_similar_hash(pdq_hex: str, dist: int) -> str:
+    """Return a pdq hash hex str with dist hamming distance away from pdq_hex"""
+    if dist > BITS_IN_PDQ or dist < 0:
+        raise ValueError("Invalid distance value")
+    order = random.sample(range(BITS_IN_PDQ), k=dist)
+    bin_list = list(hex_to_binary_str(pdq_hex))
+    for i in order:
+        bin_list[i] = str(int(bin_list[i]) ^ 1)
+    bin_str = "".join(bin_list)
+    return binary_str_to_hex(bin_str)
```
Dcallies, line 25: (alt, ignorable): Rather than converting these back and forth, you could cast this to int:

```
pdq_int = reduce(lambda x, y: x ^ (1 << y), order, int(pdq_hex, 16))
return f"{pdq_int:0{PDQ_HEX_STR_LEN}x}"
```

Dcallies (commented):
This looks reasonable to me.

As part of your test plan, can you run mypy directly on the files with more flags enabled? I mentioned `--strict` which might be too strict, but at least these:

```
--no-implicit-optional --strict-equality --disallow-any-generics --disallow-untyped-defs --disallow-incomplete-defs
``` 

Since @schatten  requested changes, will leave to him to approve

tigerchensh:
Thanks for the review. I have made those changes. I was running ```python -m mypy threatexchange``` frequently, but I missed that this file was not part of the check. And many of the functions are without typing. I will also add those type hints as well.

python-threatexchange/tests/hashing/utils.py
tigerchensh: Get the trick here. To use ^, we actually don't need to cast to binary but int will work. This reminds a lot about functional programming.

schatten (approved):

PR #1102, opened 2022-07-25
[VPDQ] Force package to return python string

tigerchensh:
Summary
---------

Decode any c_str returned in vpdq package by any C++ binding functions. Add corresponding tests.

Test Plan
---------
vpdq tests after locally install the adjusted vpdq package.


vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -28,39 +28,53 @@
...
-            assert hash_file.is_file()
-            ret = test_util.read_file_to_hash(hash_file)
-            assert ret is not None
-            sample_hashes[file] = ret
+def test_vpdq_utils():
+    sample = Path(f"{hash_folder}/{TEST_FILES[0]}.txt")
+    assert sample.is_file()
+    ret = test_util.read_file_to_hash(sample)
+    with open(sample, "r") as file:
+        lines = file.readlines()
```
Dcallies, line 43: Nit: You have read the entire file here, so you can dedent and close the file

vpdq/python/vpdq.pyx
```diff
@@ -78,14 +76,14 @@ def hash_to_hex(hash_value: "Hash256"):
     Returns:
         str: hex str of hash
     """
-    return hashToString(hash_value)
```
Dcallies, line 81: blocking: I am very confused - the signature of this function says it is returning a string. What is it returning instead, and why don't we fix it at https://github.com/facebook/ThreatExchange/blob/main/vpdq/python/vpdq.pyx#L21 ?

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -28,39 +28,53 @@
     "pattern-sd-with-small-logo-bar",
```
Dcallies, line 28: Why are these changes included in this diff? Isn't it fixing the str?

Dcallies (commented):
Hmm, interesting. This is a detail of pyx files I'm not familiar with.

blocking: However, anyone who attempts to use your python bindings is going to run into the same problem. There must be a standard solution that we can apply at the pyx file instead

Dcallies (changes requested):
Meant to request changes

vpdq/python/vpdq.pyx
tigerchensh: It is returning the c_string. The string type is from from ``` libcpp.string cimport string ``` With this cdef extern, we cannot modify codes here. It is only extracting the c function. Have to wrap it around with a python function and solve the problem there.

vpdq/python/tests/test_vpdq_hash.py
tigerchensh: Reverting those!

vpdq/python/vpdq.pyx
Dcallies: blocking: I believe there should be a solution that prevents the unencoded from leaving your python bindings, where they will later confuse folks trying to use your library.

Does either of these match what you are observing? 
1. https://stackoverflow.com/questions/12500069/ctypes-how-to-pass-string-from-python-to-c-function-and-how-to-return-string
2. https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html

Dcallies (changes requested):
blocking: I believe there should be a solution that prevents the unencoded from leaving your python bindings, where they will later confuse folks trying to use your library.

Does either of these match what you are observing?

https://stackoverflow.com/questions/12500069/ctypes-how-to-pass-string-from-python-to-c-function-and-how-to-return-string
https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html

tigerchensh:
Thanks for the review, but the current fix won't let the unencoded leave the python binding. Convert the byte string in C++ function is same as converting in python wrapper.

vpdq/python/vpdq.pyx
tigerchensh: The current output from the vpdq hash_to_hex is a binary string literal. For example, an hash "000ffff...." will be "b'000ffff.....". After this pr, the hash_to_hex will return "000ffff....".

vpdq/python/vpdq.pyx
```diff
@@ -78,14 +76,14 @@ def hash_to_hex(hash_value: "Hash256"):
...
+    return str(hashToString(hash_value), "utf-8")
 
 
 def str_to_hash(str_hash: str):
     return fromStringOrDie(str(str_hash).encode("utf-8"))
 
 
-def hamming_distance(hash1: "Hash256", hash2: "Hash256"):
```
Dcallies, line 88: (unrelated to diff): Why is this function exposed? Should we instead direct people to the PDQ bindings instead?

vpdq/python/vpdq.pyx
```diff
@@ -78,14 +76,14 @@ def hash_to_hex(hash_value: "Hash256"):
     Returns:
         str: hex str of hash
     """
-    return hashToString(hash_value)
+    return str(hashToString(hash_value), "utf-8")
 
 
 def str_to_hash(str_hash: str):
     return fromStringOrDie(str(str_hash).encode("utf-8"))
```
Dcallies, line 83: According to your typing, you are passing in `str`, so why cast to str?

Assume the return type is Hash256

Dcallies (approved):
After discussing offline, there are things in the bindings that I am surprised we are exposing or making part of the public interface:
1. Hash256
2. hamming_distance

As part of #1 I would have expected the hash inside of VpdqFeature to be hidden (_hash).

However, none of those things are part of this PR, and this does fix a discovered bug. Let's continue working on the python SignalType code before doing any work here

vpdq/python/vpdq.pyx
tigerchensh: Because the PDQ binding does not provide a hamming distance function from the C++ implementation. I believe having this function will increase performance. The translation and calculation happen in PDQ_util instead of the binding.

vpdq/python/vpdq.pyx
tigerchensh: The return type is binary literal string. For example, a hash str "000ffff...." will be "b'000ffff.....".

PR #1106, opened 2022-07-26
[PYTX] VPDQ Fix json key type when converting back

tigerchensh:
Summary
---------

A feature of json object key automatically converted to string (javascript only accepts string as key) is found and fixed. So, validate_signal_str can correctly compare the frame_number with 0.

Test Plan
---------

Tested in this pr https://github.com/facebook/ThreatExchange/pull/1107 as validating str is called.


Dcallies (changes requested):
blocking: Please add a unittest that fails before your fix, and passes after - this is called a "regression test". It turns out statistically, the things that break the most in the future are things that broke in the past. This is part of a wider strategy outlined here: https://www.guru99.com/regression-testing.html

tigerchensh:
Thanks for the review. The tests that are failing before the fix are any validate_signal_str within the tests added here https://github.com/facebook/ThreatExchange/pull/1107. I also added a test case which validated the signal_str to ensure that it is working correctly. Without the fix, validate signal_str will also fail.

Dcallies:
Talked with @tigerchensh offline, he said he accidentally merged the fix into main as part of another PR

PR #1109, opened 2022-07-27
[PYTX] Add VPDQ index default matching threshold and tests

tigerchensh:
Summary
---------

Add VPDQ query match threshold and index match threshold to filter out index match result. Any match result with a lower than threshold query/index match percentage will not return, since it is expected that we do not return every index match result.

get_zero_threshold_index is implemented in test utils to help creating index that does turn every index match result to be compared in tests.

Test Plan
---------
Tests are added to make sure filtering is working correctly
test_match_below_and_abolve_deafault_query_threshold and test_match_below_and_abolve_deafault_index_threshold are implemented to test on three cases:
1. below match percent but not zero
2. exactly at match percentage
3. above match percentage


python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -37,8 +38,10 @@
 
 if not _DISABLED:
     EXAMPLE_META_DATA = {"example_video"}
```
Dcallies, line 40: What about this one?

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
```diff
@@ -13,6 +13,8 @@
 VPDQ_TIMESTAMP_PRECISION = 3
 VPDQ_QUALITY_THRESHOLD = 50
 VPDQ_DISTANCE_THRESHOLD = 31
+VPDQ_QUERY_MATCH_THRESHOLD = 80.0
+VPDQ_INDEX_MATCH_THRESHOLD = 80.0
```
Dcallies, line 17: blocking q: I think our default in production is 0.0. 

What is your thought about using 80 here?

python-threatexchange/threatexchange/extensions/vpdq/vpdq_index.py
```diff
@@ -29,6 +31,8 @@ class VPDQIndex(SignalTypeIndex[IndexT]):
     def __init__(
         self,
         quality_threshold: int = VPDQ_QUALITY_THRESHOLD,
+        query_match_threshold: float = VPDQ_QUERY_MATCH_THRESHOLD,
+        index_match_threshold: float = VPDQ_INDEX_MATCH_THRESHOLD,
```
Dcallies, line 35: nit: Would strongly recommend using `_pct` or `_percent` here, as well as asserting it's 0 <= val <= 100 

You may also want to name the variable such that it's clear it's a lowerbound - for all other signal types, we use % difference (so 0 = exact match, 100 = accept anything). With vPDQ it's reversed.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -126,6 +131,57 @@ def test_no_match():
     assert len(res) == 0
 
 
+def test_match_below_and_abolve_deafault_query_threshold():
```
Dcallies, line 134: nit: speeling

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -126,6 +131,57 @@ def test_no_match():
...
+    video4 = pdq_hashes_to_VPDQ_features(random.sample(g3, 10))
+
+    index = VPDQIndex.build([[vpdq_to_json(video4), VIDEO4_META_DATA]])
+    res = index.query(vpdq_to_json(video1))
+    assert len(res) == 0
+
+    res = index.query(vpdq_to_json(video2))
+    assert compare_match_result(res[0], VPDQIndexMatch(100, 80, 100, VIDEO4_META_DATA))
+
+    res = index.query(vpdq_to_json(video3))
```
Dcallies, line 154: (ignorable): Many of these tests spend a lot of time converting between different formats of the same data. The index already unpacks these strings back into features. Exposing an inner function that takes features, or changing the accepted type of query to be either string or features (union type) would make these tests more straightforward.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -126,6 +131,57 @@ def test_no_match():
...
+    assert compare_match_result(res[0], VPDQIndexMatch(100, 90, 100, VIDEO4_META_DATA))
+
+
+def test_match_below_and_abolve_deafault_index_threshold():
```
Dcallies, line 158: speeling: above, default

Dcallies (changes requested):
Only have one question inline, ping me once responded

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
tigerchensh: I will replace the default threshold as 0 but still keeping this as an option to filter out low percentage matches.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
tigerchensh: I was thinking of giving an example about what the meta_data should be like.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
tigerchensh: Will do this in a separate PR!

tigerchensh:
Thanks for the review. Have changed the default back to 0 and implemented a build within vPDQIndex with thresholds.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
Dcallies: Except we don't use meta_data this way anywhere. The Value of T is always FetchedSignalMetadata

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
```diff
@@ -13,6 +13,8 @@
 VPDQ_TIMESTAMP_PRECISION = 3
 VPDQ_QUALITY_THRESHOLD = 50
 VPDQ_DISTANCE_THRESHOLD = 31
+VPDQ_QUERY_MATCH_THRESHOLD_PERCENT = 0.0
```
Dcallies, line 16: blocking: In the production the value of this 80%. 0.0 and 0.0 means to match everything even if zero frames overlap.

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
```diff
@@ -13,6 +13,8 @@
...
+VPDQ_INDEX_MATCH_THRESHOLD_PERCENT = 0.0
```
Dcallies, line 17: This probably should have a comment to explain that this won't trigger unless there's at least one matching frame (and you should confirm that this is the case in the behavior of the functions).

Dcallies (changes requested):
blocking: You changed the default for both match percentages, not just the one I flagged (INDEX)

python-threatexchange/threatexchange/extensions/vpdq/vpdq_util.py
tigerchensh: Have added a test for this.

Dcallies (approved):
shipit

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -36,9 +36,17 @@
...
 if not _DISABLED:
-    EXAMPLE_META_DATA = {"example_video"}
-    VIDEO1_META_DATA = {"video1"}
-    VIDEO2_META_DATA = {"video2"}
+    EXAMPLE_META_DATA = dict(
+        {
+            "hash_type": "vpdq",
+            "video_id": 5,
+        }
+    )
```
Dcallies, line 44: Nit: this is exactly the same as 
```suggestion
    EXAMPLE_META_DATA = {
        "hash_type": "vpdq",
        "video_id": 5,
    }
```

PR #1110, opened 2022-07-27
[PYTX]VPDQ Add support for detecting and handling empty video

tigerchensh:
Summary
---------

Before the fix in vpdq_index.py, the test which queries/indexs an empty video will fail at the stage of converting to faiss vector at faiss_index.range_search.
image

Before the fix in vpdq_util.py, the test which queries/indexs an empty signal_str will fail for json converting an empty string at function json_to_vpdq.
image

Test Plan
---------

Pass test cases for querying and indexing empty video/signal string.

### DIFF
--- a/python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
+++ b/python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
@@ -104,20 +104,23 @@ def test_serialize():
     )
 
 
-def test_empty_video_and_signal_str():
+def test_empty_video_():
     empty_video = []
-    empty_signal_str = ""
     video1 = pdq_hashes_to_VPDQ_features(g1)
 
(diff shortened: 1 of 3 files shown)


python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -104,6 +104,24 @@ def test_serialize():
     )
 
 
+def test_empty_video_and_signal_str():
```
Dcallies, line 107: blocking: these are two different cases, can you split them into two tests?

Dcallies (approved):
Thanks for adding this case!

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
tigerchensh: Have split into two

PR #1112, opened 2022-07-27
[pytx] No more dist in SignalType.compare_hash

Dcallies:
Summary
---------

This is confusing since not all SignalTypes support distance. In addition, not all signal types support the same distance. 

Mypy is smart enough to tell that new arguments are backwards compatible.

Test Plan
---------

mypy & py.test

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/raw_text.py
+++ b/python-threatexchange/threatexchange/signal_type/raw_text.py
@@ -41,15 +41,12 @@ class RawTextSignal(
 
     @classmethod
     def matches_str(
-        cls, signal: str, haystack: str, distance_threshold: t.Optional[int] = None
+        cls, signal: str, haystack: str, pct_diff_threshold: int = 5
     ) -> signal_base.HashComparisonResult:
-        threshold = 5  # Match considered if 95% match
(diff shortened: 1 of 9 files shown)


python-threatexchange/threatexchange/signal_type/raw_text.py
```diff
@@ -41,15 +41,12 @@ def get_content_types(cls) -> t.List[t.Type[ContentType]]:
 
     @classmethod
     def matches_str(
-        cls, signal: str, haystack: str, distance_threshold: t.Optional[int] = None
+        cls, signal: str, haystack: str, pct_diff_threshold: int = 5
```
tigerchensh, line 44: nit: create a constant for this default threshold 5

tigerchensh (approved):
LGTM! Love these changes for maintaining the high-level generalization while giving more flexibility in signal comparison.

python-threatexchange/threatexchange/signal_type/raw_text.py
Dcallies: I prefer not to create a constant if it's only used in one place, otherwise you are just adding indirections in the code!

PR #1114, opened 2022-07-28
[pytx] PDQ Quality filter

Dcallies:
Summary
---------

We generally consider quality score < 50 to not be usable. Sadly we don't expose the quality score in the hash (though I think we should in the future). 

When a hash is not computable, the interface is expected to return empty string (see hash_cmd), but it's undocumented, so fix that too.

Test Plan
---------

There's not a test suite for this currently. I'll plead for regression testing post 1.x, but would accept a block for test.

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq.py
@@ -5,10 +5,9 @@ Wrapper around the Photo PDQ signal type.
 """
 
 import typing as t
-import pathlib
 import re
 
-from threatexchange.hashing.pdq_hasher import pdq_from_bytes, pdq_from_file
+from threatexchange.hashing.pdq_hasher import pdq_from_bytes
 from threatexchange.content_type.content_base import ContentType
 from threatexchange.content_type.photo import PhotoContent
 from threatexchange.signal_type import signal_base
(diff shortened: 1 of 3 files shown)


python-threatexchange/threatexchange/signal_type/pdq.py
```diff
@@ -69,14 +70,11 @@ def compare_hash(
...
-    @classmethod
-    def hash_from_file(cls, file: pathlib.Path) -> str:
-        pdq_hash, _quality = pdq_from_file(file)
-        return pdq_hash
-
     @classmethod
     def hash_from_bytes(cls, bytes_: bytes) -> str:
-        pdq_hash, _quality = pdq_from_bytes(bytes_)
+        pdq_hash, quality = pdq_from_bytes(bytes_)
+        if quality < cls.QUALITY_THRESHOLD:
```
BarrettOlson, line 76: Unsure: it might be worth a warning (or at least debug log) here as this would be a hard thing to debug. Based on `hash_from_bytes` it is within the supported behavior but still feels like a "fail quietly" choice.

BarrettOlson (approved):
Sure, it's better than creating useless/misleading hashes

python-threatexchange/threatexchange/signal_type/pdq.py
Dcallies: Same problem with tlsh - it needs at least 100 bytes or it returns empty. Punting into the future.

PR #1115, opened 2022-07-28
[VPDQ] Silence ffmpeg for nonverbose option

tigerchensh:
Summary
---------

One of the todos in (https://github.com/facebook/ThreatExchange/projects/4)
Silence ffmpeg for non-verbose options. Thus, in python-binding library vpdq will produce hashes silently working with python-threatexchange.

Test
---------

Run through regression test nonverbose and see if ffmpeg prints anything.


BarrettOlson (approved):
LGTM

PR #1116, opened 2022-07-28
[Pytx] Vpdq move vpdq_index_match to vpdq_Index class

tigerchensh:
Summary
---------

vpdq_index_match should be an internal implementation detail lives in vpdq_Index class. Thus, move it to the right place.

Test Plan
---------
vpdq pytests


BarrettOlson (approved):
LGTM
(as long as the CLI commands as well as py.test still work as expected)

PR #1117, opened 2022-07-28
[pytx][exchanges] prevent datetime object from using local tz (use utc in from_xml)

BarrettOlson:
Summary
---------

#1089 seeming 'fix' broke something our ability to request. @Dcallies informed me that `_DATE_FORMAT_STR` needs the literal `"Z"`

What the `%z` was trying to fix was that `datetime` created from `content_xml.str("maxTimestamp")` is assumed to be in the local machine timezone. This change replaces that erroneous tz info to be `utc` specific.

Test Plan
---------

I ran pytest inside and outside of the python-threatexchange dev container.

This add "Z" back to `_DATE_FORMAT_STR` so it should address the breakage but I will need @Dcallies help to verify.


Dcallies (approved):
Tested on my machine to be sure, thanks for fixing!

PR #1118, opened 2022-07-28
[pytx][ncmec] Handle NCMEC API returning out-of-order entries in NCMECSignalExchangeAPI

### LINKED ISSUE #1105: [ncmec] NCMEC API does not return in timestamp order - breaks checkpointing
```
$ threatexchange fetch --only-collab ncmec
[ncmec] ncmec - Downloaded [...] updates at 2019-12-10T15:02:52
[ncmec] ncmec - Downloaded [...] updates at 2019-03-30T14:00:40
[ncmec] ncmec - Downloaded [...] updates at 2019-03-30T14:00:40
[ncmec] ncmec - Downloaded [...] updates at 2019-03-28T14:01:08
[ncmec] ncmec - Downloaded [...] updates at 2021-07-18T14:01:23
[ncmec] ncmec - Downloaded [...] updates at 2017-11-21T11:02:59
[ncmec] ncmec - Downloaded [...] updates at 2017-12-22T14:44:41
```

Dcallies:
Summary
---------

This PR is the result of stress testing on the NCMEC hash DB API, and fixes some issues discovered.

Fixes #1105

## Fixing Checkpointing Behavior
The largest one is that checkpointing behavior wasn't sane, due to the NCMEC API's behavior of not returning entries in entry order. Though https://report.cybertip.org/hashsharing/v2/documentation.pdf suggests that we can use maxTimestamp from each return in order to plan the next fetch, observed behavior doesn't seem to support this.

However, [to, from) behavior seems to be correct, (to = inclusive, from=exclusive), and so instead of using from=0 to=now() and relying on maxTimestamp, we can instead try and pick correctly-sized time ranges.

However, the actual data in the database is randomly distributed, so the "correct" time range must be discovered through probing, and if the probed time range becomes one second, we have to fetch it no matter what.

It's possible that the the next=URLs are durable, and we could have checkpointed using those instead of time ranges, as long as our end time didn't include a mutable range (such as now() or the future). However, looking at the next= parameters, I couldn't convince myself that the storage didn't use offsets for pagination, which might become invalid if an old record is updated (based on the documentation saying that to/from also considers update time), and so an offset that included a large amount of time might miss records, and paginating larger ranges for larger periods of time increase the chance that records are leaked.

There are more considerations on the probing strategy, but they are documented in the code.

Test Plan
---------

```
$ tx config collab edit ncmec -C ncmec_test_industry --environment test_Industry
$ TX_VERBOSE=1 tx fetch
<pretty scroll, exact record counts and timestamps omitted for security>
```

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
@@ -141,6 +141,25 @@ class NCMECEntryUpdate:
         )
 
 
+class GetEntriesNextInfo(t.NamedTuple):
+    """Wrapper around parsed "next" param data. Fields match param names"""
+
+    start: int  # Might be- the start ID of the next fetch
+    size: int  # How many records to fetch (always seems to be 1000)
+    max: int  # The largest id in the time range
+
+    @classmethod
+    def from_next(cls, next_: str) -> t.Optional["GetEntriesNextInfo"]:
+        s = html.unescape(next_)
+        parsed = urllib.parse.parse_qs(s)
+        start = parsed.get("start")
+        size = parsed.get("size")
+        max_ = parsed.get("max")
+        if not start or not size or not max_:
+            return None
+        return GetEntriesNextInfo(int(start[0]), int(size[0]), int(max_[0]))
+
+
 @dataclass
 class GetEntriesResponse:
     updates: t.List[NCMECEntryUpdate]
@@ -168,27 +187,21 @@ class GetEntriesResponse:
         next_ = xml.maybe("paging", "next").element.text or ""
         return cls(updates, max_ts or fallback_max_time, next_)
 
-    def get_next_info(self) -> t.Optional[t.Tuple[int, int, int]]:
+    def get_next_info(self) -> t.Optional[GetEntriesNextInfo]:
         if not self.next:
             return None
-        s = html.unescape(self.next)
-        parsed = urllib.parse.parse_qs(s)
-        start = parsed.get("start")
-        size = parsed.get("size")
-        max_ = parsed.get("max")
-        if start is None or size is None or max_ is None:
-            return None
-        return int(start[0]), int(size[0]), int(max_[0])
+        return GetEntriesNextInfo.from_next(self.next)
 
     @property
-    def entries_in_range(self) -> int:
+    def estimated_entries_in_range(self) -> int:
+        """Uses the "next" params to try and guess entries in range"""
         if not self.next:
             return len(self.updates)
         info = self.get_next_info()
         return (
-            len(self.updates) + 1
+            len(self.updates) + 1  # Parse error of some kind
             if info is None
-            else info[2] - info[0] + len(self.updates)
+            else info.max - info.start + len(self.updates)
         )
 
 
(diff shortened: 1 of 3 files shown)


Dcallies:
Hmm, test failure didn't reproduce locally, I'll take a closer look.

python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
```diff
@@ -165,6 +168,29 @@ def from_xml(cls, xml: _XMLWrapper, fallback_max_time: int) -> "GetEntriesRespon
...
+        parsed = urllib.parse.parse_qs(s)
+        start = parsed.get("start")
+        size = parsed.get("size")
+        max_ = parsed.get("max")
+        if start is None or size is None or max_ is None:
+            return None
+        return int(start[0]), int(size[0]), int(max_[0])
+
+    @property
+    def entries_in_range(self) -> int:
```
Dcallies, line 184: Should change to mention this is an estimate (observed random high numbers)

python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
```diff
@@ -165,6 +168,29 @@ def from_xml(cls, xml: _XMLWrapper, fallback_max_time: int) -> "GetEntriesRespon
...
+        if not self.next:
+            return None
+        s = html.unescape(self.next)
+        parsed = urllib.parse.parse_qs(s)
+        start = parsed.get("start")
+        size = parsed.get("size")
+        max_ = parsed.get("max")
+        if start is None or size is None or max_ is None:
+            return None
+        return int(start[0]), int(size[0]), int(max_[0])
```
Dcallies, line 181: nit: It's basically free to NamedTuple this

python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
```diff
@@ -292,33 +320,46 @@ def get_entries(
...
             next_=next_,
-            to=_date_format(int(time.time())),
+            to=_date_format(end_timestamp),
             **params,
         )
         return GetEntriesResponse.from_xml(_XMLWrapper(response), int(time.time()))
 
     def get_entries_iter(
-        self, start_timestamp: int = 0
+        self, start_timestamp: int = 0, end_timestamp: int = 0
```
Dcallies, line 347: Nit: add a `*` to match API of get_entries

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -156,21 +161,116 @@ def fetch_iter(
...
+                        log(
+                            f"est {entry.entries_in_range} is over max fetch, duration reduced"
+                        )
+                        # We want to at last shrink by our shrink factor
+                        duration = min(
+                            duration // self.FETCH_SHRINK_FACTOR,
+                            # Especially in early fetches, we are overfetching
+                            # by a huge amount, so shrink in proportion to overfetch
+                            duration * self.MAX_FETCH_SIZE // entry.entries_in_range,
+                        )
```
schatten, line 237: 🤌 Chef's kisses.

schatten (approved):
Lots of very smart optimizations. Concise, easy to follow. Love it!

Small nits below, but good to merge. Thanks!

PR #1120, opened 2022-07-28
[pytx][ncmec] Fix a few bugs with NCMEC settings

Dcallies:
Summary
---------

1. Environment was not being set by the collab config. Ironically, it's because of the same bug copy pasted from ThreatExchange
2. You can't edit ncmec configs due to a strange combination of conditions.

Test Plan
---------

regression test for 2, mypy & py.test


BarrettOlson (approved):
LGTM

PR #1121, opened 2022-07-28
[pytx][cli] Defense against time rewinds

Dcallies:
Summary
---------

The NCMEC API doesn't return entries in ascending order. We caught it because the progress time was jumping around. So add that as a feature in the CLI.

Test Plan
---------

pending

### DIFF
--- a/python-threatexchange/threatexchange/cli/fetch_cmd.py
+++ b/python-threatexchange/threatexchange/cli/fetch_cmd.py
@@ -195,6 +195,17 @@ class FetchCommand(command_base.Command):
                 next_checkpoint = delta.checkpoint
                 self._fetch_progress(len(delta.updates), next_checkpoint)
                 assert next_checkpoint is not None  # Infinite loop protection
+                if checkpoint is not None:
+                    prev_time = checkpoint.get_progress_timestamp()
+                    next_time = next_checkpoint.get_progress_timestamp()
+                    if prev_time and next_time:
+                        assert prev_time <= next_time, (
+                            "checkpoint time rewound? ",
+                            "This can indicate a serious ",
+                            "problem with the API and checkpointing",
+                        )
+                checkpoint = next_checkpoint  # Only used for the rewind check
+
                 store.merge(collab, delta)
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/cli/fetch_cmd.py
```diff
@@ -195,6 +195,17 @@ def execute_for_collab(
                 next_checkpoint = delta.checkpoint
                 self._fetch_progress(len(delta.updates), next_checkpoint)
                 assert next_checkpoint is not None  # Infinite loop protection
+                if checkpoint is not None:
+                    prev_time = checkpoint.get_progress_timestamp()
+                    next_time = next_checkpoint.get_progress_timestamp()
+                    if prev_time and next_time:
+                        assert prev_time <= next_time, (
```
schatten, line 202: blocking: HMA will probably retry `execute_for_collab` here. Can we throw a custom exception so that we do a finite number of retries (or none at all)?

HTTP errors are common and we probably want a logic like: Keep retrying this operation unless it is a certain class of problems.

If we throw a custom exception, we can enumerate the 'non-retryable' problems.

schatten: Now that I read this a bit more, it seems like HMA will need to re-implement the entirety of fetch_cmd. Proposal, I am writing the HMA side as we speak. 

How about I implement a FetchRunner class in pytx that can replace large parts of fetch_cmd AND work for HMA?

schatten (approved):
For pytx, looks good.

However, it appears like a lot of fetch_cmd would need to be re-done for HMA. Inline proposal for a re-useable FetchRunner class in pytx. I can take that up since I am copying a lot from fetch_cmd right now.

python-threatexchange/threatexchange/cli/fetch_cmd.py
Dcallies: Let's sync offline, I think defining custom exceptions might involve interface changes because exceptions might be thrown in fetch itself.

PR #1122, opened 2022-07-29
[Pytx] vPDQ index benchmark

tigerchensh:
Summary
---------

Benchmark on vPDQ with brute_force, vPDQ index with faiss and faiss search only. The difference between index and faiss is that vPDQ index stores and translates lots of data during building. Benchmark mark runs on MacBook Pro (16-inch, 2021), Apple M1 Pro, 32 GB Ram and the result is in README.MD.


python-threatexchange/benchmarks/benchmark_vpdq_index.py
```diff
@@ -0,0 +1,158 @@
...
+    ap.add_argument(
+        "--video-frames",
+        "-v",
+        type=int,
+        default=500,
+        help="How many frames in each video",
+    )
+    ap.add_argument(
+        "--dataset-size",
+        "-n",
```
Dcallies, line 130: n corresponding the number of videos, and `-v` also existing might be confusing. Would suggest either:
1. Using `-n` for video and `-m` for frames
2. Using `-v` for videos and `-f` for average frames

Dcallies (commented):
This looks roughly like what I would expect, and would accept this if you are done with it - feel free to convert to real PR

python-threatexchange/benchmarks/benchmark_vpdq_index.py
tigerchensh: Took this option "Using -v for videos and -f for average frames"

python-threatexchange/benchmarks/benchmark_vpdq_index.py
```diff
@@ -0,0 +1,174 @@
...
+    VPDQ_QUALITY_THRESHOLD,
+    VPDQ_DISTANCE_THRESHOLD,
+)
+
+import typing as t
+import random
+from threatexchange.extensions.vpdq.vpdq_index import VPDQIndex
+
+
+Oint = t.Optional[int]
```
BarrettOlson, line 26: nit: is this used anywhere?

python-threatexchange/benchmarks/benchmark_vpdq_index.py
```diff
@@ -0,0 +1,174 @@
...
+    ap.add_argument(
+        "--jitter-noise",
+        "-j",
+        type=int,
+        default=50,
+        help="How many frames varies between each video in uniformal distribution",
+    )
+    ap.add_argument(
+        "--dataset-size",
+        "-v",
```
BarrettOlson, line 146: nit: I will always read `-v` as verbose

BarrettOlson (approved):
Nice! had a few nits, nothing blocking

PR #1123, opened 2022-07-29
[PYTX]Move tests folder into threatexchange and fix import

tigerchensh:
Summary
---------
The change is moving python-threatexchange/tests to python-threatexchange/threatexchange/tests

Found a wrongly placed folder which could cause import problems. First, tests folder contain the pdq and faiss tests which are included in threatexchange as default, so it should live in  threatexchange. Secondly, tests folder is neither a package nor within threatexchange package, any import like ```tests.hashing.utils``` won't work.

But why does it not break any import in the current pytest files? It is because the pytest runs through prepend to find all the modules required(https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html). 
However, it will make any future import from the utils fail in normal python file.
(i.e. Any file that tires to use python-threatexchange/threatexchange/extensions/vpdq/tests/utils.py will fail, because it imports ```tests.hashing.utils```)
Example: https://github.com/facebook/ThreatExchange/pull/1122/files#diff-48794b01ac9b51e085edf4887addbf2ca967e2a8b67f7aab72514214a09f238b

Test Plan
---------
A temporary Test.py file at  ```python-threatexchange/threatexchange/```
Content:```import threatexchange.extensions.vpdq.tests.utils```
Before the change, the test will fail at the import.
```Traceback (most recent call last):
  File "test.py", line 1, in <module>
    import threatexchange.extensions.vpdq.tests.utils
  File "/Users/xiangnan/Documents/GitHub/ThreatExchange/python-threatexchange/threatexchange/extensions/vpdq/tests/utils.py", line 4, in <module>
    from tests.hashing.utils import get_random_hash
ModuleNotFoundError: No module named 'tests.hashing'
```
After the change, it will successfully import.


schatten (approved):
Love this change.

Pro-tip: to get more proof that the changes are valid / correct, can you paste the last line of py.test output before and after? So we can be sure that the number of tests run before and after your changes are the same.

tigerchensh:
Thanks for the suggestion. Before the change:
image
After the change:
image
In the images, it can tell that the tests folder have moved and the total number of tests passed is 117.

Dcallies:
We are moving files to threatexchange/tests/hashing instead of threatexchange/hashing/tests, but that can be a followup.

Also, your images have basically the same output - both say 117

PR #1124, opened 2022-07-29
[PYTX] VPDQ fix incorrect pytest init import

tigerchensh:
Summary
---------

Remove the unnecessary pytest init which calls prepend import from threatexchange packages. Thus, pytest can correctly check if the import library is not installed and skip. Otherwise, pytest will first prepend import and fail on vpdq import before checking & skipping it. 
(Why pytest relied on init and used prepend import before? See https://github.com/facebook/ThreatExchange/pull/1123)

Test Plan
---------

Before the change (No vpdq installed):
image

After the change:
image

### DIFF
--- a/python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
+++ b/python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
@@ -23,7 +23,7 @@ else:
         dedupe,
         quality_filter,
     )
-    from threatexchange.extensions.vpdq.tests.utils import (
+    from utils import (
         get_random_VPDQs,
         pdq_hashes_to_VPDQ_features,
     )
--- a/python-threatexchange/threatexchange/extensions/vpdq/tests/__init__.py
+++ b/python-threatexchange/threatexchange/extensions/vpdq/tests/__init__.py
@@ -1,1 +1,0 @@
-# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved


python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -23,7 +23,7 @@
         dedupe,
         quality_filter,
     )
-    from threatexchange.extensions.vpdq.tests.utils import (
+    from utils import (
```
BarrettOlson, line 26: blocking: use `threatexchange.extensions.vpdq.tests.utils` instead of just `utils`

nit: I don't love that `threatexchange.tests.hashing.utils` exist as well

BarrettOlson (approved):
LGTM but switch the relative use of `utils` to be absolute before merge, thanks!

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -23,7 +23,7 @@
         dedupe,
```
BarrettOlson, line 23: nit: this file looks like it has a lot of unnecessary imports as well

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
tigerchensh: Mypy doesn't like that because there isn't an init file in the test folder. So, mypy won't check that folder. Solution 1): type ignore here. 2): Add missing import 3) This in-place import 
I think the third one is better.

Dcallies:
Is importskip a potential solution here? https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-on-a-missing-import-dependency

tigerchensh:
https://github.com/facebook/ThreatExchange/pull/1125 should be the fix instead of deleting the __init__

PR #1126, opened 2022-08-01
Pytx vpdq replace init

tigerchensh:
Summary
---------

This should be the correct approach to solve the problem describe in https://github.com/facebook/ThreatExchange/pull/1124.
The root problem in the unempty init which will loads the VideoVPDQSignal and causes import error when testing.
Move the init files's content into manifest.py so pytest won't invoke any imports.
The change is that we do ```threatexchange config extensions add threatexchange.extensions.vpdq.manifest```
instead of ```threatexchange config extensions add threatexchange.extensions.vpdq```

Test Plan
---------

No vpdq installed:
image

Install vpdq and test:
image

CLI:
image


BarrettOlson (commented):
Works for me but I'll leave it to @Dcallies to approve as I believe you discussed this solution with him.

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/extensions/README.md

Does say that `TX_MANIFEST` doesn't have to be in `__init__.py` but it would be helpful to specify that the command then needs more than just the module path for those cases in the section `Testing Your Module from the threatexchange CLI`

PR #1127, opened 2022-08-01
[Pytx] Add support for vPDQ missing ffmpeg

tigerchensh:
Summary
---------

Add support for detecting if ffmpeg is not installed for vPDQ. Skip the test instead of throwing exception. xFail the test if the test files are missing. Anyone who fails the tests would be easier to know the root cause.

Test Plan
---------
With ffmpeg:
image

Without ffmpeg:
image

No test file:
image


Dcallies (changes requested):
I think we should fix this deeper in the stack.

I think if you use the vpdq core library, and ffmpeg is not installed it should either:
1. Throw an exception at import time
2. Throw an exception when you attempt to calculate a hash.

This will make errors of this class "noisy".

tigerchensh:
Thanks for the review. The python-ffmpeg will throw an error ```FileNotFoundError: [Errno 2] No such file or directory: 'ffmpeg'```

tigerchensh:
I have changed the behavior to check ffmpeg when importing vPDQ and raise error.
The pass for Caller CalledProcessError is because command 'ffmpeg' returned non-zero exit status 1 for its help message. And a true missing ffmpeg call will return FileNotFoundError.
image


Test Plan
----
import vpdq
----
image

run pytest and check if it still skips
----
image

install ffmpeg and import vpdq
----
image

vpdq/python/vpdq.pyx
```diff
@@ -3,12 +3,15 @@
...
 import cv2
 import typing as t
+import subprocess
 
 from dataclasses import dataclass
 from libcpp cimport bool
 from libcpp.vector cimport vector
 from libcpp.string cimport string
 
+subprocess.check_call("ffmpeg")
```
Dcallies, line 13: blocking: This could use some more comments explaining what this call means. It otherwise might be mistaken for a undeleted test statement. 

Additionally, this might be a surprising error to a caller. You may want to convert it into another error type with a more specific error message. You could provide installation instructions in the error, for example. 

FileNotFound seems like a reasonable exception to use for this case (ffmpeg's file is literally not found). You may want to catch and throw your own version with more instructions, however.

Dcallies (changes requested):
This seems like a reasonable direction to me, but you should add some more documentation explaining this rogue subprocess call!

tigerchensh:
Thanks for the review. Have thrown a custom exception to handle it.
image

tigerchensh:
Adding more support in vpdq package to detect wrong arguments and give more specific error messages:
1. Check if files exist
2. Check if parameters are within correct range

Dcallies (changes requested):
Did you mean to swallow unexpected exceptions?

vpdq/python/vpdq.pyx
```diff
@@ -3,12 +3,24 @@
...
 from dataclasses import dataclass
 from libcpp cimport bool
 from libcpp.vector cimport vector
 from libcpp.string cimport string
 
+try:
+    # A call to check if ffmpeg is installed for vPDQ
+    # FFMPEG is required to compute vPDQ hash
+    subprocess.check_call("ffmpeg", stdout=subprocess.DEVNULL, 
+    stderr=subprocess.STDOUT)
```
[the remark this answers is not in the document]
tigerchensh: Thanks for pointing out! We actually do not need to.

tigerchensh:
Thanks for the review. I have changed the command call to be "ffmpeg -L" which shows the license information and won't invoke caller exception. The code have more readability now.

vpdq/python/vpdq.pyx
```diff
@@ -3,12 +3,24 @@
...
+except (subprocess.CalledProcessError, FileNotFoundError) as e:
+    if isinstance(e, FileNotFoundError):
+        raise Exception("FFMPEG is not installed. VPDQ requires FFMPEG. Visit https://ffmpeg.org/download.html to install. ")
+    pass
```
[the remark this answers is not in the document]
tigerchensh: Changed to -L, show license.

vpdq/python/vpdq.pyx
```diff
@@ -3,12 +3,21 @@
...
 
 from dataclasses import dataclass
 from libcpp cimport bool
 from libcpp.vector cimport vector
 from libcpp.string cimport string
 
+try:
+    # A call to check if ffmpeg is installed for vPDQ
+    # FFMPEG is required to compute vPDQ hash
+    subprocess.check_call(["ffmpeg", "-L"], stdout=subprocess.DEVNULL)
```
Dcallies, line 16: blocking: also send stderr to DEVNULL

vpdq/python/vpdq.pyx
```diff
@@ -3,12 +3,21 @@
...
+except FileNotFoundError as e:
+    raise Exception("FFMPEG is not installed. VPDQ requires FFMPEG. Visit https://ffmpeg.org/download.html to install. ")
```
Dcallies, line 18: 👍

Dcallies (changes requested):
Almost there!

Can you:
1. Update your test plan with following cases:
     1. py.test with vpdq not installed
     2. py.test with vpdq installed, but ffmpeg not installed
     3. py.test with vpdq and ffmpeg installed 
2. Re-run your test plan to make sure it behaves as you expect

---

Following up on this statement:

> The python-ffmpeg will throw an error FileNotFoundError: [Errno 2] No such file or directory: 'ffmpeg' I will do the same thing to vPDQ library.

It only does this when you call the ffmpeg run() commands. While this solution is a departure from what ffmpeg does (exception on import), this is still okay for me as long as the behavior is predictable/well-defined.

tigerchensh:
With ffmpeg and vpdq
-----
image

With vpdq and no ffmpeg
-----
image

With no vpdq
-----
image

Try import vpdq without ffmpeg
----
image

Try add extension in cli without ffmpeg
----
image

vpdq/python/vpdq.pyx
```diff
@@ -3,12 +3,21 @@
...
 from libcpp cimport bool
 from libcpp.vector cimport vector
 from libcpp.string cimport string
 
+try:
+    # A call to check if ffmpeg is installed for vPDQ
+    # FFMPEG is required to compute vPDQ hash
+    subprocess.check_call(["ffmpeg", "-L"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+except FileNotFoundError as e:
+    raise Exception("FFMPEG is not installed. VPDQ requires FFMPEG. Visit https://ffmpeg.org/download.html to install. ")
```
Dcallies, line 18: nit: trailing whitespace
nit: VPDQ => vPDQ

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
```diff
@@ -10,7 +10,7 @@
     import vpdq
 
     _DISABLED = False
-except (ImportError, ModuleNotFoundError) as e:
+except Exception:
```
Dcallies, line 13: This is probably too generic, since it will catch any issue from vpdq, causing it to fail silently. 

We want to separate "expected" exceptions from "unexpected" exceptions.

Here are two solutions that would work better for me:
1. Throw ImportError, continue using import checks
2. Add test your exception class for a specific known argument so you know it's from the ffmpeg check, and not from anywhere else.

Dcallies (approved):
Not ideal, but we've been here a while, so let's just ship

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
tigerchensh: Have checked to ImportError and specify only ImportError and ModuleNotFoundError.

PR #1129, opened 2022-08-01
[VPDQ] Provide more specific error messages for invalid arguments

tigerchensh:
Summary
---------

Provide more specific error messages for invalid arguments to vPDQ computeHash.
(i.e. "Seconds_per_hash must be non-negative", "Input_video_filename doesn't exist" and etc.)
  

Test Plan
---------
Added test to check if raiseError.


vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -49,6 +49,21 @@ def test_vpdq_utils():
         assert vpdq.hash_to_hex(feature.hash) == hex_hash
 
 
+def test_error_checking():
+    video_file = Path(f"{video_folder}/{TEST_FILES[0]}.mp4")
+    with pytest.raises(ValueError):
```
Dcallies, line 54: blocking: This will only catch one exception, and the rest of these will not be executed. Additionally, if you are using messages, use the message regex feature to exactly match the expected exception

vpdq/python/vpdq.pyx
```diff
@@ -98,7 +99,7 @@ def hamming_distance(hash1: "Hash256", hash2: "Hash256") -> int:
 
 
 def computeHash(
-    input_video_filename: str,
+    input_video_filename: t.Union[str, Path],
```
Dcallies, line 102: Why not just accept one?

vpdq/python/vpdq.pyx
```diff
@@ -117,8 +118,16 @@ def computeHash(
     Returns:
         list of vpdq_feature: VPDQ hash from the video
     """
+    if not Path(input_video_filename).is_file():
```
Dcallies, line 121: If you do want to take either union or str, strongly suggest normalizing to one in the first line. I.e.

```
input_video_filename = Path(input_video_filename )
```

This will make tracking what type you expect much easier

vpdq/python/vpdq.pyx
```diff
@@ -127,7 +136,7 @@ def computeHash(
         downsample_height = vid.get(cv2.CAP_PROP_FRAME_HEIGHT)
 
     rt = hashVideoFile(
-        input_video_filename.encode("utf-8"),
+        str(input_video_filename).encode("utf-8"),
```
Dcallies, line 139: What does this encode do?

Dcallies (changes requested):
Thanks for adding a test!

vpdq/python/vpdq.pyx
tigerchensh: The C++ function takes the encoded str.

tigerchensh:
Thanks for the review! Have adjusted the error tests.

vpdq/python/vpdq.pyx
tigerchensh: I think accepting two will make the function easier to use.

vpdq/python/vpdq.pyx
tigerchensh: Thanks for the suggestion! Have added this.

vpdq/python/vpdq.pyx
Dcallies: By encoded string, do you mean `bytes` / c-style?

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -49,6 +49,31 @@ def test_vpdq_utils():
         assert vpdq.hash_to_hex(feature.hash) == hex_hash
 
 
+def test_error_checking():
+    video_file = Path(f"{video_folder}/{TEST_FILES[0]}.mp4")
+    with pytest.raises(ValueError) as error_info:
+        assert not vpdq.computeHash(
+            input_video_filename=video_file, seconds_per_hash=-1
+        )
```
Dcallies, line 57: blocking: Remove these asserts - if you expect these files to throw errors, your raises() calls with throw if they don't throw the expected exception (try it out).

As is, they are unreachable/dead code, and so might confuse a reader

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -49,6 +49,31 @@ def test_vpdq_utils():
...
+    assert str(error_info.value) == "Seconds_per_hash must be non-negative"
```
Dcallies, line 58: nit: Use match= here instead of .value - https://docs.pytest.org/en/7.1.x/how-to/assert.html

vpdq/python/vpdq.pyx
tigerchensh: Yes

Dcallies (approved):
shipit (see one blocking comment inline)

PR #1130, opened 2022-08-01
[pytx][cli] update outdated help text in match command for --hashes arg

BarrettOlson:
Summary
---------

Help text had an old name for the --hashes flag (--as-hashes) and didn't provide as much context.

TBH this could be changed to `--as-signal` but that it likely to be more confusing than helpful at this point naming is hard.

Test Plan
---------

$ tx config extensions add threatexchange.extensions.vpdq.manifest

before:
$ tx match --hashes video  -- "hi"
Too many SignalTypes for --as-hashes. Use also --only-signal

after:
$ tx match --hashes video  -- "hi"
Error: 'video' supports more than one SignalType. for '--hashes' also use '--only-signal' to specify one of ['video_md5', 'video_vpdq']


Dcallies (approved):
I plan on going through and unifying the arguments to `hash` , `match`, and `label` at some point, including synchronizing their validation.

PR #1131, opened 2022-08-02
Dynamically generate ddb-backed HMAConfigs from CollaborationConfigBase subclasses

schatten:
Summary
---
[tx_apis.py](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/hmalib/common/configs/tx_apis.py) makes it possible for users to add any number of signal exchange APIs.

These may have their own collab config objects. This change allows you to dynamically create such classes and CRUD them using the regular `hmalib.common.config.*` APIs.

Test Plan
---
Thorough unit tests.

Specifically tests FBThreatExchange and NCMEC APIs and verifies their individual attributes were accurately stored and returned.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
@@ -96,8 +96,14 @@ class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
             self.assertEqual(3, len(all_configs))
 
             ncmec_config = next(
-                iter([_ for _ in all_configs if isinstance(_, NCMECCollabConfig)])
+                iter(
+                    [
+                        conf
+                        for conf in all_configs
+                        if isinstance(conf, NCMECCollabConfig)
+                    ]
+                )
             )
             self.assertEqual(ncmec_config.environment, NCMECEnvironment.Exploitative)
-            self.assertEqual(ncmec_config.only_esp_ids, set([1, 2, 3]))
+            self.assertEqual(ncmec_config.only_esp_ids, {1, 2, 3})
             self.assertEqual(ncmec_config.name, "child-safety")
--- a/hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
@@ -1,0 +1,103 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import unittest
+from threatexchange.exchanges.clients import ncmec
+from threatexchange.exchanges.clients.ncmec.hash_api import NCMECEnvironment
+
+from threatexchange.exchanges.impl.fb_threatexchange_api import (
+    FBThreatExchangeCollabConfig,
+)
+from threatexchange.exchanges.impl.ncmec_api import (
+    NCMECCollabConfig,
+)
+
+
+from hmalib.common.models.tests.ddb_test_common import HMAConfigTestBase
+from hmalib.common.configs import tx_collab_config
+
+
+class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
+    TABLE_NAME = f"test-config-{__name__}"
+
+    def test_get_no_collab_config(self):
+        with self.fresh_dynamodb():
+            self.assertEqual([], tx_collab_config.get_all_collab_configs())
+
+    def test_write_one_config(self):
+        with self.fresh_dynamodb():
+            tx_collab_config.create_collab_config(
+                FBThreatExchangeCollabConfig(
+                    name="Threatcollaboration Something", privacy_group=1212312
+                )
+            )
+            all_configs = tx_collab_config.get_all_collab_configs()
+            self.assertEqual(1, len(all_configs))
+            self.assertEqual("Threatcollaboration Something", all_configs[0].name)
+            self.assertEqual(1212312, all_configs[0].privacy_group)
+
+    def test_write_and_retrieve(self):
+        with self.fresh_dynamodb():
+            tx_collab_config.create_collab_config(
+                FBThreatExchangeCollabConfig(
+                    name="Threatcollaboration Something", privacy_group=1212312
+                )
+            )
+            config = tx_collab_config.get_collab_config(
+                FBThreatExchangeCollabConfig, name="Threatcollaboration Something"
+            )
+            self.assertEqual("Threatcollaboration Something", config.name)
+            self.assertEqual(1212312, config.privacy_group)
+
+    def test_write_and_update(self):
+        with self.fresh_dynamodb():
+            tx_collab_config.create_collab_config(
+                FBThreatExchangeCollabConfig(
+                    name="Threatcollaboration Something", privacy_group=1212312
+                )
+            )
+            config = tx_collab_config.get_collab_config(
+                FBThreatExchangeCollabConfig, name="Threatcollaboration Something"
+            )
+            self.assertEqual("Threatcollaboration Something", config.name)
+            self.assertEqual(1212312, config.privacy_group)
+
+            # Now update
+            config.privacy_group = 444333
+            tx_collab_config.update_collab_config(config)
+
+            config = tx_collab_config.get_collab_config(
+                FBThreatExchangeCollabConfig, name="Threatcollaboration Something"
+            )
+            self.assertEqual(444333, config.privacy_group)
+
+    def test_write_multiple_configs(self):
+        with self.fresh_dynamodb():
+            tx_collab_config.create_collab_config(
+                FBThreatExchangeCollabConfig(
+                    name="ThreatCollaboration something", privacy_group=121344
+                ),
+            )
+
+            tx_collab_config.create_collab_config(
+                FBThreatExchangeCollabConfig(
+                    name="ThreatCollaboration something else", privacy_group=678876
+                )
+            )
+
+            tx_collab_config.create_collab_config(
+                NCMECCollabConfig(
+                    name="child-safety",
+                    environment=NCMECEnvironment.Exploitative,
+                    only_esp_ids=set([1, 2, 3]),
+                )
+            )
+
+            all_configs = tx_collab_config.get_all_collab_configs()
+            self.assertEqual(3, len(all_configs))
+
+            ncmec_config = next(
+                iter([_ for _ in all_configs if isinstance(_, NCMECCollabConfig)])
+            )
+            self.assertEqual(ncmec_config.environment, NCMECEnvironment.Exploitative)
+            self.assertEqual(ncmec_config.only_esp_ids, set([1, 2, 3]))
+            self.assertEqual(ncmec_config.name, "child-safety")
--- a/hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
@@ -43,7 +43,7 @@ class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
                 )
             )
             config = tx_collab_config.get_collab_config(
-                FBThreatExchangeCollabConfig, name="Threatcollaboration Something"
+                name="Threatcollaboration Something"
             )
             self.assertEqual("Threatcollaboration Something", config.name)
             self.assertEqual(1212312, config.privacy_group)
@@ -56,7 +56,7 @@ class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
                 )
             )
             config = tx_collab_config.get_collab_config(
-                FBThreatExchangeCollabConfig, name="Threatcollaboration Something"
+                name="Threatcollaboration Something"
             )
             self.assertEqual("Threatcollaboration Something", config.name)
             self.assertEqual(1212312, config.privacy_group)
@@ -66,7 +66,7 @@ class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
             tx_collab_config.update_collab_config(config)
 
             config = tx_collab_config.get_collab_config(
-                FBThreatExchangeCollabConfig, name="Threatcollaboration Something"
+                name="Threatcollaboration Something"
             )
             self.assertEqual(444333, config.privacy_group)
 
@@ -88,7 +88,7 @@ class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
                 NCMECCollabConfig(
                     name="child-safety",
                     environment=NCMECEnvironment.Exploitative,
-                    only_esp_ids=set([1, 2, 3]),
+                    only_esp_ids={1, 2, 3},
                 )
             )
 
--- a/hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
@@ -8,12 +8,13 @@ interfaces require it, keeping it as just another class for now.
 """
 
 from dataclasses import dataclass, asdict, fields
-from enum import Enum
 import typing as t
-import json
-import dacite
 
 from threatexchange.exchanges.collab_config import CollaborationConfigBase
+from threatexchange.utils.dataclass_json import (
+    dataclass_dumps,
+    dataclass_loads,
+)
 
 from hmalib.common.mappings import full_class_name, import_class
 from hmalib.common.config import HMAConfig, create_config, update_config
@@ -27,7 +28,7 @@ def _config_name(clazz: t.Type[CollaborationConfigBase], name: str):
 @dataclass
 class EditableCollaborationConfig(HMAConfig):
     """
-    Store collaboration configs as HMAConfigs. Attriutes of collaboration config
+    Store collaboration configs as HMAConfigs. Attributes of collaboration config
     are stored as serialized json.
 
     Clients are not expected to work with this class directly. Use methods
@@ -45,37 +46,11 @@ class EditableCollaborationConfig(HMAConfig):
     attributes_json_serialized: str
 
 
-class _DaciteJSONEncoder(json.JSONEncoder):
-    """
-    The actual attributes of the CollabConfig subclass are stored using a
-    serialized JSON. Since this JSON will be used with dacite to regenerate the
-    CollabConfig object, do what dacite expects.
-
-    Serialize enums to their values; and sets to lists.
-    """
-
-    def default(self, obj):
-        if isinstance(obj, Enum):
-            return obj.value
-        elif isinstance(obj, set):
-            return [x for x in obj]
-        return json.JSONEncoder.default(self, obj)
-
-
 def _build_collab_config(
     collab_config_class: str, attributes_serialized: str
 ) -> CollaborationConfigBase:
     cls = import_class(collab_config_class)
-    return dacite.from_dict(
-        cls,
-        data={
-            k: v
-            for (k, v) in json.loads(
-                attributes_serialized,
-            ).items()
-        },
-        config=dacite.Config(cast=[Enum, t.Set]),
-    )
+    return dataclass_loads(attributes_serialized, cls)
 
 
 def get_collab_config(
@@ -110,9 +85,7 @@ def create_collab_config(collab_config: CollaborationConfigBase):
         EditableCollaborationConfig(
             name=_config_name(collab_config.__class__, collab_config.name),
             collab_config_class=full_class_name(collab_config.__class__),
-            attributes_json_serialized=json.dumps(
-                asdict(collab_config), cls=_DaciteJSONEncoder
-            ),
+            attributes_json_serialized=dataclass_dumps(collab_config),
         )
     )
 
@@ -129,7 +102,5 @@ def update_collab_config(collab_config: CollaborationConfigBase):
         raise ValueError("EditableCollabConfig object does not exist as an HMAConfig.")
 
     ec.collab_config_class = full_class_name(collab_config.__class__)
-    ec.attributes_json_serialized = json.dumps(
-        asdict(collab_config), cls=_DaciteJSONEncoder
-    )
+    ec.attributes_json_serialized = dataclass_dumps(collab_config)
     update_config(ec)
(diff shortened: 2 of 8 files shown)


hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+DynamoDB backed CollaborationConfig objects.
+
+This may become a CollaborationConfigStoreBase, but since none of the pytx
```
Dcallies, line 6: I've considered deleting that API, so I think it's fine to make your own.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,66 @@
...
+def make_collab_config_subclass(
+    cls: t.Type[CollaborationConfigBase],
+) -> t.Type[HMAConfig]:
+    """
+    Generate an HMAConfig subclass on-the-fly.
+
+    Since supported APIs are dynamic, we need to allow any
+    CollaborationConfigBase to be stored as an HMAConfig object.
+    """
+    return type(f"_Dynamic{cls.__name__}_HMAConfig", (cls, HMAConfig), {})
```
Dcallies, line 28: Oof, defintely not looking forward to debugging this in the future.

Why not 

```
  class _DynamicWhatever(cls, HMAConfig):
     pass
  _DynamicWhatever.__name__ = "name"

  return _DynamicWhatever
```

Which is slightly longer, but easier to spot what's happening

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class CollabConfig(HMAConfig):
```
Dcallies, line 31: Why `is-a` and not `has-a`?

Seems like an HMA config that just contained a collab config would still work, and require less magic. You also wouldn't need to worry about collisions then.

Also, dumb question, if this class isn't a dataclass, does that mess with the `__init__` form?

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Stores and retrives CollaborationConfigBase objects as HMAConfig objects. To
+    allow for extending as CollaborationConfigBase get more complex, only known
+    and common attributes are DDB Attributes. Everything else is dumped after
+    serializing.
+    """
+
+    # The subclass of CollaborationConfigBase that we are representing
+    concrete_class: str
```
Dcallies, line 40: What form is this in? complete module path?

If we have collab_config_api, don't we have this class via `api.get_config_class()`?

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,66 @@
...
+    CollaborationConfigBase to be stored as an HMAConfig object.
+    """
+    return type(f"_Dynamic{cls.__name__}_HMAConfig", (cls, HMAConfig), {})
+
+
+class CollabConfig(HMAConfig):
+    """
+    Stores and retrives CollaborationConfigBase objects as HMAConfig objects. To
+    allow for extending as CollaborationConfigBase get more complex, only known
+    and common attributes are DDB Attributes. Everything else is dumped after
```
Dcallies, line 35: Can you explain this?

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
```diff
@@ -0,0 +1,102 @@
...
+import unittest
+from threatexchange.exchanges.clients import ncmec
+from threatexchange.exchanges.clients.ncmec.hash_api import NCMECEnvironment
+
+from threatexchange.exchanges.impl.fb_threatexchange_api import (
+    FBThreatExchangeSignalExchangeAPI,
+)
+from threatexchange.exchanges.impl.ncmec_api import (
+    NCMECCollabConfig,
+    NCMECSignalExchangeAPI,
```
Dcallies, line 12: This seems comprehensive

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
```diff
@@ -0,0 +1,102 @@
...
+from hmalib.common.configs.tx_collab_config import create_config_for_api, get_all
+
+
+def get_all_apis():
+    return [
+        api.to_concrete_class() for api in ToggleableSignalExchangeAPIConfig.get_all()
+    ]
+
+
+class TXCollabConfigsTestCase(HMAConfigTestBase, unittest.TestCase):
```
Dcallies, line 32: nit: Something I've learned after using py.test more is that unittest.TestCase is not needed for anything, and actually makes it easier to accidentally prevent access to py.test features. You can use fixtures any time you would use class state.

In general, I'd recommend not using unittest classes, and refactored away the class uses in the threatexchange library

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,66 @@
...
+    serializing.
+    """
+
+    # The subclass of CollaborationConfigBase that we are representing
+    concrete_class: str
+
+    # Attributes in CollaborationConfigBase
+    collab_config_name: str
+    collab_config_api: str
+    enabled: bool
```
Dcallies, line 45: Isn't this duplicate of values in collabconfig? Some of them are redefinitions (name => collab_config_name) and some of these will shadow (enabled)

Dcallies (dismissed):
Derp, git ate my first message.

Overall, I just want to talk through alternatives before becoming convinced this is the best one. I'm actually slightly unsure how this works correctly now given my understanding of how dataclass and default fields work, and I can forsee complexities debugging this.

My main question is why we are using "is-a" rather than "has-a" as the main relationship. If the HMA wrapper just had the original config class as an attribute, it seems like there would be less opportunity for future issues, but I think I'm missing context from a lot of problem solving and debugging you've already done.

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
schatten: Agree that unittest.TestCase is not needed, but some of the work I have done with HMAConfigTestBase makes it easier to handle creating mock dynamodb instances and handle some things auto-magically. I'd prefer to continue using them here.

Short of re-doing all of that using fixtures, there is no middle ground that I can think of. Unless it is categorically unsafe to use TestCase, I'd want to continue using it. 

Also, since you marked as a nit, I'm aware of the severity of your stance. :P

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
schatten: Oh, this is an outdated comment. Will update. Thanks for flagging.

Right now, we are storing everything as a DDB attribute.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
schatten: Whoops, this module has stuff that I was experimenting with, but did not end up using. We are not using CollabConfig, it is safe to delete.

I'll push a new version. 

Essentially, for every CollabConfig subclass, we dynamically generate an HMAConfig type using the `make_collab_config_subclass` method above.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
schatten: I guess my Java instincts kicked in. I did not know you could use the class keyword to declare classes at runtime! TIL!

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+    CollabConfig object, do what dacite expects.
+
+    Serialize enums to their values; and sets to lists.
+    """
+
+    def default(self, obj):
+        if isinstance(obj, Enum):
+            return obj.value
+        elif isinstance(obj, set):
+            return [x for x in obj]
```
Dcallies, line 61: nit: `list(obj)`

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+        return json.JSONEncoder.default(self, obj)
```
Dcallies, line 62: nit: `super().default`

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+
+    None if not.
+    """
+    ec = EditableCollaborationConfig.get(_config_name(clazz, name))
+    if ec is None:
+        return None
+    return _build_collab_config(ec.collab_config_class, ec.attributes_json_serialized)
+
+
+def get_all_collab_configs() -> t.List[CollaborationConfigBase]:
```
Dcallies, line 95: Hmm, I'm surprised this isn't a typing error, since List[] isn't covariant and they aren't strictly CollabConfigBase

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+) -> t.Optional[CollaborationConfigBase]:
+    """
+    Get CollaborationConfigBase of class and name if one exists as an HMAConfig.
+
+    None if not.
+    """
+    ec = EditableCollaborationConfig.get(_config_name(clazz, name))
+    if ec is None:
+        return None
+    return _build_collab_config(ec.collab_config_class, ec.attributes_json_serialized)
```
Dcallies, line 92: (unsure): Would this make sense as a method on EditableCollaborationConfig?

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_write_one_config(self):
+        with self.fresh_dynamodb():
+            tx_collab_config.create_collab_config(
+                FBThreatExchangeCollabConfig(
+                    name="Threatcollaboration Something", privacy_group=1212312
+                )
+            )
+            all_configs = tx_collab_config.get_all_collab_configs()
+            self.assertEqual(1, len(all_configs))
+            self.assertEqual("Threatcollaboration Something", all_configs[0].name)
```
Dcallies, line 35: I'm considering enforcing that names are only lower_under in the CLI to make them easier to parse from from bash, and have considered various conventions to make it easier to spot that something was a collab config name.

Having started this road, do you have similar feelings about constraining the name?

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
```diff
@@ -0,0 +1,103 @@
...
+                FBThreatExchangeCollabConfig(
+                    name="ThreatCollaboration something else", privacy_group=678876
+                )
+            )
+
+            tx_collab_config.create_collab_config(
+                NCMECCollabConfig(
+                    name="child-safety",
+                    environment=NCMECEnvironment.Exploitative,
+                    only_esp_ids=set([1, 2, 3]),
```
Dcallies, line 91: nit: {1, 2, 3}

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+) -> CollaborationConfigBase:
+    cls = import_class(collab_config_class)
+    return dacite.from_dict(
+        cls,
+        data={
+            k: v
+            for (k, v) in json.loads(
+                attributes_serialized,
+            ).items()
+        },
```
Dcallies, line 76: nit:
```suggestion
        data=json.loads(attributes_serialized),
```

Dcallies (changes requested):
Overall, this approach works for me. Only RC-ing in case there's more discussion from the questions, ping me for quick turnaround.

Toplevel:
* This PR goes far out of its way to namespace the HMA storage of collab configs by class name. This is surprising to me, since I had initially designed that collab names would be globally unique, and it seems like maintaining that requirement would be simpler than namespacing. 
* Worried about threatexchange.dataclass_json and this class drifting, can we merge them?

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
```diff
@@ -0,0 +1,103 @@
...
+                )
+            )
+
+            all_configs = tx_collab_config.get_all_collab_configs()
+            self.assertEqual(3, len(all_configs))
+
+            ncmec_config = next(
+                iter([_ for _ in all_configs if isinstance(_, NCMECCollabConfig)])
```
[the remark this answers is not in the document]
schatten: iter is so that next works with it. But agree about `_`. Fixing.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class _DaciteJSONEncoder(json.JSONEncoder):
+    """
+    The actual attributes of the CollabConfig subclass are stored using a
+    serialized JSON. Since this JSON will be used with dacite to regenerate the
+    CollabConfig object, do what dacite expects.
+
+    Serialize enums to their values; and sets to lists.
+    """
```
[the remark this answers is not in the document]
schatten: That. Is. An. Excellent. Point. Let me try. 

Not ignorable. :P

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+import json
+import dacite
+
+from threatexchange.exchanges.collab_config import CollaborationConfigBase
+
+from hmalib.common.mappings import full_class_name, import_class
+from hmalib.common.config import HMAConfig, create_config, update_config
+
+
+def _config_name(clazz: t.Type[CollaborationConfigBase], name: str):
```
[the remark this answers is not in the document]
schatten: I blame java. https://stackoverflow.com/questions/2529974/why-do-java-programmers-like-to-name-a-variable-clazz

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,135 @@
...
+    """Use this to make the EditableCollaborationConfig's name unique."""
```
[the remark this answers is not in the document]
schatten: There are some sources of names which are not 'controllable' by the customer. eg. ThreatExchange privacy groups. Should a second such universe come into being, perhaps a new hash-exchange, customers will not be able to use collaborations with the same name.

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
schatten: bash has a need to quote strings with spaces. The same problem does not exist on web forms and the like. I don't see HMA users running into an issue with this.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
schatten: Yeah, left over from when I tried something. Fixing.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Use these attributes to instantiate a collab_config
+    attributes_json_serialized: str
+
+
+def _build_collab_config(
+    collab_config_class: str, attributes_serialized: str
+) -> CollaborationConfigBase:
+    cls = import_class(collab_config_class)
+    return dataclass_loads(attributes_serialized, cls)
```
schatten, line 53: using dataclass_loads instead of relying on recreating dacite wrappers.

hasher-matcher-actioner/scripts/set_threatexchange_source
```diff
@@ -43,12 +43,24 @@ elif [ $1 == "local" ]; then
...
+    find . -type f \( \
+        -name "*.py" \
+        -o -name "*.rst" \
+        -o -name "*.md"  \
+        -o -name "*.txt" \
+        \) -exec cp --parents {} ../hasher-matcher-actioner/$LOCAL_LIB_LOCATION \;
+    popd
+
+    rm -r $LOCAL_LIB_LOCATION/threatexchange.egg-info
+
```
schatten, line 58: Should have made this into its own PR. Essentially this reduces the time it takes to copy pytx files for local installation of the threatexchange package by only copying the required files. 

Earlier, because of pyc files and sample files, it was taking far longer.

python-threatexchange/threatexchange/cli/tests/hash_cmd_test.py
```diff
@@ -46,11 +46,12 @@ def test_stdin(
...
-    hash_cli.assert_cli_output(
-        ("url", "-"),
-        "url_md5 6d3af727a4e7b025fd59a5469b3a9c57",
-    )
+    with tmp_file.open() as fake_stin:
+        monkeypatch.setattr("sys.stdin", fake_stin)
+        hash_cli.assert_cli_output(
+            ("url", "-"),
+            "url_md5 6d3af727a4e7b025fd59a5469b3a9c57",
+        )
```
schatten, line 54: I didn't do this change. Probably a rebase thing..

Either way, once #1157 get merged, I'll rebase this again and figure out what to do here.

hasher-matcher-actioner/hmalib/common/configs/tests/test_tx_collab_config.py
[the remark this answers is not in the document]
Dcallies: `next` works directly with a generator! i.e. `next(iter([i for i in range(5)]))` => `next(i for i in range(5))`, try it!

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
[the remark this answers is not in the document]
Dcallies: blocking q: Why aren't they controllable by the customer? They are controllable in the CLI, why not make them controllable in HMA?

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,106 @@
...
+    Update an existing CollaborationConfigBase as HMAConfig.
+    """
+    ec = EditableCollaborationConfig.get(
+        _config_name(collab_config.__class__, collab_config.name)
+    )
+
+    if ec is None:
+        raise ValueError("EditableCollabConfig object does not exist as an HMAConfig.")
+
+    ec.collab_config_class = full_class_name(collab_config.__class__)
```
Dcallies, line 104: Does this mean you can change the class of the stored aspect with this method? Should that be an error? (Force you to delete & recreate to switch classes?)

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _build_collab_config(
+    collab_config_class: str, attributes_serialized: str
+) -> CollaborationConfigBase:
+    cls = import_class(collab_config_class)
+    return dataclass_loads(attributes_serialized, cls)
+
+
+def get_collab_config(
```
Dcallies, line 56: (devil's advocate): Why not a method on EditableCollaborationConfig? Same with the others as well.

python-threatexchange/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
```
Dcallies, line 1: Hmm? Are you based on a weird version? https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/pytest.ini

python-threatexchange/threatexchange/cli/tests/hash_cmd_test.py
Dcallies: Yeah, commented on another line.

Dcallies (changes requested):
Looking good! Only same toplevel question as last time:

This PR goes far out of its way to namespace the HMA storage of collab configs by class name. This is surprising to me, since I had initially designed that collab names would be globally unique, and it seems like maintaining that requirement would be simpler than namespacing.

Your comment said that for ThreatExchange, users can't choose the collab name, which is maybe true of HMA right now, and are you making an argument we keep it that way rather than allow users to change the names to get rid of namespacing? I can see arguments for both, and I think it comes down to the user experience that makes the most sense for HMA, which I haven't thought enough about.

Also note that collab names are mutable in ThreatExchange.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
schatten: Yes. You can change the class of the stored object. Unless we do, we'll also need to check that the serialized json can be unserialized.

Since collab names are unique, I want to allow users to fully control the type of config stored against a key. I would like to keep the class mutable.

hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -0,0 +1,97 @@
...
+    return dataclass_loads(attributes_serialized, cls)
+
+
+def get_collab_config(name: str) -> t.Optional[CollaborationConfigBase]:
+    """
+    Get CollaborationConfigBase of class and name if one exists as an HMAConfig.
+
+    None if not.
+    """
+    ec = EditableCollaborationConfig.get(name)
```
schatten, line 57: @Dcallies No longer using class name in collab config names.

Dcallies (approved):
shipit

schatten:
Rebased on main a couple of times. The decisions made when @Dcallies approved have not been changed. The history may feel otherwise because of force-pushes.

PR #1139, opened 2022-08-04
[pytx][dist][stacked 3/3] IndexMatch now uses dist

Dcallies:
Summary
---------

Woof, that was a lot harder than expected due to lots of missing types in pdq index.

It's possible some of the tricky typing things aren't actually needed here - the extra typing of IndexMatch for example with Covariant might be egregious. Howevever, I was just shotgunning the typechecker

Test Plan
---------

mypy && py.test


python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -119,15 +120,18 @@ class IndexMatch(t.Generic[T]):
     NamedTuple can't be made generic, so here's an equivalent
```
Dcallies, line 120: Apparently this was fixed in 3.6 and this class can be made NamedTuple

python-threatexchange/threatexchange/signal_type/pdq/pdq_index.py
```diff
@@ -55,7 +60,12 @@ def query(self, hash: str) -> t.List[IndexMatch[IndexT]]:
 
         matches = []
         for id, _, distance in results[hash]:
-            matches.append(IndexMatch(distance, self.local_id_to_entry[id][1]))
+            matches.append(
+                IndexMatchUntyped(
+                    SignalSimilarityInfoWithSingleDistance(int(distance)),
```
BarrettOlson, line 65: nit: should this be `SignalSimilarityInfoWithIntDistance` ?

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -119,15 +120,18 @@ class IndexMatch(t.Generic[T]):
...
     """
 
-    __slots__ = ["distance", "metadata"]
-    distance: int
+    __slots__ = ["match_info", "metadata"]
+    match_info: S_Co
```
BarrettOlson, line 124: nit: similarity_info instead, less easily confused with what metadata is

first_field (match_info) -> context on how close the match is
second_field (metadata) -> details of the actaully matched object

naming is hard

offline side conversation: should some of this be type enforced? (follow up PR)

BarrettOlson (approved):
LGTM, nothing blocking but there are a few follow up details for 4/3 or 5/3  PRs :P

Dcallies:
Doing as followups:
1. rename field similarity_info
2. convert to NamedTuple

PR #1145, opened 2022-08-05
[VPDQ] Update vpdq-release.py to only generate source distribution

tigerchensh:
Summary
---------
Update vpdq-release script to only build source distribution since linux platform build wheel cannot be uploaded to PyPI.


BarrettOlson (approved):
LGTM,
Did this work when you published locally?

tigerchensh:
Yes, I have already locally published the 0.0.5 version to match repo since I am using a macos instead of linux.

PR #1147, opened 2022-08-05
[pytx] SignalType now abstract compare_hash moved to FileHasher

Dcallies:
Summary
---------

Various followups from the discussion on distance, that I was bad about keeping separate:
1. match_info => similarity_info
2. Comparing signals now lives in either MatchesStr or FileHasher, since those are the only contexts where it makes sense to do so.

Test Plan
---------

mypy && py.test

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/signal_base.py
+++ b/python-threatexchange/threatexchange/signal_type/signal_base.py
@@ -1,10 +1,10 @@
-#!/usr/bin/env python
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
 """
 Core abstractions for signal types.
 """
 
+import abc
 import pathlib
 import typing as t
 
@@ -53,7 +53,7 @@ class SignalComparisonResult(t.NamedTuple):
         )
 
 
-class SignalType:
+class SignalType(abc.ABC):
     """
     Abstraction for different signal types.
 
@@ -76,11 +76,13 @@ class SignalType:
         return common.class_name_to_human_name(cls.__name__, "Signal")
 
     @classmethod
+    @abc.abstractmethod
     def get_content_types(cls) -> t.List[t.Type[content_base.ContentType]]:
         """Which content types this Signal applies to (usually just one)"""
-        raise NotImplementedError
+        pass
 
     @classmethod
+    @abc.abstractmethod
     def get_index_cls(cls) -> t.Type[index.SignalTypeIndex]:
         """Return the index class that handles this signal type"""
         # Confused about which one to start with?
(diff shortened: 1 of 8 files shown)


python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -143,29 +148,31 @@ class TextHasher(FileHasher):
...
         return cls.hash_from_str(file.read_text())
 
 
-class MatchesStr:
+class MatchesStr(abc.ABC):
     @classmethod
+    @abc.abstractmethod
     def matches_str(cls, signal: str, haystack: str) -> SignalComparisonResult:
         """
         Compare the distance of two hashes, the key operation for matching.
```
BarrettOlson, line 170: nit/ignorable: naming is hard, it a little confusing that this doc string is the exact same as they one used for `compare_hash` might be worth updating one / clarifying the difference

python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -143,29 +148,31 @@ class TextHasher(FileHasher):
...
 
         Note that this can just be a reference/helper, and the efficient
         version of the algorithm can live in the index class.
         """
-        raise NotImplementedError
+        pass
```
BarrettOlson, line 175: does the addition of `@abc.abstractmethod` mean this will still throw or is this a logic change as well?

python-threatexchange/threatexchange/signal_type/signal_base.py
```diff
@@ -230,7 +238,7 @@ class TrivialLinearSearchHashIndex(index.SignalTypeIndex[index.T]):
 
     # You'll have to override with each usecase, because I wasn't sure
     # If pickle would behave expectedly here
-    _SIGNAL_TYPE: t.Type[SignalType]
+    _SIGNAL_TYPE: t.Type[FileHasher]
```
BarrettOlson, line 241: ignore/nonblocking: feels weird. update comment?

python-threatexchange/threatexchange/signal_type/tests/signal_type_test_helper.py
```diff
@@ -3,7 +3,12 @@
 import pytest
 import typing as t
 
-from threatexchange.signal_type.signal_base import SignalType
+from threatexchange.signal_type.signal_base import (
+    FileHasher,
+    MatchesStr,
```
BarrettOlson, line 8: nit: is this needed/used in the test?

BarrettOlson (approved):
Thanks for the rename, LGTM
The moving of `compare_hash` is a little confusing but your reasoning is sound so LGTM as well

python-threatexchange/threatexchange/signal_type/signal_base.py
Dcallies: Even better! It now fails typing and also throws on class construction.

PR #1152, opened 2022-08-06
[pytx][fb_threatexchange] Fix unxpected reaction format

Dcallies:
Summary
---------

We are parsing reactions wrong, but personally I'm kinda baffled by why it's returning in this format, since I don't think it's stored in that format. Just parse it correctly for the time being.

Test Plan
---------

Successfully parsed td id 4132471233453431 with multiple reactions.

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
@@ -146,6 +146,7 @@ class FBThreatExchangeIndicatorRecord(state.FetchedSignalMetadata):
                 td_id,
             )
 
+            # See: https://developers.facebook.com/docs/threat-exchange/reference/apis/reaction-type/
             for reaction in td_json.get("reactions", []):
                 rxn = reaction["key"]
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
```diff
@@ -148,15 +148,19 @@ def from_threatexchange_json(
...
-                owner = int(reaction["value"])
-                if rxn == "HELPFUL":
-                    implicit_opinions[
-                        owner
-                    ] = state.SignalOpinionCategory.POSITIVE_CLASS
-                elif rxn == "DISAGREE_WITH_TAGS" and owner not in implicit_opinions:
-                    implicit_opinions[
-                        owner
-                    ] = state.SignalOpinionCategory.NEGATIVE_CLASS
+                for owner in reaction["value"].split(","):
```
schatten, line 151: Can you add this link somewhere? 

It gets hard to understand the diff and the code without knowing what the JSON structure is.

https://developers.facebook.com/docs/threat-exchange/reference/apis/reaction-type/v14.0

PR #1153, opened 2022-08-06
[pytx] Add collab config to convert to signal

Dcallies:
Summary
---------

Adds the collab config to converting to signal type, which allows us to filter after the state has been fetched in places like the NCMEC API without worrying about the state being invalid.

To keep the diff small, only update the interface.

At this point, every single method of SignalExchangeAPI now takes collab. We may want to consider making it standard that one API instance = for one collab, which prevents from needing to pass it around everywhere. 

If we choose to do so, it's potentially right now, so need to sync with @schatten .

Test Plan
---------

mypy && py.test

### DIFF
--- a/python-threatexchange/threatexchange/cli/tests/fake_extension.py
+++ b/python-threatexchange/threatexchange/cli/tests/fake_extension.py
@@ -88,6 +88,7 @@ class FakeSignalExchange(
     def naive_convert_to_signal_type(
         cls,
         signal_types: t.Sequence[t.Type[SignalType]],
+        collab: FakeCollabConfig,
         fetched: t.Mapping[str, str],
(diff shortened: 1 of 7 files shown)


schatten (approved):
Approving to unblock. 

Can you give an example of a `convert_to_signal_type` call which would return different results based on the collab config passed? 

If such a use case exists (I'm sure it does, but can't put my finger on it), LGTM.

schatten:
> Can you give an example of a `convert_to_signal_type` call which would return different results based on the collab config passed?

Got my answer here. https://github.com/facebook/ThreatExchange/pull/1155/files#diff-3adadf03287e64b2af40db681e70857e7790e1fea22278a6e98d4f9106a0d411R320

Thanks!

PR #1155, opened 2022-08-06
[pytx][ncmec][stacked] Fix only esp ids

Dcallies:
Summary
---------

The implementation of this feature seems to have disappeared after being added in #1079,  #1118 looks like it removed it by accident, and without testing it was missed.

Test Plan
---------

Added a test to try and catch regressions.


python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -316,6 +317,8 @@ def naive_convert_to_signal_type(
         for entry in fetched.values():
             if entry.deleted:
                 continue  # We expect len(fingerprints) == 0 here, but to be safe
+            if collab.only_esp_ids and entry.member_id not in collab.only_esp_ids:
```
schatten, line 320: This seems to be the only change. It also addresses my question in #1153

schatten (approved):
Looks good!

PR #1156, opened 2022-08-06
[pytx] tie SignalExchangeAPI  instantiation with collab + authentication 

Dcallies:
Summary
---------

Potentially the last interface change for 1.0.0. Put up for discussion with @schatten 

Test Plan
---------

mypy && py.test

```
$ tx --factory-reset

$ tx config api fb_threatexchange -L
Couldn't authenticate fb_threatexchange, it's missing authentification!
You can fix this by:
  1. threatexchange config api fb_threatexchange --api-token
  2. populating an environment variable called TX_ACCESS_TOKEN
  3. creating (and maybe `chmod 400`) a file with credentials at ~/.txtoken

$ TX_ACCESS_TOKEN='a' tx config api fb_threatexchange -L
Invalid auth for fb_threatexchange: Invalid credentials from environment variable TX_ACCESS_TOKEN

$ TX_ACCESS_TOKEN="$(cat ~/.txtoken.bak)" tx config api fb_threatexchange -L
<works>
```

NCMEC

```
$ tx config collab edit ncmec -C ncmec --environment test_NGO

$ tx fetch
Couldn't authenticate ncmec, it's missing authentification!
You can fix this by:
  1. threatexchange config api ncmec --user --pass
  2. populating an environment variable called TX_NCMEC_CREDENTIALS
  3. creating (and maybe `chmod 400`) a file with credentials at ~/.tx_ncmec_credentials


$ tx fetch
$ code -r ~/.tx_ncmec_credentials
$ chmod 400 ~/.tx_ncmec_credentials
[ncmec] ncmec - Fetching... 
<Gets things>
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -17,23 +17,20 @@ import logging
 
 from dacite import WrongTypeError
 
-from threatexchange.exchanges.clients.stopncii import api as stopncii_api
 from threatexchange.exchanges import collab_config
 from threatexchange.exchanges.impl.stop_ncii_api import StopNCIICredentials
 from threatexchange.exchanges.signal_exchange_api import (
     SignalExchangeAPI,
     TCollabConfig,
-    TSignalExchangeAPI,
     TSignalExchangeAPICls,
 )
 from threatexchange.content_type import content_base
-from threatexchange.exchanges.fetch_state import FetchedStateStoreBase
 from threatexchange.exchanges import fetch_state
 from threatexchange.exchanges.impl.static_sample import StaticSampleSignalExchangeAPI
 from threatexchange.signal_type import signal_base
 from threatexchange.interface_validation import FunctionalityMapping
 from threatexchange.cli.cli_state import CliSimpleState, CliIndexStore
-from threatexchange.utils import dataclass_json as cli_json
+from threatexchange.utils import dataclass_json
 
 
 CONFIG_FILENAME = "config.json"
@@ -104,12 +101,12 @@ class CliState(collab_config.CollaborationConfigStoreBase):
         return self.collab_dir / f"{config.name}.json"
 
     def get_persistent_config(self) -> CLiConfig:
-        return cli_json.dataclass_load_file(
+        return dataclass_json.dataclass_load_file(
             self.config_file, CLiConfig, default=CLiConfig()
         )
 
     def update_persistent_config(self, config: CLiConfig):
-        cli_json.dataclass_dump_file(self.config_file, config)
+        dataclass_json.dataclass_dump_file(self.config_file, config)
 
     def dir_for_fetched_state(
         self,
@@ -147,7 +144,7 @@ class CliState(collab_config.CollaborationConfigStoreBase):
                         logging.warning("Ignoring collab config of unknown type: %s", f)
                         continue
                 try:
-                    config = cli_json.dataclass_load_dict(content, ctype)
+                    config = dataclass_json.dataclass_load_dict(content, ctype)
                     ret.append(config)
                 except WrongTypeError:
                     logging.exception("Failed to parse collab config: %s", f)
@@ -158,7 +155,7 @@ class CliState(collab_config.CollaborationConfigStoreBase):
         """Create or update a collaboration"""
         assert collab.api, "didn't set API?"
         path = self.path_for_collab_config(collab)
-        cli_json.dataclass_dump_file(path, collab)
+        dataclass_json.dataclass_dump_file(path, collab)
 
     def delete_collab(self, collab: collab_config.CollaborationConfigBase) -> None:
         """Delete a collaboration"""
--- a/python-threatexchange/threatexchange/cli/cli_config.py
+++ b/python-threatexchange/threatexchange/cli/cli_config.py
@@ -19,9 +19,16 @@ from dacite import WrongTypeError
 
 from threatexchange.exchanges.clients.stopncii import api as stopncii_api
 from threatexchange.exchanges import collab_config
-from threatexchange.exchanges.signal_exchange_api import SignalExchangeAPI
+from threatexchange.exchanges.impl.stop_ncii_api import StopNCIICredentials
+from threatexchange.exchanges.signal_exchange_api import (
+    SignalExchangeAPI,
+    TCollabConfig,
+    TSignalExchangeAPI,
+    TSignalExchangeAPICls,
+)
 from threatexchange.content_type import content_base
 from threatexchange.exchanges.fetch_state import FetchedStateStoreBase
+from threatexchange.exchanges import fetch_state
 from threatexchange.exchanges.impl.static_sample import StaticSampleSignalExchangeAPI
 from threatexchange.signal_type import signal_base
 from threatexchange.interface_validation import FunctionalityMapping
@@ -33,26 +40,12 @@ CONFIG_FILENAME = "config.json"
 
 
 @dataclass
-class StopNCIIKeys:
-    subscription_key: str
-    fetch_function_key: str
-    # Someday, additional keys
-
-    @property
-    def keys_are_valid(self):
-        return all(
-            stopncii_api.is_valid_key(k)
-            for k in (self.subscription_key, self.fetch_function_key)
-        )
-
-
-@dataclass
 class CLiConfig:
     """A place to store misc configuration for the CLI"""
 
     fb_threatexchange_api_token: t.Optional[str] = None
     ncmec_credentials: t.Optional[t.Tuple[str, str]] = None
-    stop_ncii_keys: t.Optional[StopNCIIKeys] = None
+    stop_ncii_keys: t.Optional[StopNCIICredentials] = None
     extensions: t.Set[str] = field(default_factory=set)
     # Every item needs a default for backwards compatibility
 
@@ -178,15 +171,25 @@ class _SignalExchangeAccessor:
 
     _parent: "CLISettings"
 
-    def get_all(self) -> t.ValuesView[SignalExchangeAPI]:
+    def get_all(self) -> t.ValuesView[TSignalExchangeAPICls]:
         return self._parent._mapping.exchange.api_by_name.values()
 
-    def get_for_collab(
-        self, collab: collab_config.CollaborationConfigBase
-    ) -> SignalExchangeAPI:
-        return self._parent._mapping.exchange.api_by_name[collab.api]
+    def get_instance_for_collab(
+        self, collab: TCollabConfig
+    ) -> SignalExchangeAPI[
+        TCollabConfig,
+        fetch_state.FetchCheckpointBase,
+        fetch_state.FetchedSignalMetadata,
+        t.Any,
+        t.Any,
+    ]:
+        api_cls = self._parent._mapping.exchange.api_by_name[collab.api]
+        # This is a hard typing challenge, and cast() doesn't seem to work
+        # By lookup by name, we are guaranteed to pick an API with the right
+        # typing signature, but of an unknown class
+        return api_cls.for_collab(collab)  # type: ignore[arg-type]
 
-    def __iter__(self) -> t.Iterator[SignalExchangeAPI]:
+    def __iter__(self) -> t.Iterator[TSignalExchangeAPICls]:
         yield from self.get_all()
 
 
@@ -209,9 +212,7 @@ class _FetchStoreAccessor:
     def get_for_collab(
         self, collab: collab_config.CollaborationConfigBase
     ) -> CliSimpleState:
-        return self.get_for_api(
-            self._parent._mapping.exchange.api_by_name[collab.api].__class__
-        )
+        return self.get_for_api(self._parent._mapping.exchange.api_by_name[collab.api])
 
 
 class CLISettings:
--- a/python-threatexchange/threatexchange/cli/config_cmd.py
+++ b/python-threatexchange/threatexchange/cli/config_cmd.py
@@ -582,7 +582,7 @@ class ConfigThreatExchangeAPICommand(command_base.Command):
         self.action(settings)
 
     def get_te_api(self) -> ThreatExchangeAPI:
-        creds = FBThreatExchangeCredentials.get()
+        creds = FBThreatExchangeCredentials.get(FBThreatExchangeSignalExchangeAPI)
         return ThreatExchangeAPI(creds.api_token)
 
     def execute_list_collabs(self, settings: CLISettings) -> None:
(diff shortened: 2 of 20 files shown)


python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -63,6 +63,38 @@ class SignalExchangeAPI(
 
     """
 
+    @classmethod
+    @abstractmethod
+    def for_collab(cls, collab: TCollabConfig) -> "SignalExchangeAPI":
+        """
```
schatten, line 69: I am fully behind this change!

I'm slightly concerned that there will now need to be fallbacks for credentials. eg. Does my collab config have credentials? Yes, use them. No, does my API have credentials?

But, I can think of ways around that in HMA. 

How are you planning on storing credentials in threatexchange-cli after this change goes in? Separately for each collab config?

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -63,6 +63,38 @@ class SignalExchangeAPI(
...
+        An instance provided by this class is expected to:
+        1. Be fully authenticated or throw an exception
+        2. All methods are callable if supported by the base implementation
+
+        For implementations that do need additional configuration or
+        authentification, this method is a good place to attempt to pull in
+        default authentification from known locations. Your general options
+        are:
+        1. From an environment variable
+        2. From a file in a known location (ideally chmod 500)
```
schatten, line 81: question: why not 400? Why would someone need to execute a credentials file?

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
Dcallies: Chatted offline, conclusions:
1. Let's go ahead and refactor to for_collab(cls, collab)
2. Explicitly rejected (for now) `for_collab(cls, collab, only_signals)` or `for_collab(cls, collab, auth=None)`, though solving reflection-based auth is something @schatten is interested in solving for HMA, and so might be a 1.1.x feature.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -277,3 +314,148 @@ def naive_convert_to_signal_type(
...
+            lines.append("You can fix this by:")
+            for i, hint in enumerate(self.hints, 1):
+                lines.append(f"  {i}. {hint}")
+        return "\n".join(lines)
+
+
+CredentialSelf = t.TypeVar("CredentialSelf", bound="CredentialHelper")
+
+
+class CredentialHelper:
```
Dcallies, line 370: Here's the majority of the new stuff.

The long term goal here is that we can use something like

```
for_collab(cls: t.Type[Self], collab: CollatT, auth: AuthT = None) -> t.Self:
```
And use reflection on AuthT to auto-generate options on the CLI.

This PR was already a nightmare to produce, and I want to avoid any stabs at fixing anything other than the interface on this pass itself.

Adding this arg won't be backwards compatible though, and the alternative was using Schatten's favorite: global state!

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -277,3 +314,148 @@ def naive_convert_to_signal_type(
...
+    """
+    Wrapper to help standardize credential sources, and tie to exceptions
+    """
+
+    API_CLS: t.ClassVar[t.Type[SignalExchangeAPI]]
+    ENV_VARIABLE: t.ClassVar[str] = ""
+    FILE_NAME: t.ClassVar[str] = ""
+
+    _DEFAULT: t.ClassVar[t.Optional["CredentialHelper"]] = None  # t.Self
+    _DEFAULT_SRC: t.ClassVar[str] = ""
```
Dcallies, line 380: The global state

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -277,3 +314,148 @@ def naive_convert_to_signal_type(
...
+        raise ex
+
+    @classmethod
+    def set_default(
+        cls: t.Type[CredentialSelf], new_default: t.Optional[CredentialSelf], src: str
+    ) -> contextlib.AbstractContextManager:
+        """
+        Set the default (highest preferred) credentials manually.
+
+        They won't be checked for validity until a future get()
```
Dcallies, line 418: Maybe note the contextlib stuff

python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
```diff
@@ -84,29 +82,26 @@ class StopNCIISignalExchangeAPI(
...
-            )
+        self.collab = collab
+        self.api = client
 
-    @property
-    def api(self) -> api.StopNCIIAPI:
-        if self._api is None:
-            raise Exception("StopNCII.org access tokens not configured.")
-        return self._api
+    @classmethod
```
schatten, line 92: blocking: We're using classmethod here, but the type in SignalExchangeAPI is staticmethod.

What's mypy good for? 🤷🏼

python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
```diff
@@ -1,12 +1,10 @@
...
 
-"""
-SignalExchangeAPI impl StopNCII.org
-
-"""
+"""SignalExchangeAPI implementation for StopNCII.org"""
 
 
 from functools import lru_cache
+import os
```
schatten, line 7: Unused import.

python-threatexchange/threatexchange/exchanges/impl/file_api.py
```diff
@@ -47,6 +47,7 @@ class FileCollaborationConfig(
     )
 
 
+@dataclass
```
schatten, line 50: nit: just to get the automatic init method?

Is there anything else 'dataclassey' about this class?

python-threatexchange/threatexchange/cli/main.py
```diff
@@ -134,133 +150,47 @@ def execute_command(settings: CLISettings, namespace) -> None:
...
+        except SignalExchangeAPIMissingAuthException as ma:
+            logging.exception("Original missing auth error")
+            raise CommandError.user(ma.pretty_str()) from ma
 
 
 class _ExtendedTypes(t.NamedTuple):
     content_types: t.List[t.Type[ContentType]]
     signal_types: t.List[t.Type[SignalType]]
-    api_instances: t.List[SignalExchangeAPI]
+    api_instances: t.List[t.Type[SignalExchangeAPI]]
```
schatten, line 193: nit: no longer instances then. api_instances → api_classes?

schatten (approved):
Only nits. 1 blocking issue where for_collab is defined as classmethods in subclasses, but as static method in interface.

python-threatexchange/threatexchange/exchanges/impl/file_api.py
Dcallies: The main thing dataclass gets you is an init method :P  'dataclassey' and "data class" don't have any formal definition to my knowledge, in which case one way we could define it is simply by what the dataclass decorator does as side effects, in which case automatic `__init__` seems like a dataclassey outcome.

python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
Dcallies: I went back and forth on whether it should be static or class (mostly trying to avoid the devilish complexity of typing the class correctly). 

Will catch you in followup conversation, but currently tending towards classmethod if we do follow through with auth.

mpypy doesn't care because the thing that comes out the other end is typing equivalent - same inputs and same outputs.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -7,7 +7,12 @@
 """
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from contextlib import contextmanager
```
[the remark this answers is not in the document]
Dcallies: vscode autocompletes and 900 PR revisions later

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -53,16 +58,50 @@ class SignalExchangeAPI(
...
     valid key, and the value can be the entire dataset.
 
-    = On Owner IDs =
-    Some APIs may not have any concept of owner ID (because the owner is the
-    API itself). In that case, it's suggested to use 0 for all IDs. If the you
-    own the API, it may also make sense to have 0 be the return of
-    get_own_owner_id, which is used by matching to tell if a signal be
-    considered "confirmed" or not. If your API does support the concept of
-    owners, make sure to implement resolve_owner and get_own_owner_id
+    = On Authentification =
```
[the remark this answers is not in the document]
Dcallies: Owner IDs are gone from a previous PR, missed this doc

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -53,16 +58,50 @@ class SignalExchangeAPI(
...
+    and in general, auth should be passible by the constructor.
+
+    If an API is constructed via the for_collab classmethod constructor,
+    it should attempt to authenticate itself via discovering it from the
+    execution environment and passing it via __init__(). If constructed
+    directly, it should not search for credentials.
 
     """
 
+    @staticmethod
```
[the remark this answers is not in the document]
Dcallies: I ran into same problem - either works, and apparently mypy doesn't care either.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -53,16 +58,50 @@ class SignalExchangeAPI(
...
+    @abstractmethod
+    def for_collab(collab: TCollabConfig) -> "SignalExchangeAPI":
```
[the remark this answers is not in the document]
Dcallies: Sadly, that solution doesn't work on the classvariable level without doing Generic[Self] as the TVariable, which looked so strange that I didn't even try and see if mypy would accept it.

Dcallies (commented):
Discussed offline:

1. @classmethod fine
2. We know we are going to want to have a typed auth
3. We want to get rid of the global state
4. 2+3 means we have to pass in auth
5. 4 means we have to have class typed auth
6. Lack of implementation support means that auth must be nullable
7. Backwards compatibility means that we are going to want to add the argument now (1.0.0)
8. Which means that we have to add it into the SignalExcangeAPI typing now

So... this lands, then I have one more PR, then 1.0.0

python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -178,15 +168,25 @@ class _SignalExchangeAccessor:
...
+        fetch_state.FetchCheckpointBase,
+        fetch_state.FetchedSignalMetadata,
+        t.Any,
+        t.Any,
+    ]:
+        api_cls = self._parent._mapping.exchange.api_by_name[collab.api]
+        # This is a hard typing challenge, and cast() doesn't seem to work
+        # By lookup by name, we are guaranteed to pick an API with the right
+        # typing signature, but of an unknown class
+        return api_cls.for_collab(collab)  # type: ignore[arg-type]
```
schatten, line 187: ack that CLI will use default args for all `SignalExchangeAPI.for_collab` invocations. 

I think HMA can reasonably start out with that. We do not need to address this to fix trunk, can add in 'editablity' later.

python-threatexchange/threatexchange/cli/config_cmd.py
```diff
@@ -428,16 +426,7 @@ def execute_add(self, settings: CLISettings) -> None:
         )
 
         # For APIs, we also need to make sure they can be instanciated without args for the CLI
```
schatten, line 428: Outdated comment? I don't think we need a no-args constructor anymore.

python-threatexchange/threatexchange/exchanges/impl/file_api.py
schatten: It also adds hash and eq methods and the ability to freeze the object. Which useful for objects that are serialized or passed along as data -- sans behavior.

But, I'm probably reading in too much from kotlin dataclasses which has a more formal defn.

python-threatexchange/threatexchange/exchanges/impl/file_api.py
```diff
@@ -58,20 +59,25 @@ class LocalFileSignalExchangeAPI(
...
+
     @classmethod
     def get_config_class(cls) -> t.Type[FileCollaborationConfig]:
         return FileCollaborationConfig
 
+    @classmethod
+    def for_collab(
+        cls, collab: FileCollaborationConfig
+    ) -> "LocalFileSignalExchangeAPI":
+        return cls(collab)
```
schatten, line 72: Simple. I like it.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -138,18 +138,29 @@ class NCMECSignalExchangeAPI(
...
         super().__init__()
+        self.collab = collab
         self._username = username
         self._password = password
 
+    @classmethod
+    def for_collab(
+        cls,
+        collab: NCMECCollabConfig,
+        credentials: t.Optional["NCMECCredentials"] = None,
```
schatten, line 159: Yay! Backwards compatible.

schatten: I think we arrived at a solution that lets HMA add support for editable credentials gradually. I like this.

schatten (approved):
Things are as I'd expect after our discussion.

blocking question: Can you remind me how I can get the 'shape' of credentials required? 

eg. How do I know that StopNCIISignalExchangeAPI.for_collab needs a collab arg of type StopNCIICredentials?

Approving because not a land blocker.

Dcallies:
> blocking question: Can you remind me how I can get the 'shape' of credentials required?
> 
> eg. How do I know that StopNCIISignalExchangeAPI.for_collab needs a collab arg of type StopNCIICredentials?
> 
> Approving because not a land blocker.

The plan is to make it the same as config_class. This diff doesn't have that change because it requires I retype everything, which will generate a lot more changes. Coming to you in a followup.

I was, however, able to sneak in the "new" signature for subclasses, since it's substitutable (Liskov substitution). Typing is fixed there because if you know the type of the instance, the type of the arg is also knowable.

PR #1157, opened 2022-08-07
Move dataclass_json in tx.cli to tx.utils

schatten:
Summary
---
Need to use dataclass_json.* methods in HMA. It feels weird to use something in namespace tx.cli because the API can change without warning.

Tests
---
py.test on local and relying on CI.


Dcallies (approved):
Works for me!

PR #1158, opened 2022-08-08
[vPDQ] Benchmark CPP and python-binding implmentations

tigerchensh:
Summary
---------

Benchmark on vPDQ with CPP and python-binding implementations. Both benchmarks run through the sample video set (fourteen 480x720 videos and one 1280x720 video). CPP implementation uses -02 flag and python-binding uses vpdq 0.0.6. Benchmark mark runs on MacBook Pro (16-inch, 2021), Apple M1 Pro, 32 GB Ram and the result is in README.MD. vPDQ performs well on 64 x 64 which comes from the pdq algorithm's support on this "magic" resolution.

### DIFF
--- a/vpdq/benchmark/README.MD
+++ b/vpdq/benchmark/README.MD
@@ -1,0 +1,63 @@
+# pytx-vpdq
+Benchmark vPDQ CPP implementation and python-binding library
+
+
+# Observed Performance
+MacBook Pro (16-inch, 2021), Apple M1 Pro, 32 GB Ram
+
+Results:
+-------
+```
+% python3 benchmark.py      
+Python hashing time: 34.1643s
+  Total hash:7069  Per hash: 4.8330ms
+CPP hashing time: 34.3673s
+  Total hash:7069  Per hash: 4.8617ms
+
+% python3 benchmark.py -r 1 
+Python hashing time: 6.9934s
+  Total hash:241  Per hash: 29.0182ms
+CPP hashing time: 7.3139s
+  Total hash:241  Per hash: 30.3483ms
+
+% python3 benchmark.py -s 540
+Python hashing time: 20.5453s
+  Total hash:7069  Per hash: 2.9064ms
+CPP hashing time: 20.9247s
+  Total hash:7069  Per hash: 2.9601ms
+Calculate deviations in downsampled hashes, since hash resolution is non-native
+Original resolution Python hashing time: 33.9340s
+  Number of mismatches:0, 0.00 percent in total.
+  Average 7.99 hamming distance away.
+
+% python3 benchmark.py -s 360
+Python hashing time: 9.8740s
+  Total hash:7069  Per hash: 1.3968ms
+CPP hashing time: 10.2460s
+  Total hash:7069  Per hash: 1.4494ms
+Calculate deviations in downsampled hashes, since hash resolution is non-native
+Original resolution Python hashing time: 34.5316s
+  Number of mismatches:0, 0.00 percent in total.
+  Average 8.57 hamming distance away.
+
+% python3 benchmark.py -s 180
+Python hashing time: 4.8490s
+  Total hash:7069  Per hash: 0.6860ms
+CPP hashing time: 5.4182s
+  Total hash:7069  Per hash: 0.7665ms
+Calculate deviations in downsampled hashes, since hash resolution is non-native
+Original resolution Python hashing time: 34.8951s
+  Number of mismatches:121, 0.02 percent in total.
+  Average 17.63 hamming distance away.
+
+
+% python3 benchmark.py -s 60 
+Python hashing time: 3.0915s
+  Total hash:7069  Per hash: 0.4373ms
+CPP hashing time: 3.3920s
+  Total hash:7069  Per hash: 0.4798ms
+Calculate deviations in downsampled hashes, since hash resolution is non-native
+Original resolution Python hashing time: 34.4134s
+  Number of mismatches:153, 0.02 percent in total.
+  Average 17.71 hamming distance away.
+```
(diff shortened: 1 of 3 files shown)


vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+        python_features = run_python(
+            ffmpegPath, secondsPerHash, downsampleFrameDimension
+        )
+    feature_size = 0
+    for feature in python_features:
+        feature_size += len(feature)
+    python_time = pt()
+    print(
+        f"  Total hash:{ feature_size}  Per hash: {1000*python_time /  feature_size:.4f}ms"
+    )
```
Dcallies, line 71: nit: print has its own formatting
nit: There's some odd spacing 
```suggestion
    print(
        "  Total hash:", feature_size, 
        "Per hash:", 
        f"{1000*python_time /  feature_size:.4f}ms"
    )
```

vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+                    "-f",
+                    str(ffmpegPath),
+                    "-r",
+                    str(secondsPerHash),
+                    "-d",
+                    str(OUTPUT),
+                    "-s",
+                    str(downsampleFrameDimension),
+                    "-i",
+                    str(VIDEOS) + "/" + file,
```
Dcallies, line 137: Nit: `str(VIDEOS / file)` is slightly safer since it will use with the equivalent of `os.path.join`

vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def run_cpp(
+    ffmpegPath: str,
+    secondsPerHash: int,
+    downsampleFrameDimension: int,
+):
+    for file in os.listdir(VIDEOS):
+        if file.endswith(".mp4"):
+            subprocess.call(
```
Dcallies, line 125: blocking: Why not check_call? This may silently fail, since you are also hiding output.

vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+                    str(secondsPerHash),
+                    "-d",
+                    str(OUTPUT),
+                    "-s",
+                    str(downsampleFrameDimension),
+                    "-i",
+                    str(VIDEOS) + "/" + file,
+                ],
+                stderr=subprocess.DEVNULL,
+                stdout=subprocess.DEVNULL,
```
Dcallies, line 140: Why hide output in this case?

vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+import subprocess
+import os
+import sys
+import argparse
+from pathlib import Path
+from contextlib import contextmanager, nullcontext
+from enum import Enum
+import time
+
+VPDQ_MATCH_DIST = 31
```
Dcallies, line 12: Nit: We've defined this same constant in many places now

vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+        print(f"{context}...")
+    start = time.perf_counter()
+    end = start
+    yield lambda: end - start
+    end = time.perf_counter()
+    print(f"{context}: {end - start:.4f}s")
+
+
+def get_argparse() -> argparse.ArgumentParser:
+    ap = argparse.ArgumentParser(description=__doc__)
```
Dcallies, line 31: nit: setting =__doc__ doesn't do anything, it's an empty string here. Either add a doc or remove the description

vpdq/benchmark/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+    secondsPerHash: int,
+    downsampleFrameDimension: int,
+):
+    with timer("Python hashing time") as pt:
+        python_features = run_python(
+            ffmpegPath, secondsPerHash, downsampleFrameDimension
+        )
+    feature_size = 0
+    for feature in python_features:
+        feature_size += len(feature)
```
Dcallies, line 67: ignorable/code golf: 
```suggestion
    feature_size = sum(len(feature) for feature in python_features)```

File: vpdq/benchmark/README.MD [hunk 8]
```diff
@@ -0,0 +1,63 @@
...
+
+# Observed Performance
+MacBook Pro (16-inch, 2021), Apple M1 Pro, 32 GB Ram
+
+Results:
+-------
+```
+% python3 benchmark.py      
+Python hashing time: 34.1643s
+  Total hash:7069  Per hash: 4.8330ms
```
Dcallies, line 13: nit: In some output you put a space between the ':' and the number, and sometimes you don't

Dcallies (changes requested):
Thanks for adding a benchmark, as well as the speed improvements! I think these are some great levels of polish at this level in the project! 

RC for better error behavior when ffmpeg fails - you probably want to use check_call

vpdq/benchmark/benchmark.py
tigerchensh: I was using an older version of CPP executables when benchmarking this, but I have found that ffmpeg output has been silenced in the CPP implementation.

vpdq/benchmark/README.MD
tigerchensh: Good catch! Have changed it.

tigerchensh:
Thanks for the review! Good catch on the silent ffmpeg fails as I was using an older version of the executables.

vpdq/benchmark/benchmark.py
tigerchensh: There were none in the vPDQ python-binding as it only includes the hashing in the library.

Dcallies (approved):
You solved my one blocking comment so I approve

PR #1163, opened 2022-08-09
[VPDQ] Update computeHash parameter second_per_hash default to 1 and release as 0.0.7

tigerchensh:
Summary
---------
Update vpdq python library computeHash function's parameter second_per_hash default value to 1, so the expected behavior won't be hashing every single frame in the video and the default setting is too slow. Release 0.0.7 for the change. Notice that the default value in pytx library is still 0 to keep the accuracy.

Test Plan
---------
Pytest and github workflow


BarrettOlson (changes requested):
Why are we keeping pytx at 0? it feels like 1 is a consistent default.

>  to keep the accuracy.

Can you provide an example of this trade off?

tigerchensh:
Videos tested here are "chair-19-sd-bar.mp4" and "chair-20-sd-bar.mp4". The latter one is a second longer in duration. The image shows the first line as using second_per_hash = 0 (i.e. hash every frame).   The second line uses second_per_hash = 1.
image

BarrettOlson:
@schatten @Dcallies thoughts?
The fact that 1.0 doesn't catch these two similar videos is an important flag but hashing every frame of a video does not scale for pytx usage. 
I think we are better off sticking to the 1.0 default with a follow up action item to make this param configurable.

Dcallies:
I think we should be using what we consider "reasonable defaults" for pytx. It sounds like we think reasonable defaults are 1.0 frame per second, given that's what all the tests are for. The matching algorithm in pytx also doesn't care about different frame speeds. We could just as easily make the frame generation flexible based on the duration. See also my comments offline about pruning sequences of similar frames.

In the long term pytx only needs to provide reasonable defaults, and the levers to adjust if the defaults don't work. My read is that we have both of those with a default of 1.0, and so suggest changing the defaults in pytx as well.

tigerchensh:
Thanks for the wonderful discussion. Have agreed on default seconds_per_hash 1 for querying index and some other values for stored hashes.  Will open issue to describe this context and further steps.

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_hash_and_match.py
```diff
@@ -29,7 +29,7 @@
 
 @pytest.mark.skipif(_DISABLED, reason="vpdq not installed")
 def test_vpdq_from_string_path():
-    computed_hash = VPDQSignal.hash_from_file(ROOTDIR / VIDEO)
+    computed_hash = VPDQSignal.hash_from_file(ROOTDIR / VIDEO, seconds_per_hash=0)
```
BarrettOlson, line 32: Could we update `test_hash.txt` instead of overriding the default?

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -81,6 +81,7 @@ def test_compare_hashes():
         assert video_file.is_file()
         ret = vpdq.computeHash(
             input_video_filename=video_file,
+            seconds_per_hash = 0
```
BarrettOlson, line 84: same question

python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_hash_and_match.py
tigerchensh: For test, it will be better if we test on every frame as we do not care about the efficiency but accuracy.

BarrettOlson: I would echo my point in #1163 to have test match production behavior.

In this case the idea of having a test that is even more granular e.g. efficiency > accuracy would be nice to have as well

Could we possibly have both (i.e. two tests/test files)?

tigerchensh: I think I will keep the vpdq test for accuracy but change the tests in pytx to match the prod parameter. The vpdq test will only trigger when the actual hashing code has changed.

BarrettOlson (approved):
I think the C++ implementations should also default to 1.0 instead of 0.0 based on the discussion in this PR but alright if we want to consider that beyond the scope of this PR

PR #1164, opened 2022-08-09
[VPDQ] Add python test github workflow and clean up test

tigerchensh:
Summary
---------

Add github work flow for vpdq python library, since now it can be easily locally installed through vpdq-release.py script. Adjust test parameters with CPP github workflow.

Test Plan
---------

github workflow

### DIFF
--- a/vpdq/python/tests/test_util.py
+++ b/vpdq/python/tests/test_util.py
@@ -1,8 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import vpdq
+import typing as t
 
 
-def read_file_to_hash(input_hash_filename: str) -> list[vpdq.VpdqFeature]:
+def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
     """Read hash file and return vpdq hash
 
     Args:
@@ -22,16 +23,3 @@ def read_file_to_hash(input_hash_filename: str) -> list[vpdq.VpdqFeature]:
         hash.append(feature)
 
     return hash
-
-
-def output_hash_to_file(output_hash_filename, hashes):
-    """Output vpdq hash to txt file
-
-    Args:
-        output_hash_filename (str): Output hash file path
-        hash (list of VpdqFeature): Vpdq Hash"""
-
-    with open(output_hash_filename, "w") as file:
-        for cur in hashes:
-            file.write(f"{cur.frame_number},{cur.quality},{hash_to_hex(cur.hash['w'])},{cur.timestamp:.3f}")
-            file.write("\n")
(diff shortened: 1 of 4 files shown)


vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -9,7 +9,7 @@
 SAMPLE_HASH_FOLDER = Path("vpdq/sample-hashes")
 SAMPLE_VIDEOS = Path("tmk/sample-videos")
 DISTANCE_TOLERANCE = 10
-QUALITY_TOLERANCE = 50
```
BarrettOlson, line 12: blocking: Why are we updating this to 80 again?

vpdq/python/tests/test_util.py
```diff
@@ -1,8 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 import vpdq
+import typing as t
 
 
-def read_file_to_hash(input_hash_filename: str) -> list[vpdq.VpdqFeature]:
+def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
```
BarrettOlson, line 6: nit: t.Sequence is a bit more flexible

BarrettOlson (changes requested):
Thanks for adding this!
Can you explain your reasoning for updating QUALITY_TOLERANCE?

vpdq/python/tests/test_vpdq_hash.py
tigerchensh: "Adjust test parameters with CPP github workflow." CPP reg test uses 80. 50 may fail on various ffmpeg versions caused by that low-quality video.

BarrettOlson: Having a test QUALITY_TOLERANCE higher than the used in prod `VPDQ_QUALITY_THRESHOLD = 50` doesn't make sense to me. (Same goes for what we do in C++ land)
I would prefer we skip the low-quality video for a regression test as it is more import the test reflects the production state of the library.

tigerchensh: The DISTANCE_TOLERANCE is lower than the used in prod. I can adjust those to be the same as the prod data. Is it better?

BarrettOlson: Sure, for the reg test, let's match prod defaults.
Was it a deliberate decision not to do so originally?

tigerchensh: Yes, we had discussion on this for smaller DISTANCE_TOLERANCE for validation process.

tigerchensh:
Thanks for the review. Has adjusted test parameters align with prod environment.

vpdq/python/tests/test_vpdq_hash.py
BarrettOlson: Ah I see, I view implementation validation as different from regression testing:
validation: two implementation of the vpdq algorithm ideally output nearly identical hashes (good to have much for strict params)
regression: adjustments, refactors, etc to code do not result in large changes in production behavior (can be more strict or just match production params)

BarrettOlson (approved):

PR #1165, opened 2022-08-09
Allow import of threatexchange.signal_type.pdq.PdqSignal

schatten:
Summary
---
This change makes tx.signal_type.pdq consistent with other modules like
tx.signal_type.md5 and tx.signal_type.raw_text.

pdq is the only module that is also a subpackage,

Test
---
CI.

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq/__init__.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/__init__.py
@@ -1,1 +1,5 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+# Allow import of threatexchange.signal_type.pdq.PdqSignal consistent with other
+# signal types.
+from .signal import PdqSignal


Dcallies (approved):
I was thinking of also moving all the "public" types to the __init__ of threatexchange itself, to try and make it more clear which APIs were stable.

Dcallies:
Can you clarify what you mean by this?
> This change makes tx.signal_type.pdq consistent with other modules like tx.signal_type.md5 and tx.signal_type.raw_text.

There's a couple of potential answers:
1. Folder structure
2. Module import path

This diff addresses 2, but I've gotten burned by strange behavior with `__init__`, as it creates side-effects during the import process that aren't always easy to spot: https://docs.python.org/3/reference/import.html#regular-packages

schatten:
> Can you clarify what you mean by this?
> 
> > This change makes tx.signal_type.pdq consistent with other modules like tx.signal_type.md5 and tx.signal_type.raw_text.

The following lines used to work in previous versions.

```python
from threatexchange.signal_type.pdq import PdqSignal
from threatexchange.signal_type.md5 import VideoMd5
```

Believing that these were 'public' and stable parts of the threatexchange API, I added these signals as [migrations](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py). While there is no evidence of these migrations having been applied in the wild, I'd like us to hold the line on what is public and stable.

This PR allows you to continue using the older names. 

If we want to explicitly create a public API like threatexchange.api and put everything into it, I'd be okay with that too.

PR #1167, opened 2022-08-10
[pytx] Version 1.0.0

Dcallies:
Summary
---------

It's upon us! Version 1.0.0! 

This is when we will start to enforce compatibility guarantees, so if you have something that needs to be changed, now is the time to request changes and flag it!

We should add branch here afterwards.


schatten (approved):
🎊

BarrettOlson (approved):

PR #1176, opened 2022-08-13
[multi] Store generic bank_info against each bank; connect tx_collab_configs to bank_ids

schatten:
Summary
---------
Please review commit by commit.

### [`55b52f38`](https://github.com/facebook/ThreatExchange/pull/1176/commits/55b52f3851a9b3203066c1fd14bad019847ef27b) Update and retrieve bank_info objects against a bank_id
We allow any serializable object to be written and retrieved from the bank. We'll start storing fetch checkpoints here.  
### [`8a673921`](https://github.com/facebook/ThreatExchange/pull/1176/commits/8a673921df9abb09a042c8a991fd25db5718d267) Collab configs now are connected to a bank_id
Each collab config must be connected to a bank_id, otherwise what do we do with this collab?  To make the API feel more natural, allow users to deal with `EditableCollaborationConfig` objects rather than the actualized `CollaborationConfigBase` object. This allows free access to the `import_as_bank_id` property.  

Test Plan
---------
Test cases for each change.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
+++ b/hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
@@ -37,40 +37,41 @@ class EditableCollaborationConfig(HMAConfig):
     # threatexchange.exchanges.impl.fb_threatexchange_api.FBThreatExchangeCollabConfig
     collab_config_class: str
 
-    # Use these attributes to instantiate a collab_config
+    # Use these attributes to instantiate a pytx collab_config object
     attributes_json_serialized: str
 
+    # A single collab configs would be connected to multiple banks.
+    #   a) One where we import signals
+    #   b) One from which we export signals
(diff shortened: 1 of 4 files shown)


hasher-matcher-actioner/hmalib/common/configs/tx_collab_config.py
```diff
@@ -94,4 +101,8 @@ def update_collab_config(collab_config: CollaborationConfigBase):
 
     ec.collab_config_class = full_class_name(collab_config.__class__)
     ec.attributes_json_serialized = dataclass_dumps(collab_config)
+
+    if import_as_bank_id:
```
Dcallies, line 105: nit: slightly confusing, since collab_config is always updated, and import_as_bank_id is only optionally updated. It's documented in the docstring, so ¯\_(ツ)_/¯

Dcallies (approved):
shipit

PR #1183, opened 2022-09-13
[pytx][setup] Set minimum numpy version

BarrettOlson:
Summary
---------

Attempt to fix an error reported when trying to install python-threatexchange with older numpy version already installed. 

Debating if we should add version requirements to the other libraries without them as well...

Test Plan
---------

`python-threatexchange $ make local_install`
without error


alutes (commented):
LGTM!

Dcallies (approved):
Yup yup yup

PR #1186, opened 2022-09-14
[py-tx] match_command -p to print query hash

Dcallies:
Summary
---------

Parking this from messing around with some hackathon code. Allows you to get back the hash that was used to query for data, which can be useful for reconstructing results.

Test Plan
---------

pending

### DIFF
--- a/python-threatexchange/threatexchange/cli/match_cmd.py
+++ b/python-threatexchange/threatexchange/cli/match_cmd.py
@@ -95,12 +95,18 @@ class MatchCommand(command_base.Command):
             ),
             help="limit to this signal type",
         )
+        ap.add_argument(
+            "--print-query-signal",
+            "-p",
+            action="store_true",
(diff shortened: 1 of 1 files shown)


sir-snaps-a-lot:
can we also consider a flag to print the reference hash that the query hash matched against?

Dcallies:
@sir-snaps-a-lot - I also thought about that one (and was my initial attempt!), the only problem is that the current SignalTypeIndex interface doesn't give us a way to do that, and my initial reaction was that doing that change would be backwards incompatible, which breaks our versioning commitments. (https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/README.md)

I didn't spend a super large amount of time trying to puzzle that out, so let me know if you have an idea.

Dcallies:
Closing for inactivity while we reconsider the approach

PR #1187, opened 2022-09-15
[pytx] SSCD extension proof of concept

Dcallies:
Summary
---------

This shows how one might make an SSCD signal type using the default suggested settings from the SSCD wiki.

Test Plan
---------

Ran against the PDQ sample data (bridge photos).


Dcallies:
Maybe again in the future

PR #1193, opened 2022-10-05
[py-tx][stopncii] Fix auth bugs

Dcallies:
Summary
---------

This is what I get for not thoroughly testing. Better late then never I suppose.

This looks like it's backwards compatible, but I'm actually swapping the order twice - so anyone who got the API fetching previously by putting them in the "wrong" order will still be okay.

Test Plan
---------
1. Put credentials in ~/.tx_stopncii_keys
2. tx config collab edit stop_ncii -C stop_ncii_prod
3. tx fetch

This uncovers a new bug - we're not picking up the vMD5s...

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
@@ -68,16 +68,16 @@ class StopNCIICredentials(auth.CredentialHelper):
     ENV_VARIABLE: t.ClassVar[str] = "TX_STOPNCII_KEYS"
     FILE_NAME: t.ClassVar[str] = "~/.tx_stopncii_keys"
 
-    function_key: str
     subscription_key: str
+    fetch_function_key: str
 
     @classmethod
     def _from_str(cls, s: str) -> "StopNCIICredentials":
-        user, _, passw = s.strip().partition(",")
-        return cls(user, passw)
+        subscription_key, _, fetch_function_key = s.strip().partition(",")
+        return cls(fetch_function_key, subscription_key)
 
     def _are_valid(self) -> bool:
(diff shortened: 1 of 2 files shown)


python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
```diff
@@ -68,16 +68,16 @@ class StopNCIICredentials(auth.CredentialHelper):
...
-    function_key: str
     subscription_key: str
+    fetch_function_key: str
 
     @classmethod
     def _from_str(cls, s: str) -> "StopNCIICredentials":
-        user, _, passw = s.strip().partition(",")
-        return cls(user, passw)
+        subscription_key, _, fetch_function_key = s.strip().partition(",")
+        return cls(fetch_function_key, subscription_key)
```
schatten, line 77: blocking: Hard to follow. The dataclass has subscription_key first and fetch_function_key second. I expect the constructor to have the same order.

Since you are using positional args, we're setting self.subscription_key to local variable fetch_function_key defined here.

schatten: If your test plan is working, things may be correctly getting tagged, but I don't know why.. :P

schatten (changes requested):
Positional arguments are out of order, but things somehow still work.

python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
Dcallies: You got confused by the same thing I did. I can reverse back the order in the constructor as long as the order in the file is maintained.

schatten (approved):
Assuming that you re-ran the test plan, Looks good to me.

PR #1198, opened 2022-10-10
[pytx] Filter out descriptors with missing owners due to API bug

Dcallies:
Summary
---------

Meta team is tracking, but this would break other functionality like filtering only specific owners, so the safest thing is to treat the record as deleted, even though this has other bad effects, as noticed in the code comment.

Test Plan
---------

Example collaboration has one inconsistent API record, before - get exception, afterwards - one less record.

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
@@ -127,10 +127,10 @@ class FBThreatExchangeIndicatorRecord(state.FetchedSignalMetadata):
             if "owner" not in td_json:
                 # Regrettable bug in threatexchange where descriptor from
                 # `descriptors` is visible but `owner` field is not.
-                # It's on Meta to fix this bug, for now, just drop record.
-                # Dropping record isn't create, because it breaks the correctness
-                # of threat_updates, since it can oscillate between visible and
-                # not separately from the threat_update order.
+                # It's on Meta to fix this bug, but for now just drop the record.
+                # This bug breaks the correctness of threat_updates, because
+                # it can oscillate between visible and not separately from the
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
```diff
@@ -124,6 +124,15 @@ def from_threatexchange_json(
...
         for td_json in te_json.raw_json["descriptors"]["data"]:
             td_id = int(td_json["id"])
+            if "owner" not in td_json:
+                # Regrettable bug in threatexchange where descriptor from
+                # `descriptors` is visible but `owner` field is not.
+                # It's on Meta to fix this bug, for now, just drop record.
+                # Dropping record isn't create, because it breaks the correctness
+                # of threat_updates, since it can oscillate between visible and
+                # not separately from the threat_update order.
+                # TODO: Should we log this?
```
BarrettOlson, line 134: yes, it would be good to set the debug log level and then check how many of the record are in this state.

python-threatexchange/threatexchange/exchanges/impl/fb_threatexchange_api.py
```diff
@@ -124,6 +124,15 @@ def from_threatexchange_json(
 
         for td_json in te_json.raw_json["descriptors"]["data"]:
             td_id = int(td_json["id"])
+            if "owner" not in td_json:
+                # Regrettable bug in threatexchange where descriptor from
+                # `descriptors` is visible but `owner` field is not.
+                # It's on Meta to fix this bug, for now, just drop record.
+                # Dropping record isn't create, because it breaks the correctness
```
BarrettOlson, line 131: create -> great ?

BarrettOlson (approved):

PR #1204, opened 2022-11-06
SignalExchangeAPI returns patch instead of applying deletes

schatten:
Version Bump?
---
Should entail a major version bump. A SignalExchangeAPI written without the 'patch' method will not work with the current version.

Summary
---
To make fetch work with a persisted networked storage, we need to stop treating SignalExchangeAPI.naive_fetch_merge as the applicator of fetch_iter() responses.

For the in-memory case, we start calling naive_fetch_merges with nothing in the accumulator and are able to reconstruct the final state.

When a fetch store is backed by network storage or a database, like in banks, the accumulator can't be provided without loading the entire bank into memory.

Instead, we add a method naive_convert_to_signal_type_patch which, instead of applying and absorbing the deletes, returns the deletes as a separate sequence.

Test Plan
---
```
$ py.test
```

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
+++ b/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
@@ -169,19 +169,46 @@ class SignalExchangeAPI(
                 old[k] = new_v
 
     @classmethod
-    @abstractmethod
     def naive_convert_to_signal_type(
         cls,
         signal_types: t.Sequence[t.Type[SignalType]],
         collab: TCollabConfig,
         fetched: t.Mapping[state.TUpdateRecordKey, state.TUpdateRecordValue],
-    ) -> t.Dict[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]]:
+    ) -> t.Mapping[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]]:
         """
         Convert the record from the API format to the format needed for indexing.
 
         This is the fallback method of creating state when there isn't a
         specialized storage for the fetch type.
+        """
+        patch = cls.naive_convert_to_signal_type_patch(signal_types, collab, fetched)
+        return patch[0]
 
+    @classmethod
+    @abstractmethod
+    def naive_convert_to_signal_type_patch(
+        cls,
+        signal_types: t.Sequence[t.Type[SignalType]],
+        collab: TCollabConfig,
+        fetched: t.Mapping[state.TUpdateRecordKey, state.TUpdateRecordValue],
+    ) -> t.Tuple[
+        t.Mapping[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]],
+        t.Sequence[state.TUpdateRecordKey],
+    ]:
+        """
+        Convert records from the API's format to a format needed for patching to
+        a persistent storage.
+
+        The returned tuple contains:
+        1. a mapping from signal_type -> signal_value -> signal_metadata
+        2. a list of keys that must be deleted.
+
+        Note the mapping already removes the deleted signals. So, if the API
+        tells you to add A and remove B, this function should return:
+        (
+            signal_type -> A -> A metadata,
+            [B]
+        )
         """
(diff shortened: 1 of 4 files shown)


python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -169,19 +169,46 @@ def naive_fetch_merge(
...
 
     @classmethod
-    @abstractmethod
     def naive_convert_to_signal_type(
         cls,
         signal_types: t.Sequence[t.Type[SignalType]],
         collab: TCollabConfig,
         fetched: t.Mapping[state.TUpdateRecordKey, state.TUpdateRecordValue],
-    ) -> t.Dict[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]]:
+    ) -> t.Mapping[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]]:
```
Dcallies, line 177: It was dict before because this isn't supposed to store internal state, which means that ownership of all objects can be transferred to the caller(Dict is mutable, Map is immutable).

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -169,19 +169,46 @@ def naive_fetch_merge(
...
+        signal_types: t.Sequence[t.Type[SignalType]],
+        collab: TCollabConfig,
+        fetched: t.Mapping[state.TUpdateRecordKey, state.TUpdateRecordValue],
+    ) -> t.Tuple[
+        t.Mapping[t.Type[SignalType], t.Dict[str, state.TFetchedSignalMetadata]],
+        t.Sequence[state.TUpdateRecordKey],
+    ]:
+        """
+        Convert records from the API's format to a format needed for patching to
+        a persistent storage.
```
Dcallies, line 200: Hmm, this comment doesn't match my intent of how the persistent storage interface works. This interface suggests that the persistent storage should be k => <black box>, but for HMA you have redefined storage as needing to be (str, str) => metadata.

I am still convinced that the best way to have fetch API storage correct is to store it in the same native form that the APIs themselves support, which tends to vary, hence a keyspace specific to that API. Any other remapping introduces edge cases that require refcounting based on that key to be correct.

I think this interface is attempting to push the requirements for refcounting back into the SignalExchangeAPI, where it will then need to be repeatedly independently implemented.

Dcallies (changes requested):
I think I need to talk to you more about this. Fundamentally, I'm having trouble making the connection for why deletes are explicitly required. Can you give me a counter example using an existing sequence of updates isn't possible to reconstruct using the existing interfaces?

The one I keep tripping up on is

```
t1: {k1: [h1, h2]}
t2: {j1: [h1]}
```

Your interface wants to be able to tell that t1 is a delete operation on h2, but _without additional storage not visible to the interface_ it's not possible to reconstruct this information. naive_fetch_merge as written is not supposed to be the only way to produce a correct state (it requires the entire database to be loadable in memory), just that it must produce a correct state for any other implementation to work. 

I've know that there has been an impedance mismatch between the form you are using in HMA (specific)  and the form on the CLI (generic), but I'm not convinced the correct change is making the CLI specific.

Dcallies (commented):
We talked a little bit more offline to try and figure out what is happening:

1. David thinks current interface is slightly unclear in that the "naive_fetch_merge" is not meant to be in the interface, but a reference of how to use the output of fetch(). It should probably be moved out of the interface to be more clear.
2. naive_convert_to_signal_type however is part of the interface... though it should be forced to work at the level of only a single key. The "naive" implementation of loading the whole dataset into memory should not be required for correctness. Comments could be updated to make this clear (or the interface changed entirely)
3. David's suggestion for HMA - try running convert_to_signal_type() on only one key at a time to get the information you need. The in-between merges are not required/part of the interface, since there shouldn't be cross-key dependencies (or else an efficient storage solution is not possible). 

Thanks for bearing with me as I try to puzzle out what I intended!

schatten:
Abandoning in favour of #1216

Issue #1211, opened 2022-11-16
[py-tx] using StopNCIIAPI.fetch_hashes will lead to ValueError due to missing field in StopNCIICSPFeedbackValue


Gnomeek:
Hey guys, I just got am `ValueError` when fetching hashes using `StopNCIIAPI.fetch_hashes`, the traceback is:
```
Traceback (most recent call last):
  File "/Users/will/Documents/Codespace/sds_pdq/cronjob/fetch_stopncii.py", line 105, in <module>
    main(x_functions_key, x_subscription_key)
  File "/Users/will/Documents/Codespace/sds_pdq/cronjob/fetch_stopncii.py", line 91, in main
    for resp in client.fetch_hashes_iter():
  File "/Users/will/Library/Caches/pypoetry/virtualenvs/pdq-cronjob-gH7F_-yJ-py3.8/lib/python3.8/site-packages/threatexchange/exchanges/clients/stopncii/api.py", line 251, in fetch_hashes_iter
    result = self.fetch_hashes(
  File "/Users/will/Library/Caches/pypoetry/virtualenvs/pdq-cronjob-gH7F_-yJ-py3.8/lib/python3.8/site-packages/threatexchange/exchanges/clients/stopncii/api.py", line 228, in fetch_hashes
    return dacite.from_dict(
  File "/Users/will/Library/Caches/pypoetry/virtualenvs/pdq-cronjob-gH7F_-yJ-py3.8/lib/python3.8/site-packages/dacite/core.py", line 63, in from_dict
    value = _build_value(type_=field.type, data=transformed_value, config=config)
  File "/Users/will/Library/Caches/pypoetry/virtualenvs/pdq-cronjob-gH7F_-yJ-py3.8/lib/python3.8/site-packages/dacite/core.py", line 90, in _build_value
    return _build_value_for_collection(collection=type_, data=data, config=config)
  File "/Users/will/Library/Caches/pypoetry/virtualenvs/pdq-cronjob-gH7F_-yJ-py3.8/lib/python3.8/site-packages/dacite/core.py", line 140, in _build_value_for_collection
    return data_type(_build_value(type_=item_type, data=item, config=config) for item in data)
  File "/Users/will/Library/Caches/pypoetry/virtualenvs/pdq-cronjob-gH7F_-yJ-py3.8/lib/python3.8/site-packages/dacite/core.py", line 140, in <genexpr>
    return data_type(_build_value(type_=item_type, data=item, config=config) for item in data)
[...]

PR #1212, opened 2022-11-16
[fix] add undefined in StopNCIICSPFeedbackValue

Gnomeek:
Summary
---------

The StopNCII.org API is emitting undocumented enum values (see #1211). This will allow them to be loadable.

Test Plan
---------

Call `StopNCIIAPI.fetch_hashes()` on the [Test | Production] instance.

Before: `ValueError: 'Undefined' is not a valid StopNCIICSPFeedbackValue`

After: 
- will not raise error to block fetching process.
- will omit `StopNCIIHashRecord` that fails to `dacite`.
- will print an error log to notice user, including error message, original json string of the omitted `StopNCIIHashRecord`.

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
@@ -44,6 +44,7 @@ class StopNCIICSPFeedbackValue(enum.Enum):
     """The feedback that a CSP has given on a hash"""
 
     Unknown = "Unknown"
+    # While undocumented, the API appears to use this value
     Undefined = "Undefined"
     None_ = "None"  # Allows you to tag without associating a state
     QualityUnknown = "QualityUnknown"  # Unsure what this is used for
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
@@ -3,6 +3,7 @@
 """Simple implementation for the StopNCII REST API"""
 
 from dataclasses import dataclass, asdict, field
+import copy
 import enum
 import logging
 import time
@@ -228,11 +229,25 @@ class StopNCIIAPI:
         logging.debug("StopNCII FetchHashes called: %s", params)
         json_val = self._get(StopNCIIEndpoint.FetchHashes, **params)
         logging.debug("StopNCII FetchHashes returns: %s", json_val)
-        return dacite.from_dict(
+        records: t.List[StopNCIIHashRecord] = []
+        for record in json_val.get("hashRecords", []):
+            try:
+                records.append(
+                    dacite.from_dict(
+                        data_class=StopNCIIHashRecord,
+                        data=record,
+                        config=dacite.Config(cast=[enum.Enum, set]),
+                    )
+                )
+            except ValueError as e:
+                logging.error("Convert response from JSON to Dataclass failed, err: %s, record: %s", e, record)
+        response: FetchHashesResponse = dacite.from_dict(
             data_class=FetchHashesResponse,
-            data=json_val,
+            data={**json_val, "hashRecords": []},
             config=dacite.Config(cast=[enum.Enum, set]),
         )
+        response.hashRecords = records
+        return response
 
     def fetch_hashes_iter(
         self, start_timestamp: int = DEFAULT_START_TIME
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
@@ -44,6 +44,7 @@ class StopNCIICSPFeedbackValue(enum.Enum):
     """The feedback that a CSP has given on a hash"""
 
     Unknown = "Unknown"
+    Undefined = "Undefined"
     None_ = "None"  # Allows you to tag without associating a state
     QualityUnknown = "QualityUnknown"  # Unsure what this is used for
     PendingReview = "PendingReview"  # There was a match
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
@@ -45,8 +45,6 @@ class StopNCIICSPFeedbackValue(enum.Enum):
     """The feedback that a CSP has given on a hash"""
 
     Unknown = "Unknown"
-    # While undocumented, the API appears to use this value
-    Undefined = "Undefined"
     None_ = "None"  # Allows you to tag without associating a state
     QualityUnknown = "QualityUnknown"  # Unsure what this is used for
     PendingReview = "PendingReview"  # There was a match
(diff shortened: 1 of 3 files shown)


Dcallies (changes requested):
Thanks a bunch for the fix @Gnomeek ! 

I'm looking at the API documentation at https://swgfl.org.uk/assets/documents/stopncii-api-reference.pdf and don't see an Undefined - where did you run into it? It might be able to ask StopNCII to try and fix the issue upstream.

No code changes needed (or I can add the comment), just want some information so I can make a bug report to StopNCII.

Please "Request Review" when you reply to get it back into my queue!

Dcallies:
Additionally, can you provide a short Summary and Test Plan?

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -44,6 +44,7 @@ class StopNCIICSPFeedbackValue(enum.Enum):
     """The feedback that a CSP has given on a hash"""
 
     Unknown = "Unknown"
+    Undefined = "Undefined"
```
Dcallies, line 47:
```suggestion
    # While undocumented, the API appears to use this value
    Undefined = "Undefined"
```

Dcallies:
Hey @Gnomeek  - apologies, I haven't used github in a while and missed your issue originally!

Can you confirm that your fix makes it so it's possible to fetch the entirety of the data? If so, I will help merge you in!

Gnomeek:
Actually, I think it's better to add some fallback statements instead of adding enum value. Printing error messages which includes the source dict when `dacite` failed is enough. This process shouldn't block the entire fetching procedure.

Gnomeek:
Unittest added

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -44,6 +45,8 @@ class StopNCIICSPFeedbackValue(enum.Enum):
     """The feedback that a CSP has given on a hash"""
 
     Unknown = "Unknown"
+    # While undocumented, the API appears to use this value
+    Undefined = "Undefined"
```
Dcallies, line 49: (blocking): Hey @Gnomeek in your comment you wrote

> Actually, I think it's better to add some fallback statements instead of adding enum value

However, it looks like the enum value is still here. Is that intentional? Your test is additionally asserting the existence.

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -226,11 +229,25 @@ def fetch_hashes(
         logging.debug("StopNCII FetchHashes called: %s", params)
         json_val = self._get(StopNCIIEndpoint.FetchHashes, **params)
         logging.debug("StopNCII FetchHashes returns: %s", json_val)
-        return dacite.from_dict(
+        records: t.List[StopNCIIHashRecord] = []
```
Dcallies, line 232: The logic below seems reasonable. Can you just add a short comment explaining why it's not a straight convert?

Maybe something like
```
# If there is a malformed record or a change that would prevent the deserialization
# of a record, skip over it instead of crashing. Please open an issue if you see the logging
# statement below.
```

python-threatexchange/threatexchange/exchanges/clients/stopncii/tests/test_api.py
```diff
@@ -163,3 +163,94 @@ def test_mocked_get_hashes_iter(api: StopNCIIAPI):
     assert_first_record(one)
     assert_second_record(two)
     assert_third_record(three)
+
+
+def mock_get_impl_error_enum(endpoint: str, **json):
```
Dcallies, line 168: Thanks for adding a unittest!

python-threatexchange/threatexchange/exchanges/clients/stopncii/tests/test_api.py
```diff
@@ -163,3 +163,94 @@ def test_mocked_get_hashes_iter(api: StopNCIIAPI):
...
+    }
+
+
+@pytest.fixture
+def error_enum_api(monkeypatch: pytest.MonkeyPatch):
+    error_enum_api = StopNCIIAPI("", "")
+    monkeypatch.setattr(error_enum_api, "_get", mock_get_impl_error_enum)
+    return error_enum_api
+
+
```
Dcallies, line 248: (ignorable): you can just inline this logic in the test, since I think it's unlikely that we'd have two tests that use it.

python-threatexchange/threatexchange/exchanges/clients/stopncii/tests/test_api.py
```diff
@@ -163,3 +163,94 @@ def test_mocked_get_hashes_iter(api: StopNCIIAPI):
...
+def test_mocked_get_hashes_with_undefined_enum(error_enum_api: StopNCIIAPI):
+    try:
+        result = error_enum_api.fetch_hashes()
+        assert result.count == 3
+        assert len(result.hashRecords) == 2
+        assert any(i.feedbackValue == StopNCIICSPFeedbackValue.Undefined for i in result.hashRecords[1].CSPFeedbacks)
+    except Exception as exc:
+        assert False, f"fetch_hashes raised an exception {exc}"
```
Dcallies, line 256: blocking: Instead I'd suggest letting this exception leak out, since the test will still show as failing, and this will only generate a deeper stack trace, where L256 isn't adding any additional information.
```suggestion
    result = error_enum_api.fetch_hashes()
    assert result.count == 3
    assert len(result.hashRecords) == 2
    assert any(i.feedbackValue == StopNCIICSPFeedbackValue.Undefined for i in result.hashRecords[1].CSPFeedbacks)
```

Dcallies (changes requested):
Thanks again for the update @Gnomeek - only one blocking concern. In your last comment you mentioned you didn't want to add the enum value, and instead add try/catch logic, but it looks like the enum value is still there, and your test asserts it. Did you change your mind?

Additionally, @shekarc - this test data shows the records with the issue.

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
Gnomeek: Both works for me. We can delete those two lines and change the test accordingly.

Gnomeek:
Hey @Dcallies , just push a commit with changes according to your CR, could you please review it again?

Gnomeek:
Hey @Dcallies , just push a commit with changes according to your CR, could you please review it again?

Dcallies (approved):
Thank you so much for the fix! I'm working with StopNCII.org to try and prevent the issue from occuring!

Dcallies:
Will just merge with the lint error, I can format before a PR to bump the version.

Gnomeek:
> Thank you so much for the fix! I'm working with StopNCII.org to try and prevent the issue from occuring!

Awesome, please let me know if there's any features/bug-fixes need help, I'm glad to contribute.

PR #1216, opened 2022-11-21
HMA FetchStore implementation backed by row-based calling of SignalExchangeAPI classmethods

schatten:
Summary
---------
Interesting files:
- bank_store.py
- fetcher.py

fetcher.py has a Fetcher class which will be invoked in a lambda. This will cause it to get all current collabs and run `Fetcher._execute_for_collab` to be called for each collab.

`_execute_for_collab` ends up calling fetch_iter for SignalExchangeAPI. Each row of the output of fetch_iter is then processed via `SignalExchangeAPI.fetch_value_merge` to determine whether it is an add/update or a delete.

Each add/update is then processed via `SignalExchangeAPI.naive_convert_to_signal_type` to identify signals to be added or updated.

While the deletes result in bank-member deletes, the adds / updates result in bank-member + bank-member-signal writes.

Test Plan
---------
```python
    import boto3
    import logging

    from hmalib.common.config import HMAConfig

    from hmalib.common.mappings import get_pytx_functionality_mapping
    from hmalib.common.models.bank import BanksTable

    HMAConfig.initialize("...-HMAConfig")
    st_mapping = get_pytx_functionality_mapping().signal_and_content
    dynamodb = boto3.resource("dynamodb")
    banks_table = BanksTable(dynamodb.Table("...-HMABanks"), st_mapping)

    # TODO: Create fetcher per
    fetcher = Fetcher(st_mapping, banks_table)
    logger.info("Will be running fetcher now.")
    fetcher.run()
```

All three banks had been manually cleared before the run.

Resulted in 3 collab configs (two threatexchange) and 1 NCMEC getting filled up.
Screen Shot 2022-11-21 at 16 23 24

Screen Shot 2022-11-21 at 16 23 19

Screen Shot 2022-11-21 at 16 23 15

While the NCMEC fetcher never terminated, subsequent calling of the threatexchange fetchers led to no new updates proving that the checkpointing (via bank_info) works.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
+++ b/hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
@@ -16,11 +16,12 @@ from hmalib.common.models.tests.test_signal_uniqueness import BanksTableTestBase
 
 
 class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._id_count = 0
+
     def _create_bank(self) -> str:
         """Create a bank with a unique id."""
-        if not hasattr(self, "_id_count"):
-            self._id_count = 0
-
         self._id_count = self._id_count + 1
 
         table_manager = BanksTable(
@@ -112,5 +113,5 @@ class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
             )
             self.assertSetEqual(
                 {PDQ_1, PDQ_2, MD5},
-                set([x.signal_value for x in all_signals]),
+                {x.signal_value for x in all_signals},
             )
--- a/hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
+++ b/hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
@@ -112,5 +112,5 @@ class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
             )
             self.assertSetEqual(
                 {PDQ_1, PDQ_2, MD5},
-                set(map(lambda x: x.signal_value, all_signals)),
+                set([x.signal_value for x in all_signals]),
             )
--- a/hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
+++ b/hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
@@ -1,10 +1,13 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
+from contextlib import contextmanager
 import typing as t
 from dataclasses import dataclass
 import unittest
 
 from threatexchange.content_type.photo import PhotoContent
+from threatexchange.signal_type.pdq import PdqSignal
+from threatexchange.signal_type.md5 import VideoMD5Signal
 
 from hmalib.common.tests.mapping_common import get_default_signal_type_mapping
 
@@ -14,41 +17,100 @@ from hmalib.common.models.tests.test_signal_uniqueness import BanksTableTestBase
 
 class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
     def _create_bank(self) -> str:
+        """Create a bank with a unique id."""
+        if not hasattr(self, "_id_count"):
+            self._id_count = 0
+
+        self._id_count = self._id_count + 1
+
         table_manager = BanksTable(
             self.get_table(), signal_type_mapping=get_default_signal_type_mapping()
         )
 
-        bank = table_manager.create_bank("TEST_Bank", "Test bank description")
+        bank = table_manager.create_bank(
+            f"TEST_Bank_{self._id_count}", "Test bank description"
+        )
         return bank.bank_id
 
+    def _add_photo_bank_member_with_key(
+        self, table_manager: BanksTable, bank_id: str, member_key: str
+    ):
+        return table_manager.add_keyed_bank_member(
+            bank_id=bank_id,
+            member_key=member_key,
+            content_type=PhotoContent,
+            storage_bucket=None,
+            storage_key=None,
+            raw_content="does not matter",
+            notes="",
+            is_media_unavailable=True,
+        )
+
     def test_add_bank_member_with_key(self):
-        with self.fresh_dynamodb():
-            table_manager = BanksTable(
-                self.get_table(), get_default_signal_type_mapping()
-            )
-
+        with self.fresh_table_manager() as table_manager:
             bank_id = self._create_bank()
-            bank_member_id = "fooofofofofoof"
+            member_key = "fooofofofofoof"
 
-            table_manager.add_bank_member_with_key(
-                bank_id=bank_id,
-                bank_member_id=bank_member_id,
-                content_type=PhotoContent,
-                storage_bucket=None,
-                storage_key=None,
-                raw_content=None,
-                notes="",
-                is_media_unavailable=True,
-            )
+            self._add_photo_bank_member_with_key(table_manager, bank_id, member_key)
 
             with self.assertRaises(KeyError):
-                table_manager.add_bank_member_with_key(
-                    bank_id=bank_id,
-                    bank_member_id=bank_member_id,
-                    content_type=PhotoContent,
-                    storage_bucket=None,
-                    storage_key=None,
-                    raw_content=None,
-                    notes="",
-                    is_media_unavailable=True,
-                )
+                self._add_photo_bank_member_with_key(table_manager, bank_id, member_key)
+
+    def test_add_keyed_bank_member_in_multiple_banks(self):
+        with self.fresh_table_manager() as table_manager:
+            bank_id = self._create_bank()
+            bank_id_2 = self._create_bank()
+
+            member_key = "a-very-unique-id-as-you-can-see"
+
+            bank_member = self._add_photo_bank_member_with_key(
+                table_manager, bank_id, member_key
+            )
+
+            bank_member_2 = self._add_photo_bank_member_with_key(
+                table_manager, bank_id_2, member_key
+            )
+
+            self.assertNotEqual(bank_member.bank_member_id, bank_member_2)
+
+    def test_add_multiple_signals_for_keyed_bank_member(self):
+        """Prove that multiple signals of the same type can be stored for the
+        same bank member. eg. MD5s from slightly altered copies of a video."""
+        with self.fresh_table_manager() as table_manager:
+            bank_id = self._create_bank()
+            member_key = "a-very-unique-id-as-you-can-see"
+
+            PDQ_1 = "f13de9de2e5d46ea51749338e45ea9891a068e17c455bca745ab17de0e7241c4"
+            PDQ_2 = "7c455bca745ab17de0e7241c4f13de9de2e5d46ea51749338e45ea9891a068e1"
+            MD5 = "7c455bca745ab17de0e7241c4e7241c4"
+
+            bank_member = self._add_photo_bank_member_with_key(
+                table_manager, bank_id, member_key
+            )
+
+            table_manager.add_bank_member_signal(
+                bank_id=bank_id,
+                bank_member_id=bank_member.bank_member_id,
+                signal_type=PdqSignal,
+                signal_value=PDQ_1,
+            )
+            table_manager.add_bank_member_signal(
+                bank_id=bank_id,
+                bank_member_id=bank_member.bank_member_id,
+                signal_type=PdqSignal,
+                signal_value=PDQ_2,
+            )
+            table_manager.add_bank_member_signal(
+                bank_id=bank_id,
+                bank_member_id=bank_member.bank_member_id,
+                signal_type=VideoMD5Signal,
+                signal_value=MD5,
+            )
+
+            all_signals = table_manager.get_signals_for_bank_member(
+                bank_member.bank_member_id
+            )
+            self.assertSetEqual(
+                {PDQ_1, PDQ_2, MD5},
+                set(map(lambda x: x.signal_value, all_signals)),
+            )
--- a/hasher-matcher-actioner/hmalib/common/models/tests/test_signal_uniqueness.py
+++ b/hasher-matcher-actioner/hmalib/common/models/tests/test_signal_uniqueness.py
@@ -1,17 +1,30 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
+from contextlib import contextmanager
 import unittest
 import typing as t
 from concurrent import futures
 
 from threatexchange.signal_type.md5 import VideoMD5Signal
 
-from hmalib.common.models.bank import BankedSignalEntry
+from hmalib.common.models.bank import BankedSignalEntry, BanksTable
+from hmalib.common.tests.mapping_common import get_default_signal_type_mapping
 
 from .ddb_test_common import DynamoDBTableTestBase
 
 
 class BanksTableTestBase(DynamoDBTableTestBase):
+    @contextmanager
+    def fresh_table_manager(self):
+        with self.fresh_dynamodb():
+            try:
+                yield BanksTable(
+                    self.get_table(),
+                    signal_type_mapping=get_default_signal_type_mapping(),
+                )
+            finally:
+                pass
+
     @classmethod
     def get_table_definition(cls) -> t.Any:
         table_name = "test-banks-table"
--- a/hasher-matcher-actioner/hmalib/common/models/tests/test_signal_uniqueness.py
+++ b/hasher-matcher-actioner/hmalib/common/models/tests/test_signal_uniqueness.py
@@ -17,13 +17,10 @@ class BanksTableTestBase(DynamoDBTableTestBase):
     @contextmanager
     def fresh_table_manager(self):
         with self.fresh_dynamodb():
-            try:
-                yield BanksTable(
-                    self.get_table(),
-                    signal_type_mapping=get_default_signal_type_mapping(),
-                )
-            finally:
-                pass
+            yield BanksTable(
+                self.get_table(),
+                signal_type_mapping=get_default_signal_type_mapping(),
+            )
 
     @classmethod
     def get_table_definition(cls) -> t.Any:
--- a/hasher-matcher-actioner/hmalib/fetching/bank_store.py
+++ b/hasher-matcher-actioner/hmalib/fetching/bank_store.py
@@ -22,15 +22,21 @@ from hmalib.common.logging import get_logger
 logger = get_logger(__name__)
 
 
-class NoImportAsBankDefined(Exception):
-    def __init__(self, collab_config: EditableCollaborationConfig):
-        self.collab_config = collab_config
-        super().__init__(
-            f"CollabConfig {collab_config.name} is not connected to an importable bank. Create a bank, and save again using tx_collab_config apis."
-        )
-
-
 class MultipleContentTypesInFetchDeltaRow(Exception):
+    """
+    At present, we can't handle if a single fetch delta update
+    (FetchDelta.updates[]), when passed through
+    SignalExchangeAPI.naive_convert_to_signal_type(...) returns signals with
+    differing content types.
+
+    Eg. imagine a ncmec id ends up returing an PDQ AND an MD5. We would not know
+    which type of bank member to create.
+
+    In the long run, this can be addressed by creating a new content-type eg.
+    unavailable which explicitly states that a bank member does not have media
+    attached.
+    """
+
     pass
 
 
@@ -41,6 +47,31 @@ class BankCollabFetchStore:
 
     Use with a single collab passed when creating an instance. Create multiple
     instances when working with more than one collab.
+
+    References
+    ===
+    Borrows heavily from python-threatexchange's fetch_cmd.py. fetch_cmd.py and
+    bank_store.py together are the strictest users of python-threatexchange's
+    SignalExchangeAPI interface methods.
+
+    fetch_cmd stores the fetched data in a SignalExchange specific format
+    defined by SignalExchangeAPI's generic typevars. Different
+    SignalExchangeAPI's data is later converted into a common type during
+    indexing by calling SignalExchangeAPI.naive_convert_to_signal_type.
+
+    bank_store OTOH, calls naive_convert_to_signal_type at fetch time itself and
+    stores data from all SignalExchangeAPIs using a consistent API defined by
+    HMA banks and bank-members. Each TUpdateRecordKey in SignalExchangeAPI is
+    stored as a bank-member in HMA.
+
+    Another key distinction is what's loaded into memory. fetch_cmd ends up
+    loading the entire dataset into memory. In pytx CLI, this is not an issue
+    because the data is expected to be co-located on the same computer where
+    fetch is running. With HMA, loading the whole bank into memory would be
+    time-consuming. So, instead we distill the results of
+    SignalExchangeAPI.fetch_value_merge and
+    SignalExchagneAPI.naive_convert_to_signal_type calls into updates and
+    deletes and issue those commands to the underlying banks.
     """
 
     def __init__(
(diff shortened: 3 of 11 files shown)


hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+            # Cases where none of the signal types in the update can be
+            # processed. eg. NCMEC returning pdna
+            return
+
+        # Ascertain that all signal_types map from the same content type.
+        # HMA at present can't store multiple content types in one bank
+        # member.
+        unique_content_types = set(
+            # Is there a signal type that may have multiple content_types?
+            list(map(lambda st: st.get_content_types()[0], signal_type_view.keys()))
```
Dcallies, line 219: nit: similar note about doing as list comprehension

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+        Simplistic 1:1 translation of opinion tag to label of same name. This
+        could be made more configurable on an API / Collab basis via the UI.
+
+        This should be an instance method even though it does not use self. This
+        is to ensure that future modifications (that require loading stuff from
+        ddb using self.banks_table) can be written.
+        """
+        return reduce(
+            lambda tags, acc: acc.union(tags), map(lambda x: x.tags, opinions)
```
Dcallies, line 288: nit: You can even generator comprehension!
```suggestion
            lambda tags, acc: acc.union(tags), (o.tags for o in opinions)
```

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
```diff
@@ -14,41 +17,100 @@
 
 class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
     def _create_bank(self) -> str:
+        """Create a bank with a unique id."""
+        if not hasattr(self, "_id_count"):
+            self._id_count = 0
```
Dcallies, line 22: Any reason not to do this in an `__init__()` with a super() call to the parent? hasatter and dynamic attribute modification always looks suspicious to me

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
```diff
@@ -14,41 +17,100 @@
 
 class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
```
Dcallies, line 18: (ignorable): I've said this in other PRs, but py.test is strictly better than unittest.TestCase, and there exists a conversion to an equivalent py.test with better error messages from any existing unittest. I think I'd hard block you for py-tx (which I had gone through a conversion at some point earlier this year), but will leave as an ignore here

Dcallies (commented):
One weird side effect from this is that bank members are no longer 1:1 with content if the source of the content was a collaboration. i.e. hashes that strictly belong to photos or videos

What should we display for a member that has a combination of photo/video signals (not possible with current APIs, but allowed by the API)?

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+
+        # Ascertain that all signal_types map from the same content type.
+        # HMA at present can't store multiple content types in one bank
+        # member.
+        unique_content_types = set(
+            # Is there a signal type that may have multiple content_types?
+            list(map(lambda st: st.get_content_types()[0], signal_type_view.keys()))
+        )
+        if len(unique_content_types) > 1:
+            raise MultipleContentTypesInFetchDeltaRow()
```
Dcallies, line 222: Ah, I see, so we just don't accept the record :P

What's the long term plan here, given that its allowed by SignalExchangeAPI? Can we split the record out by its component types and just assign it a key based on the type?

I.e. instead of <signal_exchage_key> we break up into <signal-exchange_key>::photo and ::video

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+
+        if len(signal_type_view) == 0:
+            # Cases where none of the signal types in the update can be
+            # processed. eg. NCMEC returning pdna
+            return
+
+        # Ascertain that all signal_types map from the same content type.
+        # HMA at present can't store multiple content types in one bank
+        # member.
```
Dcallies, line 216: blocking: Seems like a very important part of the implementation! Users generally have the expectation that as long as they follow the interface rules, that implementers should be able to handle it. 

if not, we should tighten down either the interface or the implementation. In this case, I think there might be a hacky solution that would handle. It could also be that we just need to add an "other" case that simply doesn't display on the UI, but still allows HMA to process it.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+        happen at most once per key per flush.
+        """
+        str_key = self._coerce_key(key)
+        member = self.banks_table.get_keyed_bank_member(
+            self._import_bank.bank_id, str_key
+        )
+        if member:
+            self.banks_table.remove_bank_member(member.bank_member_id)
+
+    def convert_opinions_to_label(self, opinions: t.Sequence[SignalOpinion]):
```
Dcallies, line 278: blocking: I'm uncomfortable with automatically taking in tags from APIs and putting them in directly as labels of the same category as the users. It's possible that the user takes a strong action in some case for matches that can be unexpectedly triggered when they join a collaboration. I think the user needs to retain control of the action label namespace the whole time, and we should either just discard the tags for now, or if you want to keep them, namespace them with a string that's really unlikely to collide with the users -

collaboration:<collab_name>:tag

Dcallies (changes requested):
Some blocking comments:
1. Current implementation only accepts some of the space allowed for by the SignalExchangeAPI, which I think can be made 1:1 (comments inline on where)
2. For me a lot of this is going to come down to whether the BankStore implementation is correct or not, and so getting a good set of test data will be key. Check out what I did for https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/tests/test_state.py - while you don't have to get this comprehensive, using standin APIs to prove that the bank storage should work for any update sequence might be useful. Since you are going one signal at a time, your possibility space is much smaller: A -> A' -> Delete A. I'm looking for a test that contains something like an update sequence of fetches that then asserts that the bank content is correct. I'm not sure how hard that test is to write (having not fully absorbed the current set).

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
schatten: Thanks! I'm not fully onboard with the functions are better than unittest.TestCase paradigm. Probably because I've not done enough work to form an opinion.

However, in this case, we have done a ton of work into designing a test classes hierarchy that makes testing dynamodb models really easy. Unless we go through a rigorous conversion in hmalib too, test classes are here to stay.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
schatten: I prefer declaration and usage of a variable be local (in the sense that they inhabit close line numbers) if it is only used in a single place. 

Can this be in init? Yes. Is any other function using it? Nope.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
```diff
@@ -14,41 +17,100 @@
...
+                table_manager, bank_id, member_key
             )
 
+            bank_member_2 = self._add_photo_bank_member_with_key(
+                table_manager, bank_id_2, member_key
+            )
+
+            self.assertNotEqual(bank_member.bank_member_id, bank_member_2)
+
+    def test_add_multiple_signals_for_keyed_bank_member(self):
```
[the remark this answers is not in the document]
schatten: Sure. I can whip something up.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
```diff
@@ -14,41 +17,100 @@
...
-                    raw_content=None,
-                    notes="",
-                    is_media_unavailable=True,
-                )
+            all_signals = table_manager.get_signals_for_bank_member(
+                bank_member.bank_member_id
+            )
+            self.assertSetEqual(
+                {PDQ_1, PDQ_2, MD5},
+                set(map(lambda x: x.signal_value, all_signals)),
```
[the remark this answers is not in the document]
schatten: Fair. Will make these changes. My Java experience trips me up sometimes. 

What's idiomatic in one language is often idiotic in another. :P

hasher-matcher-actioner/hmalib/common/models/tests/test_signal_uniqueness.py
```diff
@@ -1,17 +1,30 @@
...
+    @contextmanager
+    def fresh_table_manager(self):
+        with self.fresh_dynamodb():
+            try:
+                yield BanksTable(
+                    self.get_table(),
+                    signal_type_mapping=get_default_signal_type_mapping(),
+                )
+            finally:
+                pass
```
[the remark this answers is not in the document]
schatten: Fair point. I think I added this to put a debugger breakpoint and then forgot about it. Will remove and try.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+from hmalib.common.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class NoImportAsBankDefined(Exception):
+    def __init__(self, collab_config: EditableCollaborationConfig):
+        self.collab_config = collab_config
+        super().__init__(
+            f"CollabConfig {collab_config.name} is not connected to an importable bank. Create a bank, and save again using tx_collab_config apis."
```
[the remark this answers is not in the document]
schatten: fair. can convert to an assertion. I've never found a good theory of when to use exceptions and when not to for python. Any pointers?

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+
+
+class MultipleContentTypesInFetchDeltaRow(Exception):
```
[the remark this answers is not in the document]
schatten: Cool!

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+    pass
+
+
+class BankCollabFetchStore:
+    """
+    Translates fetch operations into bank operations. Ideally, should not need
+    to store anything outside of banks.
```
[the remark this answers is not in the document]
schatten: Agreed.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def get_checkpoint(self) -> t.Optional[FetchCheckpointBase]:
+        return self.banks_table.get_bank_info(self._import_bank.bank_id)
+
+    def set_checkpoint(self, checkpoint: FetchCheckpointBase):
+        return self.banks_table.update_bank_info(self._import_bank.bank_id, checkpoint)
+
+    def merge(self, delta: FetchDelta) -> None:
+        """
+        David's suggestion was:
```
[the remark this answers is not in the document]
schatten: Wow. This was not meant to be here at all. Lol. Removing and replacing with a comment. I jotted this down during one of our calls. :P

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+        - entry.member_id-entry.id -> multiple hashes...
+        StopNCII, FBThreatExchange FetchDelta.updates key
+        - hash-type, hash-value
+        """
+
+        if len(delta.updates) == 0:
+            logger.warning("merge() called with no updates")
+            return
+
+        delta_updates = t.cast(
```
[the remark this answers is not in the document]
schatten: Exactly my problem.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+            t.Dict[TUpdateRecordKey, t.Optional[TUpdateRecordValue]], delta.updates
+        )
+
+        for update_key, update_value in delta_updates.items():
+            # NCMECSignalExchangeAPI.fetch_iter() does not clear metadata,
```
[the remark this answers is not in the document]
schatten: This is what we were discussing for #1218. fetch_iter for ncmec does not remove the metadata when it encouters a delete. That is left for fetch_value_merge. 

If we make fetch_iter return `{key: None}` for cases where there should be a delete, then we can remove the fetch_value_merge call.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
schatten: Long term, I think we can have an "Any" content type for bank members.

hasher-matcher-actioner/webapp/src/pages/bank-management/ViewBank.tsx
```diff
@@ -55,9 +55,15 @@ function BankDetails({bankId}: BankTabProps): JSX.Element {
 
 export default function ViewBank(): JSX.Element {
   const {bankId, tab} = useParams<{bankId: string; tab: string}>();
+  const [bank, setBank] = useState<Bank>();
   const history = useHistory();
 
-  const pageTitle = 'Edit Bank';
+  useEffect(() => {
+    fetchBank(bankId).then(setBank);
```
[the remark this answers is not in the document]
schatten: You could configure that to be the case.. 

useEffect is simply telling react to do something after the element has rendered in the UI. This allows you to display empty states, loading gifs and stuff like that.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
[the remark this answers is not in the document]
schatten: This is proving to be harder than I thought. I thought the NCMEC API would be best for this test because the data model atleast allows me to return more than one signal type per row..

However, `[_get_conversion](https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py#L115)` prevents me from doing exactly that. 

I'd have to invent a new SignalExchangeAPI to get this to work. Back-burnering this ask as #1219 instead.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
Dcallies: I'm not sure if this usage is truly local if it's being attached to the class variable. Dynamic attribute setting is a source for difficult to debug errors, and so recommend it's usage in code unless absolutely necessary.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
Dcallies: If you weren't using that hierarchy, my opinion is it's worth it to convert having gone through the exercise once.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
Dcallies: What's stopping us from doing that now, and not having it displayed?

hasher-matcher-actioner/hmalib/fetching/bank_store.py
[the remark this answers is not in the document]
Dcallies: It comes down to the design of your modules. In general:
1. Use exceptions for exceptional events. If failure is likely and should always be handled (i.e. file.exists()), maybe return whether the operation was successful or not. 
2. Avoid `except Exception` except for outmost error handlers
3. Never `except BaseException` / `except:` unless you REALLY know what you are doing, since you will also catch things like interrupts, os.exit() and out-of-memory events, and your program is almost certainly boned or expected to close
4. Prefer existing exceptions (ValueError, KeyError) if they exactly map to what you are doing
5. If you do decide to use exceptions, create your own base class and catch that class at error boundaries.

The reason to assert here would be a combination of guidelines 1/4. We don't expect this state to ever be present in the application on purpose, so assert seems to map well to the expectations.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,289 @@
...
+            # need to call fetch_value_merge to clear metadata.
+            update_value = self._api_cls.fetch_value_merge(None, update_value)
+
+            if update_value is None:  # This is a delete.
+                self._unflushed_deletes.add(update_key)
+                # Why update something if it will end up getting deleted?
```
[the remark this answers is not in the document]
Dcallies: boop on updating comment

hasher-matcher-actioner/hmalib/fetching/bank_store.py
Dcallies: boop, did I miss it in your new docstring?

hasher-matcher-actioner/hmalib/fetching/bank_store.py
Dcallies: boop

hasher-matcher-actioner/hmalib/fetching/bank_store.py
Dcallies: boop

Dcallies (changes requested):
Need to follow up with you offline to figure out if the strategy is land/iterate or not.

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
```diff
@@ -112,5 +112,5 @@ def test_add_multiple_signals_for_keyed_bank_member(self):
             )
             self.assertSetEqual(
                 {PDQ_1, PDQ_2, MD5},
-                set(map(lambda x: x.signal_value, all_signals)),
+                set([x.signal_value for x in all_signals]),
```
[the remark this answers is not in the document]
schatten: fixed in a yet to be pushed commit.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -41,6 +47,31 @@ class BankCollabFetchStore:
...
+    Borrows heavily from python-threatexchange's fetch_cmd.py. fetch_cmd.py and
+    bank_store.py together are the strictest users of python-threatexchange's
+    SignalExchangeAPI interface methods.
+
+    fetch_cmd stores the fetched data in a SignalExchange specific format
+    defined by SignalExchangeAPI's generic typevars. Different
+    SignalExchangeAPI's data is later converted into a common type during
+    indexing by calling SignalExchangeAPI.naive_convert_to_signal_type.
+
+    bank_store OTOH, calls naive_convert_to_signal_type at fetch time itself and
```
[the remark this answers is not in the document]
schatten: expanded in a yet-to-be-pushed commit.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
schatten: This is a quick follow. #1220

hasher-matcher-actioner/hmalib/fetching/bank_store.py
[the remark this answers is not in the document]
schatten: blocking: I'm not certain that naive_convert_to_signal_type should decide what gets deleted. 

The SignalExchangeAPI interface (the best I can read it) does not guarantee that FetchDelta will return all signals for a changed key.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
schatten: Addressed in a yet-to-be pushed changeset.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,306 @@
...
+        # Call naive_convert_to_signal_type one update at a time, to get the
+        # full expected state of a bank-member.
+        signal_type_view = self._api_cls.naive_convert_to_signal_type(
+            self._signal_types, self.collab.to_pytx_collab_config(), {key: value}
+        )
+
+        if len(signal_type_view) == 0:
+            # Cases where none of the signal types in the update can be
+            # processed. eg. NCMEC returning pdna
+            return
```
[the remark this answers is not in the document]
schatten: I have a difference of opinion here.

My reading is:
- naive_fetch_merge decides what keys gets deleted or updated.
- naive_convert_to_signal_type converts the results of naive_fetch_merge into something that the bank can understand.

So, if naive_fetch_merge says update, that should never be converted to a delete by naive_convert_to_signal_type.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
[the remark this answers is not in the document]
schatten: After offline discussion, we decided to delete when naive_convert_to_signal_type returns no signals. 

Use-case.
`K1: [pdq: some hash, phash: some hash]`

Later, fetch_iter returns:
`K1: [phash: some hash]`

Indicating that the PDQ should be deleted. Note that the exchange in not deleting the Key K1, but asking that the pdq hash be removed.

To address this, in my local, I've now used.

```python
        if len(signal_type_view) == 0:
            # Cases where none of the signal types in the update can be
            # processed. This should result in a delete.
            #
            # Imagine an exchange returned a phash and a PDQ for an image in the
            # same key. Later, they removed the PDQ hash, but not the phash.
            # Unless we take action here, the PDQ never gets removed from the
            # bank.
            return self._handle_delete(key=key)
```

hasher-matcher-actioner/hmalib/fetching/bank_store.py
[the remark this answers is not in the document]
schatten: Talked to @Dcallies offline.

naive_fetch_merge is final and fetch_value_merge is gonna get deprecated in #1218. All key deletes should eventually be handled by fetch_iter directly().

All signal deletes would be handled by convert_to_signal_type.

In latest code, have fixed this as:

```python
        if len(signal_type_view) == 0:
            return self._handle_delete(key=key)
```

hasher-matcher-actioner/hmalib/fetching/bank_store.py
schatten: TODO: #1221

hasher-matcher-actioner/hmalib/common/models/tests/test_bank_member_with_key.py
```diff
@@ -16,11 +16,12 @@
 
 
 class BankMemberWithKeyTestCase(BanksTableTestBase, unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._id_count = 0
```
Dcallies, line 21: mrthanks

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -0,0 +1,319 @@
...
+
+        if len(signal_type_view) == 0:
+            # Cases where none of the signal types in the update can be
+            # processed. This should result in a delete.
+            #
+            # Imagine an exchange returned a phash and a PDQ for an image in the
+            # same key. Later, they removed the PDQ hash, but not the phash.
+            # Unless we take action here, the PDQ never gets removed from the
+            # bank.
+            return self._handle_delete(key=key)
```
Dcallies, line 239: I think this works!

Dcallies (approved):
All the things I was most concerned about are covered, and we have a (IMHO) a solid plan for backfilling testing and moving forward.

SHIP IT

It's time for trunk to be buildable again!

PR #1223, opened 2022-12-01
[smol] Get docker build to work and miscellaneous changes

schatten:
Summary
---
- Docker copy does not succeed without atleast one file
- add python-threatexchange to our regular devcontainer workspaces so we've got python-threatexchange to work with
- add working directory to migrations; it was failing on our CI deploy
- update keyserver key in devcontainer dockerfile - the current one stopped working and stackoverflow gave this answer

Test Plan
---
Will know after CI. Locally making these changes on our ec2 host allowed CI to work.


Dcallies (approved):
Do the thing!

PR #1224, opened 2022-12-05
Support user managed secrets using AWS Secrets Manager

schatten:
Summary
---
So far, we were using terraform to manage secrets because the number of secrets (credentials) was controlled by terraform.

With the new extensions model, we need to manage an arbitrary number of secrets. Think NCMEC, StopNCII etc. It would be best to be able to manage them from HMA UI itself. That way, updating your keys if they change or adding a new collab for a never before used signal exchange becomes possible without developer support.

Test Plan
---
```
$ terraform plan / apply.
```

Use the python REPL to test out `get_secret`, `put_secret` methods in AWS Secrets. 
1. Verify that put_secret creates OR updates secrets 
2. Verify that get_secret works for existing secrets and throws ValueError for non-existing ones.

### DIFF
--- a/hasher-matcher-actioner/hmalib/writebacker/writebacker_base.py
+++ b/hasher-matcher-actioner/hmalib/writebacker/writebacker_base.py
@@ -157,7 +157,7 @@ class ThreatExchangeWritebacker(Writebacker):
         mock_te_api = os.environ.get("MOCK_TE_API")
         if mock_te_api == "True":
             return MockedThreatExchangeAPI()
-        api_token = AWSSecrets().te_api_token()
+        api_token = AWSSecrets("").te_api_token()
         return ThreatExchangeAPI(api_token)
 
     def my_descriptor_from_all_descriptors(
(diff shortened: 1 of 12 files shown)


hasher-matcher-actioner/hmalib/writebacker/writebacker_base.py
```diff
@@ -157,7 +157,7 @@ def te_api(self) -> ThreatExchangeAPI:
         mock_te_api = os.environ.get("MOCK_TE_API")
         if mock_te_api == "True":
             return MockedThreatExchangeAPI()
-        api_token = AWSSecrets().te_api_token()
+        api_token = AWSSecrets("").te_api_token()
```
schatten, line 160: I don't see the existing writebacker infra working with the signal exchange API abstraction. We might have to rewrite this completely. 

So, I didn't invest time in bringing the prefix all the way to this module.

schatten: This will still work because as clarified in the aws_secrets.py docstring, the prefix is not used for te_api_token() or hma_api_tokens()

Dcallies (approved):
This looks like the right programming language, let's push it to prod and see what happens!

PR #1226, opened 2022-12-06
[hma] Fix indexing, matching and hashing

schatten:
Summary
---------
This is mostly doing what mypy fails to do. Tried out some common HMA flows (submitting an image, building an index, matching against an index) and solved what the logs deemed broken.

In hasher-matcher-actioner/hmalib/lambdas/common.py, we've extracted out a function that's common across all lambdas. This new function allows you to get a config backed signal_type_mapping easily.

A new addition is an S3 backed index store. The older way was to provide a mixin, but that won't fly with the new threatexchange extensions model because we want _any_ SignalTypeIndex class to be used and dynamically generated classes do not pickle.

Test Plan
---------
Terraform build / apply.

Tested out a submission..
Screen Shot 2022-12-05 at 19 11 11

tested out an index build..
- indexes are getting built regularly from a million PDQ hashes.

Both these flows were failing, but mypy never caught the errors. In its defense, some of the errors were not type errors.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/bank.py
+++ b/hasher-matcher-actioner/hmalib/common/models/bank.py
@@ -794,8 +794,9 @@ class BanksTable:
         self,
         signal_type: t.Type[SignalType],
         exclusive_start_key: t.Optional[DynamoDBCursorKey] = None,
-        limit: int = 100,
+        limit: t.Optional[int] = None,
     ) -> PaginatedResponse[BankMemberSignal]:
+        # TODO: Respect limit.
         if not exclusive_start_key:
             result = self._table.query(
                 IndexName=BankMemberSignal.BANK_MEMBER_SIGNAL_CURSOR_INDEX,
@@ -803,7 +804,6 @@ class BanksTable:
                 KeyConditionExpression=Key(
                     BankMemberSignal.BANK_MEMBER_SIGNAL_CURSOR_INDEX_SIGNAL_TYPE
                 ).eq(signal_type.get_name()),
-                Limit=limit,
             )
         else:
             result = self._table.query(
@@ -813,7 +813,6 @@ class BanksTable:
                     BankMemberSignal.BANK_MEMBER_SIGNAL_CURSOR_INDEX_SIGNAL_TYPE
                 ).eq(signal_type.get_name()),
                 ExclusiveStartKey=exclusive_start_key,
-                Limit=limit,
             )
         return PaginatedResponse(
             t.cast(DynamoDBCursorKey, result.get("LastEvaluatedKey", None)),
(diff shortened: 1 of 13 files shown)


hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -794,16 +794,16 @@ def get_bank_member_signals_to_process_page(
         self,
         signal_type: t.Type[SignalType],
         exclusive_start_key: t.Optional[DynamoDBCursorKey] = None,
-        limit: int = 100,
+        limit: t.Optional[int] = None,
```
schatten, line 797: Instead of imposing a client side arbitrary limit, we now let DynamoDB decide how big a page it can send in one query.

This allows for far better performance. With a size of 100, the round-trip latencies added up and made it impossible to build an index for a million entries in a single indexing lambda run. With removing the limit, we've reached peak performance. :)

hasher-matcher-actioner/hmalib/lambdas/common.py
```diff
@@ -0,0 +1,16 @@
...
+
+import functools
+
+from hmalib.common.mappings import (
+    HMASignalTypeMapping,
+    get_pytx_functionality_mapping,
+)
+
+
+@functools.lru_cache(maxsize=None)
```
schatten, line 11: All lambdas needed this, so extracted into a dedicated module from hashing.py

hasher-matcher-actioner/hmalib/common/models/bank.py
Dcallies: Thanks for the explainer! Will we ever get into a position where the index cannot be built in a single lambda run?

hasher-matcher-actioner/hmalib/matchers/matchers_base.py
```diff
@@ -223,32 +224,18 @@ def get_index(self, signal_type: t.Type[SignalType]) -> SignalTypeIndex:
...
-
-        for index_cls in indexes:
-            if max_custom_threshold <= index_cls.get_index_max_distance():
-                return index_cls
-
-        # if we don't have an index that supports max threshold
-        # just return the one if the highest possible max distance
-        return indexes[-1]
+        # TODO: Figure out how to switch index type when max_custom_threshold
+        # changes.
```
Dcallies, line 236: This and dynamically adjusting the SignalType configs are both hard problems at the moment (i.e. disabling signal types).

Dcallies (approved):
(ignorable): In your screenshot, the match references matching against data in ThreatExchange, which now might be against other APIs or local data.

I didn't see anything that looked crazy, so here's an accept

schatten:
> (ignorable): In your screenshot, the match references matching against data in ThreatExchange, which now might be against other APIs or local data.

The progress page screenshot has things hardcoded to ThreatExchange. Remnant of an older time. Added an item to the project https://github.com/orgs/facebook/projects/18/views/1.

PR #1227, opened 2022-12-06
Allow using config backed signal_type_mapping in all submit APIs.

schatten:
Summary
---
With new pytx, submit APIs need to know the supported signal types and content types. The way to do that is using the signal_type mapping in our middleware.py

hmalib.lambdas.middleware.jsoninator is an  bottle plugin that does automatic serialization and deserialization of request / response bodies.

Now, jsoninator takes additional signal_type_mapping parameters allowing it to deserialize objects that need signal_type_mapping.

Test Plan
---
Where all submission to the UI were failing, after doing these changes, they succeed.


Dcallies (approved):
So many layers!

PR #1232, opened 2022-12-09
stopncii submit_feedbacks will throw AttributeError

Gnomeek:
Summary
---------

Calling `StopNCII().submit_feedbacks()` will throw an `AttributeError` even though parameter is fine.
A unittest to catch this error is included in this PR.

Test Plan
---------

`StopNCII().submit_feedbacks()` should not raise error expect for `NetworkError`, etc.

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
@@ -2,8 +2,7 @@
 
 """Simple implementation for the StopNCII REST API"""
 
-from dataclasses import dataclass, asdict, field
-import copy
+from dataclasses import dataclass, field
 import enum
 import logging
 import time
@@ -62,12 +61,11 @@ class StopNCIICSPFeedback:
     tags: t.Set[str] = field(default_factory=set)  # Unstructured additional tags
     source: str = ""  # Name of the Content Service Provider (CSP)
 
-    @classmethod
-    def as_dict_for_post(cls) -> t.Dict[str, t.Any]:
+    def as_dict_for_post(self) -> t.Dict[str, t.Any]:
         """The json-friendly format to send in post requests"""
         return {
-            "tags": list(cls.tags),
-            "feedbackValue": str(cls.feedbackValue),
+            "tags": list(self.tags),
+            "feedbackValue": self.feedbackValue.value,
         }
 
 
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/tests/test_api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/tests/test_api.py
@@ -256,11 +256,32 @@ def mock_feedbacks() -> t.Dict[t.Tuple[StopNCIISignalType, str], StopNCIICSPFeed
     }
 
 
-def test_post_feedbacks(api: StopNCIIAPI):
-    try:
-        api.submit_feedbacks(mock_feedbacks())
-    except Exception as e:
-        assert (
-            repr(e)
-            == """AttributeError("type object 'StopNCIICSPFeedback' has no attribute 'tags'")"""
-        )
+def mock_submit_feedback_post_impl(endpoint: str, json):
+    desired_json = {
+        "count": 2,
+        "hashFeedbacks": [
+            {
+                "tags": ["Adult", "Nude"],
+                "feedbackValue": "Deleted",
+                "hashValue": "2afc4a5c09628a7961c14d436493bba66b89b831453baa1d556ba385554daa82",
+            },
+            {
+                "tags": ["Nude"],
+                "feedbackValue": "Blocked",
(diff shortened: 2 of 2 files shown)


Gnomeek:
Hey @dreiss , just find an issue in `StopNCII().submit_feedbacks()`. I have attached a reproducible unittest in this PR and will commit a patch to fix it.

Gnomeek:
the solution is quite straight-forward.

```python
@dataclass
class StopNCIICSPFeedback:
    """Feedback on a hash from a Content Service Provider"""

    feedbackValue: StopNCIICSPFeedbackValue  # What the feedback is
    tags: t.Set[str] = field(default_factory=set)  # Unstructured additional tags
    source: str = ""  # Name of the Content Service Provider (CSP)

    # --@classmethod
    # --def as_dict_for_post(cls) -> t.Dict[str, t.Any]:
    def as_dict_for_post(self) -> t.Dict[str, t.Any]:
        """The json-friendly format to send in post requests"""
        return {
            # --"tags": list(cls.tags),
            # --"feedbackValue": str(cls.feedbackValue),
            "tags": list(self.tags),
            "feedbackValue": str(self.feedbackValue),
        }
```

Gnomeek:
#1233 link issue for record

Dcallies (changes requested):
Hey again @Gnomeek ! Thanks for the report. 

Since you have the repo, do you want to try your hand at the fix, or should I give a shot?

Simply re-request review if you would prefer to land as is and I'll take over.

Gnomeek:
> Hey again @Gnomeek ! Thanks for the report.
> 
> Since you have the repo, do you want to try your hand at the fix, or should I give a shot?
> 
> Simply re-request review if you would prefer to land as is and I'll take over.

yea, I can do that. Just wondering how to write proper unittests since `_post()` doesn't return anything, hence it's hard to do assertion.

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -2,8 +2,7 @@
 
 """Simple implementation for the StopNCII REST API"""
 
-from dataclasses import dataclass, asdict, field
-import copy
+from dataclasses import dataclass, field
```
Gnomeek, line 5: delete unused imports

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -62,12 +61,11 @@ class StopNCIICSPFeedback:
...
 
-    @classmethod
-    def as_dict_for_post(cls) -> t.Dict[str, t.Any]:
+    def as_dict_for_post(self) -> t.Dict[str, t.Any]:
         """The json-friendly format to send in post requests"""
         return {
-            "tags": list(cls.tags),
-            "feedbackValue": str(cls.feedbackValue),
+            "tags": list(self.tags),
+            "feedbackValue": self.feedbackValue.value,
```
Gnomeek, line 68: `str(self.feedbackValue)` looks like `StopNCIICSPFeedbackValue.Blocked`, which should be `Blocked`, derived from `enum.value`

python-threatexchange/threatexchange/exchanges/clients/stopncii/tests/test_api.py
```diff
@@ -237,3 +237,51 @@ def test_mocked_get_hashes_with_undefined_enum(monkeypatch: pytest.MonkeyPatch):
...
+            },
+            {
+                "tags": ["Nude"],
+                "feedbackValue": "Blocked",
+                "hashValue": "79e07de27d7295339435d63cd31cf35a7bfa29eb2885008500a588a5ea3ae75a",
+            },
+        ],
+    }
+    assert endpoint == StopNCIIEndpoint.SubmitFeedback
+    assert json == desired_json
```
Gnomeek, line 276: assert here since `submit_feedbacks` has no return value

Dcallies:
Hey @Gnomeek, it's always a challenge to test external APIs, since the API can change even if your unittests all still pass!

If you can fix the bug and your labels now show up in the NCII database in your test plan, I'd accept that even without a unittest.

For the lint errors, I've had that on the backburned to fix - this issue is we just need to fix it to a single version (the same one that HMA uses) - they passed at one point, but mypy has gotten better at detecting historical errors.

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -62,12 +61,11 @@ class StopNCIICSPFeedback:
     tags: t.Set[str] = field(default_factory=set)  # Unstructured additional tags
     source: str = ""  # Name of the Content Service Provider (CSP)
 
-    @classmethod
-    def as_dict_for_post(cls) -> t.Dict[str, t.Any]:
```
Dcallies, line 66: Weird, I'm not sure why this was a classmethod before. Clearly I never tried to write a feedback with it (hence the bug)!

Dcallies (approved):
Thanks again for another fix! I'll roll a new version on Monday.

PR #1234, opened 2022-12-10
fix #1222: Add unit test for fetcher; fix bugs found as a result

### LINKED ISSUE #1222: [unittest] Add test for bank fetch store
Dipanjan to own the source of truth on list of edge cases.

I'm attempting to show a single update stream that can exercise all code paths, but the actual tests might break them up.

## Explainer
fetch(): SignalExchangeAPI.fetch()
merge(): SignalExchangeAPI.merge (TODO forgot the real name)
convert(): SignalExchangeAPI.naive_convert_to_signal_type (TODO forgot real name)
`K#`:  a key of the update dict returned by fetch()
`None`: if merge(fetch()[K#])  returns `None` (Signifying delete)
`S#`: a unique (type, val) handled signal type returned by filter(convert({K#: merge(fetch()[K#]})>, known_signal_types) 
`[S#, ...]`: The (unordered) result of calling filter(convert({K#: merge(fetch()[K#]})>, known_signal_types)
`[]`: If calling filter(convert({K#: merge(fetch()[K#]})>, known_signal_types) returns an empty set
`X => Y`: Redundant information - shows previous state (X) to the updated state (Y)

## Cases:
These could be presented as a sequence of fetch() updates. At the end of each step, we can assert that the index would be built with the expected signals (`[S1, S2, ..., S#]`).

| Update number  | State Change | Note |
| ------------- | ------------- | ------------- |
| 1 | `K1: None`  | |
| 2 | `K1: [] `  | |
| 3 | `K1: [S1]`  | `[] => [S1]` |
| 4 | `K1: [S1]` | `[S1] => [S1]` |
| 5 | `K1: [S1 (photo), S2 (video)]` | `[S1] => [S1 (photo), S2 (video)]` Demonstrates multi-content restriction |
| 6 | `K1: [S2]`  | `[S1, S2] => [S2]` |
| 7 | `K1: [S3]` | `[S2] => [S3]` |
| 8 | `K2: [S4]`  | |
| 9 | `K3: [S3]` | |
| 10 | `K1: []`  | `[S3] => []` |
| 11 | `K3: None` | |

schatten:
Summary
---
Most of the new code sets up a Fake exchange that can return whatever data we command. This is stolen from test_state in python-threatexchange.

Expresses 4 of the behaviors in the issue. I'm a little confused about what the others are so will add them after discussing with the team.

Test Plan
---
New unit test, py.test, black, mypy yada yada.

### DIFF
--- a/hasher-matcher-actioner/hmalib/common/models/bank.py
+++ b/hasher-matcher-actioner/hmalib/common/models/bank.py
@@ -678,10 +678,12 @@ class BanksTable:
 
     def remove_bank_member(self, bank_member_id: str):
         """
-        Removes the bank member and associated signals from the bank. Merely
-        marks as removed, does not physically delete from the store. DOES NOT
-        stop matching until index is updated. DOES NOT undo any actions already
-        taken.
+        Removes the bank member from the bank. Merely marks as removed, does not
+        physically delete from the store.
+
+        DOES NOT stop matching until index is updated.
+        DOES NOT undo any actions already taken.
+        DOES NOT delete associated signals. Consider bank_operations.remove_bank_member.
         """
         bank_member_keys = self._table.query(
             IndexName=BankMember.BANK_MEMBER_ID_INDEX,
--- a/hasher-matcher-actioner/hmalib/fetching/bank_store.py
+++ b/hasher-matcher-actioner/hmalib/fetching/bank_store.py
@@ -17,6 +17,7 @@ from threatexchange.exchanges.fetch_state import (
 from hmalib.common.configs.tx_collab_config import EditableCollaborationConfig
 from hmalib.common.configs.tx_apis import ToggleableSignalExchangeAPIConfig
 from hmalib.common.models.bank import BanksTable
+from hmalib.banks import bank_operations as bank_ops
 from hmalib.common.logging import get_logger
 
 logger = get_logger(__name__)
@@ -303,7 +304,7 @@ class BankCollabFetchStore:
             self._import_bank.bank_id, str_key
         )
         if member:
-            self.banks_table.remove_bank_member(member.bank_member_id)
+            bank_ops.remove_bank_member(self.banks_table, member.bank_member_id)
 
     def convert_opinions_to_label(self, opinions: t.Sequence[SignalOpinion]):
         """
@@ -315,5 +316,5 @@ class BankCollabFetchStore:
         ddb using self.banks_table) can be written.
         """
         return reduce(
-            lambda tags, acc: acc.union(tags), map(lambda x: x.tags, opinions)
+            lambda tags, acc: acc.union(tags), map(lambda x: x.tags, opinions), []
         )
--- a/hasher-matcher-actioner/hmalib/fetching/fetcher.py
+++ b/hasher-matcher-actioner/hmalib/fetching/fetcher.py
@@ -91,6 +91,13 @@ class Fetcher:
             if c.to_pytx_collab_config().enabled
         ]
 
+    def get_store(
+        self, collab: tx_collab_config.EditableCollaborationConfig
+    ) -> BankCollabFetchStore:
+        return BankCollabFetchStore(
+            self.signal_type_mapping.signal_types, self.banks_table, collab
+        )
+
     def _execute_for_collab(
         self, collab: tx_collab_config.EditableCollaborationConfig
     ) -> bool:
@@ -99,9 +106,7 @@ class Fetcher:
             collab.to_pytx_collab_config().api
         ].for_collab(collab=collab.to_pytx_collab_config())
 
-        store = BankCollabFetchStore(
-            self.signal_type_mapping.signal_types, self.banks_table, collab
-        )
+        store = self.get_store(collab)
 
         checkpoint = store.get_checkpoint()
         if checkpoint:
--- a/hasher-matcher-actioner/hmalib/fetching/tests/__init__.py
+++ b/hasher-matcher-actioner/hmalib/fetching/tests/__init__.py
@@ -1,0 +1,2 @@
+# Not part of the hmalib public interface, however, because of how
+# SignalExchangeAPI classes are handled, we need this to be reachable by python
--- a/hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
+++ b/hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
@@ -1,0 +1,310 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+"""
+Most classes unapologetically stolen from
+threatexchange.exchanges.tests.test_state
+
+Verifies (using a mock bank) that the operations issued on a bank when fetching
+a collab are what we'd expect.
+"""
+
+import typing as t
+import contextlib
+from collections import defaultdict
+from dataclasses import dataclass
+from datetime import datetime
+import pytest
+import boto3
+from moto import mock_dynamodb2
+
+from threatexchange.signal_type.signal_base import SignalType
+from threatexchange.exchanges.signal_exchange_api import TSignalExchangeAPICls
+from threatexchange.exchanges.collab_config import (
+    CollaborationConfigBase,
+    CollaborationConfigWithDefaults,
+)
+from threatexchange.utils.dataclass_json import dataclass_dumps
+from threatexchange.exchanges.signal_exchange_api import SignalExchangeAPI
+from threatexchange.exchanges.fetch_state import (
+    FetchCheckpointBase,
+    TUpdateRecordKey,
+    TUpdateRecordValue,
+    SignalOpinion,
+    SignalOpinionCategory,
+    FetchedSignalMetadata,
+    FetchDelta,
+)
+from threatexchange.signal_type.md5 import VideoMD5Signal
+from threatexchange.signal_type.pdq.signal import PdqSignal
+from threatexchange.content_type.text import TextContent
+from threatexchange.content_type.content_base import ContentType
+
+from hmalib.fetching.fetcher import Fetcher
+from hmalib.fetching.bank_store import BankCollabFetchStore
+from hmalib.common.mappings import HMASignalTypeMapping
+from hmalib.common.models.bank import (
+    BanksTable,
+    Bank,
+    BankMember,
+    BankMemberSignal,
+    PaginatedResponse,
+)
+from hmalib.common.config import HMAConfig, create_config
+from hmalib.common.configs.tx_collab_config import EditableCollaborationConfig
+from hmalib.common.configs.tx_apis import ToggleableSignalExchangeAPIConfig
+
+from hmalib.common.tests.mapping_common import get_default_signal_type_mapping
+from hmalib.common.models.tests.ddb_test_common import (
+    DynamoDBTableTestBase,
+    HMAConfigTestBase,
+)
+from hmalib.common.models.tests.test_signal_uniqueness import BanksTableTestBase
+
+
+@dataclass
+class FakeCheckpoint(FetchCheckpointBase):
+    update_time: int
+
+
+@dataclass
+class FakeUpdateRecord:
+    # Interpret all values None as delete
+    signal_type: t.Optional[t.Type[SignalType]]
+    signal_value: t.Optional[str]
+    tag: t.Optional[str]
+
+
+@dataclass
+class FakeSignalMetadata(FetchedSignalMetadata):
+    tags: t.Set[str]
+
+    def get_as_opinions(self) -> t.Sequence[SignalOpinion]:
+        return [
+            SignalOpinion(True, SignalOpinionCategory.INVESTIGATION_SEED, self.tags)
+        ]
+
+
+# terrible-design, but it is a test. module level state to be overriden before
+# creating a fetcher instance. Since FakeAPI is instantiated via a classmethod
+# and FakeAPI needs to be discoverable as hmalib.foo.bar.tests.FakeAPI, I can't
+# use any kind of dynamic class gen.
+FETCH_RESPONSES: t.Sequence[
+    t.Dict[TUpdateRecordKey, t.Optional[TUpdateRecordValue]]
+] = []
+
+
+class ImaginarySignalType(SignalType):
+    @classmethod
+    def get_content_types(cls) -> t.List[t.Type[ContentType]]:
+        return [TextContent]
+
+
+class FakeAPI(
+    SignalExchangeAPI[
+        CollaborationConfigBase,
+        FakeCheckpoint,
+        FakeSignalMetadata,
+        str,
+        FakeUpdateRecord,
+    ]
+):
+    @classmethod
+    def for_collab(cls, collab: CollaborationConfigBase) -> SignalExchangeAPI:
+        return FakeAPI()
+
+    def fetch_iter(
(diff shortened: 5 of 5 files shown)


hasher-matcher-actioner/hmalib/common/models/bank.py
```diff
@@ -678,10 +678,12 @@ def update_bank_member(
 
     def remove_bank_member(self, bank_member_id: str):
         """
-        Removes the bank member and associated signals from the bank. Merely
-        marks as removed, does not physically delete from the store. DOES NOT
-        stop matching until index is updated. DOES NOT undo any actions already
-        taken.
+        Removes the bank member from the bank. Merely marks as removed, does not
```
schatten, line 681: Nasty bug. Documentation used to say that associated signals are removed, when they are not.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -303,7 +304,7 @@ def _handle_delete(self, key: TUpdateRecordKey):
             self._import_bank.bank_id, str_key
         )
         if member:
-            self.banks_table.remove_bank_member(member.bank_member_id)
+            bank_ops.remove_bank_member(self.banks_table, member.bank_member_id)
```
schatten, line 307: Use bank_ops.remove_bank_member which does the coordination of removing signals and the member itself.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
```diff
@@ -315,5 +316,5 @@ def convert_opinions_to_label(self, opinions: t.Sequence[SignalOpinion]):
         ddb using self.banks_table) can be written.
         """
         return reduce(
-            lambda tags, acc: acc.union(tags), map(lambda x: x.tags, opinions)
+            lambda tags, acc: acc.union(tags), map(lambda x: x.tags, opinions), []
```
schatten, line 319: Errored out if opinions was empty. I'm sure there is a comprehension that would work, but that's a fight for another day.

hasher-matcher-actioner/hmalib/fetching/fetcher.py
```diff
@@ -91,6 +91,13 @@ def _get_all_active_collabs(
             if c.to_pytx_collab_config().enabled
         ]
 
+    def get_store(
```
schatten, line 94: This style allows overriding when we are setting up tests.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+    # Helper to convert bank member signals response to something that can be
+    # compared with TEST_CASE data
+    return {item.signal_value for item in response.items}
+
+
+@pytest.mark.usefixtures(
+    "mock_enabled_signal_exchange_apis",
+)
+@pytest.mark.parametrize("i", range(len(TEST_CASES)))
+def test_for_case_until(i, banks_table):
```
schatten, line 280: I swallowed the pill!

functional tests are, in fact, much more natural and conducive to fixtures, and parameters and all the good things.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+        {VideoMD5Signal: {md5(1)}},
+    ),
+    (
+        "Must be able to replace signals of same key..",
+        {
+            K1: FakeUpdateRecord(VideoMD5Signal, md5(2), "tag-two"),
+            K2: FakeUpdateRecord(VideoMD5Signal, md5(3), "tag-three"),
+        },
+        {VideoMD5Signal: {md5(3)}},
+        NotImplemented,
```
schatten, line 260: Captures behaviors that are unsupported too so that we can fix them.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+
+# Each test case is a 3 value tuple. For each case n,
+# tuple[0] is a human readable test name. We'll print this if tests fail.
+# tuple[1] is the update to be emitted by fetch_iter()
+# tuple[2] is the expected end state after all n tuple[1]s have been emitted.
+# tuple[3] if NotImplemented, indicates unsupported behavior.
+#
+# Note: the test cases are cumulative. Each test case will behave as if all the
+# previous tests have been performed on this store.
+TEST_CASES = [
```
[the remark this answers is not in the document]
Dcallies: Yeah, let's sync over the next week

hasher-matcher-actioner/hmalib/common/models/bank.py
Dcallies: What's the expected long term behavior here? For a multitude of reasons, I assume we want these to be scrubbed from the bank at some point.

hasher-matcher-actioner/hmalib/fetching/bank_store.py
Dcallies: Strongly enough, the inner map is already basically a comprehension, just needs the map removed.

As written, I think this is going to create a lot of intermediary objects.

```
return {t for o in opinions for t in o.tag}
```
Since the return type is sequence, you may need wrap with `list()`

hasher-matcher-actioner/hmalib/fetching/fetcher.py
Dcallies: (ignorable): This would make a good docstring

hasher-matcher-actioner/hmalib/fetching/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Not part of the hmalib public interface, however, because of how
+# SignalExchangeAPI classes are handled, we need this to be reachable by python
```
Dcallies, line 2: Confused about why this comment is here.
Python packaging expects an `__init__.py` if you want to import any files from this folder, I don't think this has anything to do with SignalExchangeAPI.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+from hmalib.common.configs.tx_collab_config import EditableCollaborationConfig
+from hmalib.common.configs.tx_apis import ToggleableSignalExchangeAPIConfig
+
+from hmalib.common.tests.mapping_common import get_default_signal_type_mapping
+from hmalib.common.models.tests.ddb_test_common import (
+    DynamoDBTableTestBase,
+    HMAConfigTestBase,
+)
+from hmalib.common.models.tests.test_signal_uniqueness import BanksTableTestBase
+
```
Dcallies, line 62: 49 lines of imports!

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+        ]
+
+
+# terrible-design, but it is a test. module level state to be overriden before
+# creating a fetcher instance. Since FakeAPI is instantiated via a classmethod
+# and FakeAPI needs to be discoverable as hmalib.foo.bar.tests.FakeAPI, I can't
+# use any kind of dynamic class gen.
+FETCH_RESPONSES: t.Sequence[
+    t.Dict[TUpdateRecordKey, t.Optional[TUpdateRecordValue]]
+] = []
```
Dcallies, line 93: The secret here is to move this into a dedicated python fixture, or to use monkeypatch (a py.test fixture equivalent to mock.patch)

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+
+        return {
+            k: {ik: FakeSignalMetadata(iv) for ik, iv in v.items()}
+            for k, v in result.items()
+        }
+
+
+class FakeFetcher(Fetcher):
+    # Allows overriding store. Since the real fetcher must instantiate a store
+    # per collab, we need this override.
```
Dcallies, line 163: Seems also fine as a docstring

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+
+def md5(n: int) -> str:
+    return f"{n:032x}"
+
+
+def pdq(n: int) -> str:
+    return f"{n:064x}"
+
+
+# Each test case is a 3 value tuple. For each case n,
```
Dcallies, line 234: 3-4?

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+# tuple[0] is a human readable test name. We'll print this if tests fail.
+# tuple[1] is the update to be emitted by fetch_iter()
+# tuple[2] is the expected end state after all n tuple[1]s have been emitted.
+# tuple[3] if NotImplemented, indicates unsupported behavior.
```
Dcallies, line 238: python now supports shape-like dicts if you want to use that for special options. 

What does unsupported mean in this context?

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+        {K1: None},
+        {VideoMD5Signal: {md5(3)}},
+    ),
+]
+
+
+def unwrap_bank_signals_response(response: PaginatedResponse) -> t.Set[str]:
+    # Helper to convert bank member signals response to something that can be
+    # compared with TEST_CASE data
+    return {item.signal_value for item in response.items}
```
Dcallies, line 273: You did it!

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
Dcallies: For me, having the assert functions provide more context was the thing that motived me to start rewriting the old ones.

Otherwise, anything you can do with functional tests you can do with test case if you are motivated enough, but mix and matching is very nice.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+# Each test case is a 3 value tuple. For each case n,
+# tuple[0] is a human readable test name. We'll print this if tests fail.
+# tuple[1] is the update to be emitted by fetch_iter()
+# tuple[2] is the expected end state after all n tuple[1]s have been emitted.
+# tuple[3] if NotImplemented, indicates unsupported behavior.
+#
+# Note: the test cases are cumulative. Each test case will behave as if all the
+# previous tests have been performed on this store.
```
Dcallies, line 241: 🤔

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+        "Must be able to handle a delete for a known key.",
+        {K1: None},
+        {VideoMD5Signal: {md5(3)}},
+    ),
+]
+
+
+def unwrap_bank_signals_response(response: PaginatedResponse) -> t.Set[str]:
+    # Helper to convert bank member signals response to something that can be
+    # compared with TEST_CASE data
```
Dcallies, line 272: docstring

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+@pytest.mark.parametrize("i", range(len(TEST_CASES)))
+def test_for_case_until(i, banks_table):
+    signal_type_mapping = get_default_signal_type_mapping()
+    bank = banks_table.create_bank("TEST_IMPORT_COLLAB", "Description")
+
+    fake_config = get_fake_collab_config(bank.bank_id)
+    create_config(fake_config)
+
+    global FETCH_RESPONSES
+    FETCH_RESPONSES = [tc[1] for tc in TEST_CASES[: i + 1]]
```
Dcallies, line 288: Oof, this is going to be super hard to debug. The mostest rightest thing to do seems to be to break out this functionality with controllable SignalTypeAPI to its own unittest fixture, then provide an accessor on the returned object to manipulate the returns. An HMA fixture that wrapped that guy (or provided an accessor to connect them) would then enable a lot of future tests.

At the very least, why isn't this class state?

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
```diff
@@ -0,0 +1,310 @@
...
+
+    # fake_table = FakeBanksTable()
```
Dcallies, line 290: Dangling comment?

Dcallies (approved):
I'm getting the impression that this is a rush job, especially given its the weekend :P 

I think we're going to want to come back here and clean this up to some extent. Will sync with you when you are in town.

hasher-matcher-actioner/hmalib/fetching/tests/__init__.py
schatten: Typically, we don't make tests findable. Unless it is a case of re-use. 

I would not have needed to add the tests in the `hmalib.fetching.tests` module to the hmalib package, but because of the way SignalExchangeAPI works, I had to bite the bullet and add it in. 

This is not new, however, we have done it earlier to make some test utils re-useable. Anyhoo. Can revise this comment.

hasher-matcher-actioner/hmalib/common/models/bank.py
schatten: The recommended way is to use bank_operations.remove_bank_member which does the coordination for you. I wanted to keep table operations atomic as much as possible in this class.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
schatten: Added a comment. 

Unsupported means a missing feature that is not catastrophic but should be addressed soon. An example is fetch_iter K1 [PDQ] followed by K1 [PDNA] will not remove the PDQ.

This is problematic in the long run, but in the current crowd of signal exchanges, this is unlikely to cause a problem. Because we'd need the same key to resolve to multiple hashes of the same content type for this problem to have consequences.

hasher-matcher-actioner/hmalib/fetching/tests/test_fetcher_and_bank_store.py
schatten: <facepalm> Totally can be a class variable. Made it so.

PR #1235, opened 2022-12-10
UI to add new collab configs

schatten:
Summary
---
Adds a new modal that allows the user to add a new collab config. Some of this modal can be hidden. Eg why should the user have to type out the api name again? Why not enable by default.

Test Plan
---
Tested for NCMEC and FB ThreatExchange.

# Empty Collabs page
Screen Shot 2022-12-10 at 14 27 05

# Add Collabs Modal with no type selected
Screen Shot 2022-12-10 at 14 27 14

# NCMEC Collab selected
Screen Shot 2022-12-10 at 14 58 01

# Adding a FB ThreatExchange PG
https://user-images.githubusercontent.com/217056/206873136-bab7fedb-5ef6-4c19-96a1-2dcb43c7e196.mp4

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
@@ -5,6 +5,7 @@ APIs to power viewing and editing of collaborations.
 """
 
 import bottle
+import json
 from dataclasses import dataclass, fields
 import typing as t
 from enum import Enum
@@ -12,8 +13,8 @@ from enum import Enum
 from mypy_boto3_dynamodb.service_resource import Table
 
 from hmalib.common.config import HMAConfig
-from hmalib.common.configs import tx_collab_config
-from hmalib.common.mappings import import_class
+from hmalib.common.configs import tx_collab_config, tx_apis
+from hmalib.common.mappings import import_class, full_class_name
 from hmalib.common.mappings import HMASignalTypeMapping
 from hmalib.common.models.bank import Bank, BanksTable
 
@@ -21,6 +22,7 @@ from hmalib.lambdas.api.middleware import SubApp, jsoninator, JSONifiable
 
 from threatexchange.exchanges.collab_config import CollaborationConfigBase
 from threatexchange.utils.dataclass_json import dataclass_loads
+from threatexchange.exchanges.impl.file_api import LocalFileSignalExchangeAPI
 
 
 @dataclass
@@ -41,6 +43,10 @@ def _serialize_type(f):
         return f.type.__name__
 
 
+def _to_schema(config_class: t.Type[CollaborationConfigBase]):
+    return {f.name: _serialize_type(f) for f in fields(config_class)}
+
+
 def get_collabs_api(
     hma_config_table: str, bank_table: Table, signal_type_mapping: HMASignalTypeMapping
 ) -> bottle.Bottle:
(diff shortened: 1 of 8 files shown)


hasher-matcher-actioner/webapp/src/pages/collaborations/AddCollaborationModal.tsx
```diff
@@ -0,0 +1,209 @@
...
+                {Object.keys(existingSchemas).length === 0 ? (
+                  <p>
+                    No Signal Exchange APIs enabled. Go to{' '}
+                    <Link to="/settings/exchanges">Settings</Link> and add one.
+                  </p>
+                ) : (
+                  Object.entries(existingSchemas).map(
+                    ([className, thisSchema]) => (
+                      <Button
+                        key={className}
```
Dcallies, line 159: What is adding the comma in the class name? (From the screenshots)

hasher-matcher-actioner/webapp/src/pages/collaborations/AddCollaborationModal.tsx
```diff
@@ -0,0 +1,209 @@
...
+                        variant="link"
```
Dcallies, line 160: This doesn't make it super clear that this is selectable. Is there a selector type available (dropdown?)

hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
```diff
@@ -56,17 +62,19 @@ def get_available_schemas():
 
         A future version could pull from a static location or scan all packages
         for subclasses.
-
-        PS: Is schemata the correct plural?
         """
-        configs = tx_collab_config.get_all_collab_configs()
+        apis = tx_apis.ToggleableSignalExchangeAPIConfig.get_all()
```
Dcallies, line 66: Why not prevent this from returning LocalFileSignalExchangeAPI?

hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
```diff
@@ -56,17 +62,19 @@ def get_available_schemas():
...
+        config_classes = [
+            api.to_concrete_class().get_config_cls()
+            for api in apis
+            # Do not want to handle local files without supporting s3 uploads
+            # for those files..
```
Dcallies, line 71: We'd probably just implement S3 exchange as its own style in hmalib (or as an extension).

hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
```diff
@@ -78,4 +86,53 @@ def get_collab_configs():
...
+        except ModuleNotFoundError:
+            bottle.abort(404)
+
+        return {
+            "class": full_class_name(pytx_collab_config_class),
+            "schema": _to_schema(pytx_collab_config_class),
+        }
+
+    @collabs_api.post("/add-collab-config")
+    def add_collab_config():
```
Dcallies, line 107: Is there an edit, or only an add?

Dcallies (approved):
As discussed offline, we're in rush mode at the moment, but hoping we can come back for a level of polish on some of these elements.

hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
schatten: So far, only an add. Even the UI does not allow one. This is because the behaviour when you change something like the privacy group id is undefined. 

Unless we add a bunch of validations, enable / disable would be the only change supported. That can / should be done via the bank.

hasher-matcher-actioner/hmalib/lambdas/api/collabs.py
schatten: Fair. Moved it into `ToggleableSignalExchangeAPIConfig`

hasher-matcher-actioner/webapp/src/pages/collaborations/AddCollaborationModal.tsx
schatten: There is a comma after `humanizeClassName(className)` in line 166. But we're changing that to selects per your suggestion. Which, makes total sense. :) Weekend me is not too smart.

PR #1236, opened 2022-12-10
Hook in secrets and the new fetcher to the fetching lambda

schatten:
Summary
---
Use AWSSecrets to store signal exchange credentials and use them during fetching. Also, hook up the new hmalib.fetching.fetcher modules into the lambda and cleanup the older threatexchange focussed fetching code.

Test Plan
---
Ran against a threatexchange privacy group and verified that banks are getting populated.


Dcallies (approved):
As discussed offline, moving quickly to get some last fixes in. Let's discuss polish for this week.

PR #1237, opened 2022-12-12
[py-tx] 1.0.9

Dcallies:
Contains fix for writefeedback (#1232)


BarrettOlson (approved):
LGTM. CI lint already broken due to a lib version it seems

PR #1241, opened 2023-01-02
Delete docs directory. We never ended up using this fully. Recommend using github wiki going forward

schatten:


BarrettOlson (approved):
It was @schatten that killed the docusaurus :) 

Thanks for the clean up!

Issue #1245, opened 2023-01-09
[Documentation] README for StopNCII credentials out of order


thedanielsun:
SUBSCRIPTION_KEY, FUNCTION_KEY are reversed in the README

README section: https://github.com/facebook/ThreatExchange/blob/31a044d80766998101ec7ac8a3afe87ea75aff64/python-threatexchange/README.md?plain=1#L198

API impl: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py#L79

BarrettOlson:
Thanks! 
Looked through the code and it appears consistent. The variable order was just swapped for the README. The only place in the code I see the order from the README is when keyword args are used, therefore I think we are safe just updating the README but I will leave to @Dcallies to confirm.

BarrettOlson:
Addressed in #1246

PR #1246, opened 2023-01-09
Update StopNCII key order in README.md

thedanielsun:
Summary
---------

Addresses #1245
Order of keys in the README doesn't match what is expected in the application
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py#L79
 
Test Plan
---------

Changes the README, not a functional change

### DIFF
--- a/python-threatexchange/README.md
+++ b/python-threatexchange/README.md
@@ -195,13 +195,13 @@ $ threatexchange fetch
 
 ```
 # Step 1 - TX_ACCESS_TOKEN Environment variable - comma separated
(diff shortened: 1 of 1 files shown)


Dcallies (approved):
Easy to make mistakes here, especially since it's hard to differentiate the two key types.

Before we can merge, can you confirm that the way the credentials are stored maps to the correct property names of the structure?

If you aren't sure how to do this, let me know and I can test it.

It's super hard to tell the keys apart as is, and I've already made one similar fix in the area (that I'm now worried was a breakage).

Dcallies:
Now that I'm not on my phone: the "fix" was on #1193

However,

PR #1250, opened 2023-01-18
Update file copyright headers and minor cleanup

BarrettOlson:
Summary
---------

New header is "Copyright (c) Meta Platforms, Inc. and affiliates." 
I updated existing files and added to a few missing ones (I am sure I missed a few)

Test Plan
---------

Changes should only be comments and formatting.

PR #1251, opened 2023-01-18
Add missing copyright headers and bump pyjwt version in hmalib

BarrettOlson:
Summary
---------

#1250 Updated the header for existing files, this adds to most of the files missing any form of a copyright header

Also updated pyjwt to 2.4.0 because of potential vulnerability

Test Plan
---------

Changes should only be comments and formatting.

Did an HMA test deploy to make sure "pyjwt[crypto]==2.4.0" didn't break anything.
 
I was able to query API from postman and UI so hmalib still builds correctly with version bump


Dcallies (approved):
This looks like the right repository!

PR #1254, opened 2023-01-19
[pytx] fix broken gh actions for pytest and lint

BarrettOlson:
Summary
---------

Unfortunately the test breaking were not showing up locally, so I used a dev container to get a close as possible to the gh action ENV so I could repro and fix.

Test Plan
---------
py.test and gh actions

### DIFF
--- a/python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
+++ b/python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
@@ -49,7 +49,7 @@ class ThreatExchangeCLIE2eHelper:
                 inner_main(args, state_dir=self._state_dir)
             except SystemExit as se:
                 if se.code != 0:
-                    raise E2ETestSystemExit(se.code)
+                    raise E2ETestSystemExit(t.cast(int, se.code))
             return fake_out.getvalue()
 
     def assert_cli_output(
@@ -72,7 +72,7 @@ class ThreatExchangeCLIE2eHelper:
             assert lines[line] == expected_line_output
 
     def assert_cli_usage_error(
-        self, args: t.Iterable[str], msg_regex: str = None
+        self, args: t.Iterable[str], msg_regex: str = ''
     ) -> None:
         with pytest.raises((CommandError, E2ETestSystemExit), match=msg_regex) as ex:
             self.cli_call(*args)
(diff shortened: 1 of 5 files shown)


python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -272,7 +272,7 @@ class SignalExchangeAPIWithSimpleUpdates(
         state.TFetchCheckpoint,
         state.TFetchedSignalMetadata,
         t.Tuple[str, str],
-        state.TFetchedSignalMetadata,
+        t.Any,
```
BarrettOlson, line 275: @Dcallies I'm not sure on the typing here. All changes besides `t.Any` expanded to 10 more errors where `SignalExchangeAPIWithSimpleUpdates` is being used.

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -72,7 +72,7 @@ def assert_cli_output(
             assert lines[line] == expected_line_output
 
     def assert_cli_usage_error(
-        self, args: t.Iterable[str], msg_regex: str = None
+        self, args: t.Iterable[str], msg_regex: str = ""
     ) -> None:
         with pytest.raises((CommandError, E2ETestSystemExit), match=msg_regex) as ex:
```
Dcallies, line 77: TIL: https://github.com/pytest-dev/pytest/blob/main/src/_pytest/python_api.py#L783

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
Dcallies: Hmm, this is definitely not right here, the other option is to just # type: ignore for now

Dcallies (approved):
blocking: That Any change is a bit on the sketch end, and its on one of the key APIs. 

If ignoring it will get the typechecker okay, I can come back to it in the future, which I like better than swapping to Any

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
BarrettOlson: I can't get mypy to understand the `# type: ignore` directive here. No matter where I put it (above, same line, below as part of class) it get a version of the following:
```
threatexchange/exchanges/signal_exchange_api.py:269: error: Variance of TypeVar "state.TFetchedSignalMetadata" incompatible with variance in parent type  [type-var]
threatexchange/exchanges/signal_exchange_api.py:277: error: Unused "type: ignore" comment
```

BarrettOlson: Synced offline, as ignore was not working I will merge as t.Any but created high-pri issue https://github.com/facebook/ThreatExchange/issues/1255

Issue #1259, opened 2023-02-06
pdq_hasher error for B/W png


thedanielsun:
threatexchange hash photo https://i.redd.it/4shux9eu3mga1.png
  File "/Users/daniel.sun/.pyenv/versions/3.8.12/lib/python3.8/site-packages/threatexchange/signal_type/signal_base.py", line 195, in hash_from_file
    return cls.hash_from_bytes(file.read_bytes())
  File "/Users/daniel.sun/.pyenv/versions/3.8.12/lib/python3.8/site-packages/threatexchange/signal_type/pdq/signal.py", line 77, in hash_from_bytes
    pdq_hash, quality = pdq_from_bytes(bytes_)
  File "/Users/daniel.sun/.pyenv/versions/3.8.12/lib/python3.8/site-packages/threatexchange/signal_type/pdq/pdq_hasher.py", line 33, in pdq_from_bytes
    return _pdq_from_numpy_array(np_array)
  File "/Users/daniel.sun/.pyenv/versions/3.8.12/lib/python3.8/site-packages/threatexchange/signal_type/pdq/pdq_hasher.py", line 37, in _pdq_from_numpy_array
    hash_vector, quality = pdqhash.compute(array)
  File "pdqhash/bindings.pyx", line 67, in pdqhash.bindings.compute
IndexError: index 2 is out of bounds for axis 2 with size 2 

----

Possible root cause:

i’m just looking at the ndarray size which result for the image
threatexchange hash photo https://i.redd.it/4shux9eu3mga1.png

```
_check_dimension_and_expand_if_needed size
(2048, 2004, 2)
_pdq_from_numpy_array
(2048, 2004, 2)
```
Pillow version: 9.3.0
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py#L50-L56

maybe this not working as intended? 
`ndim=3` here but maybe B/W conversion is not calculating dimension properly

Dcallies:
Thanks for the report Daniel! The logging output and links really help as well.

PR #1260, opened 2023-02-07
Fix pdq_hasher for LA images

thedanielsun:
Summary
---------
LA images: Luminance with Alpha returns a 3 dimensional ndarray but is a grayscale image.
This causes the image to fail to hash with pdqhash. See full error trace in #1259 

This PR fixes the issue by converting from LA to RGB before hashing LA images.

This PR preserves performance hack for L images, although this code can be refactored if necessary to make possible forks for different image types more consistent.

Test Plan
---------
Added resources folder with example image and unit test demonstrating pdq hash is successfully computed.


python-threatexchange/threatexchange/tests/hashing/test_pdq_hasher.py
```diff
@@ -59,7 +59,15 @@ def test_pdq_from_file(self):
             assert pdq_hash == RANDOM_IMAGE_PDQ
 
     def test_pdq_from_bytes(self):
-        """Runs the pdq hashser directly on bytes"""
+        """Runs the pdq hasher directly on bytes"""
```
Dcallies, line 62: 👍

python-threatexchange/threatexchange/tests/hashing/test_pdq_hasher.py
```diff
@@ -59,7 +59,15 @@ def test_pdq_from_file(self):
...
         bytes_ = base64.b64decode(RANDOM_IMAGE_BASE64)
         pdq_hash = pdq_hasher.pdq_from_bytes(bytes_)[0]
         assert pdq_hash == RANDOM_IMAGE_PDQ
+
+    def test_pdq_from_file_la_png(self):
```
Dcallies, line 67: (ignorable): For goofballs like me that have never seen la before, can you add a dobloc to reference "LA images (luminance with alpha)"

python-threatexchange/threatexchange/tests/hashing/test_pdq_hasher.py
```diff
@@ -59,7 +59,15 @@ def test_pdq_from_file(self):
...
+        file_path = pathlib.Path("threatexchange/tests/hashing/resources/LA.png")
+        expected_pdq_output = (
+            "5dbc6c369dc4476538bcf307e61d80994c5ed0a5790efaf999bc499ad3b02421",
```
Dcallies, line 70: (ignorable): We've seen some cases of drift between operating systems (or that might be video specific with videoPDQ), in which case it makes sense to do a distance check instead, but it should show up in a test run (if only it weren't blocked by other errors).

Dcallies (approved):
Thanks a ton for contributing the fix @thedanielsun ! 

Your lint errors are unrelated, but makes it possible to generate new ones. Since I'm in the area I can try and clean those up.

The test error looks unrelated at first glance (more oncall tasks for me!), but the code format issue is probably something you can fix by installing black (https://github.com/psf/black). 

The black issue also looks unrelated (likely a pinned version change), but unfortunately it blocks the typing (I should change the order).

PR #1261, opened 2023-02-07
[py-tx] Black 23.1.0

Dcallies:
Summary
---------

A new version of black is out, which adds some new rules! Surprisingly, doing `pip install --upgrade -e .[all]` didn't upgrade the version, I had to do it manually. 

It may be getting close to where we want to try and set a py env

Test Plan
---------

`black` in the root directory, see no changes.


BarrettOlson (approved):
Looks like either a version update or a weird formatting has made the CI test fail. Can you address before merge?

Dcallies:
That's not the reason that test is failing (it's also failing in trunk), will talk to you offline.

PR #1264, opened 2023-02-09
Add force_rgb option for pdq_hasher

thedanielsun:
Summary
---------
Found a new image type which fails to get hashed by existing pdq_hasher logic

```
% threatexchange hash --signal-type pdq photo https://i.redd.it/vum5cw55ryga1.png
  File "/Users/daniel.sun/.pyenv/versions/threatexchange/bin/threatexchange", line 33, in <module>
    sys.exit(load_entry_point('threatexchange', 'console_scripts', 'threatexchange')())
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 319, in main
    inner_main()
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 312, in inner_main
    execute_command(settings, namespace)
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 157, in execute_command
    command.execute(settings)
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/cli/hash_cmd.py", line 106, in execute
    hash_str = hasher.hash_from_file(file)
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/signal_type/signal_base.py", line 195, in hash_from_file
    return cls.hash_from_bytes(file.read_bytes())
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/signal_type/pdq/signal.py", line 77, in hash_from_bytes
    pdq_hash, quality = pdq_from_bytes(bytes_)
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py", line 28, in pdq_from_bytes
    return _pdq_from_image(Image.open(io.BytesIO(file_bytes)), force_rgb)
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py", line 33, in _pdq_from_image
    return _pdq_from_numpy_array(np_array)
  File "/Users/daniel.sun/src/ThreatExchange/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py", line 37, in _pdq_from_numpy_array
    hash_vector, quality = pdqhash.compute(array)
  File "pdqhash/bindings.pyx", line 71, in pdqhash.bindings.compute
ValueError: Buffer dtype mismatch, expected 'char' but got 'int'
```

```
>>> from PIL import Image
>>> f = open("/Users/daniel.sun/Downloads/vum5cw55ryga1.png", "rb")
>>> image = Image.open(f)
>>> image.mode
'I'
```
`I` image mode is 32-bit pixels which makes sense why error message says `expected 'char' but got 'int'`

This PR adds an option to force RGB conversion of input image for PDQ hashing
According to #465 this caused a perf regression, so making this feature an option for those who can trade-off latency for stability.

Test Plan
---------

Added this image as `I.png` in resources and add unit test which demonstrates calling `pdq_from_file` with `force_rgb` returns a PDQ hash instead of throwing above error

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
@@ -13,22 +13,23 @@ PDQOutput = t.Tuple[
 ]  # hexadecimal representation of the Hash vector and a numerical quality value
 
 
-def pdq_from_file(path: pathlib.Path) -> PDQOutput:
+def pdq_from_file(path: pathlib.Path, force_rgb=False) -> PDQOutput:
     """
     Given a path to a file return the PDQ Hash string in hex.
     Current tested against: jpg
     """
-    np_array = _convert_image_to_correct_array_dimension(Image.open(path))
-    return _pdq_from_numpy_array(np_array)
+    return _pdq_from_image(Image.open(path), force_rgb)
 
 
-def pdq_from_bytes(file_bytes: bytes) -> PDQOutput:
+def pdq_from_bytes(file_bytes: bytes, force_rgb=False) -> PDQOutput:
     """
     For the bytestream from an image file, compute PDQ Hash and quality.
     """
-    np_array = _convert_image_to_correct_array_dimension(
-        Image.open(io.BytesIO(file_bytes))
-    )
+    return _pdq_from_image(Image.open(io.BytesIO(file_bytes)), force_rgb)
+
+
+def _pdq_from_image(image: Image, force_rgb) -> PDQOutput:
+    np_array = _convert_image_to_correct_array_dimension(image, force_rgb)
     return _pdq_from_numpy_array(np_array)
 
 
@@ -46,11 +47,11 @@ def _pdq_from_numpy_array(array: np.ndarray) -> PDQOutput:
     return hex_str, quality
 
 
-def _convert_image_to_correct_array_dimension(image: Image) -> np.ndarray:
+def _convert_image_to_correct_array_dimension(image: Image, force_rgb) -> np.ndarray:
     """
     Handle possible image format conversion or
     """
-    if image.mode == "LA":
+    if force_rgb or image.mode == "LA":
         # LA images (luminance with alpha) return 3 dimensional ndarray
         # which is incompatible with pdqhash
         image = image.convert("RGB")
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
@@ -13,7 +13,7 @@ PDQOutput = t.Tuple[
 ]  # hexadecimal representation of the Hash vector and a numerical quality value
 
 
-def pdq_from_file(path: pathlib.Path, force_rgb=False) -> PDQOutput:
+def pdq_from_file(path: pathlib.Path, force_rgb: bool = False) -> PDQOutput:
     """
     Given a path to a file return the PDQ Hash string in hex.
     Current tested against: jpg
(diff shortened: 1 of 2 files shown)


python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
```diff
@@ -46,11 +47,11 @@ def _pdq_from_numpy_array(array: np.ndarray) -> PDQOutput:
     return hex_str, quality
 
 
-def _convert_image_to_correct_array_dimension(image: Image) -> np.ndarray:
+def _convert_image_to_correct_array_dimension(image: Image, force_rgb) -> np.ndarray:
```
Dcallies, line 50: type

python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
```diff
@@ -13,22 +13,23 @@
 ]  # hexadecimal representation of the Hash vector and a numerical quality value
 
 
-def pdq_from_file(path: pathlib.Path) -> PDQOutput:
+def pdq_from_file(path: pathlib.Path, force_rgb=False) -> PDQOutput:
```
Dcallies, line 16: blocking: This needs a type (: bool) - thought we added a flag to the mypy settings to enforce this, but it looks like it didn't catch it. 

Additionally, in what circumstances will the caller know the correct option but that code won't be able to figure it out?

python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
```diff
@@ -13,22 +13,23 @@
...
     Given a path to a file return the PDQ Hash string in hex.
     Current tested against: jpg
     """
-    np_array = _convert_image_to_correct_array_dimension(Image.open(path))
-    return _pdq_from_numpy_array(np_array)
+    return _pdq_from_image(Image.open(path), force_rgb)
 
 
-def pdq_from_bytes(file_bytes: bytes) -> PDQOutput:
+def pdq_from_bytes(file_bytes: bytes, force_rgb=False) -> PDQOutput:
```
Dcallies, line 24: add type hint

Dcallies (changes requested):
Thanks again for the PR Daniel! Really appreciate all the help!

blocking: Something I forgot to ask about in the previous PR - do we have rights to these images? We want to make sure they are in the public domain before adding them. You can additionally generate images using a variety of tools that would be rights free.

Additionally, there might be a better place to add test images - in https://github.com/facebook/ThreatExchange/tree/main/pdq/data/misc-images might be a good destination (though it's tricky to refer to from the test.

python-threatexchange/threatexchange/signal_type/pdq/pdq_hasher.py
thedanielsun: Updated PR with typehints.

> Additionally, in what circumstances will the caller know the correct option but that code won't be able to figure it out?

I think our code should try to do the correct thing by default, but I think we need to confirm (have tests) around a few things:

1. How bad are the performance tradeoffs of always converting to RGB first? In our experience, eating the perf for converting to RGB isn't a problem. But I think there are performance benefits to resizing before conversion for example. I'm not innately familiar with pdq implementation so I'm not sure if pdq C bindings expect RGB.
2. Do we have test coverage for all "image types"? I'm hoping we (maybe I) can just add test coverage for all of these https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes There's also a difference between not raising exceptions (which currently this library is) and this library doing the right thing ie computing the correct hash given the visual appearance.

thedanielsun:
I'll look into generating new image types based on existing image via Pillow API and conversion to see if we can not worry about rights issue. Will do that in a different PR

Dcallies (changes requested):
Can you link to this image's license information? We shouldn't merge this if it is a licensed image.

In your summary can you confirm you've run the py.test tests locally (since we can't trust the contbuild tests at the moment).

Dcallies:
Or maybe get permission from [/u/ehogg377](https://www.reddit.com/user/ehogg377/)? It seems like the subreddit is a bit ambiguous about copyright.

thedanielsun:
Opened #1265 
then I will convert all the tests for pdq_hasher here to try to read from that directory to run tests against which shouldn't have license issues

Dcallies:
Hey @thedanielsun, I'm going to close this PR for inactivity, feel free to re-open if you still want to try and get it merged!

PR #1265, opened 2023-02-13
Add same image in different formats to verify test data

thedanielsun:
Summary
---------

Added a bunch of test images to pdq/data/misc-images which are generated using Pillow for testing whether pdq hashing works with different modes.

These are all based off `b.jpg` in the repo for which we shouldn't need to worry about rights

Test Plan
---------


BarrettOlson (approved):
Looks reasonable to me, thanks!

Dcallies:
Thanks @thedanielsun , I love this approach!

PR #1266, opened 2023-02-16
[pytx] pin setup.py dependency lib versions to prevent test regressions

BarrettOlson:
Summary
---------

I did this for `hasher-matcher-actioner/setup.py` awhile ago but we hadn't yet done it for `python-threatexchange/setup.py`

This resulted in a regression in our CI tests because `dacite==1.8.0` was overzealous with their new caching functionality.  (not to mention a debugging nightmare)
See https://github.com/konradhalas/dacite/issues/215 
as they have yet to release 1.8.1 can should contains the fix I have pinned 1.7.0 for now.

To prevent this or related issue from happening in the future, I went ahead and pinned the other dependencies as well. We should periodically update them but confirm there are no test breakages when doing so to avoid this regression. 

Test Plan
---------

Previously failing pytest runs in CI and locally should now pass.

### DIFF
--- a/python-threatexchange/setup.py
+++ b/python-threatexchange/setup.py
@@ -23,10 +23,14 @@ for extension_dir in extensions_dir.iterdir():
 
 all_extras = set(sum(extras_require.values(), []))
 # We might not get any value from splitting these all out
-extras_require["test"] = sorted({"pytest"} | all_extras)
-extras_require["package"] = ["wheel"]
-extras_require["lint"] = ["black"]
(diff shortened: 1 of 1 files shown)


Dcallies (approved):
Pinning all these versions to make them reproducible is a great idea in general, and makes it easier to move these forward. Thanks a ton for the hard work! Getting these tests working again will help the external folks like @thedanielsun making commitments, since we can trust continuous build again.

PR #1278, opened 2023-03-23
hma: devcontainer fixes

dougneal:
Summary
---------
Fixed a few issues I ran into when building the dev container:
1. The vscode-managed "Docker in Docker" install was conflicting with the Docker CLI already installed inside the the container - removed it.
2. Fixed the build on arm64. This should improve the developer experience by allowing the dev container to run at native speeds on newer Apple computers and other arm64 machines.

Test Plan
---------

Rebuild and run devcontainer from VSCode.

I don't have an amd64 machine handy to test this on - anybody else?

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/Dockerfile
+++ b/hasher-matcher-actioner/.devcontainer/Dockerfile
@@ -4,7 +4,7 @@
 # TODO: pin version later. Otherwise starting up will be too slow.
 # Note: We specify AMD64 architecture as it is currently the only one supported for this devcontainer
 # relevant for ARM architecture machines as using `python:3.8-buster` will have it try to use that instead
-FROM amd64/python:3.9-buster
+FROM python:3.9-buster
 
 # [Unixname wrestling]
 # Some of our script (docker-related) are dependent on the unixname. Would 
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/.devcontainer/Dockerfile
```diff
@@ -4,7 +4,7 @@
 # TODO: pin version later. Otherwise starting up will be too slow.
 # Note: We specify AMD64 architecture as it is currently the only one supported for this devcontainer
 # relevant for ARM architecture machines as using `python:3.8-buster` will have it try to use that instead
-FROM amd64/python:3.9-buster
+FROM python:3.9-buster
```
BarrettOlson, line 7: Unfortunately the specificity of `FROM amd64/python:3.9-buster` is at least for now required for a successful HMA deployment. This is because various parts of the build out (specific I think we are just down to the webapp) don't have a supported runtime and error. 

If you know of a clean way to make this toggle-able I would be happy to merge it but I think we want sometime that will be closer to working by default (even if slower/not always native)

dougneal: OK, I'll follow the rest of the process through and see what else breaks, if it's not easily fixable, I'll close this.

BarrettOlson:
#1285 supplanted this

Issue #1249, opened 2023-01-13
HMA migrations that are hmalib scripts do not run successfully


dmukhg:
Right now, migrations that are simply `python -m hmalib.....`commands do not run successfully. 

This is because terraform is not able to run the command from the hasher-matcher-actioner directory which prevents python -m from finding the hmalib package.

Attempts at using working_directory inside the local_exec directive have not worked out. 

See a sample error:
```
╷
│ Error: local-exec provisioner error
│
│   with module.migrations.null_resource.default_signal_content_type_configs,
│   on migrations/main.tf line 9, in resource "null_resource" "default_signal_content_type_configs":
│    9:   provisioner "local-exec" {
│
│ Error running command 'python -m hmalib.scripts.cli.main migrate 2022_04_02_default_content_signal_type_configs --config-table barrett-HMAConfig': exit status 1. Output: /Users/barrett/.venv/hma/bin/python: Error while finding module specification for 'hmalib.scripts.cli.main'
│ (ModuleNotFoundError: No module named 'hmalib')
│
╵
```

PR #1281, opened 2023-03-24
[hma][setup] Move migrations that are actually core setup steps to root main.tf

BarrettOlson:
Summary
---------

#1249 Flags that migrations do not run successfully. My first thought was to disable them by default (as our most common usages is still fresh installs) however looking at the scripts themselves these "migrations" are actually a core apart of post deploy config set up. Meaning they are always required regardless.   

So instead of disabling I did the following:

1) moved the existing migration commands to the root `main.ft` as they are setup critical
2) update the command that was failing to first install `hmacli` via pip  instead of trying to run the command and hope everything is in the right place
 - This does have the added side effect of tf apply updating the python libraries installed, therefore I updated the README.md to included a recommendation for using python's `venv` (something I hope folks most were already doing).

I also went ahead anded added some error checking to one of the scripts run in the python setup command so both they can be rerun without issue.

Unfortunately I can't just delete `terraform/migrations` because then anyone already deployed will get an error. So I just emptied the .tf files and updated the local README (We will likely have a usage for this module for actual migrations eventually anyway.) 

Test Plan
---------
Made sure `make dev_apply_with_newest_docker_image` can be run without error and deploys cleanly.

Validated install and error handling on setup scripts:
```
python3 -m venv ~/.venv/hma
source ~/.venv/hma/bin/activate 
# force terraform to rerun the python scripts
terraform -chdir=terraform taint null_resource.default_system_config
<redacted>
Apply complete! Resources: 1 added, 1 changed, 1 destroyed.

Outputs:
<redacted>

```

### DIFF
--- a/hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
+++ b/hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
@@ -2,6 +2,7 @@
 
 import typing as t
 
+from botocore.exceptions import ClientError
 from threatexchange.content_type.photo import PhotoContent
 from threatexchange.content_type.video import VideoContent
 from threatexchange.signal_type.md5 import VideoMD5Signal
@@ -13,40 +14,62 @@ from hmalib.common.mappings import (
     ToggleableSignalTypeConfig,
     full_class_name,
 )
-
+from hmalib.common.logging import get_logger
 from hmalib.scripts.migrations.migrations_base import MigrationBase
+
+logger = get_logger(__name__)
 
 
 class _Migration(MigrationBase):
     def do_migrate(self):
-        create_config(
-            ToggleableContentTypeConfig(
-                name=ToggleableContentTypeConfig.get_name_from_type(VideoContent),
-                content_type_class=full_class_name(VideoContent),
-                enabled=True,
+        """
+        if any of the 4 default mappings do not exist, create them
+        """
+        try:
+            create_config(
+                ToggleableContentTypeConfig(
+                    name=ToggleableContentTypeConfig.get_name_from_type(VideoContent),
+                    content_type_class=full_class_name(VideoContent),
+                    enabled=True,
+                )
             )
-        )
-
-        create_config(
-            ToggleableContentTypeConfig(
-                name=ToggleableContentTypeConfig.get_name_from_type(PhotoContent),
-                content_type_class=full_class_name(PhotoContent),
-                enabled=True,
+        except ClientError:
+            logger.warning(
+                "Attempted to add ToggleableContentTypeConfig for VideoContent, but it already exists."
             )
-        )
-
-        create_config(
-            ToggleableSignalTypeConfig(
-                name=ToggleableSignalTypeConfig.get_name_from_type(VideoMD5Signal),
-                signal_type_class=full_class_name(VideoMD5Signal),
-                enabled=True,
+        try:
+            create_config(
+                ToggleableContentTypeConfig(
+                    name=ToggleableContentTypeConfig.get_name_from_type(PhotoContent),
+                    content_type_class=full_class_name(PhotoContent),
+                    enabled=True,
+                )
             )
-        )
-
-        create_config(
-            ToggleableSignalTypeConfig(
-                name=ToggleableSignalTypeConfig.get_name_from_type(PdqSignal),
-                signal_type_class=full_class_name(PdqSignal),
-                enabled=True,
+        except ClientError:
+            logger.warning(
+                "Attempted to add ToggleableContentTypeConfig for PhotoContent, but it already exists."
             )
-        )
+        try:
+            create_config(
+                ToggleableSignalTypeConfig(
+                    name=ToggleableSignalTypeConfig.get_name_from_type(VideoMD5Signal),
+                    signal_type_class=full_class_name(VideoMD5Signal),
+                    enabled=True,
+                )
+            )
+        except ClientError:
+            logger.warning(
+                "Attempted to add ToggleableSignalTypeConfig for VideoMD5Signal, but it already exists."
+            )
+        try:
+            create_config(
+                ToggleableSignalTypeConfig(
+                    name=ToggleableSignalTypeConfig.get_name_from_type(PdqSignal),
+                    signal_type_class=full_class_name(PdqSignal),
+                    enabled=True,
+                )
+            )
+        except ClientError:
+            logger.warning(
+                "Attempted to add ToggleableSignalTypeConfig for PdqSignal, but it already exists."
+            )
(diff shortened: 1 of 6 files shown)


hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
```diff
@@ -13,40 +14,62 @@
...
     full_class_name,
 )
-
+from hmalib.common.logging import get_logger
 from hmalib.scripts.migrations.migrations_base import MigrationBase
 
+logger = get_logger(__name__)
+
 
 class _Migration(MigrationBase):
```
Dcallies, line 23: Should we take a step to rename all these migration things to try and separate "Setup" from "Migration for compatibility?"

I'm especially worried about folks getting state re-applied to them after they explicitly opt out. 

If this is already handled in some way, can we separate these two in the codebase by calling one InitialSetup and another MigrateOldDeployments or somesuch?

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
```diff
@@ -13,40 +14,62 @@
...
     def do_migrate(self):
-        create_config(
-            ToggleableContentTypeConfig(
-                name=ToggleableContentTypeConfig.get_name_from_type(VideoContent),
-                content_type_class=full_class_name(VideoContent),
-                enabled=True,
+        """
+        if any of the 4 default mappings do not exist, create them
```
Dcallies, line 26: This is slightly sparse, what do you think of:

```
Loads default core interfaces and extensions:
1. ContentTypes (Photo, Video)
2. SignalTypes (PDQ, vMD5)
```

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
```diff
@@ -13,40 +14,62 @@
...
-        )
-
-        create_config(
-            ToggleableContentTypeConfig(
-                name=ToggleableContentTypeConfig.get_name_from_type(PhotoContent),
-                content_type_class=full_class_name(PhotoContent),
-                enabled=True,
+        except ClientError:
+            logger.warning(
+                "Attempted to add ToggleableContentTypeConfig for VideoContent, but it already exists."
```
Dcallies, line 38: blocking: We do not check the error code of this to see if this is actually the case, and so could accidentally eat some serious error. Is there something in the exception we can get?

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
```diff
@@ -13,40 +14,62 @@
...
+                )
             )
-        )
-
-        create_config(
-            ToggleableSignalTypeConfig(
-                name=ToggleableSignalTypeConfig.get_name_from_type(PdqSignal),
-                signal_type_class=full_class_name(PdqSignal),
-                enabled=True,
+        except ClientError:
```
Dcallies, line 48: (and for all the others as well)

hasher-matcher-actioner/terraform/migrations/README.md
```diff
@@ -1,8 +1,8 @@
 # What are migrations
 
-Migrations are commands to run when installing a new version of HMA. This includes both a fresh install or an upgrade of an older version. 
+Migrations are commands to run when installing a new version of HMA.
```
Dcallies, line 3: blocking q: Wait, is the note about being a fresh install vs upgrade not the case anymore?

hasher-matcher-actioner/terraform/migrations/README.md
```diff
@@ -1,8 +1,8 @@
 # What are migrations
 
-Migrations are commands to run when installing a new version of HMA. This includes both a fresh install or an upgrade of an older version. 
+Migrations are commands to run when installing a new version of HMA. 
 
-In both cases, there may be things that can't be done from within terraform, like creating some database entries, or running an hmalib.cli command, etc.
+There may be things that can't be done from within terraform, like creating some database entries, or running an hmalib.cli command, etc.
```
Dcallies, line 5: There's a missing connecting statement here somewhere like "We need to use scripts because there are certain actions..."

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
BarrettOlson: It is handled another way, if they use terraform it'll be added once but can be removed. If removed it will not be applied because the terraform data won't have changed.

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
BarrettOlson: I believe only kind of ClientError we can see is this one, I believe everything else should be different as it would be an AccessError or Network error instead. This is matching the existing checks used elsewhere in this flow: https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/hmalib/common/configs/tx_apis.py#L81

hasher-matcher-actioner/terraform/migrations/README.md
BarrettOlson: This README used to say it was for both, right now the README is about an empty module but if used it would be for upgrading older version.

hasher-matcher-actioner/terraform/migrations/README.md
BarrettOlson: yeah, I'll redraft it more.

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
BarrettOlson: As for the rename, we probably want to take the step eventually however I didn't want to make that drastic a rewrite without a lot more testing and I wanted to un-break the regular deploy first anyway. I am happy to create or have you create and issue to follow through with a rewrite.

hasher-matcher-actioner/hmalib/scripts/migrations/2022_04_02_default_content_signal_type_configs.py
Dcallies: The existence of other checks with this same problem is not reassuring to me!

Where do ClientErrors come from? Can we add codes to them?

Dcallies (approved):
Discussed offline, I think we need to come back to this and at least rename. I think the current state of the name (Migration), READMEs, and code make it hard for me to predict what the behavior of this functionality is.

Accepting to unbork the clean deployment.

PR #1284, opened 2023-03-27
fix typos in SignalExchangeAPI class

NikolayOG:
While reading through the docs in SignalExchangeAPI found some typos.


python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -107,7 +107,7 @@ def for_collab(cls, collab: TCollabConfig) -> "SignalExchangeAPI":
         2. All methods are callable if supported by the base implementation
 
         For implementations that do need additional configuration or
-        authentification, @see threatexchange.exchanges.auth
```
BarrettOlson, line 110: TIL: "authentification" not a real word.

PR #1285, opened 2023-03-27
Update devcontainer config: rebase on newer upstream and simplify

dougneal:
Summary
---------

The existing devcontainer config was a couple of years old and presumably put together before there was a working off-the-shelf container available. I've mostly just copied and pasted from [here](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/javascript-node/.devcontainer). It seems to work pretty nicely.

Note the post install command, `pip install --editable .`, this results in the CLI command `threatexchange` being served by the in-repo code rather than the installed pip.

Test Plan
---------

* Rebuild devcontainer from VSCode
* Make trivial edit to `main.py` to alter the help text in some recognisable way
* Run `threatexchange` from the container CLI and look for help text modification


python-threatexchange/.devcontainer/Dockerfile
```diff
@@ -1,60 +1,5 @@
...
-  && ZSH_SNIPPET="HISTFILE=/commandhistory/.zsh_history" \
-  && mkdir -p /commandhistory \
-  && touch /commandhistory/.bash_history \
-  && touch /commandhistory/.zsh_history \
-  && chown -R $unixname /commandhistory \
-  && echo $BASH_SNIPPET >> "/home/$unixname/.bashrc" \
-  && echo $ZSH_SNIPPET >> "/home/$unixname/.zshrc" \
-  && echo $ZSH_SNIPPET >> "/home/$unixname/.profile"
-# Also appends ZSH_SNIPPET to ~/.profile in case users want to overwrite zshrc file
-# e.g. https://code.visualstudio.com/docs/remote/containers#_personalizing-with-dotfile-repositories
```
BarrettOlson, line 54: This is something I do because I prefer `zsh` I don't feel too strongly about it but one thing that I would love to be able to keep around in the command history between stopping and starting the container. Do you know if there is a way can be preserved using this "off-the-shelf" option?

Dcallies:
Woah, awesome find!

python-threatexchange/.devcontainer/Dockerfile
dougneal: Checking!

BarrettOlson (approved):
LGTM. @Dcallies feel free to weigh in but I am in favor of shipping this as I have heard from both of you that the existing `python-threatexchange/.devcontainer/Dockerfile` is broken.

I mentioned this inline but if we can find a way to preserve sessions history I think that is valuable functionality for the dev container to have. However I am okay leaving this for a future PR.

dougneal:
Persistent command history fixed.

PR #1289, opened 2023-03-28
[py-tx] Align `tx match --hash` format with the output of `tx hash` 

### LINKED ISSUE #1188: inconsistency in hash format
`threatexchange hash photo <path-to-image-file>` outputs hashes in the form

`pdq f5b44e4f1673f4e44e57fd6d021201effd8b40d6b54906d3dc302df672160708`

however

`threatexchange match --hashes photo <path-to-hash-file>` expects a single file without a "pdq " prefix

this causes an inconsistency where hashes generated by `threatexchange hash` can't be directly used in `threatexchange match --hashes`

Sam-Freeman:
Summary
---------
Fixes #1188 
This pr allows for the usage of the hash format produced by the hash command to be used in the match command (i.e. prefixed with 'pdq'). In addition, allows for multiple hashes in a single file of varying signal types. 

Will throw exceptions when:
- You specify (-S) a signal type, and provide hashes of another type
- When there are multiple signal type prefixes with a mixture of no prefixes -- i.e. we are unable to 'infer' the signal type from the hash
- Provide an invalid signal type 
- Provide an invalid hash for the signal type

Test Plan
---------
Wrote unit tests and they all pass. Also ran a few manual tests, and piped the command from hash to match which works.
`tx hash photo <photo location> | tx match --hashes -S pdq photo -`

### DIFF
--- a/python-threatexchange/threatexchange/cli/match_cmd.py
+++ b/python-threatexchange/threatexchange/cli/match_cmd.py
@@ -173,13 +173,15 @@ class MatchCommand(command_base.Command):
         )
 
         if self.as_hashes:
-            hashes_grouped_by_prefix = dict()
+            hashes_grouped_by_prefix: t.Dict[
+                t.Optional[t.Type[SignalType]], t.Set[str]
+            ] = dict()
             # Infer the signal types from the prefixes (None is used as key for hashes with no prefix)
             for path in self.files:
                 _group_hashes_by_prefix(path, settings, hashes_grouped_by_prefix)
             # Validate the SignalType and append the None prefixes to the correct SignalType
             self.validate_hashes_signal_type(hashes_grouped_by_prefix, signal_types)
-            signal_types = list(hashes_grouped_by_prefix.keys())
+            signal_types = [key for key in hashes_grouped_by_prefix.keys() if key]
         indices: t.List[t.Tuple[t.Type[SignalType], SignalTypeIndex]] = []
         for s_type in signal_types:
             index = settings.index.load(s_type)
@@ -220,9 +222,9 @@ class MatchCommand(command_base.Command):
 
     def validate_hashes_signal_type(
         self,
-        hashes_grouped_by_prefix: t.Dict[t.Optional[SignalType], t.Set[str]],
+        hashes_grouped_by_prefix: t.Dict[t.Optional[t.Type[SignalType]], t.Set[str]],
         signal_types: t.List[t.Type[SignalType]],
-    ) -> bool:
+    ) -> None:
         if (
             len(hashes_grouped_by_prefix) > 2
             and None in hashes_grouped_by_prefix.keys()
@@ -257,7 +259,7 @@ class MatchCommand(command_base.Command):
             # Based on the validations, we know that there will only be one key here or one defined in settings
             hashes_grouped_by_prefix.clear()
             if not self.only_signal:
-                key = signal_types[0]
+                key: t.Optional[t.Type[SignalType]] = signal_types[0]
                 if len(keys) > 0:
                     key = keys[0]
             else:
@@ -277,7 +279,7 @@ def _match_file(
 def _group_hashes_by_prefix(
     path: pathlib.Path,
     settings: CLISettings,
-    hashes_grouped_by_prefix: t.Dict[t.Optional[SignalType], t.Set[str]],
+    hashes_grouped_by_prefix: t.Dict[t.Optional[t.Type[SignalType]], t.Set[str]],
 ) -> None:
     for line in path.read_text().splitlines():
         line = line.strip()
@@ -300,13 +302,15 @@ def _group_hashes_by_prefix(
                 )
             except Exception as e:
                 logging.exception(
-                    "%s failed verification on %s", signal_type.get_name(), hash
+                    "%s failed verification on %s",
+                    signal_type.get_name() if signal_type else "None!",
+                    hash,
                 )
                 hash_repr = repr(hash)
                 if len(hash_repr) > 50:
                     hash_repr = hash_repr[:47] + "..."
                 raise CommandError(
-                    f"{hash} from {path} is not a valid hash for {signal_type.get_name()}",
+                    f"{hash} from {path} is not a valid hash for {signal_type.get_name() if signal_type else 'None!'}",
                     2,
                 )
         else:
--- a/python-threatexchange/threatexchange/cli/match_cmd.py
+++ b/python-threatexchange/threatexchange/cli/match_cmd.py
@@ -173,15 +173,16 @@ class MatchCommand(command_base.Command):
         )
 
         if self.as_hashes:
-            hashes_grouped_by_prefix: t.Dict[
-                t.Optional[t.Type[SignalType]], t.Set[str]
-            ] = dict()
+            hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]] = {}
             # Infer the signal types from the prefixes (None is used as key for hashes with no prefix)
             for path in self.files:
                 _group_hashes_by_prefix(path, settings, hashes_grouped_by_prefix)
             # Validate the SignalType and append the None prefixes to the correct SignalType
-            self.validate_hashes_signal_type(hashes_grouped_by_prefix, signal_types)
-            signal_types = [key for key in hashes_grouped_by_prefix.keys() if key]
+            hashes_grouped_by_signal = self.validate(
+                settings, hashes_grouped_by_prefix, signal_types
+            )
+            signal_types = list(hashes_grouped_by_signal)
+
         indices: t.List[t.Tuple[t.Type[SignalType], SignalTypeIndex]] = []
         for s_type in signal_types:
             index = settings.index.load(s_type)
@@ -199,13 +200,13 @@ class MatchCommand(command_base.Command):
                 seen = set()  # TODO - maybe take the highest certainty?
                 if self.as_hashes:
                     results = _match_hashes(
-                        hashes_grouped_by_prefix[s_type], s_type, index
+                        hashes_grouped_by_signal[s_type], s_type, index
                     )
                 else:
                     results = _match_file(path, s_type, index)
 
                 for r in results:
-                    # TODO Improve visualisation of a single multiple hash query
+                    # TODO Improve output of a single multiple hash query
                     metadatas: t.List[t.Tuple[str, FetchedSignalMetadata]] = r.metadata
                     for collab, fetched_data in metadatas:
                         if not self.all and collab in seen:
@@ -220,51 +221,89 @@ class MatchCommand(command_base.Command):
                             fetched_data,
                         )
 
-    def validate_hashes_signal_type(
+    def validate(
         self,
-        hashes_grouped_by_prefix: t.Dict[t.Optional[t.Type[SignalType]], t.Set[str]],
+        settings: CLISettings,
+        hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]],
         signal_types: t.List[t.Type[SignalType]],
-    ) -> None:
-        if (
-            len(hashes_grouped_by_prefix) > 2
-            and None in hashes_grouped_by_prefix.keys()
-        ):
-            raise CommandError(
-                f"Error: Provided more than one SignalType and some hashes are missing a prefix",
-                2,
-            )
-        if self.only_signal:
-            if (
-                self.only_signal not in hashes_grouped_by_prefix.keys()
-                and None not in hashes_grouped_by_prefix.keys()
-            ):
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        """
+        Takes the hashes grouped by optional string prefix, performs all required validations.
+        Command line arguments, SignalType, hashes, and ambiguous SignalTypes are validated.
+        Returns a dict of validated SignalType, and validated hashes of each SignalType.
+        """
+
+        # There is a signal without a prefix and it's ambiguous (fix: -S)
+        if None in hashes_grouped_by_prefix:
+            if len(hashes_grouped_by_prefix) > 2 or len(signal_types) > 1:
+                # Ambiguous -- throw
                 raise CommandError(
-                    f"Error: SignalType '{self.only_signal} was provided, but inferred more from provided hashes."
-                    f"Inferred signal types: {', '.join(s_type.get_name() for s_type in hashes_grouped_by_prefix.keys() if s_type)}"
+                    f"Error: The SignalType is ambiguous for '{self.content_type.get_name()}'"
+                    "For '--hashes' please use '-S' or '--only-signal' to specify one of"
+                    f"{[s.get_name() for s in signal_types]}",
+                    2,
                 )
-        if (
-            len(signal_types) > 1
-            and len(hashes_grouped_by_prefix) == 1
-            and None in hashes_grouped_by_prefix.keys()
+            # unambiguous -- set
+            hashes_grouped_by_prefix.setdefault(
+                signal_types[0].get_name(), set()
+            ).update(hashes_grouped_by_prefix.pop(None))
+
+        # Validate that the SignalTypes and hashes are valid (fix: remove or update hash)
+        hashes_grouped_by_signal = self.validate_signal_type_and_hash(
+            settings, hashes_grouped_by_prefix
+        )
+
+        # There is a signal not valid for your arguments (fix: remove or change args)
+        if self.only_signal and (
+            self.only_signal not in hashes_grouped_by_signal
+            or len(hashes_grouped_by_signal) > 1
         ):
             raise CommandError(
-                f"Error: '{self.content_type.get_name()}' supports more than one SignalType"
-                "No prefix applied to the hashes, cannot infer correct SignalType"
+                f"Error: SignalType '{self.only_signal} was specified, but attempting to query with additional/different SignalTypes"
+                f"Please remove or change your arguments, or remove the additional SignalTypes"
             )
-        # As well as the above validations, also need to combine the None prefixes into the correct SignalType
-        if None in hashes_grouped_by_prefix.keys():
-            values = set().union(*hashes_grouped_by_prefix.values())
-            keys = list(hashes_grouped_by_prefix.keys())
-            keys.remove(None)
-            # Based on the validations, we know that there will only be one key here or one defined in settings
-            hashes_grouped_by_prefix.clear()
-            if not self.only_signal:
-                key: t.Optional[t.Type[SignalType]] = signal_types[0]
-                if len(keys) > 0:
-                    key = keys[0]
-            else:
-                key = self.only_signal
-            hashes_grouped_by_prefix[key] = values
+
+        return hashes_grouped_by_signal
+
+    def validate_signal_type_and_hash(
+        self,
+        settings: CLISettings,
+        hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]],
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        hashes_grouped_by_signal: t.Dict[t.Type[SignalType], t.Set[str]] = {}
+        for possible_signal, hashes in hashes_grouped_by_prefix.items():
+            if not possible_signal:
+                # This shouldn't be hit as we filter out the None key before
+                continue
+            # Validate the signal
+            try:
+                signal_type = settings.get_signal_type(possible_signal)
+            except:
+                logging.exception("Signal type %s is invalid.", possible_signal)
+                raise CommandError(
+                    f"Error: '{possible_signal}' is not a valid Signal Type."
+                    "Please remove or update this hash from your query",
+                    2,
+                )
+            # Validate the hashes
+            for hash in hashes:
+                try:
+                    hash = signal_type.validate_signal_str(hash.strip())
+                except:
+                    logging.exception(
+                        "%s failed verification on %s",
+                        signal_type.get_name() if signal_type else "None!",
+                        hash,
+                    )
+                    hash_repr = repr(hash)
+                    if len(hash_repr) > 50:
+                        hash_repr = hash_repr[:47] + "..."
+                    raise CommandError(
+                        f"{hash_repr} is not a valid hash for {signal_type.get_name() if signal_type else 'None!'}",
+                        2,
+                    )
+            hashes_grouped_by_signal[signal_type] = hashes
+        return hashes_grouped_by_signal
 
 
 def _match_file(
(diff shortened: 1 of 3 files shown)


python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -196,11 +198,14 @@ def execute(self, settings: CLISettings) -> None:
...
                 if self.as_hashes:
-                    results = _match_hashes(path, s_type, index)
+                    results = _match_hashes(
+                        hashes_grouped_by_prefix[s_type], s_type, index
+                    )
                 else:
                     results = _match_file(path, s_type, index)
 
                 for r in results:
+                    # TODO Improve visualisation of a single multiple hash query
```
Dcallies, line 208: nit, ignorable: 'output' is probably more accurate than visualization

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -167,19 +167,21 @@ def execute(self, settings: CLISettings) -> None:
...
-
         logging.info(
             "Signal types that apply: %s",
             ", ".join(s.get_name() for s in signal_types) or "None!",
         )
 
+        if self.as_hashes:
+            hashes_grouped_by_prefix: t.Dict[
+                t.Optional[t.Type[SignalType]], t.Set[str]
+            ] = dict()
```
Dcallies, line 178: nit: prefer using `{}` to `dict()`

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -167,19 +167,21 @@ def execute(self, settings: CLISettings) -> None:
...
+            # Infer the signal types from the prefixes (None is used as key for hashes with no prefix)
+            for path in self.files:
+                _group_hashes_by_prefix(path, settings, hashes_grouped_by_prefix)
+            # Validate the SignalType and append the None prefixes to the correct SignalType
+            self.validate_hashes_signal_type(hashes_grouped_by_prefix, signal_types)
+            signal_types = [key for key in hashes_grouped_by_prefix.keys() if key]
```
Dcallies, line 184: nit: The default is to iterate by keys, so this `.keys()` is not needed

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+        hashes_grouped_by_prefix: t.Dict[t.Optional[t.Type[SignalType]], t.Set[str]],
+        signal_types: t.List[t.Type[SignalType]],
+    ) -> None:
+        if (
+            len(hashes_grouped_by_prefix) > 2
+            and None in hashes_grouped_by_prefix.keys()
+        ):
+            raise CommandError(
+                f"Error: Provided more than one SignalType and some hashes are missing a prefix",
+                2,
```
Dcallies, line 234: Consider putting what the user should fix in these cases - it's clear that something is wrong, but what the user should do to fix it should be in the message.

For this one however, you might not be adding anything by this message - it's likely captured by one of the other messages below (add a prefix)

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+            )
+        if self.only_signal:
+            if (
+                self.only_signal not in hashes_grouped_by_prefix.keys()
+                and None not in hashes_grouped_by_prefix.keys()
+            ):
+                raise CommandError(
+                    f"Error: SignalType '{self.only_signal} was provided, but inferred more from provided hashes."
```
Dcallies, line 242: For this one, "inferred" might be hard

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+            len(hashes_grouped_by_prefix) > 2
+            and None in hashes_grouped_by_prefix.keys()
+        ):
+            raise CommandError(
+                f"Error: Provided more than one SignalType and some hashes are missing a prefix",
+                2,
+            )
+        if self.only_signal:
+            if (
+                self.only_signal not in hashes_grouped_by_prefix.keys()
```
Dcallies, line 238: nit: You query is on a dict, which by default `in` looks at keys with O(1), this .keys() will create an iterator that takes O(n) to evaluate

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+                and None not in hashes_grouped_by_prefix.keys()
+            ):
+                raise CommandError(
+                    f"Error: SignalType '{self.only_signal} was provided, but inferred more from provided hashes."
+                    f"Inferred signal types: {', '.join(s_type.get_name() for s_type in hashes_grouped_by_prefix.keys() if s_type)}"
+                )
+        if (
+            len(signal_types) > 1
+            and len(hashes_grouped_by_prefix) == 1
+            and None in hashes_grouped_by_prefix.keys()
```
Dcallies, line 248: nit: no .keys()

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+        ):
+            raise CommandError(
+                f"Error: '{self.content_type.get_name()}' supports more than one SignalType"
+                "No prefix applied to the hashes, cannot infer correct SignalType"
+            )
+        # As well as the above validations, also need to combine the None prefixes into the correct SignalType
+        if None in hashes_grouped_by_prefix.keys():
```
Dcallies, line 255: It might be easier to instead bring this case up to the top and handle it first. 

1. Check for None
     1. Check if none is ambiguous, and if so throw (fix: use -S)
     2. `hashes_grouped.setdefault(unambiguous_signal, []).extend(hashes_grouped.pop(None))`
2. Check if unhandled signals added (fix: remove unhandled signals)

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+            values = set().union(*hashes_grouped_by_prefix.values())
+            keys = list(hashes_grouped_by_prefix.keys())
+            keys.remove(None)
+            # Based on the validations, we know that there will only be one key here or one defined in settings
+            hashes_grouped_by_prefix.clear()
+            if not self.only_signal:
```
Dcallies, line 261: `signals` will only contain one signal if `self.only_signal` is true, so you don't have to refer to it again.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+                key: t.Optional[t.Type[SignalType]] = signal_types[0]
+                if len(keys) > 0:
+                    key = keys[0]
+            else:
+                key = self.only_signal
+            hashes_grouped_by_prefix[key] = values
```
Dcallies, line 267: What is the intention of this block?

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +220,52 @@ def execute(self, settings: CLISettings) -> None:
...
+            and len(hashes_grouped_by_prefix) == 1
+            and None in hashes_grouped_by_prefix.keys()
+        ):
+            raise CommandError(
+                f"Error: '{self.content_type.get_name()}' supports more than one SignalType"
+                "No prefix applied to the hashes, cannot infer correct SignalType"
+            )
+        # As well as the above validations, also need to combine the None prefixes into the correct SignalType
+        if None in hashes_grouped_by_prefix.keys():
+            values = set().union(*hashes_grouped_by_prefix.values())
```
Dcallies, line 256: blocking: Danger! This will merge potentially unrelated hashes together. There is a simpler way to merge this: 

```
if len(signal) != 1:
  throw Exception()
hashes_grouped.setdefault(signals[0], []).extend(hashes_grouped.pop(None))
```

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -225,23 +276,72 @@ def _match_file(
...
+            continue
+        components = line.split()
+        signal_type = None
+        if len(components) > 1:
+            # Assume it has a prefix
+            possible_type = components[0]
+            hash = components[1].strip()
+            try:
+                signal_type = settings.get_signal_type(possible_type)
+                hash = signal_type.validate_signal_str(hash)
```
Dcallies, line 296: blocking: I don't think we are guaranteed that validate_signal_str won't throw a key error for some reason (we just say "throws an exception" in the interface). Separate the two catch blocks to prevent a collision.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -167,19 +167,21 @@ def execute(self, settings: CLISettings) -> None:
...
             ", ".join(s.get_name() for s in signal_types) or "None!",
         )
 
+        if self.as_hashes:
+            hashes_grouped_by_prefix: t.Dict[
+                t.Optional[t.Type[SignalType]], t.Set[str]
+            ] = dict()
+            # Infer the signal types from the prefixes (None is used as key for hashes with no prefix)
+            for path in self.files:
+                _group_hashes_by_prefix(path, settings, hashes_grouped_by_prefix)
```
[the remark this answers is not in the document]
Dcallies: Nah, I think this is fine

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
                             fetched_data,
                         )
 
+    def validate(
```
Dcallies, line 224: nit: This is still only called with `--hashes`, not in all calls will use it

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+        self,
+        settings: CLISettings,
+        hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]],
+        signal_types: t.List[t.Type[SignalType]],
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        """
+        Takes the hashes grouped by optional string prefix, performs all required validations.
+        Command line arguments, SignalType, hashes, and ambiguous SignalTypes are validated.
+        Returns a dict of validated SignalType, and validated hashes of each SignalType.
```
Dcallies, line 233: Thanks for adding a docstring!

One detail (similar to above comment) is that this validates when the input is hashes

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+        )
+
+        # There is a signal not valid for your arguments (fix: remove or change args)
+        if self.only_signal and (
+            self.only_signal not in hashes_grouped_by_signal
+            or len(hashes_grouped_by_signal) > 1
+        ):
+            raise CommandError(
+                f"Error: SignalType '{self.only_signal} was specified, but attempting to query with additional/different SignalTypes"
+                f"Please remove or change your arguments, or remove the additional SignalTypes"
```
Dcallies, line 263: nit: f-string not needed here

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        """
+        Takes the hashes grouped by optional string prefix, performs all required validations.
+        Command line arguments, SignalType, hashes, and ambiguous SignalTypes are validated.
+        Returns a dict of validated SignalType, and validated hashes of each SignalType.
+        """
+
+        # There is a signal without a prefix and it's ambiguous (fix: -S)
+        if None in hashes_grouped_by_prefix:
+            if len(hashes_grouped_by_prefix) > 2 or len(signal_types) > 1:
```
Dcallies, line 238: ignorable: I think this is being overly specific about the input. This will throw if you do

```
cat 'pdq facefaceface...
facefaceface...' > file.txt
$ tx match -S pdq photo file.txt
```
but I think it's unlikely that you'll get a non-prefixed value here that will pass the signal type validation on L247

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+        # Validate that the SignalTypes and hashes are valid (fix: remove or update hash)
+        hashes_grouped_by_signal = self.validate_signal_type_and_hash(
+            settings, hashes_grouped_by_prefix
+        )
+
+        # There is a signal not valid for your arguments (fix: remove or change args)
+        if self.only_signal and (
+            self.only_signal not in hashes_grouped_by_signal
+            or len(hashes_grouped_by_signal) > 1
+        ):
```
Dcallies, line 260: You don't need this block if you pass in the expected signal types into 252 (rather than all settings)

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+        hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]],
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        hashes_grouped_by_signal: t.Dict[t.Type[SignalType], t.Set[str]] = {}
+        for possible_signal, hashes in hashes_grouped_by_prefix.items():
+            if not possible_signal:
+                # This shouldn't be hit as we filter out the None key before
+                continue
+            # Validate the signal
+            try:
+                signal_type = settings.get_signal_type(possible_signal)
```
Dcallies, line 280: The only thing you pass the settings god-object in for is the signal types. Why not pass in the signal types and use those? `{st.get_name(): st for st in signal_types}`

Another reason I don't like using the full set of signal types is that this will accept types that are not valid for the command (pdq for video for example).

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -167,19 +167,22 @@ def execute(self, settings: CLISettings) -> None:
...
+        if self.as_hashes:
+            hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]] = {}
+            # Infer the signal types from the prefixes (None is used as key for hashes with no prefix)
+            for path in self.files:
+                _group_hashes_by_prefix(path, settings, hashes_grouped_by_prefix)
+            # Validate the SignalType and append the None prefixes to the correct SignalType
+            hashes_grouped_by_signal = self.validate(
+                settings, hashes_grouped_by_prefix, signal_types
+            )
+            signal_types = list(hashes_grouped_by_signal)
```
Dcallies, line 184: blocking: danger! I think this will allow you jump the tracks of the signal type validation done already above. Unless this was intentional, since I actually can't think of strong benefits of narrowing it when the user clearly is passing you hashes and we enforce uniqueness on all the names. However, this will come as a surprise to the reader of the code.

```
cat pdq faceface... | tx match video -  # Surprising?
```

^ Is the above an intentional behavior change?

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -167,19 +167,22 @@ def execute(self, settings: CLISettings) -> None:
         if self.as_hashes:
             types = (BytesHasher, TextHasher, FileHasher)
         signal_types = [s for s in signal_types if issubclass(s, types)]
```
Dcallies, line 169: (In reference to below comments) This list of signal types have already been narrowed down the valid set given the input.

It might be that we should consider moving `--hashes` to a toplevel argument (tx match photo|video|hash)` and allowing cross matching might make more sense, but now we've exceeded the scope of the pr.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+    def validate_signal_type_and_hash(
+        self,
+        settings: CLISettings,
+        hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]],
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        hashes_grouped_by_signal: t.Dict[t.Type[SignalType], t.Set[str]] = {}
+        for possible_signal, hashes in hashes_grouped_by_prefix.items():
+            if not possible_signal:
+                # This shouldn't be hit as we filter out the None key before
+                continue
```
Dcallies, line 277: blocking: Convert this to `assert` if you are sure

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+                raise CommandError(
+                    f"Error: '{possible_signal}' is not a valid Signal Type."
+                    "Please remove or update this hash from your query",
+                    2,
+                )
+            # Validate the hashes
+            for hash in hashes:
+                try:
+                    hash = signal_type.validate_signal_str(hash.strip())
+                except:
```
Dcallies, line 292: blocking: This will catch CTRL+C interrupt and out of memory, which I don't think you intended. use `except Exception` to narrow down to the usual set.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+    ) -> t.Dict[t.Type[SignalType], t.Set[str]]:
+        hashes_grouped_by_signal: t.Dict[t.Type[SignalType], t.Set[str]] = {}
+        for possible_signal, hashes in hashes_grouped_by_prefix.items():
+            if not possible_signal:
+                # This shouldn't be hit as we filter out the None key before
+                continue
+            # Validate the signal
+            try:
+                signal_type = settings.get_signal_type(possible_signal)
+            except:
```
Dcallies, line 281: blocking: This will catch CTRL+C interrupt and out of memory, which I don't think you intended. use `except Exception` to narrow down to the usual set.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -215,6 +221,90 @@ def execute(self, settings: CLISettings) -> None:
...
+                logging.exception("Signal type %s is invalid.", possible_signal)
+                raise CommandError(
+                    f"Error: '{possible_signal}' is not a valid Signal Type."
+                    "Please remove or update this hash from your query",
+                    2,
+                )
+            # Validate the hashes
+            for hash in hashes:
+                try:
+                    hash = signal_type.validate_signal_str(hash.strip())
```
Dcallies, line 291: This may actually change the string for some types (i.e. for some binary types we accept octal, hex, etc). You probably want to save this back.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -225,23 +315,48 @@ def _match_file(
     return index.query(s_type.hash_from_file(path))
 
 
+def _group_hashes_by_prefix(
+    path: pathlib.Path,
+    settings: CLISettings,
```
Dcallies, line 320: nit: unused - also, I prefer to avoid passing this around because it is somewhat of a god object

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -225,23 +315,48 @@ def _match_file(
...
+    hashes_grouped_by_prefix: t.Dict[t.Optional[str], t.Set[str]],
+) -> None:
+    for line in path.read_text().splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        components = line.split()
```
Dcallies, line 327: I'm torn over what to do when len(component) > 2 - this must mean that there are spaces in the hash, which some future type could allow, but it means our naive parsing here will fail oddly.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -225,23 +315,48 @@ def _match_file(
...
+        signal_type = None
+        if len(components) == 2:
+            # Assume it has a prefix
+            signal_type, hash = components
+        else:
+            # Assume it doesn't have a prefix and is a raw hash
+            hash = components[0]
```
Dcallies, line 334: blocking: What about len(component) > 2?

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -225,23 +315,48 @@ def _match_file(
...
+        hashes = hashes_grouped_by_prefix.get(signal_type, set())
+        hashes.add(hash)
+        hashes_grouped_by_prefix[signal_type] = hashes
```
Dcallies, line 337: You can use setdefault again here to simplify (some folks also like defaultdict, but I have burned before!)

```
hashes_grouped_by_prefix.setdefault(signal_type, set()).add(hash)  # you are done
```
```

Dcallies (changes requested):
Almost there! Couple of blocking comments for bits I'm not sure about

python-threatexchange/threatexchange/cli/match_cmd.py
Sam-Freeman: Cool, will update -- should we also be validating the non `--hashes` calls as well? I can update either in this pr or a follow up one.

python-threatexchange/threatexchange/cli/match_cmd.py
Sam-Freeman: ohhh, this is a good catch. I see what you mean. Thanks!

python-threatexchange/threatexchange/cli/match_cmd.py
Sam-Freeman: Completely agree, and had a similar thought -- on re-evaluation I'm thinking for when len(component) > 1 it's probably better to assume that [0] is the prefix and concat [1:] into the hash. 

The only issue here is that if there is a future hash with a space, without a prefix the parsing will fail still. I'll have a think on this

Dcallies: I think your intuition is right - if there are hashes with a space, providing the prefix allows for a non-ambiguous parse. Check out `lpartition` or `split(limit=2)`

Dcallies:
Hey @Sam-Freeman, I'm going to close this PR for inactivity, feel free to re-open if you still want to try and get it merged!

PR #1290, opened 2023-03-28
Define 'tx' as additional CLI entry point

dougneal:
Summary
---------

To save those precious keystrokes

Test Plan
---------

`tx -h` - it works

### DIFF
--- a/python-threatexchange/setup.py
+++ b/python-threatexchange/setup.py
@@ -66,5 +66,10 @@ setup(
         "numpy==1.24.2",  # faiss
     ],
     extras_require=extras_require,
-    entry_points={"console_scripts": ["threatexchange = threatexchange.cli.main:main"]},
+    entry_points={
+        "console_scripts": [
+            "threatexchange = threatexchange.cli.main:main",
+            "tx = threatexchange.cli.main:main",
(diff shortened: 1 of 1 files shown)


Dcallies:
Only problem here is that there is another unix tool named `tx`: https://www.unix.com/man-page/linux/1/tx/

Do you have any familiarity with this guy? Should we just stomp it?

dougneal:
> Only problem here is that there is another unix tool named `tx`: https://www.unix.com/man-page/linux/1/tx/
> 
> Do you have any familiarity with this guy? Should we just stomp it?
> 
> We just tell folks to use `alias`

Oh. Well, from reading on what it actually is, the chances of it being installed alongside ThreatExchange are pretty low, I'd say. I don't feel too strongly though. Maybe `thrx` instead of `tx`?

BarrettOlson:
Personally I don't think, `thrx` is worth it. I am fine with leaving it was an alias suggestion but I also agree that the changes for both being installed alongside would be very rare.

Dcallies:
I say let's take `tx`, we can always walk it back to alias later.

Dcallies (approved):

PR #1291, opened 2023-03-28
Dockerize python-threatexchange

dougneal:
Summary
---------

Reduce user friction by providing a Docker container for the CLI. This eliminates the need for the user to have to care about Python.

To make for a clean Docker image layout, the persistent state directory is now configurable. While the default remains `~/.threatexchange`, it can now be overridden by setting a `TX_STATEDIR` environment variable.

Test Plan
---------

```
$ docker build --tag threatexchange .
$ docker run threatexchange --help
usage: threatexchange [-h] [--factory-reset] [--version] [--verbose]
                      {config,fetch,match,label,dataset,hash} ...

Wrapper for the `threatexchange` library, can serve as a simple e2e solution.
...
```

### DIFF
--- a/python-threatexchange/threatexchange/cli/main.py
+++ b/python-threatexchange/threatexchange/cli/main.py
@@ -273,7 +273,9 @@ def _setup_logging(level_str: str, *, initial: bool = False) -> None:
 
 def inner_main(
     args: t.Optional[t.Sequence[t.Text]] = None,
-    state_dir: pathlib.Path = pathlib.Path("~/.threatexchange"),
+    state_dir: pathlib.Path = pathlib.Path(
+        os.getenv("TX_STATEDIR", "~/.threatexchange")
+    ),
 ) -> None:
     """The main called by tests"""
     config = CliState(
--- a/python-threatexchange/Dockerfile
+++ b/python-threatexchange/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.11-bullseye
+
+WORKDIR /usr/src/threatexchange
+COPY . .
+RUN pip install .
+RUN rm -rf /usr/src/threatexchange
+
+ENV TX_STATEDIR=/var/lib/threatexchange
+
+VOLUME ["/var/lib/threatexchange"]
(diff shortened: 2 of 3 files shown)


BarrettOlson (commented):
Awesome! here on in another PR I think it is worth updated the README.md with details/expected usage

dougneal:
The README changes got stuck on my machine. Fixed

BarrettOlson (approved):

python-threatexchange/threatexchange/cli/main.py
```diff
@@ -273,9 +273,9 @@ def _setup_logging(level_str: str, *, initial: bool = False) -> None:
 
 def inner_main(
     args: t.Optional[t.Sequence[t.Text]] = None,
-    state_dir: pathlib.Path = pathlib.Path("~/.threatexchange"),
```
Dcallies, line 276: Why this change?

Dcallies (commented):
I don't use a lot of docker - what does this allow that users might like?

python-threatexchange/threatexchange/cli/main.py
dougneal: This was a bad move. I thought it would be cleaner to move into the method body because it appeared unused. I hadn't considered the tests.

dougneal:
> I don't use a lot of docker - what does this allow that users might like?

The value to the user is that they don't have to be Python developers to run it, or even care about their Python install at all. Just `docker build` followed by `docker run` and you have a functional CLI. The value will be even greater when you can spin up the API in the same manner.

If we were to publish the Docker image to the Docker hub or another container registry, they wouldn't even need the build step, just `docker run docker.example.com/threatexchange`.

python-threatexchange/threatexchange/cli/main.py
BarrettOlson: This seems reasonable, looks like the format/lint `black` is complaining though

Dcallies (approved):
So close! Just need to run black!

PR #1294, opened 2023-03-28
[pytx] CLI separation: move config and state classes out of cli module

dougneal:
Summary
---------

The library's core functionality and the CLI are separate concerns, so code which doesn't concern the CLI, shouldn't have the appearance of doing so. Move and rename the config and state management classes to reflect this. This will be fundamental to making the API parts make sense.

Test Plan
---------

See what GH CI says...

### DIFF
--- a/python-threatexchange/threatexchange/state.py
+++ b/python-threatexchange/threatexchange/state.py
@@ -17,7 +17,6 @@ import logging
 
 from threatexchange.signal_type.index import SignalTypeIndex
 from threatexchange.signal_type.signal_base import SignalType
-from threatexchange.cli.exceptions import CommandError
 from threatexchange.exchanges.collab_config import CollaborationConfigBase
 from threatexchange.exchanges.fetch_state import (
     FetchDelta,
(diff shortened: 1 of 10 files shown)


python-threatexchange/threatexchange/state.py
```diff
@@ -148,12 +153,12 @@ def _read_state(
...
             return delta
-        except Exception:
+        except Exception as exc:
             logging.exception("Failed to read state for %s", collab_name)
-            raise CommandError(
+            raise TXStateException(
                 f"Failed to read state for {collab_name}. "
                 "You might have to delete it with `threatexchange fetch --clear`"
-            )
+            ) from exc
```
Dcallies, line 161: I think this is captured by default in modern python, and might explicitly hint you don't want the full stack printed (or maybe that's `from None`

python-threatexchange/threatexchange/state.py
```diff
@@ -32,9 +31,15 @@
 from threatexchange.signal_type import index
 
 
-class CliIndexStore:
+class TXStateException(Exception):
```
Dcallies, line 34: This is going to collide with Nikolay's changes if he has any in flight,

python-threatexchange/threatexchange/state.py
dougneal: Oh ok, I was following a linter recommendation but it may have been an out of date rule

dougneal:
> I think it might be better at this exact time to just import the CLI libraries as they are named - this is something that is easy for us to do after the hackathon, but may be disruptive during the hackathon.
> 
> That, plus the other considerations on whether we want to promote these interfaces fully make this not clear cut to me.
> 
> Is there anything being blocked by these changes that requires them?

No prob. I'll close this.

PR #1297, opened 2023-03-29
[RFC] Introducing DBM for storage

NikolayOG:
### Summary
Draft commit, please focus on `DMBFetchedStateStore` class. 

### Test Plan

1. Delete the pdq index
2. Run

```
threatexchange match photo https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/b.jpg?raw=true
No data to match against
```

3. Run `threatexchange fetch`
5. Run `threatexchange match photo https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/b.jpg?raw=true`
pdq 16 (Sample Signals) INVESTIGATION_SEED

Now we are finding matches


python-threatexchange/threatexchange/cli/cli_config.py
```diff
@@ -198,13 +201,16 @@ def empty(self) -> bool:
             collab for collab in collabs if self.get_for_collab(collab).exists(collab)
         )
 
-    def get_for_api(self, api: t.Type[SignalExchangeAPI]) -> CliSimpleState:
-        return CliSimpleState(api, self._parent._state.dir_for_fetched_state(api))
+    def get_for_api(self, api: t.Type[SignalExchangeAPI], storage=None) -> fetch_state.FetchedStateStoreBase:
+        print("PASSED STORAGE??", storage)
+        # if storage == "dbm":
+        return DMBFetchedStateStore(api)
```
NikolayOG, line 207: Ignore this for now, 

Next steps would be to propagate storage flag from the cli terminal and select the correct storage interface class

python-threatexchange/threatexchange/exchanges/helpers.py
```diff
@@ -145,3 +148,91 @@ def get_for_signal_type(
                 if by_signal:
                     ret[collab.name] = by_signal
         return ret
+
+    def exists(self, collab: CollaborationConfigBase) -> bool:
+        raise NotImplementedError
+
+
+class DMBFetchedStateStore(fetch_state.FetchedStateStoreBase):
```
NikolayOG, line 156: Please focus on this class, thanks!

Dcallies: DMB or DBM?

python-threatexchange/threatexchange/exchanges/helpers.py
```diff
@@ -145,3 +148,91 @@ def get_for_signal_type(
...
+        """
+
+        ret = {}
+        for collab in collabs:
+            with dbm.open(collab.name, 'c') as db:
+                k = db.firstkey()
+                data = {}
+                while k:
+                    print(k, db[k])
+                    _k, _v = map(pickle.loads, [k, db[k]])
```
Dcallies, line 185: nit: underscores usually denote that the variable is unused (not that it's private as in other language conventions), so `k`, `v` is probably preferred.

python-threatexchange/threatexchange/exchanges/helpers.py
```diff
@@ -145,3 +148,91 @@ def get_for_signal_type(
...
+
+    def exists(self, collab: CollaborationConfigBase) -> bool:
+        raise NotImplementedError
+
+
+class DMBFetchedStateStore(fetch_state.FetchedStateStoreBase):
+    def __init__(
+        self,
+        api_cls: TSignalExchangeAPICls,
+    ) -> None:
```
Dcallies, line 160: All of these methods end up taking a collaboration implicitly, and we could probably provide a wrapper that could cover the equivalent of get_for_signal_type.

Dcallies (commented):
Awesome work!

Thanks for filling out a test plan.

Based on your work here, I think we can actually simplify the base interface to make it easier to implement as an extension. Catch me at the hackathon and we can talk through it!

python-threatexchange/threatexchange/exchanges/storage.py
```diff
@@ -0,0 +1,81 @@
...
+T = t.TypeVar('T')
+
+class GenericStorage(ABC):
+
+    @abstractmethod
+    def connect(self, *args, **kwargs):
+        raise NotImplementedError
+    
+    @abstractmethod
+    def exists(self, *args, **kwargs):
```
Dcallies, line 17: Having *args and **kwargs is probably overkill.

python-threatexchange/threatexchange/exchanges/storage.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import TypeVar 
+import typing as t
+import pickle
+
+T = t.TypeVar('T')
+
+class GenericStorage(ABC):
+
+    @abstractmethod
+    def connect(self, *args, **kwargs):
```
Dcallies, line 13: If we require that a constructing an instance of this object make it functional, do we need this method?

Dcallies (changes requested):
Awesome progress! I think this is starting to narrow in on what I think will solve the GIFCT's issues and give us some flexibility in future designs.

Toplevel thoughts:
1. Let's continue iterating on the GenericStorage interface - this feel like it will play out in the long term more than specializing for the whole
2. Can you add some docstrings to the GenericStorage index? 
3. (Doesn't yet need to be solved / future PR) Consider also how it might be possible to port the old pickle + dict storage forward. We make a compatibility guarantee for users of the CLI

Dcallies:
Hey @NikolayOG, I'm going to close this PR for inactivity, feel free to re-open if you still want to try and get it merged!

PR #1302, opened 2023-03-30
[RFC][HMA-in-a-bottle] Possible example of creating a modular matcher

Sam-Freeman:
Summary
---------

An example of decoupling the matcher logic from the CLI into its own module and generic base class which allows for creation of specific matcher 'types' -- this was the simplest way to separate the logic for file matching and raw hash matching. 

Test Plan
---------

This is an RFC, but ran it locally and it all works.

### DIFF
--- a/python-threatexchange/threatexchange/cli/match_cmd.py
+++ b/python-threatexchange/threatexchange/cli/match_cmd.py
@@ -16,7 +16,7 @@ from threatexchange.cli.fetch_cmd import FetchCommand
 from threatexchange.cli.helpers import FlexFilesInputAction
 from threatexchange.exchanges.fetch_state import FetchedSignalMetadata
 
-from threatexchange.signal_type.index import IndexMatch, SignalTypeIndex
+from threatexchange.signal_type.index import IndexMatch
 from threatexchange.cli.exceptions import CommandError
 from threatexchange.signal_type.signal_base import BytesHasher, SignalType
 from threatexchange.cli.cli_config import CLISettings
@@ -24,6 +24,7 @@ from threatexchange.content_type.content_base import ContentType
 
 from threatexchange.signal_type.signal_base import MatchesStr, TextHasher, FileHasher
 from threatexchange.cli import command_base
+from threatexchange.matcher import matcher, file_matcher, hash_matcher
 
 
 TMatcher = t.Callable[[pathlib.Path], t.List[IndexMatch]]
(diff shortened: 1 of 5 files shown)


python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -145,104 +146,41 @@ def __init__(
...
+            for r in results:
+                metadatas: t.List[t.Tuple[str, FetchedSignalMetadata]] = r.metadata
+                for collab, fetched_data in metadatas:
+                    if not self.all and collab in seen:
+                        continue
+                    seen.add(collab)
+                    # Supposed to be without whitespace, but let's make sure
+                    distance_str = "".join(r.similarity_info.pretty_str().split())
+                    print(
+                        # s_type.get_name(),
```
BarrettOlson, line 182: hmm, knowing the type matched is valuable. One options though I'm not sure it is worth the overhead is to have it as part of the metadata(s)

Sam-Freeman: Yeah, agreed, and something that we/I need to come back to. I commented it out for now as it didn't exist in this scope, and didn't want to forget that I removed it.

python-threatexchange/threatexchange/matcher/matcher.py
```diff
@@ -0,0 +1,54 @@
...
+from threatexchange.content_type.content_base import ContentType
+from threatexchange.signal_type.index import IndexMatch, SignalTypeIndex
+from threatexchange.signal_type.signal_base import SignalType
+
+T = t.TypeVar("T")
+
+
+class Matcher(t.Generic[T]):
+    # question (sa) -- how do we want to handle the settings? In the CLI it's considered
+    # a god object -- do we want to pass it to the init of the class?
```
dougneal, line 13: Maybe a singleton?

python-threatexchange/threatexchange/matcher/matcher.py
```diff
@@ -0,0 +1,54 @@
...
+            for i in input:
+                results.extend(self._match_impl(i, s_type, index))
+        return results
+
+    def _match_impl(
+        self, i: T, s_type: t.Type[SignalType], index: SignalTypeIndex
+    ) -> t.Sequence[IndexMatch]:
+        raise NotImplementedError
+
+    def _get_indicies(self) -> t.List[t.Tuple[t.Type[SignalType], SignalTypeIndex]]:
```
dougneal, line 46: s/indicies/indices/

dougneal:
Nice, this is pretty much what I had in mind also.

PR #1303, opened 2023-03-30
Modularise the Hasher code from the hasher CLI 

Sam-Freeman:
Summary
---------

Pretty much as title, this is quite simple. It's the first iteration, and likely not the final form. 

Test Plan
---------

Ran locally, produces same result. All tests pass


Sam-Freeman:
Note: some of these were just merged in #1302 but are still showing up here.

python-threatexchange/threatexchange/hasher/hasher.py
```diff
@@ -0,0 +1,36 @@
...
+from threatexchange.cli.cli_config import CLISettings
+
+from threatexchange.content_type.content_base import ContentType
+from threatexchange.signal_type.signal_base import FileHasher, SignalType
+
+
+class Hasher:
+    def __init__(
+        self,
+        settings: CLISettings,
```
Dcallies, line 12: It looks like the only thing you pass in the settings type for is the list of signal types. What if you passed in the full list of signal types? 

Alternatively, you only use content type to narrow down to the applicable signal types. What if you only passed in those?

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -88,21 +89,7 @@ def __init__(
...
-                    f"does not apply to {self.content_type.get_name()}"
-                )
-            hashers = [self.signal_type]  # type: ignore  # can't detect intersection types
-
-        for file in self.files:
-            for hasher in hashers:
-                hash_str = hasher.hash_from_file(file)
-                if hash_str:
-                    print(hasher.get_name(), hash_str)
+        hasher = Hasher(settings, self.content_type, self.signal_type)
```
Dcallies, line 92: blocking: Unfortunately, anything that is being passed in the settings god object is still heavily tied in with the CLI.

Dcallies (changes requested):
I like the general thrust of this, and I think we are unequivocally going to do something very similar to this in pulling out components for HMAiaB.

However, I'm not convinced at some of the top level choices on the interface yet:
1. I don't think passing in the settings object is needed for this functionality, and may be detrimental. 
2. Stepping back, what does a Hasher object represent? What would be the advantage over a functional version?
3. Which things might be details of the CLI, and which things might be generalized?
4. Looking at SignalType interface, we support 3 versions: bytes hasher, str hasher, and then a superclass - file hasher. How should that map to the hasher instance/helper?
5. Which version will HMAiaB need, assuming that it will be sent original content?
6. Is there anything we can learn from UnifiedHasher in AWS HMA? https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/hmalib/hashing/unified_hasher.py

Dcallies:
Hey @Sam-Freeman , I'm going to close this PR for inactivity, feel free to re-open if you still want to try and get it merged!

Issue #1301, opened 2023-03-29
/matches/for-hash/ gives AttributeError: 'IndexMatchUntyped' object has no attribute 'distance'


kyuilgo:
This happens in the file hmalib/lambdas/api/matches.py 
at line 533 -  match_distance=int(match.distance)

Dcallies:
Thanks again for the report! This looks like another bug from the SignalType refactor :/

Issue #1299, opened 2023-03-29
[hma] Fetcher policy fails to access index


Dcallies:
An error occurred (AccessDeniedException) when calling the Query operation: User: arn:aws:sts::####:assumed-role/xxx-ts_fetche####/xxx-ts_fetcher is not authorized to perform: dynamodb:Query on resource: arn:aws:dynamodb:us-east-1:####:table/xxx-ts-HMABanks/index/BankMemberIdIndex because no identity-based policy allows the dynamodb:Query action

PR #1304, opened 2023-03-30
[hma] address missing distance field on `/for-hash/` result objects 

BarrettOlson:
Summary
---------

Had to resolve a set of errors to get to the `for-hash` api issue (would normally be different PR but the first two are 1 liners)

1. update python-threatexchange version because of fetch error handling
2. fetcher IAM policy updated to enable to query of bank member to sub parts of  banks_datastore. 
3. Patch signal type check of field check on `distance` to remove error for `for-hash` API
Test Plan
---------

Confirmed that with changes fetcher can get values from API for sample set and build index and with the python code change that a valid matches data will return.

### DIFF
--- a/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
+++ b/hasher-matcher-actioner/hmalib/lambdas/api/matches.py
@@ -15,6 +15,7 @@ from mypy_boto3_sqs.client import SQSClient
 from threatexchange.exchanges.clients.fb_threatexchange.descriptor import (
     ThreatDescriptor,
 )
+from threatexchange.signal_type.index import IndexMatch
 from threatexchange.signal_type.signal_base import SignalType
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/hmalib/lambdas/api/matches.py
```diff
@@ -539,6 +540,13 @@ def _matches_for_hash(
 
         return match_objects
 
+    def _distance_from_match_if_exits(match: IndexMatch) -> int:
```
Dcallies, line 543: blocking: Add a docstring explaining the compatibility issue and whether you think we should move to off-int distance at some future point. Feel free to shorthand to an instance.

Dcallies (approved):
Thanks again for all your hard work on these issues!

BarrettOlson:
This should address #1301 and #1299, I will wait for @kyuilgo to validate

PR #1320, opened 2023-06-27
[vpdq] Fix resource leak during hashing

ianwal:
### Files are leaked because they're never closed

The video files are being read using `popen()`, but they are not closed before going out of scope. Unlike fstream, which automatically releases resources, `popen()` requires a corresponding `pclose()`.

I fixed this by calling `pclose()` before the opened file goes out of scope in order to release the files.

---

Additionally, `avformat_open_input()` and `avformat_alloc_context()` require releasing after allocation.

`avformat_open_input()` is released by `avformat_close_input()`, which will also call `avformat_free_context()`. If `avformat_open_input()` fails, the context will be automatically freed during the failure so nothing more has to be done.

I fixed the leak by calling `avformat_close_input()` before the returns of functions that opened input.

---

Finally, in the Python binding, video files are read using OpenCV, but they are not immediately released after reading.

I fixed this by calling vid.release() after the video is read and the attributes are extracted.

This does not appear to directly affect the file leaks above because they are released by the VideoCapture destructor, but it would allow other processes access to file during hashVideoFile which could take some time to run.

### Test Plan

Get number of files open:

```
lsof | wc -l
```

You can see that if you run computeHash multiple times it will never clean up the files and the number of files open will continually increase.

### DIFF
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -60,6 +60,7 @@ bool hashVideoFile(
       vpdq::hashing::FrameBufferHasherFactory::createFrameHasher(height, width);
   if (phasher == nullptr) {
     fprintf(stderr, "Error: Phasher is null");
+    pclose(inputFp);
     return false;
   }
 
@@ -96,6 +97,7 @@ bool hashVideoFile(
             "%s: failed to hash frame buffer. Frame width or height smaller than minimum hashable dimension. %d.\n",
             argv0,
             fno);
+        pclose(inputFp);
         return false;
       }
       // Push to pdqHashes vector
@@ -114,6 +116,7 @@ bool hashVideoFile(
           (int)fread_rc);
     }
   }
+  pclose(inputFp);
   return true;
 }
 
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -100,8 +100,19 @@ bool hashVideoFile(
         pclose(inputFp);
         return false;
       }
+
+      // Avoid divide by zero when calculating timestamp
+      // The timestamp will equal the frame number of the hash
+      // if framesPerSec <= 0. This is common on very short videos.
+      double timestamp = 0;
+      if (framesPerSec > 0) {
+        timestamp = (double)fno / framesPerSec;
+      } else {
+        timestamp = fno;
+      }
+
       // Push to pdqHashes vector
-      pdqHashes.push_back({pdqHash, fno, quality, (double)fno / framesPerSec});
+      pdqHashes.push_back({pdqHash, fno, quality, timestamp});
       if (verbose) {
         printf("PDQHash: %s \n", pdqHash.format().c_str());
       }
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -100,19 +100,8 @@ bool hashVideoFile(
         pclose(inputFp);
         return false;
       }
-
-      // Avoid divide by zero when calculating timestamp
-      // The timestamp will equal the frame number of the hash
-      // if framesPerSec <= 0. This is common on very short videos.
-      double timestamp = 0;
-      if (framesPerSec > 0) {
-        timestamp = (double)fno / framesPerSec;
-      } else {
-        timestamp = fno;
-      }
-
       // Push to pdqHashes vector
-      pdqHashes.push_back({pdqHash, fno, quality, timestamp});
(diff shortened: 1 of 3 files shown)


Dcallies:
Hey @ianwal, sorry for the super slow response here, I stopped checking the github as part of my daily rounds, and I've been let down by notifications!

Dcallies (approved):
Thanks a ton for the help here!

vpdq/cpp/io/vpdqio.cpp
```diff
@@ -140,6 +140,7 @@ bool readVideoStreamInfo(
         "%s: could not find video stream info \"%s\".\n",
         programName,
         inputVideoFileName.c_str());
+    avformat_close_input(&pFormatCtx)
```
Dcallies, line 143: blocking: Missing semicolon

vpdq/python/vpdq.pyx
```diff
@@ -144,6 +144,8 @@ def computeHash(
 
     if downsample_height == 0:
         downsample_height = vid.get(cv2.CAP_PROP_FRAME_HEIGHT)
+    
+    vid.release()
```
Dcallies, line 148: (more familiar with python): There's probably a nice way to do this with a with/finally, but this is fine for the purpose of this CLI.

Dcallies (changes requested):
Continuous integration is smarter than I am - you should be able to run this locally with make.

You can also run the python CLI once on any file just to make sure it produces something.

ianwal:
Wow, I forgot about the missing semicolon on that fork! Fixed.

I also found another issue for the edge case where calculating the timestamp for a hash would cause the program to fail due to a divide by zero error. I fixed it by checking for a zero framesPerSec and setting the timestamp to the actual frame number if that occurs. This is a real issue and occurs on very short videos with low frame counts and high framerates.

The main brute-force comparing algorithm doesn't depend on the timestamp at all, but if one was using a more advanced comparison that depended on the timestamp then it could potentially be inaccurate. Perhaps the hashing should gracefully exit if framesPerSec==0 then instead? The scope of this implementation is unclear to me.

vpdq/python/vpdq.pyx
```diff
@@ -144,6 +144,8 @@ def computeHash(
 
     if downsample_height == 0:
         downsample_height = vid.get(cv2.CAP_PROP_FRAME_HEIGHT)
+
```
Dcallies, line 147: L141 is the source of our 0 frames_per_sec, I assume that is from ffmpeg

vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -96,10 +97,22 @@ bool hashVideoFile(
...
       }
+
+      // Avoid divide by zero when calculating timestamp
+      // The timestamp will equal the frame number of the hash
+      // if framesPerSec <= 0. This is common on very short videos.
+      double timestamp = 0;
+      if (framesPerSec > 0) {
+        timestamp = (double)fno / framesPerSec;
+      } else {
+        timestamp = fno;
```
Dcallies, line 111: Can you tell me a little bit more about the types of files that generate this? 

By any chance are they all single frame gifs or something like that?

I'm worried that it will distort the meaning of timestamp if multi-frame 0 fps items exist, but I don't know what a multi-frame 0 fps video might be other than invalid :P 

If they are all single frame only, we can make it more clear what edge case we are catching and crash / except for multi-frame 0fps since it's not going to produce a valid vPDQ hash as we've defined it.

Dcallies (commented):
Thanks again for the hard work! 

By any chance do you have a copyright-free video we can add to the test cases for the future?

Dcallies:
If you want to try and land the closing fix separate from the 0 fps discussion, I'm good to merge the file close one.

ianwal:
Sure, I reverted the divide by zero and I'll create a new PR for it. Should I make a corresponding issue as well?

vpdq/python/vpdq.pyx
ianwal: The Python bindings get this information here from OpenCV, but for cpp it gets this in `vpdqio.cpp` from readVideoStreamInfo(). It's unclear to me why the Python bindings don't reuse readVideoStreamInfo(), which would avoid importing OpenCV entirely.

OpenCV has issues with some video types, even common ones like [AV1](https://github.com/opencv/opencv/issues/22304). [PyAV](https://pypi.org/project/av/) offers excellent bindings to FFmpeg for this kind of thing.

Dcallies (approved):
Thanks for splitting the PR!

Dcallies:
> Sure, I reverted the divide by zero and https://github.com/facebook/ThreatExchange/pull/1326 for it. Should I make a corresponding issue as well?

No need to make an issue unless you think it's helpful, I'm fine just continuing discussion in that PR. Thanks again for all the help!

PR #1322, opened 2023-07-09
[pdq] Avoid function call by loading pixels before iteration

ianwal:
Summary
---------

Accessing individual pixels with Image.getpixel() is extremely slow, so by loading all the pixels before iterating there will be a huge performance gain.

This speeds up hashing by around 10% every image!

Also, there is a redundant `numCols, numRows = img.size()` for some reason that I also removed. This does not affect noticeably affect speed, and it does not affect functionality at all.

Test Plan
---------
Tested with a variety of image types. All of them passed.

No further tests will needed to be added because this just loads the image's pixels before iteration.

### DIFF
--- a/pdq/python/pdqhashing/hasher/pdq_hasher.py
+++ b/pdq/python/pdqhashing/hasher/pdq_hasher.py
@@ -131,10 +131,11 @@ class PDQHasher:
     def fillFloatLumaFromBufferImage(self, img, luma):
         numCols, numRows = img.size
         rgb_image = img.convert("RGB")
-        numCols, numRows = img.size
+        pixels = rgb_image.load()
+
         for i in range(numRows):
             for j in range(numCols):
-                r, g, b = rgb_image.getpixel((j, i))
+                r, g, b = pixels[j, i]
                 luma[i * numCols + j] = (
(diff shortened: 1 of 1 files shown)


Dcallies (changes requested):
Woah! Nice find! Sorry again for the delay, will check at least daily for the next few weeks.

I don't see anything concerning with the code, but I'm worried about an accidental behavior change. When you say variety of images, can you try everything under the
https://github.com/facebook/ThreatExchange/tree/main/pdq/data directory and make sure that it has the same hash before/after?

If you are looking at performance, see also https://github.com/faustomorales/pdqhash-python

We should really automate this as a test...  I see some tests under /testing in this directory, but I  notice they are not being triggered by CI...

ianwal:
Just tested on all 51 files in https://github.com/facebook/ThreatExchange/tree/main/pdq/data and the hashes are identical.

Thanks for the link to the bindings, but unfortunately the author stopped building for Linux.

Dcallies:
Dang, the bindings seemed like a much more sustainable way to handle maintaining many PDQ implementations. It's a tight ship right now, but still thinking about the long term. 

Thanks again for the help here!

Dcallies (approved):

PR #1328, opened 2023-07-12
[vpdq] Regression test improvements for local development

ianwal:
Summary
---------

Improvements to the regression test for local runs:

- Create output dir if it doesn't exist

- Create output files if they do not exist

- Changed default directories so that they work locally

- Remove duplicate code just for verbosity option

With these changes you can just run regtest.py and it will run locally with the defaults.

Before, the defaults were set to stuff like "/ThreatExchange/vpdq/output-hashes". I don't know why.

You can also run `regtest.py` from anywhere, you don't have to be in cpp/build.

I also changed all of the os.path stuff to Path. It's much nicer to work with.

I don't see a reason why the cpp CI test working directory shouldn't be set to the whole directory, especially since it accesses it anyway to get the sample hashes.

Test Plan
---------

Run the CI test and regtest.py locally. See that it outputs correctly.

Perhaps take a close look at the things I changed since it's a regression test. But, the test "passes" anyway even if it fails as long as it compiles and executes so the changes can't do too much harm...

<details>
<summary>Example regtest.py run</summary>

```sh

Writing output to temp directory: /tmp/tmpshrfc681

Hashing file chair-19-sd-bar.mp4


Hashing file chair-20-sd-bar.mp4


Hashing file chair-22-sd-grey-bar.mp4


Hashing file chair-22-sd-sepia-bar.mp4


Hashing file chair-22-with-large-logo-bar.mp4


Hashing file chair-22-with-small-logo-bar.mp4


Hashing file chair-orig-22-fhd-no-bar.mp4


Hashing file chair-orig-22-hd-no-bar.mp4


Hashing file chair-orig-22-sd-bar.mp4


Hashing file doorknob-hd-no-bar.mp4


Hashing file pattern-hd-no-bar.mp4


Hashing file pattern-longer-no-bar.mp4


Hashing file pattern-sd-grey-bar.mp4


Hashing file pattern-sd-with-large-logo-bar.mp4


Hashing file pattern-sd-with-small-logo-bar.mp4


Matching Video chair-19-sd-bar.txt from hash file chair-19-sd-bar.txt
100.000000 Percentage  matches


Matching Video chair-20-sd-bar.txt from hash file chair-20-sd-bar.txt
100.000000 Percentage  matches


Matching Video chair-22-sd-grey-bar.txt from hash file chair-22-sd-grey-bar.txt
100.000000 Percentage  matches


Matching Video chair-22-sd-sepia-bar.txt from hash file chair-22-sd-sepia-bar.txt
100.000000 Percentage  matches


Matching Video chair-22-with-large-logo-bar.txt from hash file chair-22-with-large-logo-bar.txt
100.000000 Percentage  matches


Matching Video chair-22-with-small-logo-bar.txt from hash file chair-22-with-small-logo-bar.txt
100.000000 Percentage  matches


Matching Video chair-orig-22-fhd-no-bar.txt from hash file chair-orig-22-fhd-no-bar.txt
100.000000 Percentage  matches


Matching Video chair-orig-22-hd-no-bar.txt from hash file chair-orig-22-hd-no-bar.txt
100.000000 Percentage  matches


Matching Video chair-orig-22-sd-bar.txt from hash file chair-orig-22-sd-bar.txt
100.000000 Percentage  matches


Matching Video doorknob-hd-no-bar.txt from hash file doorknob-hd-no-bar.txt
100.000000 Percentage  matches


Matching Video pattern-hd-no-bar.txt from hash file pattern-hd-no-bar.txt
100.000000 Percentage  matches


Matching Video pattern-longer-no-bar.txt from hash file pattern-longer-no-bar.txt
100.000000 Percentage  matches


Matching Video pattern-sd-grey-bar.txt from hash file pattern-sd-grey-bar.txt
100.000000 Percentage  matches


Matching Video pattern-sd-with-large-logo-bar.txt from hash file pattern-sd-with-large-logo-bar.txt
100.000000 Percentage  matches


Matching Video pattern-sd-with-small-logo-bar.txt from hash file pattern-sd-with-small-logo-bar.txt
100.000000 Percentage  matches

```
</details>

### DIFF
--- a/.github/workflows/vpdq-ci-cpp.yml
+++ b/.github/workflows/vpdq-ci-cpp.yml
@@ -17,7 +17,7 @@ on:
 
 defaults:
   run:
-    working-directory: vpdq/cpp
+    working-directory: vpdq
 
 jobs:
   build-and-test:
@@ -31,11 +31,11 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y make cmake ffmpeg libavdevice-dev libavfilter-dev libavformat-dev libavcodec-dev libswresample-dev libswscale-dev libavutil-dev
-          mkdir -p build
-          cd build
+          mkdir -p cpp/build
+          cd cpp/build
           cmake ..
           make
       - name: regression test
         run: |
-          mkdir -p output-hashes
-          python regtest.py -d ${GITHUB_WORKSPACE}/vpdq/cpp/output-hashes -i ${GITHUB_WORKSPACE}/tmk/sample-videos
+          mkdir -p cpp/output-hashes
+          python cpp/regtest.py -d ${GITHUB_WORKSPACE}/vpdq/cpp/output-hashes -i ${GITHUB_WORKSPACE}/tmk/sample-videos
--- a/vpdq/cpp/regtest.py
+++ b/vpdq/cpp/regtest.py
@@ -11,6 +11,7 @@ VPDQ_DIR = DIR.parent
 SAMPLE_HASHES_DIR = VPDQ_DIR / "sample-hashes"
 EXEC_DIR = VPDQ_DIR / "cpp/build"
 
+
 def get_argparse() -> argparse.ArgumentParser:
     DEFAULT_OUTPUT_HASHES_DIR = VPDQ_DIR / "output-hashes"
     DEFAULT_SAMPLE_VIDEOS_DIR = VPDQ_DIR.parent / "tmk/sample-videos"
@@ -108,6 +109,11 @@ def main():
     distanceTolerance = str(args.matchDistanceTolerance)
     qualityTolerance = str(args.qualityTolerance)
     verbose = args.verbose
+
+    if not inputVideoFolder.exists():
+        raise argparse.ArgumentTypeError(
+            f"inputVideoFolder: {inputVideoFolder} does not exist."
+        )
 
     outputHashFolder.mkdir(parents=True, exist_ok=True)
 
--- a/vpdq/cpp/regtest.py
+++ b/vpdq/cpp/regtest.py
@@ -15,6 +15,17 @@ SAMPLE_HASHES_DIR = VPDQ_DIR / "sample-hashes"
 EXEC_DIR = VPDQ_DIR / "cpp/build"
 
 
+def get_os():
+    if platform.system() == "Windows":
+        return "Windows"
+    elif platform.system() == "Darwin":
+        return "Darwin"
+    elif platform.system() == "Linux":
+        return "Linux"
+    else:
+        raise Exception("Unsupported OS")
+
+
 def get_argparse() -> argparse.ArgumentParser:
     DEFAULT_SAMPLE_VIDEOS_DIR = VPDQ_DIR.parent / "tmk/sample-videos"
 
@@ -79,7 +90,9 @@ def main():
     hashVideoExecutable = EXEC_DIR / "vpdq-hash-video"
     matchHashesExecutable = EXEC_DIR / "match-hashes-byline"
 
-    if platform.system() == "Windows":
+    OS = get_os()
+
+    if OS == "Windows":
         hashVideoExecutable = Path(f"{hashVideoExecutable}.exe")
         matchHashesExecutable = Path(f"{matchHashesExecutable}.exe")
 
@@ -105,11 +118,9 @@ def main():
         )
 
     with TemporaryDirectory() as tempOutputHashFolder:
-        """
-        Write the files to temp
-        If outputHashFolder is specified then
-        copy the output files there, overwriting existing files
-        """
+        # Write the files to temp
+        # If outputHashFolder is specified then
+        # copy the output files there, overwriting existing files
 
         tempOutputHashFolder = Path(tempOutputHashFolder)
         if outputHashFolder is not None:
@@ -123,22 +134,21 @@ def main():
         for fileStr in glob.iglob(f"{inputVideoFolder}/**/*.mp4", recursive=True):
             file = Path(fileStr)
 
-            """
-            Create output hash file or overwrite existing file
-            This is hardcoded in cpp, and it
-            does not create the file if it does not exist:
-
-            vpdq-hash-video.cpp:
-
-            // Strip containing directory:
-            std::string b = basename(inputVideoFileName, "/");
-            // Strip file extension:
-            b = stripExtension(b, ".");
-            outputHashFileName = outputDirectory + "/" + b + ".txt";
-            """
+            # Create output hash file or overwrite existing file
+            # This is hardcoded in cpp, and it
+            # does not create the file if it does not exist:
+            #
+            # vpdq-hash-video.cpp:
+            #
+            # // Strip containing directory:
+            # std::string b = basename(inputVideoFileName, "/");
+            # // Strip file extension:
+            # b = stripExtension(b, ".");
+            # outputHashFileName = outputDirectory + "/" + b + ".txt";
             outputHashFile = tempOutputHashFolder / f"{file.stem}.txt"
             with open(outputHashFile, "x+t"):
                 pass
+
             print(f"\nHashing file {file.name}")
             command = [
                 hashVideoExecutable,
@@ -156,9 +166,20 @@ def main():
                 command.insert(1, "-v")
 
             try:
-                subprocess.check_call(command)
+                hash_proc = subprocess.run(
+                    command,
+                    check=True,
+                    capture_output=True,
+                    shell=(OS == "Windows"),
+                )
+                if verbose:
+                    # This will print all PDQHashes e.g.
+                    # PDQHash: ebcc8b06b0666ea34cf9b85972a983a4f94668af05fc9d52aa9662f975499514
+                    # selectframe 563
+                    print(str(hash_proc.stdout, "utf-8"))
             except subprocess.CalledProcessError as e:
-                print(e)
+                print(e.cmd)
+                print(str(e.stderr, "utf-8"))
                 sys.exit(1)
 
             if outputHashFolder is not None:
@@ -183,9 +204,20 @@ def main():
                 command.insert(1, "-v")
 
             try:
-                subprocess.check_call(command)
+                match_proc = subprocess.run(
+                    command,
+                    check=True,
+                    capture_output=True,
+                    shell=(OS == "Windows"),
+                )
+                if verbose:
+                    # This will print all PDQHashes e.g.
+                    # Line 201 Hash1: da4b380330b725b4a5f08ff03d0f6949da4fd2d3e7c8e4930fa7b80662a17c4e
+                    # Hash2: da4b380330b725b4a5f08ff03d0f6949da4fd2d3e7c8e4930fa7b80662a17c4e match
+                    print(str(match_proc.stdout, "utf-8"))
             except subprocess.CalledProcessError as e:
-                print(e)
+                print(e.cmd)
+                print(str(e.stderr, "utf-8"))
                 sys.exit(1)
 
 
--- a/vpdq/cpp/regtest.py
+++ b/vpdq/cpp/regtest.py
@@ -4,6 +4,8 @@ import subprocess
 import sys
 import argparse
 from pathlib import Path
+from tempfile import TemporaryDirectory
+from shutil import copyfile
 import glob
 
 DIR = Path(__file__).parent
@@ -13,7 +15,6 @@ EXEC_DIR = VPDQ_DIR / "cpp/build"
 
 
 def get_argparse() -> argparse.ArgumentParser:
-    DEFAULT_OUTPUT_HASHES_DIR = VPDQ_DIR / "output-hashes"
     DEFAULT_SAMPLE_VIDEOS_DIR = VPDQ_DIR.parent / "tmk/sample-videos"
 
     ap = argparse.ArgumentParser(description=__doc__)
@@ -30,8 +31,7 @@ def get_argparse() -> argparse.ArgumentParser:
         "--outputHashFolder",
         metavar="OUTPUT_HASH_Folder_PATH",
         help="Output Hash Folder's Name",
-        default=DEFAULT_OUTPUT_HASHES_DIR,
-        type=dir_path,
+        type=Path,
     )
     ap.add_argument(
         "-i",
@@ -39,7 +39,7 @@ def get_argparse() -> argparse.ArgumentParser:
         metavar="INPUTPUT_VIDEO_FOLDER_PATH",
         help="Input Video Folder",
         default=DEFAULT_SAMPLE_VIDEOS_DIR,
-        type=dir_path,
+        type=Path,
     )
     ap.add_argument(
         "-s",
@@ -74,14 +74,6 @@ def get_argparse() -> argparse.ArgumentParser:
     return ap
 
 
-def dir_path(string):
-    if Path(string).is_dir():
-        return string
-    raise argparse.ArgumentTypeError(
-        f"readable_dir: {string} is not a valid directory path"
-    )
-
-
 def main():
     hashVideoExecutable = EXEC_DIR / "vpdq-hash-video"
     matchHashesExecutable = EXEC_DIR / "match-hashes-byline"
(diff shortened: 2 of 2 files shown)


.github/workflows/vpdq-ci-cpp.yml
```diff
@@ -17,7 +17,7 @@ on:
 
 defaults:
```
Dcallies, line 18: (unsure): I remember something about this being weird, where maybe the old CI would run instead of the new one until merge. Since it seems like you are planning on staying with the project a bit longer, we can just make sure it doesn't stop failing afterwards :P

vpdq/cpp/regtest.py
```diff
@@ -1,9 +1,17 @@
...
 import argparse
+from pathlib import Path
+import glob
+
+DIR = Path(__file__).parent
+VPDQ_DIR = DIR.parent
+SAMPLE_VIDEOS_DIR = VPDQ_DIR.parent / "tmk/sample-videos"
+SAMPLE_HASHES_DIR = VPDQ_DIR / "sample-hashes"
+OUTPUT_HASHES_DIR = VPDQ_DIR / "output-hashes"
+EXEC_DIR = VPDQ_DIR / "cpp/build"
```
Dcallies, line 14: (nit): Since we use as these as the defaults for argparse, and not definitively, it might be helpful to prefix these with `DEFAULT_` so a casual future author doesn't reference them directly.

(ignoreable, alt): One reason to put these in get_argparse rather than as globals here is it makes it harder to make that mistake. I think it's harmless, with the above fix, but you could hide these in `get_argparse` or a helper function to try and avoid it.

vpdq/cpp/regtest.py
```diff
@@ -73,90 +81,91 @@ def get_argparse() -> argparse.ArgumentParser:
...
+        print(
+            "Error: Hashing executable/s not found. Build vpdq before running regtest."
+        )
+        sys.exit(1)
+
     ap = get_argparse()
     args = ap.parse_args()
-    inputVideoFolder = args.inputVideoFolder
-    outputHashFolder = args.outputHashFolder
+    inputVideoFolder = Path(args.inputVideoFolder)
```
Dcallies, line 103: nit: Another approach we can sue here is to use argparse's `type=Path` to force it to be parsed there. We could go a step further and also check things like existence and throw exceptions if they aren't perfect, though I'm fine not thoroughly validating given the nature of this script. 

Both click and argparse modules have pre-built filetypes, but we chose when building this repo to take a draconian approach to not using any external dependencies if we didn't need to, though click is quite nice.

vpdq/cpp/regtest.py
```diff
@@ -73,90 +81,91 @@ def get_argparse() -> argparse.ArgumentParser:
...
+        # vpdq-hash-video.cpp:
+        #
+        # // Strip containing directory:
+        # std::string b = basename(inputVideoFileName, "/");
+        # // Strip file extension:
+        # b = stripExtension(b, ".");
+        # outputHashFileName = outputDirectory + "/" + b + ".txt";
+
+        with open(outputHashFolder / f"{file.stem}.txt", "w"):
+            pass
```
Dcallies, line 131: maybe blocking: I'm somewhat spooked by this script writing to files in an existing directory, this could be quite dangerous - what we can instead do is use pythons temporary directory functionality: https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

NamedTemporaryFile allows you to specify `delete=False` to retain the data, but there's no equivalent - instead you can use /tmp/ `gettmpprefix()` https://docs.python.org/3/library/tempfile.html#tempfile.gettempprefix to get a random name that will get cleaned up. 

You can then output the created directory while the command is running as a hint of where to look at the output.

Dcallies (changes requested):
> Before, the defaults were set to stuff like "/ThreatExchange/vpdq/output-hashes". I don't know why.

This looks like hardcoding of paths for the docker image development. Agree with you that it's clowny :P

I had one blocking question that might make the script safter, which is to use the /tmp/ directory for output, otherwise that truncation looks a little scary (and writing ine repo and relying on what I assumping is .gitignore is overly strange). 

I wasn't able to quite do a full review, have to board my flight. If you want to put an example execution in the test plan (with the output to screen), that can help with review.

vpdq/cpp/regtest.py
ianwal: Sure, I moved the parameters that be changed into get_argparse(). I only add DEFAULT to those because the others can't be changed. If parameters are added to change any of those as well they can be moved into get_argparse() and have a DEFAULT prefix.

vpdq/cpp/regtest.py
ianwal: I believe what you're explaining is already done through the `type=dir_path` and `dir_path` function. I'll add a check for existence for the input folder. The output hash folder should not exist before running, but this won't matter if I use tempfile.

vpdq/cpp/regtest.py
ianwal: Okay, I'll use a tempdir by default, but there's still an outputHashFolder parameter. If you want to use this instead of the tempdir, you would still have to write and potentially overwrite files in the directory.

I don't see how using a tempdir fixes this. Do you want me to use tempfiles if the user specifies outputHashFolder as well? But then if you want to inspect the files manually (I've done this multiple times, it's useful) you would have to not clean them up after a run. Then you have a bunch of tempfiles with random prefixes in your outputdir you have to delete manually. Yuck.

I would say check if a file exists before writing to it but then you can't run it multiple times without manually deleting all the files which is bad. Very annoying.

I think if we're using a tempdir it should be filled with regular files, not tempfiles. That way the file names retain their original structure as they should.

If the user specifies a folder, it should not be a tempdir, obviously.

And for either case there's no reason to clean up the files between runs because then you can't inspect them, so they should overwrite existing files (won't happen for tempdir) as they already do.

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
-        raise argparse.ArgumentTypeError(f"readable_dir: {string} is not a valid path")
+def main():
+    hashVideoExecutable = EXEC_DIR / "vpdq-hash-video"
+    matchHashesExecutable = EXEC_DIR / "match-hashes-byline"
 
+    if not hashVideoExecutable.exists() or not matchHashesExecutable.exists():
+        print(
+            "Error: Hashing executable/s not found. Build vpdq before running regtest."
+        )
+        sys.exit(1)
```
Dcallies, line 85: (ignorable): You can turn these into a validator on argparse and raise ArgumentError to get slightly more consistent behavior: https://docs.python.org/3/library/argparse.html#type

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
     qualityTolerance = str(args.qualityTolerance)
-    for file in os.listdir(sample):
-        if file.endswith(".txt"):
-            print("\nMatching File " + file)
-            sampleFile = f"{sample}/{file}"
-            outputFile = f"{output}/{file}"
+    verbose = args.verbose
+
+    if not inputVideoFolder.exists():
+        raise argparse.ArgumentTypeError(
```
Dcallies, line 98: nit: This is a misplaced ArgumentTypeError that will turn into a full stack trace. You can embed this into parsing for nice error messages by turning this into a function like

```
def _validate_path(string_path: str) -> Path:
   ret = Path(string_path)
   if not ret.exists():
      raise argparse.ArgumentTypeError("path does not exist")
   return ret

...
     ...
     type=_validate_path,
   )
```

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
+            does not create the file if it does not exist:
+
+            vpdq-hash-video.cpp:
+
+            // Strip containing directory:
+            std::string b = basename(inputVideoFileName, "/");
+            // Strip file extension:
+            b = stripExtension(b, ".");
+            outputHashFileName = outputDirectory + "/" + b + ".txt";
+            """
```
Dcallies, line 133: blocking: This becomes a string in python, it looks like this wants to be a comment. Use

```
# Create output hash file or overwrite existing file
# This is hardcoded in cpp, and it
# ...
```

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
+            outputHashFile = tempOutputHashFolder / f"{file.stem}.txt"
+            with open(outputHashFile, "x+t"):
+                pass
```
Dcallies, line 136: This might be surprising enough to a reader that something like

```
pass # opening a file in write mode truncates it
```

might be helpful for code historians of the future

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
+            ]
+
+            if verbose:
+                command.insert(1, "-v")
+
+            try:
+                subprocess.check_call(command)
+            except subprocess.CalledProcessError as e:
+                print(e)
+                sys.exit(1)
```
Dcallies, line 184: ditto

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
-                        qualityTolerance,
-                    ]
-                )
+                command.insert(1, "-v")
+
+            try:
+                subprocess.check_call(command)
+            except subprocess.CalledProcessError as e:
+                print(e)
+                sys.exit(1)
```
Dcallies, line 157: blocking:  I think this is the same as not catching the exception (that check_call throws exception is well known). Am I missing something?

Dcallies (changes requested):
Sorry this is taking so many passes! I can try and find a channel to make it easier to contact me.

blocking: Can you paste the example output of one run into the test plan?

Should look like

```
# in my xyz directory
$ python3 regtest.py --maybe-a-flag-i-dunno
This is the output of the run!
More output
```

vpdq/cpp/regtest.py
ianwal: I actually changed my mind and if a tempdir is used then just clean it up. This is probably the best option for 99% of tests where you just want to see it pass.

You can create TemporaryDirectory's with the context manager that aren't deleted starting in Python 3.12, but not now. Not super eager to use tempfile.mkdtemp directly.

vpdq/cpp/regtest.py
ianwal: These aren't CLI parameters. Do you mean just use common function for every check for file/folder validity?

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
-        return string
-    else:
-        raise argparse.ArgumentTypeError(f"readable_dir: {string} is not a valid path")
+def main():
+    hashVideoExecutable = EXEC_DIR / "vpdq-hash-video"
+    matchHashesExecutable = EXEC_DIR / "match-hashes-byline"
 
+    if not hashVideoExecutable.exists() or not matchHashesExecutable.exists():
+        print(
+            "Error: Hashing executable/s not found. Build vpdq before running regtest."
```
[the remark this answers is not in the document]
ianwal: Not a typo, it was meaning one or more executables.

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
-            sampleFile = f"{sample}/{file}"
-            outputFile = f"{output}/{file}"
+    verbose = args.verbose
+
+    if not inputVideoFolder.exists():
+        raise argparse.ArgumentTypeError(
+            f"inputVideoFolder: {inputVideoFolder} does not exist."
+        )
+
+    with TemporaryDirectory() as tempOutputHashFolder:
```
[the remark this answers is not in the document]
ianwal: It will delete the tempdir but not the directory it copies files to. I changed my mind on  not cleaning up if using tempdir.

vpdq/cpp/regtest.py
ianwal: I'm not sure what the difference is. [PEP257](https://peps.python.org/pep-0257/) states "String literals occurring elsewhere in Python code may also act as documentation." Even [the creator of Python suggests](https://twitter.com/gvanrossum/status/112670605505077248) using these. But I'll change it for style reasons.

vpdq/cpp/regtest.py
ianwal: No, it does not truncate it. `x` does not truncate, `w` does. Additionally, this is for the temporary directory, so if there is somehow a name collision there are bigger problems than truncating the file.

If an output directory is specified, _copying_ the file will replace it if it exists in the outputHashFolder, which I cannot imagine anyone would expect otherwise. If it was a regular CLI program it would probably be helpful to add some warning messages, but this is a simple regression test script, and it would be very irritating to have to delete the output folder contents every time the test needs to be run.

vpdq/cpp/regtest.py
ianwal: Yeah looking at it this is stupid. I think I was looking at `vpdq-release.py` as a model, but that is definitely NOT the file I want to use as a reference.

Also check_call is old and replaced by check=True in subprocess.run(). I'll fix this.

ianwal:
Added all requested changes and an example output run with no arguments from my machine to the Test Plan.

Dcallies (approved):
Sorry it took so many passes! Left some comments inline, let me know if you want to just merge now.

vpdq/cpp/regtest.py
ianwal: It could be just x because it's not read from. It could also be w.

I just used t to specify it's a text file. Doesn't affect anything since nothing is written to it.

Writing to temp then copying does mean on a very early failure it will be cleaned up.

Yes you could use touch. I didn't know that was a Pathlib method. I'll change it.

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
+        else:
+            print(f"Writing output to temp directory: {tempOutputHashFolder}")
+        # TODO: Add more general options for other video extensions.
+        for fileStr in glob.iglob(f"{inputVideoFolder}/**/*.mp4", recursive=True):
+            file = Path(fileStr)
+
+            """
+            Create output hash file or overwrite existing file
+            This is hardcoded in cpp, and it
+            does not create the file if it does not exist:
```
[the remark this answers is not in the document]
ianwal: Yeah having the separate output `outputVPDQFeatureToFile` thing makes no sense.

vpdq/cpp/regtest.py
```diff
@@ -72,91 +74,114 @@ def get_argparse() -> argparse.ArgumentParser:
...
+                outputHashFolder.mkdir(parents=True, exist_ok=True)
+            print(f"Writing output to directory: {outputHashFolder}")
+        else:
+            print(f"Writing output to temp directory: {tempOutputHashFolder}")
+        # TODO: Add more general options for other video extensions.
+        for fileStr in glob.iglob(f"{inputVideoFolder}/**/*.mp4", recursive=True):
+            file = Path(fileStr)
+
+            """
+            Create output hash file or overwrite existing file
```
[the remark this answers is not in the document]
ianwal: No. copyfile can overwrite existing files with the same name if you specify the output directly. Otherwise you have to delete the files every run if you use -d which is a huge pain.

Dcallies:
Since I see you working on this guy right now, and I never mentioned it previously, so here's what I mean with some of the comment prefixes:

**blocking:** I think this should be changed before landing (if you think I'm wrong, like the comment/docstring example, you can teach me something)
**blocking question**: I am not sure if this should be changed, but I want to know the answer to the question before landing
**nit**: I have this as a style preference or it's something minor. Always ignorable if you prefer yours
**alt**: Here's another way to do it/code golf. Always ignorable if you prefer yours
**ignorable**: Just underlining you can choose to keep what you have if you like yours better

I aim to get all the hills I'll die on in a single review, so hopefully new blockers that were in previous commits don't appear. If I didn't give you a complete review, I'll try and mention it in the summary. Additionally, I try and check github at least every 24 hours on weekdays to respond to comments.

ianwal:
Ready to merge

Dcallies:
New code looks fine!

PR #1329, opened 2023-07-13
[vpdq] Cleanup matchTwoHash

ianwal:
Summary
---------

Use idiomatic C++ in general:
 - Iterate over vectors with range based for loops
 - Use std::cout instead of printf for verbose output. 

Reduce repeated code by moving them into two functions: filterFeatures and findMatches.

I also fixed casting to the wrong type for the result. It should be a double but it was cast to a float. C++ doesn't need the dividend to be cast at all.
If you use 100.0 it's guaranteed to be a double since C++11, otherwise you have to use the suffix "f".

I made the parameters const because that's what `filehasher.h` does, even if they're passed by value.

Test Plan
---------

Compare old output to new output and see that they're the same.

regtest.py is not very useful with these changes because it will be using this function to test. There should probably be some false-positive tests added because right now it only shows 100.0% for all. There should be some test videos that should not be matches.

Timed with [hyperfine](https://github.com/sharkdp/hyperfine) and performance is the same as before the changes.

### DIFF
--- a/vpdq/cpp/hashing/matchTwoHash.cpp
+++ b/vpdq/cpp/hashing/matchTwoHash.cpp
@@ -2,6 +2,8 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 // ================================================================
 
+#include <iostream>
+
 #include <pdq/cpp/io/hashio.h>
 #include <vpdq/cpp/hashing/vpdqHashType.h>
 #include <vpdq/cpp/io/vpdqio.h>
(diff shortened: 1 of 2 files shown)


vpdq/cpp/hashing/matchTwoHash.cpp
```diff
@@ -2,82 +2,99 @@
...
+ *
+ * @param features1 Features to match
+ * @param features2 Features to match
+ * @param distanceTolerance Distance tolerance of considering a match. Lower is
+ * more similar.
+ * @param verbose Print features with matching hashes
+ *
+ * @return Number of matches
+ */
+static unsigned int findMatches(
```
Dcallies, line 55: I last used C++ seriously in 2014, so I might be a yokel :P 

Rather than `unsigned int` does `size_type` align better, since it should be the same range as the number of elements in the vectors.

vpdq/cpp/hashing/matchTwoHash.cpp
```diff
@@ -2,82 +2,99 @@
...
+      if (feature1.pdqHash.hammingDistance(feature2.pdqHash) <
           distanceTolerance) {
-        tMatchCnt++;
+        matchCnt++;
         if (verbose) {
-          printf(
-              "Query Hash: %s Target Hash: %s match \n",
-              queryFiltered[j].pdqHash.format().c_str(),
-              targetFiltered[i].pdqHash.format().c_str());
+          std::cout << "Query Hash: " << feature1.pdqHash.format()
```
Dcallies, line 67: (pls ignore): In an alternate universe, we could consider switching to use whatever the GLOG has turned into, since libraries writing directly to screen is a little odd. Way out of scope for this guy, but something for us to consider in the future.

Dcallies (approved):
Looks much cleaner, thanks!

Leaving it unmerged for the `size_type` question, but let me know if you prefer merged as is.

vpdq/cpp/hashing/matchTwoHash.cpp
ianwal: Good catch. I'll change it.

ianwal:
Fixed the size_type. Should be ready to merge.

PR #1331, opened 2023-07-16
[vpdq] Link Python binding with libav*

ianwal:
Summary
---------

Link Cython extension against libav* so that Python binding can use the new hashing function and avoid piping from FFmpeg.

Before my refactor, the binding uses a separate Makefile which does not include files that needed libav* functions (only `vpdqio.cpp` I think) to avoid library issues. It also didn't build the 3 CLI programs in cpp/bin.

Now `setup.py` uses CMake to build the library. CMake sets ENV variables for the library and include dirs, and then `setup.py` links the Cython extension to the libav* libraries. This is so that they don't have to be hardcoded in `setup.py`.

Building this requires new packages on Ubuntu:
- libavcodec-dev
- libavformat-dev 
- libavdevice-dev 
- libavutil-dev 
- libavfilter-dev
- libswscale-dev 
- libswresample-dev 

I also made the "Install dependencies" CI Python test fail if any `vpdq-release` commands fail rather than having it fail during the pytest step.

It currently builds and appears to work and produce the same hashes and the CPP implementation.

Test Plan
---------

Run on the same files as CPP implementation and compare vpdqFeatures.

---

This is still a WIP but I'm creating the draft to allow for comments.

I don't know yet if FFmpeg is still required for building or running the program.

If some CMake wizard wants to improve what I've done so far I would greatly appreciate it.

The build system for the Python binding has some questionable things, even before these changes, but if it works it works.


Dcallies:
Ran out of time today, but I definitely see your changes, will put more time in reviews tomorrow! Thanks again for the hard work!

vpdq/cpp/bin/vpdq-hash-video.cpp
```diff
@@ -32,45 +34,39 @@ static void usage(char* argv0, int rc) {
...
  *
  * @param path Path of the target file
- * @param delimiter The character that marks the beginning or end
  *
  */
-string basename(const string& path, const string& delimiter) {
-  size_t n = path.length();
-  size_t i = path.rfind(delimiter, n);
-  if (i == string::npos) {
+std::string basename(const std::string& path) {
```
ianwal, line 46: I removed the delimiter parameter because it's useless. What file system doesn't use "." as the extension delimiter??

vpdq/cpp/bin/vpdq-hash-video.cpp
```diff
@@ -32,45 +34,39 @@ static void usage(char* argv0, int rc) {
...
+  size_t i = path.find_last_of("\\/");
```
ianwal, line 47: This allows the function to work on Windows and Linux.

"\\/" is used because both backslash and forward slash are special characters in C++, so they need to be escaped with backslashes.

vpdq/cpp/bin/vpdq-hash-video.cpp
```diff
@@ -32,45 +34,39 @@ static void usage(char* argv0, int rc) {
...
+ * @param filename Path of the target file
  *
  */
-string stripExtension(const string& path, const string& delimiter) {
-  size_t n = path.length();
-  size_t i = path.rfind(delimiter, n);
-  if (i == string::npos) {
-    return path;
+std::string stripExtension(const std::string& filename) {
+  size_t n = filename.rfind('.');
```
ianwal, line 65: I removed the delimiter parameter because it's useless. What OS doesn't use "." as the file extension delimiter??

C++17 removes the need for stripExtension() and basename() with `std::filesystem` which provides similar file ops to Python, but this project is C++14.

vpdq/cpp/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.17)
 
-set(CMAKE_CXX_FLAGS "-O2")
+set(CMAKE_CXX_FLAGS "-O2 -fPIC")
```
ianwal, line 3: Shared library means it needs to be position-independent code

vpdq/cpp/CMakeLists.txt
```diff
@@ -63,18 +63,28 @@ add_executable(vpdq-hash-video
...
 
+
 target_link_libraries(vpdqlib
     PkgConfig::LIBAV
 )
 
-target_link_libraries(vpdq-hash-video
+target_link_libraries(match-hashes-byline
     vpdqlib
+    PkgConfig::LIBAV
```
ianwal, line 73: Fairly certain you don't need PkgConfig::LIBAV because vpdqlib already has it, but from what I read it's better to specify this.

vpdq/cpp/CMakeLists.txt
```diff
@@ -63,18 +63,28 @@ add_executable(vpdq-hash-video
...
-target_link_libraries(match-hashes-byline
+target_link_libraries(vpdq-hash-video
     vpdqlib
+    PkgConfig::LIBAV
 )
+
+# Pass the include and libav* library directories to setup.py for Python binding
+set_target_properties(vpdqlib
+    PROPERTIES
+    ENVIRONMENT "LIBRARY_DIRS=$<TARGET_PROPERTY:PkgConfig::LIBAV,IMPORTED_LOCATION>;INCLUDE_DIRS=$<TARGET_PROPERTY:PkgConfig::LIBAV,INTERFACE_INCLUDE_DIRECTORIES>"
```
ianwal, line 89: `LIBRARY_DIRS` and `INCLUDE_DIRS` are passed as env variables to `setup.py`

This is how setup.py gets the libav* library locations.

vpdq/python/vpdq.pyx
```diff
@@ -137,27 +161,30 @@ def computeHash(
...
+    
+    width, height, frames_per_sec = read_video_stream_info(
+        str_path,
+        "vpdqPY",
+    )
 
-    if downsample_height == 0:
-        downsample_height = vid.get(cv2.CAP_PROP_FRAME_HEIGHT)
+    if downsample_width > 0:
+        width = downsample_width
```
ianwal, line 172: Inverted logic because width is retrieved from `read_video_stream_info`() regardless if downsamples are 0. Also the original was not really proper naming since

BTW I think read_video_stream_info() should probably just be a function in `filehasher.cpp` that is called by `hashVideoFile()` because there's no need for the programmer to have to get these attributes. It would reduce complexity.

vpdq/setup.py
```diff
@@ -6,33 +6,55 @@
...
 DIR = Path(__file__).parent
 read_me = DIR / Path("vpdq/python/README.md")
 long_description = read_me.read_text()
 version = (DIR / "vpdq/version.txt").read_text(encoding="utf-8").strip()
 
+# Get the library directories and include directories from the environment variables
+# These variables should be set in the CMakeLists.txt file
+lib_dirs = os.getenv("LIBRARY_DIRS", "").split(":")
+include_dirs = os.getenv("INCLUDE_DIRS", "").split(":")
+include_dirs.extend(["vpdq/pdq", "./"])
```
ianwal, line 21: include_dirs is weird. I couldn't get it to include pdq on Windows for some reason, even though vpdqlib compiles if you manually run the commands.

Something is off with it. I think it works even if the "vpdq/pdq" part is just any random string, but it won't work if it's not there.

Also not sure if "./" should be "." or not.

But on Linux this works fine.

vpdq/setup.py
```diff
@@ -6,33 +6,55 @@
 import sys
 import subprocess
 from pathlib import Path
+import os
+import logging
 
 DIR = Path(__file__).parent
 read_me = DIR / Path("vpdq/python/README.md")
```
ianwal, line 13: This hardcoding "vpdq/..." is very odd. You can't run setup.py directly because of it. You have to run `vpdq-release.py`. I do not understand this decision.

vpdq/vpdq-release.py
```diff
@@ -55,6 +56,7 @@ def run_command(command, cwd="."):
         subprocess.check_call(command, cwd=cwd)
     except subprocess.CalledProcessError as e:
         print(e.output)
+        sys.exit(1)
```
ianwal, line 59: I don't know why this check_call is caught. It means you can "install" the package, with no visible output, but it actually didn't install at all and will error if you do `python -c "import vpdq"`. Very confusing.

vpdq/cpp/CMakeLists.txt
Dcallies: Outside of my area of knowledge, I'm going to trust you :P

vpdq/cpp/bin/vpdq-hash-video.cpp
Dcallies: I would guess it was copied from the TMK code - I don't know enough about why it would be included.

vpdq/python/vpdq.pyx
Dcallies: I open to these changes if you want to make them! If you aren't sure, feel free to leave a comment to that effect to prompt someone in the future.

vpdq/setup.py
Dcallies: I didn't either, and complained about this exact problem at one point.

vpdq/setup.py
Dcallies: Something with environment variable component? 

It behaving better on linux over windows usually has me looking for shell features or file behavior differences. 

I also think this entire file is just brute forcing something that would generate the right output. Watching the original author write this python extension was my first introduction to them, and I tried to learn enough to make some suggestions, but overall I didn't have a good model to pull off of.

vpdq/vpdq-release.py
Dcallies: I don't either...

Maybe trying to emulate a bash script without `-e` ? But that's not a great model to work off of. 

I wonder if you can just replace this entirely with subprocess directly at this point, I don't think there's any need for `run_command`

Dcallies (commented):
I'm not sure how much help I'm going to be on this. I'm happy to accept any version of this that:
1. Looks like its an improvement
2. Still passes existing CI 
3. Has the vPDQ pypy package buildable in linux

I'm leaning heavily on your test plans, and am only looking for things I think are hard blockers (will create challenges in the future). If I were being more thorough, I'd be patching your changes and testing them on my side as well, but due to low bandwidth, and the fact that you've said you'll stick around for a bit, I'm willing to try a bit of iteration. 

Let me know when you think this guy is ready, and I'll take another look - sorry I didn't have more insightful comments!

vpdq/setup.py
ianwal: Yes, it's probably that. LIBRARY_DIRS seemed like it was missing. I'm not worried about it for this PR, but I wanted to document it at the source.

ianwal:
Leaving my notes for an incomplete list of what has to be done to build on Windows here since it's relevant. I don't care if the Python binding builds on Windows for this PR. I want to get the binding working on Windows though just for fun, so I've been doing things in a portable way the whole time.

The CPP impl. runs on Windows right now because of the PDQ changes. Cool.

Windows stuff:
PDQ:
- [x] hashio.cpp getline() is POSIX, so it needs to be reimplemented or removed
- [x] random() is a POSIX function, so it needs to be replaced with C++ <random> or removed
  - Needs to not have -DUSE_BUILTIN_POPCOUNT

VPDQ:
Dependencies:


- Install CMake
- Install MinGW toolchain `pacman -S --needed base-devel mingw-w64-x86_64-toolchain`
- Install MinGW FFmpeg (for libav*) `pacman -S mingw-w64-x86_64-ffmpeg`
- Add MinGW bin folder to path
- `pip install wheel`

Building vpdq:
- Instead of "cmake ..", have to do "cmake .. -G "MinGW Makefiles"
  - This has to be added in `setup.py` as well in the build_ext for Python binding. Also run shell=True for Windows for subprocess when this is added.

- Install Microsoft Visual C++ >=14.0 Build Tools to build the extension (8GB??????????)
  - Also install "Windows 10/11 SDK"

ianwal:
I improved the logging for errors in my last commit and made sure if the compile for vpdq.pyx or the vpdq source fails then the CI fails. This works when installing with `vpdq-release.py -i` or from the distribution package with `-r`]

Other than that I'm ready to merge.

vpdq/vpdq-release.py
```diff
@@ -35,27 +36,51 @@ def get_argparse() -> argparse.ArgumentParser:
...
+                ["pip3", "install", "-e", "."],
+                cwd=PARENTDIR,
+                check=True,
+                capture_output=True,
+            )
+            logger.info(str(install_proc.stdout, "utf-8"))
+        except subprocess.CalledProcessError as e:
+            logger.critical(str(e.stderr, "utf-8"))
+            logger.critical("Failed to install vpdq library.")
+            sys.exit(1)
```
ianwal, line 79: Fail if it doesn't compile!!!!

vpdq/python/vpdq.pyx
```diff
@@ -44,7 +43,7 @@ cdef extern from "vpdq/cpp/hashing/vpdqHashType.h" namespace "facebook::vpdq::ha
 
 
 cdef extern from "vpdq/cpp/hashing/filehasher.h" namespace "facebook::vpdq::hashing":
-    bool hashVideoFileFFMPEG(
+    bool hashVideoFile(
         string input_video_filename,
         vector[vpdqFeature]& pdqHashes,
         string ffmpeg_path,
```
ianwal, line 49: I'm planning on removing the FFmpeg path residuals from CPP and then Python after this PR is merged. I want to ask here if for backwards compatibility for both the CPP CLI and Python function computeHash() the parameters should remain and do nothing, or if they can be removed which might cause errors.

ianwal:
I updated and improved the three README's.

Ready to merge.

ianwal:
By the way, I wanted to mention that the libraries from FFmpeg that are being linked to are licensed under LGPL V2.1. 

From what I have read, this is not a problem since you're not distributing binaries and because it's dynamically linked. That's the whole point of LGPL vs GPL.

If the binary is distributed then it can remain under the license in LICENSE.txt as long as the program is provided in object format, and a link to the LGPL V2.1 license and libraries source code is provided. See [this](https://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic) for info on linking.

I noticed the license in LICENSE.txt is not named, but it appears to be the 2-Clause BSD license, while ThreatExchange is under the 3-Clause BSD license for some reason.

I added the notice [FFmpeg](https://ffmpeg.org/legal.html) requests in the README.md.

This is not legal advice.

Dcallies:
The licensing situation has confused the heck out of me in the past, the intent is for everything distributed in this library to be under BSD (though which version I'm unsure). Thanks again for all the work.

vpdq/python/vpdq.pyx
```diff
@@ -119,7 +135,7 @@ def computeHash(
 
     Args:
         input_video_filename: Input video file path
-        ffmpeg_path: ffmpeg path
+        ffmpeg_path: ffmpeg path (this is not used anymore)
```
Dcallies, line 138: We should probably plan to remove this in some future release.

Dcallies (approved):
Let's land and iterate!

PR #1332, opened 2023-07-16
[vpdq] Add test for downsampled hash file for Python binding

ianwal:
Summary
---------

- Add test for downsampled hash files
  - If a file has a downsampled hash file, calculate the hash for the downsample and compare it to that hash file
- Add a way to generate hash files with the Python binding
  - Useful for adding more test videos
  - Toggled by env var `GENERATE_HASHES`. Default is False
  - Does not overwrite existing files unless env var `OVERWRITE_HASHES` is set

Test Plan
---------

If I replace some random line of the downsampled hash file with 0's (meaning all black frame):

chair-20-sd-bar-128x128.txt:
```
550,100,0000000000000000000000000000000000000000000000000000000000000000,18.333
```

Test for chair-20-sd-bar.mp4:
```
if h1.quality >= QUALITY_TOLERANCE and h2.quality >= QUALITY_TOLERANCE:
>                       assert h1.hamming_distance(h2) < DISTANCE_TOLERANCE
E                       assert 128 < 31
E                        +  where 128 = <bound method VpdqFeature.hamming_distance of VpdqFeature()>(VpdqFeature())
E                        +    where <bound method VpdqFeature.hamming_distance of VpdqFeature()> = VpdqFeature().hamming_distance

python/tests/test_vpdq_hash.py:163: AssertionError
```

The test will fail which is the expected result.

### DIFF
--- a/vpdq/python/tests/test_util.py
+++ b/vpdq/python/tests/test_util.py
@@ -2,7 +2,7 @@
 import vpdq
 import typing as t
 from pathlib import Path
-from typing import Union
+from typing import List, Union
 
 
 def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
@@ -31,7 +31,7 @@ def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
 
 def generate_hashes(
     output_folder: Union[Path, str],
-    input_filepaths: list[Union[Path, str]],
+    input_filepaths: List[Union[Path, str]],
     overwrite: bool = False,
     downsample_width: int = 0,
     downsample_height: int = 0,
--- a/vpdq/python/tests/test_util.py
+++ b/vpdq/python/tests/test_util.py
@@ -1,6 +1,8 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 import vpdq
 import typing as t
+from pathlib import Path
+from typing import Union
 
 
 def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
@@ -25,3 +27,49 @@ def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
         hash.append(feature)
 
     return hash
+
+
+def generate_hashes(
+    output_folder: Union[Path, str],
+    input_filepaths: list[Union[Path, str]],
+    overwrite: bool = False,
+    downsample_width: int = 0,
+    downsample_height: int = 0,
+) -> None:
+    """
+    Generate vpdq hashes for the test videos
+    and write them to files in output_folder
+
+    If overwrite is set, it will overwrite existing hashes in the folder.
+    """
+    output_folder = Path(output_folder)
+    if not output_folder.exists():
+        print(f"Creating output directory at {output_folder}")
+        output_folder.mkdir(parents=True, exist_ok=True)
+
+    for fileStr in input_filepaths:
+        file = Path(fileStr)
+        if downsample_width > 0 and downsample_height > 0:
+            output_hash_filename = (
+                output_folder
+                / f"{file.stem}-{downsample_width}x{downsample_height}.txt"
+            )
+        else:
+            output_hash_filename = output_folder / f"{file.stem}.txt"
+
+        if not overwrite and output_hash_filename.exists():
+            print(f"Hash file {output_hash_filename.name} already exists. Skipping.")
+            continue
+
+        vpdq_hash = vpdq.computeHash(
+            input_video_filename=file,
+            seconds_per_hash=0,
+            downsample_width=downsample_width,
+            downsample_height=downsample_height,
+        )
+
+        with open(output_hash_filename, "w") as output_file:
+            for feature in vpdq_hash:
+                output_file.write(
+                    f"{feature.frame_number},{feature.quality},{vpdq.hash_to_hex(feature.hash)},{feature.timestamp:.3f}\n"
+                )
--- a/vpdq/python/tests/test_util.py
+++ b/vpdq/python/tests/test_util.py
@@ -1,5 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
-import vpdq
+import vpdq # type: ignore
 import typing as t
 from pathlib import Path
 from typing import List, Union
--- a/vpdq/python/tests/test_util.py
+++ b/vpdq/python/tests/test_util.py
@@ -1,8 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
-import vpdq # type: ignore
+import vpdq  # type: ignore
 import typing as t
-from pathlib import Path
-from typing import List, Union
 
 
 def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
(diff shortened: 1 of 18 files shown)


vpdq/python/tests/test_util.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 import vpdq
 import typing as t
+from pathlib import Path
+from typing import List, Union
```
Dcallies, line 5: (ignorable, YMMV): Newer versions of python allow you to use `list[]` in places where we previously used List. I often find myself using `import typing as t` in older version because your list of imported types usually grows to need many lines

vpdq/python/tests/test_util.py
```diff
@@ -25,3 +27,49 @@ def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
         hash.append(feature)
 
     return hash
+
+
+def generate_hashes(
```
Dcallies, line 32: (ignorable, YMMV) This looks like a good place to force kwarg arguments by doing
```
def generate_hashes(
    *,
    output_folder: Union[Path, str],
```

This will prevent callers from doing 

```
generate_hashes("foo", ["bar"], true, 5, 10)
```


---

EDIT: Building off the question abou tgenerating the hashes as a CLI, the arguments of that CLI are probably just this, and then you do something like:

```
generate_vpdqs <list of files> --overwrite -d 250
```

Having it be a list of files then allow you to do shell expansion like dir/a/*.mp4

vpdq/python/tests/test_util.py
```diff
@@ -25,3 +27,49 @@ def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
...
+) -> None:
+    """
+    Generate vpdq hashes for the test videos
+    and write them to files in output_folder
+
+    If overwrite is set, it will overwrite existing hashes in the folder.
+    """
+    output_folder = Path(output_folder)
+    if not output_folder.exists():
+        print(f"Creating output directory at {output_folder}")
```
Dcallies, line 47: (ignorable): I usually try to avoid calling print() directly from library functions - `logging` is often preferred if you need chatty libraries, but then you also have to set up logging.

vpdq/python/tests/test_util.py
```diff
@@ -25,3 +27,49 @@ def read_file_to_hash(input_hash_filename: str) -> t.List[vpdq.VpdqFeature]:
...
+        output_folder.mkdir(parents=True, exist_ok=True)
+
+    for fileStr in input_filepaths:
```
Dcallies, line 50: Super dumb toplevel comment (I can't remember if I knew the answer at one point):

Why do we output to files rather than to screen? Is this because we intend to generate the answer once and then compare against the known answer for future runs of the test?

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -28,15 +33,24 @@
     "pattern-sd-with-small-logo-bar",
 ]
 
-project_dir = Path(__file__).parents[3]
-hash_folder = project_dir / SAMPLE_HASH_FOLDER
-video_folder = project_dir / SAMPLE_VIDEOS
```
Dcallies, line 33: (ignorable): Ugh, I'm starting to remember this mess now

Dcallies (changes requested):
At a toplevel, I don't think generating files should be a test, we don't expect tests to have side-effects in the project directories, and it showing as skipped all the time will be surprising. 

For your types, are you using mypy or pyre or anything to confirm them? I don't think we properly set up linting for this repo, so no CI may be running (though types are still appreciated)

vpdq/python/tests/test_util.py
ianwal: Yes they do allow list, but this commit actually reverts this because I'm developing on Python 3.11 and used list[] which has been allowed since 3.9 but the CI runs a test on 3.8.

vpdq/python/tests/test_util.py
ianwal: Not looking to create a CLI, just trying to add another test.

vpdq/python/tests/test_vpdq_hash.py
ianwal: If it makes you feel better it gets even worse if you look at `setup.py` and `vpqd-release.py` for vpdq.

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -91,3 +131,62 @@ def test_compare_hashes():
...
+            hash2 = sample_hashes[downsampled_hash_file]
+            assert len(hash1) == len(hash2)
+            for h1, h2 in zip(hash1, hash2):
+                if h1.quality >= QUALITY_TOLERANCE and h2.quality >= QUALITY_TOLERANCE:
+                    assert h1.hamming_distance(h2) < DISTANCE_TOLERANCE
+                    assert h1.frame_number == h2.frame_number
+
+
+@pytest.mark.skipif(
+    os.getenv("GENERATE_HASHES") != "1", reason="Skip generating new hashes."
```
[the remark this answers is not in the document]
ianwal: Sure that makes more sense. I don't want to create another CLI though. If someone wants to configure it they can do it through env variables or through the code.

vpdq/python/tests/test_util.py
ianwal: I'm not sure exactly, but I'm just adding another test that closely follows how the old one functioned.

vpdq/python/tests/test_util.py
ianwal: Agreed

vpdq/python/tests/test_vpdq_hash.py
```diff
@@ -28,15 +33,24 @@
...
 sample_hashes = {}
 
+TEST_FILES = []
+for fileStr in glob.iglob(f"{VIDEO_FOLDER}/**/*.mp4", recursive=True):
+    file = Path(fileStr)
+    if not (VIDEO_FOLDER / f"{file.name}").is_file():
+        print(f"Video file {file.name} doesn't exist. Skipping.")
+        continue
+    TEST_FILES.append(file)
+assert len(TEST_FILES) > 0
```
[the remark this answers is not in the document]
ianwal: Sure, and then I will also remove the hardcoded test names to allow more flexibility for adding videos.

The regtest might also break if more videos are added to tmk, but I don't have the energy to fix that right now and that was the same as the old one.

I wish the sample videos were in the base directory instead of in tmk the repo but it is what it is.

Also, is there a reason why Git LFS wasn't used to store the videos?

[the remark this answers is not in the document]
Dcallies: > but I don't have the energy to fix that right now and that was the same as the old one.

No worries, that sounds fine to me

> Also, is there a reason why Git LFS wasn't used to store the videos?

I would guess because we didn't know about Git LFS 😅. I've been soapboxing internally about not putting large files into git into the past.

vpdq/python/tests/test_vpdq_hash.py
Dcallies: I think we ran into a time/polish tradeoff with the original author. We're running a small staff. You are currently talking at the only maintainer, and I also have the ThreatExchange API and our hash import/export pipelines on my duties at Meta. Would love to spend more time on our open source (and am blocking some time out in preparation for a hackathon in October).

Hopefully hiring goes well and we'll be able to spend more time. Other than that, I'm incredibly grateful for folks like you who can spend some time.

Dcallies (approved):
Thanks a ton for splitting the generation out of the test! Another approach that I think would have been fine, would be a docstring telling a user to call the function with the repl, but this hopefully is more clear for future folks.

PR #1333, opened 2023-07-17
[pdq] Replace POSIX dependent functions with STL equivalent

ianwal:
Summary
---------

While I was testing vpdq on Windows I had to make a couple changes to get it to build with pdq.

They're solid changes in general, so I thought I would upload them.

Use C++11 STL over OS dependent functions in general:

- Change file pointers to streams 
  - File pointers were used with getline() which is POSIX only. Replaced file pointers **_only wherever necessary_** with streams and use std::getline(). I tried to change as little as possible here.
- Use C++ random device instead of random() which is POSIX only 
  - This could have been done with rand() but I just used the C++11 one because it's a C++ project
- Fix __builtin_popcount where hamming*() should have been used
  - There's a define in `pdqhamming.h` which chooses either a lookup table or __builtin_popcount based on the `USE_BUILTIN_POPCOUNT`  directive but some functions were directly calling __builtin_popcount regardless.

None of these changes should affect bindings. The Python binding doesn't call any of the changed functions since they're almost all file IO.

I'm not sure why __builtin_popcount is being used at all. [std::bitset::count](https://godbolt.org/z/UpN8Qh) and [__builtin_popcount](https://godbolt.org/z/C4tY8W) generate the same assembly. I can replace __builtin_popcount with std::bitset::count if you would like.

The getline problem is a really good example of why it's not good to do `using namespace std;` since std::getline and getline are both functions.

I vaguely remember reading something about Buck for building and if you guys use that for PDQ I'm not sure how these changes would be affected by that.

Test Plan
---------

Run included PDQ tests and vPDQ tests.

### DIFF
--- a/pdq/cpp/bin/clusterize256x.cpp
+++ b/pdq/cpp/bin/clusterize256x.cpp
@@ -6,6 +6,9 @@
 #define _GNU_SOURCE
 #endif
 
+#include <fstream>
+#include <iostream>
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -47,7 +50,7 @@ const int DEFAULT_PDQ_DISTANCE_THRESHOLD = 31;
 
 // ----------------------------------------------------------------
 static void handle_fp(
-    FILE* fp,
+    std::istream& in,
     MIH256<std::string>& mih,
     std::map<Hash256, int>& centersToIndices,
     int distanceThreshold,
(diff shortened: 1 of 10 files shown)


Dcallies (approved):
Thanks for the detailed explanation in the summary. I had a few questions that normally would have forced a roundtrip, but I was able to answer them from reading your summary.

As for the test plan, I don't know if there are dedicated unittests for it, the blast radius for the C++ implementation might be slightly larger if it is the one linked to the python bindings, though you mentioned it's not in the summary.

ianwal:
Did you have any thoughts on using std::bitset::count?

Dcallies:
If it produces the right output, then I'm game!

PR #1337, opened 2023-07-20
[vpdq] Cleanup FFmpeg residuals and other general code cleanup

ianwal:
Summary
---------

Style and general non-breaking changes:
- Remove unused function `readVideoDuration()`
  - I saw the commit for this and it was never used or really had a purpose
- Convert some file pointers to file streams
- Use exception in hashVideoFile to catch `processFrame()` errors instead of using `-1`
  - Using -1 muddies the meaning of the return. No performance impact expected
  - This function is only used for debugging
- **Remove namespace std**
- Use `<cstdlib>`'s `EXIT_SUCCESS` and `EXIT_FAILURE` macros for exits instead of 1 and 0 for clarity and consistency
- Remove FFmpeg path parameter functionality
  - Does nothing if passed and doesn't error
  - Does not show in `--help` for CLI
  - Still a function parameter in Python binding for backwards compatibility
- Set libav* log level with verbose and change the default from INFO to FATAL
  - Removes ugly useless messages like `[libdav1d @ 0x5576493b62c0] libdav1d 1.2.1`
  - verbose changes it to WARNING
- Add compiler warning flags 

All of these should be self-explanatory and are mostly contained within individual commits. I messed up on not staging one or two like vpdq.pyx so some might be out of logical order.

Breaking changes:
- Remove `readVideoStreamInfo()` and move it into `hashVideoFile()`
  - Breaks old code that had to call this and pass the results to `hashVideoFile()`

Potentially breaking changes:
- Remove `ffmpegPath`, `framesPerSec`, `programName` parameters from `hashVideoFile()`
- Remove `programName` parameter from `hashVideoFile()`
- Remove `programName` parameter from `loadHashesFromFileOrDie()`
- Remove `programName` parameter from `outputVPDQFeatureToFile()`

I didn't see a reason to keep the program name as function parameters for these functions. The three executables are standalone executables so they wouldn't benefit much printing the name while that for debugging. Printing the name in hashVideoFile() I could understand if you're calling it from another program. I could add something like printing the `__FILE__` and/or `__func__`, which are macros, but they're at least portable. But, if you want me to add them back in I can. 

I'm assuming there's much better logging frameworks for C++ that you guys use, I saw you mention glog, but for this project it probably doesn't matter. I haven't written anything substantial in C++ to warrant the use of these so let me know if my assumptions above are incorrect.

I use `std::endl` instead of `"\n"` for printing which does flush the stream, but that should be such a tiny impact when printing with verbose compared to the time it takes to decode a frame and hash it.

Let me know if you want me to split some of this up into different PR's. Almost all of it is just cleanup so I just lumped it together because I didn't want to create multiple branches in case I changed shared things between them.

Test Plan
---------

Run `regtest.py` (which now works as expected!) and see the output is ok.

Also run `vpdq-hash-video` and test with `--ffmpeg` to see it does not give an error.

Also tested with other weird video files I have like 2 frame GIFs and odd video codecs and it hashes them fine.


vpdq/cpp/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.17)
 
-set(CMAKE_CXX_FLAGS "-O2 -fPIC")
+set(CMAKE_CXX_FLAGS "-O2 -fPIC -Wall -Wextra -Werror -Wno-unused-function")
```
Dcallies, line 3: 👍

vpdq/cpp/bin/match-hashes-brute.cpp
```diff
@@ -45,15 +47,13 @@ int main(int argc, char** argv) {
   qualityTolerance = atoi(argv[argi + 3]);
   vector<facebook::vpdq::hashing::vpdqFeature> qHashes;
   vector<facebook::vpdq::hashing::vpdqFeature> tHashes;
-  bool ret =
-      facebook::vpdq::io::loadHashesFromFileOrDie(argv[argi], qHashes, argv[0]);
+  bool ret = facebook::vpdq::io::loadHashesFromFileOrDie(argv[argi], qHashes);
   if (!ret) {
-    return -1;
```
Dcallies, line 51: -1 is not a valid return code -_-

vpdq/cpp/bin/vpdq-hash-video.cpp
```diff
@@ -73,15 +71,14 @@ std::string stripExtension(const std::string& filename) {
...
-  string outputDirectory = "";
+  std::string inputVideoFileName = "";
+  std::string outputHashFileName = "";
+  std::string outputDirectory = "";
   double secondsPerHash = 0;
   int downsampleFrameDimension = 0;
 
   while ((argi < argc) && argv[argi][0] == '-') {
-    string flag(argv[argi++]);
+    std::string flag(argv[argi++]);
```
Dcallies, line 81: I wonder why we didn't use gflags?

vpdq/cpp/bin/vpdq-hash-video.cpp
```diff
@@ -90,42 +87,42 @@ int main(int argc, char** argv) {
...
     }
     if (flag == "-o" || flag == "--output-hash-file-name") {
       if ((argc - argi) < 1) {
         usage(argv[0], 1);
       }
-      outputHashFileName = string(argv[argi++]);
+      outputHashFileName = std::string(argv[argi++]);
       continue;
     }
     if (flag == "-f" || flag == "--ffmpeg-path") {
```
Dcallies, line 100: I'm also open to nuking the option

vpdq/cpp/hashing/filehasher.h
```diff
@@ -18,27 +18,22 @@ namespace hashing {
...
  * @param secondsPerHash The time period of picking frames in vpdq
- * @param width Specified width of the input video
- * @param height Specified height of the input video
- * @param framesPerSec Frames per second of the input video
- * @param downsampleFrameDimension The down-scale dimensions for vpdq
- * @param programName The name of executable program which invokes the function
+ * @param downsampleWidth Width to downsample to before hashing. 0 means no
+ * downsample
+ * @param downsampleHeight Height to downsample to before hashing. 0 means no
+ * downsample
```
Dcallies, line 26: Dumb question - is there ever a reason for these values to be different? PDQ wants a square aspect ratio at the end, of I fixed size IIRC

vpdq/python/vpdq.pyx
```diff
@@ -125,12 +98,12 @@ def hamming_distance(hash1: "Hash256", hash2: "Hash256") -> int:
 
 def computeHash(
     input_video_filename: t.Union[str, Path],
-    ffmpeg_path: str = "ffmpeg",
+    ffmpeg_path: t.Union[str, None] = None,
```
Dcallies, line 101: Question for you - I had a devil of a time getting any python tooling to run on the .pyx file. Did you get typechecking to run properly, or just keeping them updated by eye?

Dcallies (approved):
I haven't hard of any users that are depending on these flags, and it's early enough in the library's life that we might be able to get away with these scott free.

Once we've finished all the changes, we should release a new version of the pypy library for vpdq

vpdq/cpp/hashing/filehasher.h
ianwal: Not sure. I'll read the PDQ doc and if not I'll change it.

vpdq/python/vpdq.pyx
ianwal: I've just been keeping them updated by eye. Some of them could be wrong, but I doubt it.

PR #1338, opened 2023-07-21
[vpdq] Add multithreaded hashing

ianwal:
Summary
---------

Decoding is already optionally multithreaded by libav*, but it was limited by the hash speed because hashing was blocking. 

This provides a nice performance boost proportional to the number of frames in the video and number of CPU cores.

It spawns consumer threads before decoding any frames, then the decoder copies frames to a queue, and the consumer threads take frames from the queue and hash them and fill the vector passed in to be returned. The threads can return frames out of order, so they are sorted by frame number before returning. Sorting is probably faster than the alternative, which is having a separate buffered unordered queue with a thread that inserts hashes as they come in in order.

Also made some general changes to the hasher:
* Wrapped AVFrame* in a unique ptr with a custom destructor to avoid complicated freeing.
* Moved most of the av stuff into an AVVideo class

I also might add an option to set the number of threads as a function parameter to `hashVideoFile()` and for the CLI in vpdq-hash-video. Functionality is the same as before if `num_consumers` variable in class `vpdqHasher` is set to 1.

Test Plan
---------

Compiled and tested with `-fsanitize=address` to ensure there's no memory leaks during regular usage.

I tested on some long video files and memory usage is low.

I tested the Python bindings using its tests and it works. But, the Python CI test doesn't run even if the C++ source code is changed. This should probably be fixed. I had errors happen that only occurred during the Python tests even if the C++ regtest passed.

I have not tested for memory leaks during odd errors and unrecoverable errors such as bad_alloc exceptions. Anything that throws bad_alloc just crashes the program. But I tried to design it such that regular runtime exceptions could be caught.

Added reg test to make sure frame numbers are positive and strictly increasing. I tested the test to make sure it does indeed fail if they are out of order or two frames have the same frame number.

<details>

<summary>Example regtest run for non-increasing:</summary>

```
  File "/home/ianwal/ThreatExchange/vpdq/cpp/regtest.py", line 242, in <module>
    main()
  File "/home/ianwal/ThreatExchange/vpdq/cpp/regtest.py", line 236, in main
    assert (oldFrameNumber + 1) == frameNumber
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

</details>

### DIFF
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -300,11 +300,9 @@ class vpdqHasher {
             targetFrame->data,
             targetFrame->linesize);
 
-        std::unique_lock<std::mutex> lock(queue_mutex);
+        std::lock_guard<std::mutex> lock(queue_mutex);
         FatFrame fatFrame{std::move(targetFrame), frameNumber};
-
         hash_queue.push(std::move(fatFrame));
-        lock.unlock();
         queue_condition.notify_one();
       }
       frameNumber += 1;
@@ -329,22 +327,22 @@ class vpdqHasher {
           std::string(
               " Frame width or height smaller than the minimum hashable dimension"));
     }
+    if (verbose) {
+      std::cout << "PDQHash: " << pdqHash.format() << std::endl;
+    }
 
     // Write frame to file here for debugging:
     // This is not thread safe. Use one thread when writing to file
     // saveFrameToFile(frame, "frame.rgb");
 
     // Append vpdq feature to pdqHashes vector
-    std::lock_guard<std::mutex> lock(pdqHashes_mutex);
     vpdqFeature feature = {
         pdqHash,
         fatFrame.frameNumber,
         quality,
         static_cast<double>(fatFrame.frameNumber) / video->frameRate};
+    std::lock_guard<std::mutex> lock(pdqHashes_mutex);
     pdqHashes.push_back(feature);
-    if (verbose) {
-      std::cout << "PDQHash: " << pdqHash.format() << std::endl;
-    }
   }
 
   void consumer() {
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -405,7 +405,7 @@ bool hashVideoFile(
   }
 
   // Read frames in a loop and process them
-  int ret;
+  int ret = 0;
   int frameNumber = 0;
   bool failed = false;
   while (av_read_frame(formatContext, base_packet) == 0) {
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -2,9 +2,9 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 // ================================================================
 
-#include <math.h>
 #include <algorithm>
 #include <cassert>
+#include <cmath>
 #include <condition_variable>
 #include <cstdio>
 #include <fstream>
@@ -398,7 +398,7 @@ bool hashVideoFile(
     av_log_set_level(AV_LOG_FATAL);
   }
 
-  std::unique_ptr<AVVideo> video = nullptr;
+  std::unique_ptr<AVVideo> video;
   try {
     video = std::make_unique<AVVideo>(inputVideoFileName);
   } catch (const std::runtime_error& e) {
--- a/vpdq/cpp/hashing/filehasher.cpp
+++ b/vpdq/cpp/hashing/filehasher.cpp
@@ -35,6 +35,9 @@ namespace facebook {
 namespace vpdq {
 namespace hashing {
 
+// Pixel format for the image passed to PDQ
+constexpr AVPixelFormat pixelFormat = AV_PIX_FMT_RGB24;
+
 /* @brief Writes an AVFrame to a file
  *
  * Useful for debugging.
@@ -69,6 +72,113 @@ static void saveFrameToFile(AVFrame* frame, const char* filename) {
   outfile.close();
 }
 
+class AVVideo {
+ public:
+  AVCodecContext* codecContext = nullptr;
+  SwsContext* swsContext = nullptr;
+  AVFormatContext* formatContext = nullptr;
+  int videoStreamIndex = -1;
+  int width;
+  int height;
+  double frameRate;
+
+  AVVideo(const std::string filename) {
+    // Open the input file
+    if (avformat_open_input(
+            &formatContext, filename.c_str(), nullptr, nullptr) != 0) {
+      throw std::runtime_error("Cannot open video");
+    }
+
+    // Retrieve stream information
+    if (avformat_find_stream_info(formatContext, nullptr) < 0) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("Cannot find video stream info");
+    }
+
+    // Find the first video stream
+    for (unsigned int i = 0; i < formatContext->nb_streams; ++i) {
+      if (formatContext->streams[i]->codecpar->codec_type ==
+          AVMEDIA_TYPE_VIDEO) {
+        videoStreamIndex = i;
+        break;
+      }
+    }
+
+    if (videoStreamIndex == -1) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("No video stream found");
+    }
+
+    // Get the video codec parameters
+    AVCodecParameters* codecParameters =
+        formatContext->streams[videoStreamIndex]->codecpar;
+
+    width = codecParameters->width;
+    height = codecParameters->height;
+    if (width == 0 || height == 0) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("Width or height equals 0");
+    }
+
+    // Find the video decoder
+    const AVCodec* codec = avcodec_find_decoder(codecParameters->codec_id);
+    if (!codec) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("Video codec id not found");
+    }
+
+    // Create the codec context
+    codecContext = avcodec_alloc_context3(codec);
+    if (avcodec_parameters_to_context(codecContext, codecParameters) < 0) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("Cannot copy codec parameters to context");
+    }
+
+    // Determine the number of threads to use and multithreading type
+    codecContext->thread_count = 0;
+
+    if (codec->capabilities & AV_CODEC_CAP_FRAME_THREADS) {
+      codecContext->thread_type = FF_THREAD_FRAME;
+    } else if (codec->capabilities & AV_CODEC_CAP_SLICE_THREADS) {
+      codecContext->thread_type = FF_THREAD_SLICE;
+    } else {
+      codecContext->thread_count = 1;
+    }
+
+    // Open the codec context
+    if (avcodec_open2(codecContext, codec, nullptr) < 0) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("Cannot open video codec context");
+    }
+
+    // Get the framerate
+    AVRational avframeRate =
+        formatContext->streams[videoStreamIndex]->avg_frame_rate;
+
+    // if avg_frame_rate is 0, fall back to r_frame_rate which is the
+    // lowest framerate with which all timestamps can be represented accurately
+    if (avframeRate.num == 0 || avframeRate.den == 0) {
+      avframeRate = formatContext->streams[videoStreamIndex]->r_frame_rate;
+    }
+
+    frameRate = static_cast<double>(avframeRate.num) /
+        static_cast<double>(avframeRate.den);
+    if (frameRate == 0) {
+      avformat_close_input(&formatContext);
+      throw std::runtime_error("Video framerate is zero");
+    }
+  }
+  ~AVVideo() {
+    if (codecContext != nullptr)
+      avcodec_free_context(&codecContext);
+    if (swsContext != nullptr)
+      sws_freeContext(swsContext);
+    if (formatContext != nullptr) {
+      avformat_close_input(&formatContext);
+    }
+  }
+};
+
 class vpdqHasher {
  public:
   struct FatFrame {
@@ -84,13 +194,12 @@ class vpdqHasher {
   std::mutex done_mutex;
   bool done_hashing = false;
   int num_consumers = std::thread::hardware_concurrency();
-  AVCodecContext* codecContext;
-  SwsContext* swsContext;
-  AVFormatContext* formatContext;
-  int width;
-  int height;
-  double frameRate;
+
   int frameMod;
+
+  std::unique_ptr<AVVideo> video;
+
+  vpdqHasher(std::unique_ptr<AVVideo> video) : video(std::move(video)) {}
 
   // Decode and add vpdqFeature to the hashes vector
   // Returns the number of frames processed (this is can be more than 1!)
@@ -98,10 +207,10 @@ class vpdqHasher {
     AVFrame* frame = av_frame_alloc();
     assert(frame != nullptr);
     // TODO: check for frame good alloc
-    AVFrame* targetFrame = createFrame(width, height);
+    AVFrame* targetFrame = createFrame(video->width, video->height);
     assert(targetFrame != nullptr);
     // Send the packet to the decoder
-    int ret = avcodec_send_packet(codecContext, packet) < 0;
+    int ret = avcodec_send_packet(video->codecContext, packet) < 0;
     // std::cout << codecContext->frame_num << std::endl;
     if (ret < 0) {
       throw std::runtime_error("Cannot send packet to decoder");
@@ -109,7 +218,7 @@ class vpdqHasher {
 
     // Receive the decoded frame
     while (ret >= 0) {
-      ret = avcodec_receive_frame(codecContext, frame);
+      ret = avcodec_receive_frame(video->codecContext, frame);
       if (ret == AVERROR(EAGAIN) || ret == AVERROR_EOF) {
         break;
       } else if (ret < 0) {
@@ -119,16 +228,16 @@ class vpdqHasher {
       if (frameNumber % frameMod == 0) {
         // Resize the frame and convert to RGB24
         sws_scale(
-            swsContext,
+            video->swsContext,
             frame->data,
             frame->linesize,
             0,
-            codecContext->height,
+            video->codecContext->height,
             targetFrame->data,
             targetFrame->linesize);
 
         std::unique_lock<std::mutex> lock(queue_mutex);
-        AVFrame* newTargetFrame = createFrame(width, height);
+        AVFrame* newTargetFrame = createFrame(video->width, video->height);
         av_frame_copy(newTargetFrame, targetFrame);
         FatFrame fatFrame = {newTargetFrame, frameNumber};
 
@@ -145,9 +254,6 @@ class vpdqHasher {
   }
 
   AVFrame* createFrame(int width, int height) {
-    // Pixel format for the image passed to PDQ
-    constexpr AVPixelFormat pixelFormat = AV_PIX_FMT_RGB24;
-
     // Create a frame for resizing and converting the decoded frame to RGB24
     AVFrame* targetFrame = av_frame_alloc();
     if (targetFrame == nullptr) {
@@ -205,7 +311,7 @@ class vpdqHasher {
         pdqHash,
         frameNumber,
         quality,
-        static_cast<double>(frameNumber) / frameRate};
+        static_cast<double>(frameNumber) / video->frameRate};
     pdqHashes.push_back(feature);
     if (verbose) {
       std::cout << "PDQHash: " << pdqHash.format() << std::endl;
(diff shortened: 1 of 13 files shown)


Dcallies:
Oh man, now we are in hard engineering!

The first thing I looked for in your PR was usage of omp (https://bisqwit.iki.fi/story/howto/openmp/), which I learned about as part of tmk work by @dxdc in prs like #959. It helps with giving your architectural-specific tools and giving you a bunch of helpers for common patterns. 

The other thing I'll be looking for is whether there is good separation from the single threaded version - multithreading can add a lot of complexity, and a source of many subtle bugs.

vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -62,74 +92,285 @@ static void saveFrameToFile(AVFrame* frame, const char* filename) {
...
+    for (unsigned int i = 0; i < formatContext->nb_streams; ++i) {
+      if (formatContext->streams[i]->codecpar->codec_type ==
+          AVMEDIA_TYPE_VIDEO) {
+        videoStreamIndex = i;
+        break;
       }
+    }
+
+    if (videoStreamIndex == -1) {
+      avformat_close_input(&formatContext);
```
Dcallies, line 146: Hmm, I wonder if separating construction and opening will let you lean on your destructor for correctness, otherwise you have to carefully partially undo everything allocated.

Another solution might be to wrap the pointers with another object with a destructor, and if the constructor fails it can still destruct the inner object normally.  

Something I read while looking at this: https://isocpp.org/wiki/faq/exceptions#selfcleaning-members

Dcallies (commented):
Oh man, github failures at my review! 

Trying to peproduce some of the comments from memory, only a partial review:
1. I like your separation of concerns on the object layer: video, hasher, frame seem like good cut points
2. The usage of raw pointers we should try and avoid if possible. I had some inline comments that had some suggested, but I don't see them anymore, but tl;dr: wrap the pointers in the hasher with a second object with a destructor to avoid needing to manage the pointers from thrown exceptions in the constructor
3. I need to read more of the docs, but we need a way to control overall thread use - it's okay if the default is unconstrained, but we need a way to limit to X threads (where one version is X=1).

ianwal:
> The first thing I looked for in your PR was usage of omp (https://bisqwit.iki.fi/story/howto/openmp/), which I learned about as part of tmk work by @dxdc in prs like #959. It helps with giving your architectural-specific tools and giving you a bunch of helpers for common patterns.

I looked a little into OpenMP, but it seemed a bit unnecessary from what I read (I could be very wrong!). I read one of the major benefits from that is fast thread creation, but I'm creating my threads at the start of hashing and reusing them. I'm not sure how much more optimized the threading could get, but I'm not too knowledgeable about distributed programming in general.

> The other thing I'll be looking for is whether there is good separation from the single threaded version - multithreading can add a lot of complexity, and a source of many subtle bugs.

Yep, I made sure to make it easy to separate. Just comment out the bottom locked section in processFrame and make it call hasher directly. This could be done by the preprocessor.

in processFrame():
```cpp
...
            targetFrame->linesize);

        // single threaded
        FatFrame fatFrame{std::move(targetFrame), frameNumber};
        hasher(std::move(fatFrame));
        // multithreaded
        /* 
         * std::lock_guard<std::mutex> lock(queue_mutex);
         * FatFrame fatFrame{std::move(targetFrame), frameNumber};
         * hash_queue.push(std::move(fatFrame));
         * queue_condition.notify_one();
         */
      }
      frameNumber += 1;

```

The core loop is very much the same as the original. I thought about limiting the queue_size to the number of threads on a hunch that it might reduce context switching, but I haven't tried implementing that. It would also be useful to not blow up memory if for whatever reason the decoder was much faster than hashing, but that didn't seem very realistic to me for IO and computational complexity reasons.

By the way, multithreaded is not much faster if the results are downsampled since hashing speed depends on the size of the image and decoding can take much longer depending on the codec (h265 ouch!). If hardware decoding is used that could may change but that's definitely not a given.

vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -153,271 +394,117 @@ bool hashVideoFile(
...
     std::cerr << "Cannot create sws context" << std::endl;
-    av_freep(targetFrame->data);
-    av_frame_free(&targetFrame);
-    av_frame_free(&frame);
-    avcodec_free_context(&codecContext);
-    avformat_close_input(&formatContext);
     return false;
   }
 
   AVPacket* packet = av_packet_alloc();
```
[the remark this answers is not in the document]
ianwal: Yes, that's a good idea. But packet's are not entirely made simple by the RAII since I'm reusing the same one for reading frames and each iteration its buffer MUST be freed with av_packet_unref(packet); so that might complicate things. I think I might have to use a shared_ptr.

vpdq/cpp/hashing/filehasher.cpp
ianwal: I could move constructor to an init, but I have a hunch that might not be a good idea. An AVVideo instance shouldn't exist without belonging to an opened video.

It might just be more readable, although verbose, to just free everything manually. The open video sequence is very straightforward.

If you move the complexity outside of the class with init, then that responsibility falls on them to handle it. 

Wrapping the pointers would be a must if they were used at all outside of the AVVideo class other than reads, but since they're not right now I think it would just add more complexity than it would solve.

Your call though, either way makes sense.

Dcallies: I think this might not be too much work:

## Option 1: Swap all pointers to smartpointers with custom deallocs
Before:
```
AVCodecContext* codecContext = nullptr;
SwsContext* swsContext = nullptr;
AVFormatContext* formatContext = nullptr;
```
After:
```
# lots of custom desctructors
unique_ptr<AVCodecContext> codecContext = nullptr;
unique_ptr<SwsContext> swsContext = nullptr;
unique_ptr<AVFormatContext> formatContext = nullptr;
```

## Option 2: Inner object to manage pointers:
class AVContext {
  AVCodecContext* codecContext = nullptr;
  SwsContext* swsContext = nullptr;
  AVFormatContext* formatContext = nullptr;

  open() {
    codecContext = alloc(...) || throw()
    swsContext = alloc(...) || throw()
    formatContext = alloc(...) || throw()
  }

  ~Foo(){
    if (codecContext != nullptr) {
      dealloc(...)
    }
   if (swsContext != nullptr) {
      dealloc(...)
    }
    ...
  }
}

class AVVideo {
 public:
  AVContext context{}; // Might be forgetting the exact syntax

  AVVideo() {
    // any exceptions will lead to dealloc called on context, so don't have to manage order
    context.open();
  }

vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -62,322 +104,408 @@ static void saveFrameToFile(AVFrame* frame, const char* filename) {
...
-    return false;
+    // Determine the number of threads to use and multithreading type
+    codecContext->thread_count = 0;
+
+    if (codec->capabilities & AV_CODEC_CAP_FRAME_THREADS) {
+      codecContext->thread_type = FF_THREAD_FRAME;
+    } else if (codec->capabilities & AV_CODEC_CAP_SLICE_THREADS) {
+      codecContext->thread_type = FF_THREAD_SLICE;
+    } else {
+      codecContext->thread_count = 1;
```
Dcallies, line 200: Dumb question - should we be passing the thread count passed in to the hash function to AV?

vpdq/cpp/regtest.py
```diff
@@ -127,96 +128,125 @@ def main():
...
-            except subprocess.CalledProcessError as e:
-                print(" ".join([str(i) for i in e.cmd]))
-                print(str(e.stderr, "utf-8"))
-                sys.exit(1)
-
-            # Copy hash files to output directory if it is specified
-            # This will overwrite existing files with the same
-            # name as outputHashFile in the directory
+    # Run the hashing and matching tests for single and multithreaded
+    for thread_count in range(0, 2):
```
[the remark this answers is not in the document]
ianwal: Yeah... this regtest and CLI programs should probably be rewritten from the ground up, but I thought adding this was better than nothing for now.

vpdq/cpp/hashing/filehasher.cpp
ianwal: I was on the fence. I'm pretty sure that for decoding it won't spawn a ton of threads because that's not very helpful for decoding, but if you want me to I can pass it. I haven't tried setting it to anything but 0 or 1 though so I would need to do some testing.

ianwal:
Okay, I made wrapped the rest of the av stuff in smart pointers. No memory leaks tested with -fsanitize=address.

Only thing left is deciding if decoding thread count should be set to the thread_count parameter.

ianwal:
I'm getting a data race warning from `--fsanitize=thread`. Working a fix. I'm going to make the CI tests run with several `--fsanitize` flags.


<details>

<summary>Data race warning:</summary>

```
==================
WARNING: ThreadSanitizer: data race (pid=6752)
  Write of size 8 at 0x7b0400005000 by thread T27:
    #0 pipe /usr/src/debug/gcc/gcc/libsanitizer/tsan/tsan_interceptors_posix.cpp:1814 (libtsan.so.2+0x4760f) (BuildId: 773101dc26300568d4d7039dc84897e22bbc2ca0)
    #1 __sanitizer::IsAccessibleMemoryRange(unsigned long, unsigned long) /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_posix_libcdep.cpp:276 (libubsan.so.1+0x22273) (BuildId: cff607e0df2cb7f60ed900a621e70fce450eddd0)
    #2 std::default_delete<std::thread::_State>::operator()(std::thread::_State*) const /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/unique_ptr.h:99 (libstdc++.so.6+0xe194b) (BuildId: a4da590505ed6ce81e1e49a17dc233f20f9da07c)
    #3 std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >::~unique_ptr() /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/unique_ptr.h:404 (libstdc++.so.6+0xe194b)
    #4 execute_native_thread_routine /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/thread.cc:106 (libstdc++.so.6+0xe194b)

  Previous write of size 8 at 0x7b0400005000 by thread T28:
    #0 pipe /usr/src/debug/gcc/gcc/libsanitizer/tsan/tsan_interceptors_posix.cpp:1814 (libtsan.so.2+0x4760f) (BuildId: 773101dc26300568d4d7039dc84897e22bbc2ca0)
    #1 __sanitizer::IsAccessibleMemoryRange(unsigned long, unsigned long) /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_posix_libcdep.cpp:276 (libubsan.so.1+0x22273) (BuildId: cff607e0df2cb7f60ed900a621e70fce450eddd0)
    #2 std::default_delete<std::thread::_State>::operator()(std::thread::_State*) const /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/unique_ptr.h:99 (libstdc++.so.6+0xe194b) (BuildId: a4da590505ed6ce81e1e49a17dc233f20f9da07c)
    #3 std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >::~unique_ptr() /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/unique_ptr.h:404 (libstdc++.so.6+0xe194b)
    #4 execute_native_thread_routine /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/thread.cc:106 (libstdc++.so.6+0xe194b)

  Thread T27 (tid=6780, running) created by main thread at:
    #0 pthread_create /usr/src/debug/gcc/gcc/libsanitizer/tsan/tsan_interceptors_posix.cpp:1036 (libtsan.so.2+0x44219) (BuildId: 773101dc26300568d4d7039dc84897e22bbc2ca0)
    #1 __gthread_create /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/x86_64-pc-linux-gnu/bits/gthr-default.h:663 (libstdc++.so.6+0xe1a29) (BuildId: a4da590505ed6ce81e1e49a17dc233f20f9da07c)
    #2 std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)()) /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/thread.cc:172 (libstdc++.so.6+0xe1a29)
    #3 facebook::vpdq::hashing::hashVideoFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<facebook::vpdq::hashing::vpdqFeature, std::allocator<facebook::vpdq::hashing::vpdqFeature> >&, bool, double, int, int, unsigned int) <null> (vpdq-hash-video+0x563b2) (BuildId: 4fa2c63fb16850bec2895a73db56c45b11c3881c)
    #4 main <null> (vpdq-hash-video+0x4925c) (BuildId: 4fa2c63fb16850bec2895a73db56c45b11c3881c)

  Thread T28 (tid=6781, finished) created by main thread at:
    #0 pthread_create /usr/src/debug/gcc/gcc/libsanitizer/tsan/tsan_interceptors_posix.cpp:1036 (libtsan.so.2+0x44219) (BuildId: 773101dc26300568d4d7039dc84897e22bbc2ca0)
    #1 __gthread_create /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/x86_64-pc-linux-gnu/bits/gthr-default.h:663 (libstdc++.so.6+0xe1a29) (BuildId: a4da590505ed6ce81e1e49a17dc233f20f9da07c)
    #2 std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)()) /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/thread.cc:172 (libstdc++.so.6+0xe1a29)
    #3 facebook::vpdq::hashing::hashVideoFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<facebook::vpdq::hashing::vpdqFeature, std::allocator<facebook::vpdq::hashing::vpdqFeature> >&, bool, double, int, int, unsigned int) <null> (vpdq-hash-video+0x563b2) (BuildId: 4fa2c63fb16850bec2895a73db56c45b11c3881c)
    #4 main <null> (vpdq-hash-video+0x4925c) (BuildId: 4fa2c63fb16850bec2895a73db56c45b11c3881c)

SUMMARY: ThreadSanitizer: data race /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_posix_libcdep.cpp:276 in __sanitizer::IsAccessibleMemoryRange(unsigned long, unsigned long)
==================
ThreadSanitizer: reported 1 warnings
```

</details>

ianwal:
Man... it's odd that an abstract class and factory method were added in `bufferhasher` for such a small program. Adding support for other image phashers as they're invented seems more logical to me (is there even any other than wHash, pHash, and PDQ?). I'm guessing this is a remnant from your internal implementation?

ianwal:
I messed around with OpenMP and it's possible to get a huge speed gain out of pdq and vpdq without doing any manual multithreading.

Just adding `#pragma omp parallel for` in `boxAlongRowsFloat` and `boxAlongColsFloat` I achieved a huge speed boost and it fully saturated my CPU:
```cpp
void boxAlongRowsFloat(
    const float* in, // matrix as num_rows x num_cols in
                     // row-major order
    float* out, // matrix as num_rows x num_cols in row-major
                // order
    int numRows,
    int numCols,
    int window_size) {
#pragma omp parallel for
  for (int i = 0; i < numRows; i++) {
    box1DFloat(&in[i * numCols], &out[i * numCols], numCols, 1, window_size);
  }
}
```

So if you want a huge performance gain for PDQ without changing anything major there it is.

But, from my tests it's still not as good as multithreading vpdq hashing for most videos:

regtest.py Ryzen 5800x, x86-64 Linux 6.4.4:
```
multithread
real    0m10.646s
user    1m35.586s
sys     0m15.485s

multithread pdq by openmp (no queue)
real    0m17.288s
user    4m12.790s
sys     0m2.892s

singlethread
real    0m27.139s
user    0m33.989s
sys     0m1.897s
```

I think unless there is a significant gap I think it's better to multithread vpdq because it's easier to control the number of threads and appears to be faster.

My multithreading uses more allocs, but it can call `fillFloatLumaFromRGB()` and the rest of the functions for hashing in multiple threads which seems to make up the portion not spent on the above functions.

perf report:
```c
Samples: 689K of event 'cycles:u', Event count (approx.): 732410451613
Overhead  Command          Shared Object            Symbol
  64.58%  vpdq-hash-video  vpdq-hash-video          [.] facebook::pdq::downscaling::box1DFloat
   6.65%  vpdq-hash-video  vpdq-hash-video          [.] facebook::pdq::hashing::fillFloatLumaFromRGB
   1.32%  vpdq-hash-video  libc.so.6                [.] 0x000000000015758e
   1.28%  vpdq-hash-video  libc.so.6                [.] 0x0000000000157584
   0.66%  vpdq-hash-video  libswscale.so.7.1.100    [.] 0x000000000007a6a9
...

File: vpdq/cpp/hashing/filehasher.cpp (hunk 3 above)
Comment (Dcallies) [reply vpdq/cpp/hashing/filehasher.cpp:213]: I'm happy to leave as a future improvement if it doesn't seem like the answer is straightforward

File: vpdq/cpp/hashing/filehasher.cpp [hunk 5]
```diff
@@ -386,38 +523,39 @@ bool hashVideoFile(
...
   }
 
-  av_packet_free(&packet);
-  sws_freeContext(swsContext);
-  av_freep(targetFrame->data);
-  av_frame_free(&targetFrame);
-  av_frame_free(&frame);
-  avcodec_free_context(&codecContext);
-  avformat_close_input(&formatContext);
+  if (thread_count != 1) {
```
Dcallies, line 533: nit, ignorable: In some cases you use the multithreaded variable, in some cases you use this check. Seems like consistently using this  version of the check might be less error prone than using the separate `multithreaded` variable.

vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -62,322 +109,412 @@ static void saveFrameToFile(AVFrame* frame, const char* filename) {
...
+ public:
+  struct FatFrame {
+    AVFramePtr frame;
+    int64_t frameNumber;
+  };
+
+  std::condition_variable queue_condition;
+  std::mutex queue_mutex;
+  std::queue<FatFrame> hash_queue;
+  std::atomic<bool> done_hashing{false};
```
Dcallies, line 267: You are using std::atomic for this variable, but everywhere you are using it, you are holding the queue_mutex, so you don't need an std::atomic.

Furthermore, I think you are relying on mutex to make sure the condition variable reads correctly (it's set before notification), so I think better to convert to just a bool to prevent a future author thinking its threadsafe to use as is.

There was a pattern I used in a previous project that wrapped variables in a mutex which only made them accessible if the mutex was held, which make things like this much less error prone, but I can't find it in the core CPP libraries, so it must have been bespoke.

vpdq/cpp/hashing/filehasher.cpp
```diff
@@ -62,322 +109,412 @@ static void saveFrameToFile(AVFrame* frame, const char* filename) {
...
+  void consumer() {
+    while (true) {
+      std::unique_lock<std::mutex> lock(queue_mutex);
+      queue_condition.wait(
+          lock, [this] { return !hash_queue.empty() || done_hashing; });
+      if (hash_queue.empty() && done_hashing)
+        break;
+      FatFrame fatFrame(std::move(hash_queue.front()));
+      hash_queue.pop();
+      lock.unlock();
```
Dcallies, line 416: (unsure, alt): If you add one level of scoping you won't need the explicit unlock, i.e.

```
  { 
      std::unique_lock<std::mutex> lock(queue_mutex);
      queue_condition.wait(
          lock, [this] { return !hash_queue.empty() || done_hashing; });
      if (hash_queue.empty() && done_hashing)
        return;
      FatFrame fatFrame(std::move(hash_queue.front()));
      hash_queue.pop();
  }
  hasher(std::move(fatFrame));
```

Dcallies (approved):
> Man... it's odd that an abstract class and factory method were added in bufferhasher for such a small program. Adding support for other image phashers as they're invented seems more logical to me (is there even any other than wHash, pHash, and PDQ?). I'm guessing this is a remnant from your internal implementation?

No idea - original original author was a java refugee, so abstract factory patterns maybe was just standard fare.

Agree that someday would love a standard model for image-based hashing algorirthms. vPDQ's hash strategy & lookup works for any image-based hash algorithm.

---

If all your correctness checks are passing, then I think you are good to go. Thanks for including threadsan as part of your approach, and continuing to add error checking on the compiles, that definitely makes me more confident in your approach.

One or two comments, leaving open to you on whether to adjust them or merge now.

ianwal:
There's still a data race somewhere, probably relating to the decode packet.

CI tests weren't working properly because the linker didn't have the sanitizers when passed from CLI. When I added the sanitizers directly to CMakeLists.txt locally they caught the data race, which is strange.

ianwal:
Ready to merge

Dcallies (approved):
Changes look good, merging

PR #1339, opened 2023-07-21
[vpdq] Add downsample test hashes

ianwal:
Summary
---------

Add 128x128 downsample hashes to utilize the previously added downsample regression test.

Generated using `generate_hashes.py` on the main branch.

Test Plan
---------

Test with videos and see that when downsampled they pass.


Dcallies (approved):
Thanks for doing this!

PR #1341, opened 2023-08-04
Add Postgres to dev environment

dougneal:
Summary
---------

Switch the devcontainer config to use docker-compose to bring up a Postgres container in addition to the Flask app container.

Test Plan
---------

Rebuild dev container environment in VSCode

### DIFF
--- a/open-media-match/.devcontainer/omm_config.py
+++ b/open-media-match/.devcontainer/omm_config.py
@@ -1,0 +1,6 @@
+# Database configuration
+DBUSER = 'media_match'
+DBPASS = 'hunter2'
+DBHOST = 'db'
+DBNAME = 'media_match'
+DATABASE_URI = f"postgresql+psycopg2://{DBUSER}:{DBPASS}@{DBHOST}/{DBNAME}"
--- a/open-media-match/src/OpenMediaMatch/__init__.py
+++ b/open-media-match/src/OpenMediaMatch/__init__.py
@@ -1,5 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 
+import os
 import flask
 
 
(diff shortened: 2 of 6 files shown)


open-media-match/.devcontainer/omm_config.py
```diff
@@ -0,0 +1,6 @@
+# Database configuration
+DBUSER = 'media_match'
+DBPASS = 'hunter2'
```
Dcallies, line 3: lul

open-media-match/.devcontainer/omm_config.py
```diff
@@ -0,0 +1,6 @@
+# Database configuration
```
Dcallies, line 1: I'm inherently suspicious of turing complete configuration files, but I think this is fine in the short term. 

I've read enough about YAML to be scared, and tend towards just JSON, but this is all irrelevant in first pass.

open-media-match/src/OpenMediaMatch/__init__.py
```diff
@@ -1,13 +1,15 @@
...
 import flask
 
 
 def create_app():
     """
     Create and configure the Flask app
     """
-    app = flask.Flask(__name__, instance_relative_config=True)
+    app = flask.Flask(__name__)
+    app.config.from_envvar('OMM_CONFIG')
```
Dcallies, line 12: (ignorable): I think having this be a harcoded path at this stage is also fine, one less indirection to follow

Dcallies (approved):
This all looks fine, some popcorn gallery comments that are ignorable. Up to you to merge as is.

open-media-match/.devcontainer/omm_config.py
dougneal: Fair point, and Flask provides a few methods by which to load config.

PR #1343, opened 2023-08-09
OMM: Roles and blueprints

dougneal:
Summary
---------

Define the hasher, matcher and curator roles. Allow them to be enabled or disabled via config. Separate those role functions out into discrete Flask "Blueprints" with URL prefixes for easy Layer 7 routing. Add stubs for a handful of example functions from the README.

The index endpoint ('/') is distinct in function from '/status',  so separate it out. Add a basic HTML page here so that user is greeted with something human-readable to show they're up and running.

Test Plan
---------

Local `curl` commands

### DIFF
--- a/open-media-match/.devcontainer/omm_config.py
+++ b/open-media-match/.devcontainer/omm_config.py
@@ -5,3 +5,7 @@ DBPASS = 'hunter2'
 DBHOST = 'db'
 DBNAME = 'media_match'
 DATABASE_URI = f"postgresql+psycopg2://{DBUSER}:{DBPASS}@{DBHOST}/{DBNAME}"
+
+ROLE_HASHER = True
+ROLE_MATCHER = True
+ROLE_CURATOR = True
--- a/open-media-match/.devcontainer/omm_config.py
+++ b/open-media-match/.devcontainer/omm_config.py
@@ -1,4 +1,5 @@
 # Database configuration
+PRODUCTION = False
 DBUSER = 'media_match'
 DBPASS = 'hunter2'
 DBHOST = 'db'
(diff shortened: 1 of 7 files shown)


open-media-match/.devcontainer/omm_config.py
```diff
@@ -1,6 +1,11 @@
...
+PRODUCTION = False
 DBUSER = 'media_match'
 DBPASS = 'hunter2'
 DBHOST = 'db'
 DBNAME = 'media_match'
 DATABASE_URI = f"postgresql+psycopg2://{DBUSER}:{DBPASS}@{DBHOST}/{DBNAME}"
+
+ROLE_HASHER = True
+ROLE_MATCHER = True
+ROLE_CURATOR = True
```
Dcallies, line 11: (ignorable): The reason to split these into individual settings vs an enum is for the dev container to respond to all of these roles?

My initial reaction is if it was simpler to just have all roles have all APIs, and just route traffic based on roles, but this is simple and straightforward enough that I don't think that's needed.

open-media-match/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -0,0 +1,6 @@
+from flask import Blueprint
```
Dcallies, line 1: A future PR to explain what APIs should be in these files might be helpful

Dcallies (approved):
LGTM - think blueprints help with long term organization. 

Think we should put as little work as possible into status page, but it's nice to have it available.

dougneal:
> Think we should put as little work as possible into status page, but it's nice to have it available.

Agreed, I don't think it needs anything more at the moment. I'm thinking it will head off "is that it? is it working?" questions at the hackathon when people are setting up.

PR #1349, opened 2023-08-17
OMM: Bring in SQLAlchemy, get database working

dougneal:
Summary
---------

Bring in SQLAlchemy, Alembic, and the Flask integrations for them.

I've added a couple of models/dataclasses/tables - banks and hashes - really just something super basic to get us started, I expect this to change a lot.

There was some discussion about whether to use plain dataclasses with separate ORM classes; we can still do that if we want to.

Test Plan
---------

Rebuild the dev container and watch the tables be created!


Dcallies (approved):
I think the Model stuff influenced the later python dataclass design. 

Looks good

dougneal:
I should probably update the `CONTRIBUTING.md` with details of the db migrations etc and how to change the data model.

Issue #1351, opened 2023-08-18
[OMM] Set up basic Continuous Integration git workflows for OMM


Dcallies:
@juanmrad will be taking on (can't assign because I'm not smart enough)

1. py.test
2. black
3. mypy (or pyre if it's ready)

Feel free to make a case to add others

Dcallies:
Our crappy inexpert attempt at this was here: https://github.com/facebook/ThreatExchange/blob/main/.github/workflows/python-threatexchange-ci.yaml

PR #1352, opened 2023-08-24
[Open Media Matcher] GitHub Actions Setup

juanmrad:
Summary
---------

Adding lint and test dependencies to the project as well as initial test and GitHub Actions settings for the Open Media Matcher.

Test Plan
---------

Test run on CI now. 

cc/ #1351


open-media-match/pyproject.toml
```diff
@@ -19,6 +19,15 @@ dependencies = [
     "flask",
 ]
 
+[project.optional-dependencies]
```
Dcallies, line 22: Lol, @dougneal , who I think removed the all alias in a previous version.

To toss my hat in the ring:
1. I think it's helpful to separate test and production dependencies
2. Where I had landed on with Doug is that it would be fine to just have installing anything be the default
3. I am okay with having the test/all split brought back, but defer to @dougneal  for additional opinons.

open-media-match/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -0,0 +1,23 @@
...
+
+    yield app
+
+
+@pytest.fixture()
+def client(app):
+    return app.test_client()
+
+
+def test_status_response(client):
```
Dcallies, line 20: Ooo, nice, e2e tests

open-media-match/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -0,0 +1,23 @@
+import os
+import pytest
+
+from OpenMediaMatch import create_app
+
+
+@pytest.fixture()
```
Dcallies, line 7: <3 pytest fixtures

open-media-match/src/OpenMediaMatch/__init__.py
```diff
@@ -5,38 +5,40 @@
...
     """
     Create and configure the Flask app
     """
     app = flask.Flask(__name__)
-    app.config.from_envvar('OMM_CONFIG')
-    
-    @app.route('/')
+    app.config.from_envvar("OMM_CONFIG")
+
+    @app.route("/")
```
Dcallies, line 16: @dougneal  - this is going to create collisions with your outstanding diffs, but there's no way around it, and we should be innoculated going forward.

Dcallies (approved):
This looks good to me, thanks for doing it! 

Let me just check with Doug to make sure there isn't additional considerations or things he sees that I can't

open-media-match/src/OpenMediaMatch/tests/test_api.py
juanmrad: We can definitively have both e2e and unit ( which I prefer. We can also split the test jobs if it becomes slow), but given the current state e2e sounded the way to go for a good sample for now.

open-media-match/pyproject.toml
dougneal: I would definitely expect the dev mode dependencies to be different to the production mode dependencies, so this is expected in some form or another. 

The `pyproject.toml` way of doing things has a built-in concept of [build-time dependencies](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-time-dependencies).  This means these tools are listed under the `requires` section of the `build-system` stanza (above). This makes it more explicit that "these deps are for dev/build time" whereas putting them under optional dependencies is a bit more ambiguous in its intent.

The `pyproject.toml` can also contain configuration stanzas for each of these tools, see [this example](https://github.com/pamelafox/flask-db-quiz-example/blob/main/pyproject.toml).

open-media-match/src/OpenMediaMatch/tests/test_api.py
dougneal, line 1: Love that we're getting some tests in place early on.

open-media-match/src/OpenMediaMatch/__init__.py
dougneal: It's ok

open-media-match/.devcontainer/devcontainer.json
```diff
@@ -46,7 +46,7 @@
     }
   },
   "remoteUser": "vscode",
-  "postCreateCommand": "pip install --editable .",
+  "postCreateCommand": "pip install --editable .[all]",
```
dougneal, line 49: Not necessary if we use `build-system.requires` (see later comment)

open-media-match/mypy.ini
dougneal, line 1: It should be possible to roll this into a `tool` stanza of `pyproject.toml`.

dougneal (changes requested):
This is great, thanks! Just a few feedback points on the dependencies stuff.

.github/workflows/omm-ci.yaml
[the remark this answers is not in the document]
Dcallies: As far as I understand, workflows have to be under .github, and symlink support doesn't work on windows :/ 

https://docs.github.com/en/actions/using-workflows/about-workflows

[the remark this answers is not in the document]
juanmrad: yeah this will need to be inside the repositories root `.github` one of the downsides of monolith apps or repos with multiple projects.

open-media-match/pyproject.toml
juanmrad: but wouldn't adding them all to the `build-system` mean it will always be installed? we wouldn't want them installed on the prod container tho. unless you mean to create one for prod that overrides the dependencies with a smaller subset?

juanmrad: @dougneal I'm not sure I am able to follow how to get this working without the use of optional dependencies. I'm having a hard time understanding how to get the dependencies to be part of the build-system such as they are installed on `--editable` 😓 .

PR #1355, opened 2023-09-08
[omm] Hash api implementation first draft

Dcallies:
Summary
---------

Use the storage interfaces to pull the currently configured SignalType / ContentTypes, then download the content from the given URL and then process it.

Test Plan
---------

```bash
curl 'localhost:5000/h/hash?'\
'url=https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg?raw=true&'\
'content_type=photo'
```
Response:
```
{
  "pdq": "f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22"
}
```

### DIFF
--- a/open-media-match/src/OpenMediaMatch/__init__.py
+++ b/open-media-match/src/OpenMediaMatch/__init__.py
@@ -1,5 +1,8 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 
+import os
+import sys
+
 import flask
 import flask_migrate
 import flask_sqlalchemy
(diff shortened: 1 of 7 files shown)


open-media-match/src/OpenMediaMatch/__init__.py
```diff
@@ -16,7 +18,14 @@ def create_app():
...
     app = flask.Flask(__name__)
-    app.config.from_envvar("OMM_CONFIG")
+    if "OMM_CONFIG" in os.environ:
+        app.config.from_envvar("OMM_CONFIG")
+    elif sys.argv[0].endswith("/flask"):  # Default for flask CLI
+        # The devcontainer settings. If you are using the CLI outside
+        # the devcontainer and getting an error, just override the env
+        app.config.from_pyfile("/workspace/.devcontainer/omm_config.py")
+    else:
+        raise RuntimeError("No flask config given - try populating OMM_CONFIG env")
```
dougneal, line 28: Why was this needed?

open-media-match/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -10,18 +26,65 @@ def hash_media():
     Fetch content and return its hash.
     TODO: implement
     """
+
+    content_type = _parse_request_content_type()
```
dougneal, line 30: How about we get the content type from the response headers (below) and simplify the interface?
`download_resp.headers['content-type']`

dougneal (changes requested):

dougneal (approved):

Dcallies:
Discussed offline, doing content detection in a followup and removing as an argument for now.

PR #1357, opened 2023-09-08
[pytx] 1.0.11

Dcallies:
* Add the type checking marker


Chuwiey (approved):
a monumental change that looks great to me.

PR #1358, opened 2023-09-12
The computation of intGradientSum was found to be inaccurate.

rusan:
Summary
---------

Within the second loop, it consistently referenced the last $d value from the first loop, potentially causing adverse effects on images with a small intGradientSum, particularly when the final $d value from the first loop was substantial.


Dcallies (approved):
Thanks for the help here @rusan! Just sign the CLA and I will merge as is (though see note below on other tests you can try). 

Ooof, the lack of testing here is probably killing us - if you have some known imagery, you can compare with the cpp implementation to see if it's producing near results.

The test that we came up with (due to differences in the underlying os libraries for getting the image bytes out), that the same image with quality = 100 should produces a distance <= 8. 

So comparing the cpp implementation the php implementation on the same high-quality image should be within that threshold.

PR #1361, opened 2023-09-13
[omm][devcontainer] Add home mount for ease of access to local files

Dcallies:
Summary
---------

I keep wanting to touch a local install of python-threatexchange in my devcontainer, but it's not in the folder. I have my local checkout in my home, for this works for me, and this lets others do either something similar or add symlinks for workarounds.

Test Plan
---------

Rebuild devcontainer, navigate to /host-home-mount

### DIFF
--- a/open-media-match/.devcontainer/devcontainer.json
+++ b/open-media-match/.devcontainer/devcontainer.json
@@ -3,7 +3,10 @@
   "dockerComposeFile": "docker-compose.yaml",
   "service": "app",
   "workspaceFolder": "/workspace",
-  "forwardPorts": [8080, 5432],
+  "forwardPorts": [
+    8080,
+    5432
+  ],
   "customizations": {
     "vscode": {
       "extensions": [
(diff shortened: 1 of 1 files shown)


dougneal:
Problem with doing this kinda thing is that while it's convenient for iterating quickly it will reintroduce the kinds of "works on my machine" problems that containers are supposed to solve. If there is something specifically that needs to be accessible from within the dev container we should look at why that is and how to solve it sustainably.

dougneal:
This one makes me very nervous...

Dcallies:
1. It's not required by any actual functionality, and would agree to block anything that had a functional dependency on this
2. There's no workaround other than locally edit the devcontainer or not use this devcontainer
3. It should work for mac , linux, and windows.

In my research, mounting to local file system was the suggested sustainable solution, and I copied the line from a tips & tricks thread. 

Do you know of a better alternative or just dislike this for aesthetic reasons?

dougneal (approved):

PR #1362, opened 2023-09-14
OMM: bank CRUD functions

dougneal:
Summary
---------

With this commit, OMM is now a legit REST API.
Banks can be indexed, created, updated and deleted through the API.
When implementing CRUD functions for other data entities, we now have some reference code.

Test Plan
---------

I used the `humao.rest-client` VSCode extension to compose HTTP JSON requests in the editor and fire them off to the API. I recommend it.

```
POST http://localhost:5000/c/banks
content-type: application/json

{
    "name": "terrorism"
}

###

GET http://localhost:5000/c/banks

###

GET http://localhost:5000/c/bank/2

###

GET http://localhost:5000/c/bank/lloyds

###

PUT http://localhost:5000/c/bank/2
content-type: application/json

{
    "enabled": false
}

###

DELETE http://localhost:5000/c/bank/3
```

### DIFF
--- a/open-media-match/src/OpenMediaMatch/__init__.py
+++ b/open-media-match/src/OpenMediaMatch/__init__.py
@@ -7,9 +7,7 @@ import flask
 import flask_migrate
 import flask_sqlalchemy
 
-from .blueprints import hashing, matching, curation
-
-database = flask_sqlalchemy.SQLAlchemy()
+db = flask_sqlalchemy.SQLAlchemy()
 migrate = flask_migrate.Migrate()
 
 
@@ -30,8 +28,10 @@ def create_app():
         SQLALCHEMY_DATABASE_URI=app.config.get("DATABASE_URI"),
         SQLALCHEMY_TRACK_MODIFICATIONS=False,
     )
-    database.init_app(app)
-    migrate.init_app(app, database)
+    app.db = db
+
+    db.init_app(app)
+    migrate.init_app(app, db)
 
     @app.route("/")
     def index():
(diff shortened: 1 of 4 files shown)


open-media-match/src/OpenMediaMatch/__init__.py
```diff
@@ -5,9 +5,7 @@
 import flask_migrate
```
Dcallies, line 5: (Unrelated to diff): I'm always suspicious of usage of `__init__` (it's very easy to accidentally create problems in `foo.__init__` that trigger when doing `import foo.bar`, and I'm doubly suspicious of files that do work on import.

For example, we may find it hard to write unittests on files under OpenMediaMatch because all these lines are being executed each test.

Is there is an equivalent of 
```
if __name__ == '__main__':
```
for flask apps?

open-media-match/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -1,6 +1,72 @@
...
+        current_app.db.select(models.Bank)
+    ).scalars().all()
+    return jsonify(banks)
+
+
+@bp.route("/bank/<int:bank_id>", methods=["GET"])
+def bank_show_by_id(bank_id: int):
+    bank = models.Bank.query.get(bank_id)
+    if not bank:
+        return jsonify({'message': 'bank not found'}), 404
```
Dcallies, line 23: Ah, this was what I needed for proper error messages instead of `abort()`

open-media-match/src/OpenMediaMatch/models.py
```diff
@@ -1,15 +1,17 @@
...
+from dataclasses import dataclass
+from . import db
 
+@dataclass
 class Bank(db.Model):
     __tablename__ = "banks"
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
-    name = db.Column(db.String(255), nullable=False)
-    enabled = db.Column(db.Boolean, nullable=False)
+    id: int = db.Column(db.Integer, primary_key=True, autoincrement=True)
```
Dcallies, line 9: Wait this works? I'd have expected that typing doesn't complain that int != db.Column

If so, there's an easy way to make this bank class and a neutral interface class have the least amount of duplication.

Dcallies (approved):
Any idea why the typecheck and lint CI didn't run?

open-media-match/src/OpenMediaMatch/__init__.py
dougneal: This is a great point and I spent far too long yesterday fighting circular import problems. I don't know the answer but I'll research a bit more. I'm not a Flask expert 😅

open-media-match/src/OpenMediaMatch/models.py
dougneal: I'm as surprised as you. There's probably some disgusting reflection magic happening somewhere. I'll dig into it a bit more.

PR #1364, opened 2023-09-15
[omm] Add requests package to pyproject.toml

dougneal:
Summary
---------

Whoops, should've caught this in https://github.com/facebook/ThreatExchange/pull/1355

Test Plan
---------

`black ./src`


Dcallies (approved):
It's a transitive dependency, it's included in threatexchange

PR #1368, opened 2023-09-22
[omm] deplopulate __init__ and delint

Dcallies:
Summary
---------

## Depopulating `__init__.py`
A bunch of the lint tools were behaving strangely because they were getting confused about contents of the project based on the behavior of the `__init__.py` module. `__init__.py` is special in that its executed whenever you import subprojects (import a.b executes a.__init__.py).

Putting the flask application creation factory (`create_app`) in `__init__.py` is initially convenient because it means the factory is always at the top level of the init, however it introduces a lot of potential for errors (I've cut myself with `__init__.py` enough that I now prefer to leave them empty).

How flask finds the right app is described in https://flask.palletsprojects.com/en/2.3.x/cli/#application-discovery

The other common place to put it is in app.py.

### 
As part of this work, I also tried to remove all the module-level work that was being done, but specifically for the "db" object, this doesn't appear to be possible due to the design of flask-sqlalchemy.

I decided to just arbitrary shuffle the database information all into one place (models and all). 

## Delint
emptying `__init__.py` allowed me to delint the rest of app, which were mostly small fixes. 

Test Plan
---------
1. Rebuild the docker container from fresh, see that everything boots up correctly.
2. Manually run startup.sh (needed when you accidentally fatal your original copy), note it runs clean.

mypy . && py.test

LINT SHOULD NOW BE GREEN!


Dcallies:
Request for reviewers is that unless I've made some fatal error in this step, if there are minor changes to be made that I can do them in followups, since I think getting the CI green is useful!

PR #1382, opened 2023-09-30
[omm] Configure test DB for pytest

juanmrad:
Summary
---------

Change postgresql config to allow for multiple dbs on setup.
Change back to use docker image for testing.

*note:* You will need to recreate the postgresql for it to grab the init script updates and create both dbs. 

Test Plan
---------

pytest.


open-media-match/.devcontainer/docker-compose.yaml
```diff
@@ -16,13 +16,14 @@ services:
   db:
     image: postgres:latest
     restart: unless-stopped
+    volumes:
+        - ./pg-scripts/init-multiple-pg-dbs.sh:/docker-entrypoint-initdb.d/init-multiple-pg-dbs.sh
     ## Uncomment below lines for persistent database
-    # volumes:
```
Dcallies, line 20: Why this change?

.github/workflows/omm-ci.yaml
```diff
@@ -59,8 +59,7 @@ jobs:
         image: postgres
         # Provide the env variables
         env:
-          POSTGRES_HOST: localhost
-          POSTGRES_DB: media_match
+          POSTGRES_DB: media_match_test
```
Dcallies, line 62: Are we certain this DB is being used? If the docker image itself contains a database, and includes the instructions for creating the test database, and the docker run below doesn't have permission to talk to this one, do we need this copy?

.github/workflows/omm-ci.yaml
```diff
@@ -73,18 +72,10 @@ jobs:
...
-          python3 -m pip install -e .[all]
-      - name: Create tables
-        run: flask --app src/OpenMediaMatch/app.py create_tables
-        env:
-          OMM_CONFIG: ../../.devcontainer/omm_config.py
-      - name: Run tests
-        run: py.test src/OpenMediaMatch
-        env:
-          OMM_CONFIG: ../../.devcontainer/omm_config.py
+          docker run --network="host" local sh -c "python -m pip install --upgrade pip ;
```
Dcallies, line 79: Is --network the fix that allows it to contact the above docker?

If so, is there a Postgres being stood up by docker itself as well?

open-media-match/src/OpenMediaMatch/migrations/versions/9d85c5274e1f_.py
```diff
@@ -0,0 +1,88 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "9d85c5274e1f"
+down_revision = "b44643ee7ef0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
Dcallies, line 19: Are we relying on database.upgrade() to create our tables instead of create_tables()?

Dcallies (approved):
I'm worried I still don't know enough to give you a good review here, but at least it isolated the test database.

I'm also worried we're going to have to undo some amount of this when we figure out the production deployment.

See inline comments, I'm not understanding the  scope of the interaction between the docker-compose and the explicitly specified Postgres.

PR #1389, opened 2023-10-03
add clip

halfpete:
Summary
---------

Test Plan
---------


Dcallies:
Hey - we talked offline, we're hoping this can be in its own repo, let me show you an example!

PR #1390, opened 2023-10-03
[omm] Fix dev container startup script (path case sensitive issue)

benoit-yubo:
Summary
---------

[omm] Fix dev container startup script (path case sensitive issue)


Dcallies:
Talked with @benoit-yubo IRL and he showed me the signed SLA, so it's just delayed.

PR #1392, opened 2023-10-03
Initial hash method for posting files.

jafischer:
Summary
---------

Added new hash route for POST method that will accept multipart/form-data file uploads.

Currently returns `501 Not Implemented`.

Test Plan
---------

No testing yet.


Dcallies:
@jafischer Showed me IRL that the CLA is signed, merging as soon as CI passes.

PR #1394, opened 2023-10-03
[omm] Allow digits in bank names except as first character

benoit-yubo:
Summary
---------

Bank names should allow digits, except in first char

Test Plan
---------

Added tests to test_api.py to validate that invalid names result in HTTP 500 response codes

### DIFF
--- a/open-media-match/src/OpenMediaMatch/tests/test_api.py
+++ b/open-media-match/src/OpenMediaMatch/tests/test_api.py
@@ -15,12 +15,27 @@ def test_banks_empty_index(client: FlaskClient):
 
 
 def test_banks_create(client: FlaskClient):
+    # Must not start with number
     post_response = client.post(
         "/c/banks",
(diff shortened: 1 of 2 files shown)


open-media-match/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -15,12 +15,29 @@ def test_banks_empty_index(client: FlaskClient):
 
 
 def test_banks_create(client: FlaskClient):
+    # Must not start with number
     post_response = client.post(
         "/c/banks",
-        json={"name": "MY_TEST_BANK"},
+        json={"name": "01_TEST_BANK"},
+    )
+    assert post_response.status_code == 500
```
Dcallies, line 23: This should be a 400 (user error) not 500 (server error). 

We would have to figure out how to validate ahead of time to get the right error type.

Dcallies (approved):

PR #1398, opened 2023-10-03
[omm] Index query yield

juanmrad:
Summary
---------

Builds a way to query DB for all Content Signals of type while not overloading the DB in case of it being too big. 

Test Plan
---------

local testing.

### DIFF
--- a/open-media-match/src/OpenMediaMatch/storage/default.py
+++ b/open-media-match/src/OpenMediaMatch/storage/default.py
@@ -161,7 +161,33 @@ class DefaultOMMStore(interface.IUnifiedStore):
         raise Exception("Not implemented")
 
     def bank_yield_content(
-        self, signal_type: t.Optional[t.Type[SignalType]] = None
+        self, signal_type: t.Optional[t.Type[SignalType]] = None, batch_size: int = 100
     ) -> t.Iterator[t.Sequence[t.Tuple[t.Optional[str], int]]]:
-        # TODO
-        return MockedUnifiedStore().bank_yield_content(signal_type)
+        # Start the initial offset at 0
+        offset = 0
+        while True:
+            # Query the ContentSignal table
+            query = database.db.session.query(database.ContentSignal)
+
+            # Conditionally apply the filter if signal_type is provided
+            if signal_type is not None:
+                query = query.filter(
+                    database.ContentSignal.signal_type == signal_type.get_name()
+                )
+
+            # Batch the query by offset to avoid loading all results at ones overloading the DB.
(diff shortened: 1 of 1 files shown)


open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -161,7 +161,33 @@ def bank_remove_content(self, bank_name: str, content_id: int) -> None:
...
+            query = database.db.session.query(database.ContentSignal)
+
+            # Conditionally apply the filter if signal_type is provided
+            if signal_type is not None:
+                query = query.filter(
+                    database.ContentSignal.signal_type == signal_type.get_name()
+                )
+
+            # Batch the query by offset to avoid loading all results at ones overloading the DB.
+            query = query.offset(offset).limit(batch_size).yield_per(batch_size)
```
Dcallies, line 179: Is this ordering stable, or can you skip hashes if you insert into the table while querying?

open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -161,7 +161,33 @@ def bank_remove_content(self, bank_name: str, content_id: int) -> None:
...
-        # TODO
-        return MockedUnifiedStore().bank_yield_content(signal_type)
+        # Start the initial offset at 0
+        offset = 0
+        while True:
+            # Query the ContentSignal table
+            query = database.db.session.query(database.ContentSignal)
+
+            # Conditionally apply the filter if signal_type is provided
+            if signal_type is not None:
```
Dcallies, line 173: This filter may cause a table scan if you have a block of non-matching filters

open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -161,7 +161,33 @@ def bank_remove_content(self, bank_name: str, content_id: int) -> None:
...
+            # If there are no more results, break the loop
+            if not results:
+                break
+
+            # Increment the offset for the next batch
+            offset += batch_size
+
+            # Yield the results as tuples (signal_val, content_id)
+            for content_signal in results:
+                yield [(content_signal.signal_val, content_signal.content_id)]
```
Dcallies, line 193: Nit:

yield [(content_signal.signal_val, content_signal.content_id) for content_signal in results]

Dcallies (commented):
Hmm, I have two concerns:
1. The ordering won't be stable - inserts will potentially go near the front of the ordering rather than the back
2. The query won't be efficient - the offset may lead to a table scan to find the Nth result.

Let me read if there is native support for cursoring.

Dcallies (approved):
lgtm

PR #1400, opened 2023-10-03
[omm] bank update modified to take bank name as input

benoit-yubo:
Summary
---------

- bank update takes bank name as parameter
- catch ValueError and IntegrityError to return HTTP 400 or 404 instead of 500

Test Plan
---------

- Modified existing test plan to match new error codes and to check update method


Dcallies (approved):
Nice work! Thanks for adding tests!

PR #1401, opened 2023-10-03
[omm] bank delete by name

benoit-yubo:
Summary
---------

- bank delete by name and not by id
- always return JSON in case of error for bank API

Test Plan
---------

- bank delete test plan added

### DIFF
--- a/open-media-match/src/OpenMediaMatch/blueprints/curation.py
+++ b/open-media-match/src/OpenMediaMatch/blueprints/curation.py
@@ -18,6 +18,7 @@ bp = Blueprint("curation", __name__)
 
 
 @bp.route("/banks", methods=["GET"])
+@utils.abort_to_json
 def banks_index():
     storage = persistence.get_storage()
     return list(storage.get_banks().values())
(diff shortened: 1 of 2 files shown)


open-media-match/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -75,12 +77,13 @@ def bank_update(bank_name: str):
...
-def bank_delete(bank_id: int):
+@bp.route("/bank/<bank_name>", methods=["DELETE"])
+@utils.abort_to_json
+def bank_delete(bank_name: str):
     # TODO - rewrite using persistence.get_storage()
-    bank = database.Bank.query.get(bank_id)
+    bank = database.Bank.query.filter_by(name=bank_name).one_or_none()
     if not bank:
-        return jsonify({"message": "bank not found"}), 404
+        abort(404, "Bank not found")
```
Dcallies, line 86: blocking: Deleting a bank that doesn't exist should return success (it may be the result of a retry).

benoit-yubo: updated code and tests

PR #1402, opened 2023-10-03
Adding Signal and Content Types to index page

htracey-godaddy:
Summary
---------

Added signal and content types to index page.

Test Plan
---------
Screenshot 2023-10-03 at 3 02 12 PM

### DIFF
--- a/open-media-match/src/OpenMediaMatch/app.py
+++ b/open-media-match/src/OpenMediaMatch/app.py
@@ -4,6 +4,8 @@ import logging
 import os
 import sys
 import warnings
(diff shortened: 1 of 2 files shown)


Dcallies (commented):
LGTM

I think an alternative that would also work is doing something like:

```
Signal Types
  * PDQ: Enabled
  * video_md5: Disabled
```

But this is fine too

htracey-godaddy:
> LGTM
> 
> I think an alternative that would also work is doing something like:
> 
> ```
> Signal Types
>   * PDQ: Enabled
>   * video_md5: Disabled
> ```
> 
> But this is fine too

ok updated the ui

PR #1404, opened 2023-10-03
Implemented build_index

Tsytrytskyi:
Summary
---------

Test Plan
---------

### DIFF
--- a/open-media-match/src/OpenMediaMatch/background_tasks/build_index.py
+++ b/open-media-match/src/OpenMediaMatch/background_tasks/build_index.py
@@ -41,8 +41,11 @@ def build_index(
     Build one index from scratch with the current bank contents and persist it.
     """
     logger.info("Building index for %s", for_signal_type.get_name())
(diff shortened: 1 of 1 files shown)


open-media-match/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -41,8 +41,11 @@ def build_index(
...
     """
     logger.info("Building index for %s", for_signal_type.get_name())
-    # TODO
-
+    index_cls = for_signal_type.get_index_cls()
+    list = []
     for batch in bank_store.bank_yield_content(for_signal_type):
         for signal, bc_id in batch:
-            logging.debug("TODO remove me - %d: %s", bc_id, signal)
+            if signal:
```
Dcallies, line 48: This is probably okay since empty string is unlikely to be a valid signal type, but if we wanted to explicit we could use `if signal is not None`

PR #1407, opened 2023-10-03
Bank list page

Tsytrytskyi:
Summary
---------

Add banks to the ui

Test Plan
---------

image


Dcallies (changes requested):
Can you put this at the bottom of the index instead?

For now, I think that having a single page is beneficial

Dcallies (approved):
Yay!

PR #1410, opened 2023-10-03
[omm] extension support for signal types

benoit-yubo:
Summary
---------

- support extension of signal types via omm_config.py

Test Plan
---------

- TBD

### DIFF
--- a/open-media-match/src/OpenMediaMatch/storage/default.py
+++ b/open-media-match/src/OpenMediaMatch/storage/default.py
@@ -6,8 +6,6 @@ The default store for accessing persistent data on OMM.
 
 import typing as t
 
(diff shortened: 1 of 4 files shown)


benoit-yubo:
left code in default storage as moving it to persistence.py creates a circular dependency

open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -6,6 +6,8 @@
 
 import typing as t
 
+from importlib import import_module
```
evank28, line 9: Let's remove this

evank28 (approved):
Looks good - this should work

PR #1413, opened 2023-10-04
[OMM] Index building and querying

juanmrad:
Summary
---------

Allow to store index properly and querying of data from it. 

Test Plan
---------

Added test. pytest.

Screenshot 2023-10-03 at 5 21 56 PM


Dcallies (approved):
LGTM!

PR #1414, opened 2023-10-04
[omm] use only storage interface for bank API

benoit-yubo:
Summary
---------

Removed direct usage of SQLAlchemy in API and use storage interface instead

Test Plan
---------

- Nothing changed

### DIFF
--- a/open-media-match/src/OpenMediaMatch/blueprints/curation.py
+++ b/open-media-match/src/OpenMediaMatch/blueprints/curation.py
@@ -72,6 +72,8 @@ def bank_update(bank_name: str):
             bank.name = data["name"]
         if "enabled" in data:
             bank.matching_enabled_ratio = 1 if bool(data["enabled"]) else 0
+        if "enabled_ratio" in data:
+            bank.matching_enabled_ratio = data["enabled_ratio"]
 
         storage.bank_update(bank, rename_from=rename_from)
     except ValueError as e:
--- a/open-media-match/src/OpenMediaMatch/blueprints/curation.py
+++ b/open-media-match/src/OpenMediaMatch/blueprints/curation.py
@@ -58,8 +58,9 @@ def bank_create():
 @utils.abort_to_json
 def bank_update(bank_name: str):
     # TODO - rewrite using persistence.get_storage()
+    storage = persistence.get_storage()
     data = request.get_json()
-    bank = database.Bank.query.filter_by(name=bank_name).one_or_none()
+    bank = storage.get_bank(bank_name)
     if bank is None:
         abort(404, "Bank not found")
 
@@ -67,24 +68,20 @@ def bank_update(bank_name: str):
         if "name" in data:
             bank.name = data["name"]
         if "enabled" in data:
-            bank.enabled = bool(data["enabled"])
+            bank.matching_enabled_ratio = 1 if bool(data["enabled"]) else 0
 
-        database.db.session.commit()
+        storage.bank_update(bank, name=bank_name)
     except ValueError as e:
         abort(400, *e.args)
-    return jsonify(bank.as_storage_iface_cls())
+    return jsonify(bank)
 
 
 @bp.route("/bank/<bank_name>", methods=["DELETE"])
 @utils.abort_to_json
 def bank_delete(bank_name: str):
-    # TODO - rewrite using persistence.get_storage()
-    bank = database.Bank.query.filter_by(name=bank_name).one_or_none()
-    if not bank:
-        return {"message": "Done"}, 200
-    database.db.session.delete(bank)
-    database.db.session.commit()
-    return jsonify({"message": f"Bank {bank.name} ({bank.id}) deleted"})
+    storage = persistence.get_storage()
+    storage.bank_delete(bank_name)
+    return jsonify({"message": "Done"})
 
 
 @bp.route("/bank/<bank_name>/content", methods=["POST"])
--- a/open-media-match/src/OpenMediaMatch/blueprints/curation.py
+++ b/open-media-match/src/OpenMediaMatch/blueprints/curation.py
@@ -1,4 +1,5 @@
 import re
+from typing import Optional
 
 from flask import Blueprint
 from flask import request, jsonify, abort
@@ -61,16 +62,18 @@ def bank_update(bank_name: str):
     storage = persistence.get_storage()
     data = request.get_json()
     bank = storage.get_bank(bank_name)
+    rename_from: Optional[str] = None
     if bank is None:
         abort(404, "Bank not found")
 
     try:
         if "name" in data:
+            rename_from = bank.name
             bank.name = data["name"]
         if "enabled" in data:
             bank.matching_enabled_ratio = 1 if bool(data["enabled"]) else 0
 
-        storage.bank_update(bank, name=bank_name)
+        storage.bank_update(bank, rename_from=rename_from)
     except ValueError as e:
         abort(400, *e.args)
     return jsonify(bank)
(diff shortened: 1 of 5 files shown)


open-media-match/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -60,33 +60,30 @@ def bank_create():
 @utils.abort_to_json
 def bank_update(bank_name: str):
     # TODO - rewrite using persistence.get_storage()
+    storage = persistence.get_storage()
     data = request.get_json()
-    bank = database.Bank.query.filter_by(name=bank_name).one_or_none()
+    bank = storage.get_bank(bank_name)
     if bank is None:
         abort(404, "Bank not found")
```
Dcallies, line 67: Theres a helper for this in this file somewhere

open-media-match/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -60,33 +60,30 @@ def bank_create():
...
-    # TODO - rewrite using persistence.get_storage()
-    bank = database.Bank.query.filter_by(name=bank_name).one_or_none()
-    if not bank:
-        return {"message": "Done"}, 200
-    database.db.session.delete(bank)
-    database.db.session.commit()
-    return jsonify({"message": f"Bank {bank.name} ({bank.id}) deleted"})
+    storage = persistence.get_storage()
+    storage.bank_delete(bank_name)
+    return jsonify({"message": "Done"})
```
Dcallies, line 86: nit: You don't need to call jsonify here

open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -113,21 +113,23 @@ def _get_bank(self, name: str) -> t.Optional[database.Bank]:
...
             database.db.session.add(database.Bank.from_storage_iface_cls(bank))
         else:
-            database.db.session.execute(
-                update(database.Bank)
-                .where(database.Bank.name == bank.name)
-                .values(enabled_ratio=bank.matching_enabled_ratio)
-            )
+            previous = database.Bank.query.filter_by(name=name).one_or_404()
+            previous.name = bank.name
+            previous.enabled_ratio = 1 if bank.enabled else 0
```
Dcallies, line 124: blocking: Shouldn't this be be 

```
previous.enabled_ratio = bank.enabled_ratio
```
?

open-media-match/src/OpenMediaMatch/storage/interface.py
```diff
@@ -267,12 +267,15 @@ def get_bank(self, name: str) -> t.Optional[BankConfig]:
...
-    def bank_update(self, bank: BankConfig, *, create: bool = False) -> None:
+    def bank_update(
+        self, bank: BankConfig, *, create: bool = False, name: t.Optional[str] = None
+    ) -> None:
         """
         Update a bank config in the backing store.
 
         If create is false, will throw an exception if not already existing.
         If create is true, will throw an exception if it already exists
+        If create is false, name has to be provided
```
Dcallies, line 278: blocking: name is only required if you are changing the name. 

Suggest changing "name" to "rename_from", looking up the old record with bank.name unless rename is true.

Dcallies (changes requested):
blocking: You have a merge conflict to work through sadly

open-media-match/src/OpenMediaMatch/storage/default.py
benoit-yubo: actually it can be previous.enabled_ration  = bank.matching_enabled_ratio I think

benoit-yubo:
@Dcallies I updated the PR, let me know

open-media-match/src/OpenMediaMatch/storage/interface.py
Dcallies: nit: Comment needs updating

open-media-match/src/OpenMediaMatch/storage/mocked.py
```diff
@@ -98,8 +98,17 @@ def get_collab_data(
...
+        self,
+        bank: interface.BankConfig,
+        *,
+        create: bool = False,
+        rename_from: t.Optional[str] = None,
+    ) -> None:
+        if create:
+            self.banks[bank.name] = bank
+        else:
+            self.banks[rename_from if rename_from is not None else bank.name] = bank
```
Dcallies, line 111: blocking: this isn't quite correct - when rename from is true, you need to pop it out.

Try this:
```suggestion
        if rename_from is not None:
            self.banks.pop(rename_from)
        self.banks[bank.name] = bank
```

open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -149,21 +149,29 @@ def _get_bank(self, name: str) -> t.Optional[database.Bank]:
...
             database.db.session.add(database.Bank.from_storage_iface_cls(bank))
         else:
-            database.db.session.execute(
-                update(database.Bank)
-                .where(database.Bank.name == bank.name)
-                .values(enabled_ratio=bank.matching_enabled_ratio)
-            )
+            previous = database.Bank.query.filter_by(
+                name=rename_from if rename_from is not None else bank.name
+            ).one_or_404()
```
Dcallies, line 164: nice find

open-media-match/src/OpenMediaMatch/storage/default.py
```diff
@@ -149,21 +149,29 @@ def _get_bank(self, name: str) -> t.Optional[database.Bank]:
...
+            previous.name = bank.name
+            previous.enabled_ratio = bank.matching_enabled_ratio
```
Dcallies, line 166: I'm surprised this works - you don't have to make an explicit update?

open-media-match/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -58,33 +59,32 @@ def bank_create():
...
     if bank is None:
         abort(404, "Bank not found")
 
     try:
         if "name" in data:
+            rename_from = bank.name
             bank.name = data["name"]
         if "enabled" in data:
-            bank.enabled = bool(data["enabled"])
+            bank.matching_enabled_ratio = 1 if bool(data["enabled"]) else 0
```
Dcallies, line 74: nit: This should be taking a float enabled ratio, but this can be a followup PR for someone else

Dcallies (changes requested):
mocked impl needs one more pass.

I am surprised this works - can you either:
1. Add a new test to exercise updating without rename & update with rename
2. In the test plan, show a series of curl commands and their outputs which show the functionality works

open-media-match/src/OpenMediaMatch/storage/mocked.py
benoit-yubo: yeah good catch :+1:

open-media-match/src/OpenMediaMatch/storage/default.py
benoit-yubo: I was surprised as well, but it seems that because previous is a model instance, SQLAlchemy tracks changes you do on it and does the update automatically when you commit.

open-media-match/src/OpenMediaMatch/blueprints/curation.py
benoit-yubo: I added it as well

open-media-match/src/OpenMediaMatch/storage/default.py
benoit-yubo: also, previous code doing the .update().where().values() was somehow not triggering name validation, that's why I changed it to this format.

Dcallies: Oh man, that's good to know, good catch

Dcallies (approved):

PR #1421, opened 2023-10-04
[Tech Coalition Hackathon 2023] add entrypoint for manifest extensions

halfpete:
Summary
---------

Creates a new function called `entrypoint` under Manifest Extensions, which Extensions will implement. 
`entrypoint` will be used to bootstrap anything for the extension at when the `add` command is called:

`tx config extensions add [extension_name]`

Usage Example: https://github.com/TechnologyCoalitionOrg/tx-extension-clip/commit/afe2ff18f1de8ba107fd2109afa8d3ad38581f77

```python
class CLIPExtensionManifest(ThreatExchangeExtensionManifest):
    """
    CLIP Extension Manifest.
    """

    @classmethod
    def entrypoint(cls):
        """
        Download the model on extension install.
        """
        clip_hasher: CLIPHasher = CLIPHasher(
            model_name=OPEN_CLIP_MODEL_NAME,
            pretrained=OPEN_CLIP_PRETRAINED,
            normalized=CLIP_NORMALIZED,
        )
        clip_hasher.init_model_and_transforms()
 ```

Test Plan
---------

We tested by adding a model load for `tx_extension_clip`: https://github.com/TechnologyCoalitionOrg/tx-extension-clip
Removed the model in our cache, added the load in the extension's `entrypoint` override, and ensured that the model was properly loaded, and loaded only once. 

image

### CODE REVIEW

**Review by Dcallies — CHANGES_REQUESTED** [review]
Thanks for adding additional functionality!

Discussed offline, can you:

1. Rename "entrypoint()" to "bootstrap()", with instructions that it can 
2. Add a second method called "verify()"

load should not call either of these methods

File: python-threatexchange/threatexchange/extensions/manifest.py [hunk 1]
```diff
@@ -48,4 +48,35 @@ def load_from_module_name(
...
+        try:
+            manifest.bootstrap()
+        except Exception as exc:
+            raise ValueError(f'Manifest failed to bootstrap: {module_name}') from exc
+
+        try:
+            manifest.verify()
+        except Exception as exc:
+            raise ValueError(f'Manifest failed verification: {module_name}') from exc
+
```
Dcallies, line 61: blocking: Per previous comment these shouldn't be added here.

Instead, in the CLI, call these two functions in the extensions command

Dcallies (changes requested):

python-threatexchange/threatexchange/extensions/manifest.py
halfpete: oop - apologies missed this the first time. Just updated

Dcallies (approved):

PR #1438, opened 2023-10-04
[omm] fix issues with ui addcontent

benoit-yubo:
Summary
---------

- fix add content ui (limitation on content type, work only for addcontent as photo)

Test Plan
---------

### DIFF
--- a/open-media-match/src/OpenMediaMatch/templates/index.html.j2
+++ b/open-media-match/src/OpenMediaMatch/templates/index.html.j2
@@ -139,6 +139,12 @@
                 // Handle the response from the server here
                 renderAddBankResult(data);
             })
+            .then(() => {
+                return fetch('/ui/rebuild_index', {
+                    method: 'POST',
+                    body: new FormData()
+                })
+            })
             .catch(error => {
                 console.error('Error:', error);
             });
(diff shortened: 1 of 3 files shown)


Dcallies:
@benoit-yubo, @juanmrad did a similar fix in #1439, but I think I want to add your version as well, because I think auto index rebuilding is super useful. 

Do you want to:
1. Rebase on #1439 and switch the call over to what you wrote
2. Call it a day, and I can take over making the same fixes based on your PR?

open-media-match/src/OpenMediaMatch/templates/index.html.j2
```diff
@@ -43,7 +43,7 @@
         <!-- Add Content to Bank -->
         <h3> Add Content to Bank: </h3>
         <form action = "/ui/addcontent" method = "post" enctype="multipart/form-data">  
-            <input type="file" name="file" /> Banks
+            <input type="file" name="photo" /> Banks
```
Dcallies, line 46: We should switch this over the a selector.

You might be able to rewrite the request parameters to populate it how the hash expects.

Dcallies (changes requested):
RC to rebase / decide what to do.

benoit-yubo:
I’ll rebase it. I’m on my way to the airport now so will do it as soon as I
can

Le jeu. 5 oct. 2023 à 09:17, David Callies ***@***.***> a
écrit :

benoit-yubo:
Hello @Dcallies I added the automatic call to rebuild_index in javascript

PR #1448, opened 2023-10-13
[pytx] Add back support for Python 3.7

huythehoang:
Summary
---------
Our team relies heavily on Python 3.7 and would like to use `threatexchange` to fetch hash updates. This is my attempt at making this library compatible with 3.7 as well. 

It seems like the only things we strictly need 3.8 for are `final` and `Protocol` which were introduced in 3.8. To make things compatible with 3.7, I'm importing these from `typing-extesions`.

Test Plan
---------
Tested in python `3.10` and `3.7` virtual environments and confirmed that `threatexchange` was installed and imported successfully.

### DIFF
--- a/.github/workflows/python-threatexchange-ci.yaml
+++ b/.github/workflows/python-threatexchange-ci.yaml
@@ -42,7 +42,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.x']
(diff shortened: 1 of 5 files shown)


Dcallies:
Hey @huythehoang, thanks for reaching out. As 3.7 has [reached end of life](https://devguide.python.org/versions/), we are no longer able to support. We're hoping to adopt new features that are not backwards compatible in the future, so suggest freezing your version or upgrading to a new version of python.

PR #1453, opened 2023-10-31
[TMK] Add support to build on Mac M2

sahilx13:
Summary
---------

Updates the TMK Makefile to add support for building on Mac M2 machines

### DIFF
--- a/tmk/cpp/Makefile
+++ b/tmk/cpp/Makefile
@@ -146,41 +146,41 @@ libtmk.a: $(LIBOBJ)
 
 # ================================================================
 tmk-hash-video: ./bin/tmk-hash-video.cpp $(LIBHDR) libtmk.a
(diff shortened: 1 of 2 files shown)


Dcallies (changes requested):
Thanks for submitting a PR! 

Looks like the CI can't build with the new flags (as you suspected).

Check out the approach the PDQ makefile uses for this: #1016 / https://github.com/facebook/ThreatExchange/blame/main/pdq/cpp/Makefile

Moving the common flags into the top, and then using if statements (or just a comment...) to select the right one may be as best as we can do.

Let me know if you have the time to take a crack at this, or if this is as far as you can make with your current bandwidth!

Dcallies:
Hey @sahilx13, I'm going to close this PR for inactivity, feel free to re-open if you still want to try and get it merged!

PR #1474, opened 2023-12-06
Fix labeler

Dcallies:
Seems like the syntax changed - looking at https://github.com/actions/labeler to find the new stuff


Dcallies:
I could checkout as a step, but I'm too dumb, so just going to try and brute force it.

Issue #1346, opened 2023-08-17
Windows?


therreher:
Does this software work with windows 10. I have ffmpeg and cmake installed but when I try to install I get this error:
pip install vpdq
Collecting vpdq
  Using cached vpdq-0.0.8.tar.gz (86 kB)
  Preparing metadata (setup.py) ... done
Requirement already satisfied: cython in c:\users\user\appdata\local\programs\python\python311\lib\site-packages (from vpdq) (3.0.0)
Requirement already satisfied: opencv-python in c:\users\user\appdata\local\programs\python\python311\lib\site-packages (from vpdq) (4.8.0.76)
Collecting opencv-python-headless (from vpdq)
  Obtaining dependency information for opencv-python-headless from https://files.pythonhosted.org/packages/70/78/7a13730745684584db53e8aa3c3bd84beef2dcb32bebf627bda0d6df461e/opencv_python_headless-4.8.0.76-cp37-abi3-win_amd64.whl.metadata
  Using cached opencv_python_headless-4.8.0.76-cp37-abi3-win_amd64.whl.metadata (20 kB)
Requirement already satisfied: numpy>=1.21.2 in c:\users\user\appdata\local\programs\python\python311\lib\site-packages (from opencv-python->vpdq) (1.25.2)
Using cached opencv_python_headless-4.8.0.76-cp37-abi3-win_amd64.whl (38.0 MB)
Building wheels for collected packages: vpdq
  Building wheel for vpdq (setup.py) ... error
  error: subprocess-exited-with-error

  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [4 lines of output]
      running bdist_wheel
      running build
      running build_ext
      error: [WinError 2] The system cannot find the file specified
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for vpdq
  Running setup.py clean for vpdq
Failed to build vpdq
ERROR: Could not build wheels for vpdq, which is required to install pyproject.toml-based projects

Dcallies:
Hmm, @ianwal had done a series of diffs to add windows support, but I don't think we built a new version at the end, which you've tried to pip install. Let me release a new version and see if that fixes it. 

I have a windows machine, so should be able to test it.

ianwal:
The Python bindings I didn't have time to get working fully with Windows. It's a build issue. But, vpdq itself does build and work with Windows.

[...]

PR #1508, opened 2024-01-17
[vpdq] Upgrade Python binding to modern build system

ianwal:
Summary
---------

These changes update the Python binding to the modern standard Python build system (PEP 621).

Overview:

* Move vpdq cpp files into own directory to preserve include file paths
    * Rather than going UP two directories in the build system to the root of this repo in `vpdq-release.py`, just move everything DOWN two folders.

* Vendorize PDQ
    * Having all source files in the same folder _greatly_ simplifies CMake and Python binding script.
    * Changes to PDQ don't run vpdq unit tests, so this allows changes to PDQ without breaking things.  
    * One alternative is copying the pdq folder while running the CMake and the Python scripts, but this adds a relative path dependency and complicates both scripts. 
    * A git submodule or one source of truth would be preferable to copying the files, but you can't make a submodule of the same repo.
    * No changes were made to PDQ when copying it.

* Use `build` for building and packaging the Python binding

* Create pyproject.toml for packaging and installing Python binding

* Fix vpdq cpp compiler warning for wrong printf datatype when compiling with MinGW

* Fix misnamed vpdq cpp include guard, move library includes inside of the include guard

* Bump GitHub Actions versions

* Update documentation

Some file paths still rely on files in the repo outside of vpdq, such as the the regression test videos. But these do not affect the C++ or Python builds.

This PR could be split up if it is too large, but most of the changes are small cleanups, moving files, and large changes that don't really make sense on their own.

New build and packaging instructions are in CONTRIBUTING.md (it's 3 commands total).

Test Plan
---------

Tested building on Ubuntu and Arch Linux. Tested packaging from one and installing on the other.

All CI tests pass.

The only thing left untested is the release GitHub action and the result. Worst case it can be yanked if something goes terribly wrong.

This PR does not create a new release for the Python binding, so more testing could be done for the package before it is published.


vpdq/python/README.md
```diff
@@ -5,28 +5,41 @@ It is written in Cython.
 
 Development is hosted on [GitHub](https://github.com/facebook/ThreatExchange/tree/main/vpdq).
 
+Windows is currently not supported by the Cython bindings.
```
ianwal, line 8: I tried to get Windows working through MinGW. I actually got it to build with a few more changes that I wrote about [here](https://github.com/facebook/ThreatExchange/issues/1346#issuecomment-1682189882). But it gives errors about loading DLLs at runtime.

vpdq cpp implementation still builds and runs on Windows fine, though.

vpdq/python/vpdq.pyx
```diff
@@ -52,23 +52,23 @@ class VpdqFeature:
     frame_number: int
     hash: Hash256
     hex: str
-    timestamp: double
+    timestamp: float
```
ianwal, line 55: These typehint changes are warnings I spotted in the output of the Cython build and fixed.

.github/workflows/vpdq-release.yaml
```diff
@@ -17,24 +17,31 @@ jobs:
...
-          pip install -r requirements.txt
-          python vpdq-release.py -i
-      - name: Package threatexchange
+          python -m pip install -r packaging-requirements.txt
+
+      - name: Package
+        run: |
+          python -m build --sdist
+
+      - name: Ensure packaged build runs
```
ianwal, line 33: Maybe move this check to the other test file?

Still this is an important sanity check and doing it before uploading seemed logical. If I had this check in other projects I wouldn't have published broken builds.

vpdq/cpp/CMakeLists.txt
```diff
@@ -15,78 +15,84 @@ pkg_check_modules(LIBAV REQUIRED IMPORTED_TARGET
...
 )
 
-# TODO: Add more specific directories
-include_directories (
-    ../..
-)
+include_directories( vpdq/cpp/bin vpdq/cpp/hashing vpdq/cpp/io ${CMAKE_SOURCE_DIR} )
 
-add_library(vpdqlib
+add_library(vpdq
```
ianwal, line 54: Renamed `vpdqlib` to just `vpdq`. The output file before this was `libvpdqlib.a`, now it is just `libvpdq.a` like every other library is.

Dcallies:
Woah, this is great, thanks @ianwal !

As you can probably tell, we didn't have much internal knowledge of the best practices, and did our best with what we could figure out live. 

Let me see if I can figure out if I can dryrun the pip release. Since you've done all the work so far I'm happy to work on that if it has issues.

vpdq/cpp/pdq/README.md
```diff
@@ -0,0 +1,5 @@
+Vendorized PDQ CPP implementation
```
Dcallies, line 1: What does Vendorized mean?

vpdq/cpp/pdq/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+PDQ is vendorized in order to ease the complexity of the VPDQ's build system and to avoid version conflicts.
+
+If VPDQ or PDQ are moved to their own repositories in the future, VPDQ's build system will still continue to function without modification.
```
Dcallies, line 5: Ugh, hearing that submodules don't work if its to the same repo is the strongest argument I've heard yet to start breaking up the ThreatExchange repo.

vpdq/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+requires = [
+    "setuptools >= 61.0",
+    "wheel",
+    "cython",
+]
+build-backend = "setuptools.build_meta"
+
+[project]
+name="vpdq"
+authors=[{name="Facebook", email="threatexchange@fb.com"}]
```
Dcallies, line 11: The new address is `threatexchange@meta.com`

vpdq/python/README.md
Dcallies: That's fine for me

vpdq/python/vpdq.pyx
Dcallies: I couldn't figure out how to actually run any of the typehints on .pyx files, so much appreciated.

vpdq/setup.py
```diff
@@ -6,50 +6,77 @@
 import sys
```
Dcallies, line 6: I was never happy with the way that this `setup.py` worked, glad you could get a crack at it.

Dcallies (approved):
Overall this looks like an improvement, your comprehensive writeup helped a lot.

While copying PDQ directory isn't ideal, given the situation with submodules I think it's acceptable, though I'm worried we'll forgot to copy it if we ever make a chance to PDQ (though it's been pretty stable).

I can try doing a versioned release to see if that all still works.

Let me know if you have any other thoughts, otherwise I'm fine trying to merge.

vpdq/cpp/pdq/README.md
ianwal: I meant to write [vendoring](https://stackoverflow.com/questions/26217488/what-is-vendoring)! Will update in all places.

vpdq/pyproject.toml
ianwal: Should author name be updated to `Meta` as well?

Dcallies:
> Changing PDQ without testing VPDQ is not a good thing

+1 to this, I think in my ideal world we'd have a pinned version of PDQ in the directory, which hopefully would give us the best of both worlds. However, your comments + random googling doesn't look like the tech is there yet without splitting up the ThreatExchange repo into multiple repos.

vpdq/pyproject.toml
Dcallies: I don't think I've gotten any guidance on it, I think both are currently accepted.

vpdq/cpp/pdq/README.md
ianwal: I think there is a way to do it, or with git subtrees, but that's beyond me. I would rather not ruin the repo doing some weird tricks.

ianwal:
Ready to merge

Dcallies:
Well, yolo build failed: https://github.com/facebook/ThreatExchange/actions/runs/7570151016/job/20614934802

```
Building wheels for collected packages: vpdq
  Building wheel for vpdq (pyproject.toml): started
  Building wheel for vpdq (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Building wheel for vpdq (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [[26](https://github.com/facebook/ThreatExchange/actions/runs/7570151016/job/20614934802#step:7:27) lines of output]
      INFO:root:running bdist_wheel
      INFO:root:running build
      INFO:root:running build_ext
      INFO:setup.py:Removing CPP build directory /tmp/pip-req-build-o5od4zvw/cpp/build...
      INFO:setup.py:Removing compiled Cython files /tmp/pip-req-build-o5od4zvw/python/vpdq.cpp...
      INFO:setup.py:Removing libraries-dir.txt file /tmp/pip-req-build-o5od4zvw/cpp/libraries-dirs.txt...
      INFO:setup.py:Creating CPP build directory...
      INFO:setup.py:Running CMake...
      CRITICAL:setup.py:CMake Error at /usr/local/share/cmake-3.[28](https://github.com/facebook/ThreatExchange/actions/runs/7570151016/job/20614934802#step:7:29)/Modules/FindPkgConfig.cmake:619 (message):
        The following required packages were not found:
      
         - libavdevice
         - libavfilter
         - libavformat
         - libavcodec
         - libswresample
         - libswscale
         - libavutil
      
      Call Stack (most recent call first):
        /usr/local/share/cmake-3.28/Modules/FindPkgConfig.cmake:8[41](https://github.com/facebook/ThreatExchange/actions/runs/7570151016/job/20614934802#step:7:42) (_pkg_check_modules_internal)
        CMakeLists.txt:6 (pkg_check_modules)
      
Failed to build vpdq
      
      
      CRITICAL:setup.py:Failed to compile vpdq library.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for vpdq
ERROR: Could not build wheels for vpdq, which is required to install pyproject.toml-based projects
```

.github/workflows/vpdq-release.yaml
Dcallies: This step is the one that fails post-diff, I'm not sure how to test it other than creating a temporary copy that runs on contbuild.

I think it's missing a step to install ffmpeg, let me see if I can get a repro

ianwal:
> Well, yolo build failed: https://github.com/facebook/ThreatExchange/actions/runs/7570151016/job/20614934802

Yes, the workflow needs to be updated to install the dependencies listed in the README. There hasn't been a new deployment since libav* replaced the FFmpeg binary. I will open a new PR for it that.

ianwal:
Another thing: I could see the value in making the frame decoder separate from the hashing entirely. If users wanted to supply the frames directly then they should probably be able to do that. It would make testing easier too.

Then, it could be optional (but probably the default) to install with the libav* or use FFmpeg or whatever. I use pyav in my Python for another project because I download video files into memory using SpooledTemporaryFile and hashing files without a path to it is a giant pain for FFmpeg or libav.

Issue #1513, opened 2024-01-19
[vpdq] pypi publish failing


Dcallies:
After the build tool refresh from #1508 , something is off about the publish step. 

Ex: https://github.com/facebook/ThreatExchange/actions/runs/7586086044/job/20663427609

Currently saying version already used, here's what we have: https://pypi.org/project/vpdq/#history - however, 0.1.1 is not used, so the version is probably being picked up from somewhere other than setup.py
Grepping through the codebase, I don't see any strings for "0\.0\.[1-8]" so more investigation needed.

Note to self: retry build instead of bumping veresion.

ianwal:
Version `0.1.0` and several others are listed here: https://libraries.io/pypi/vpdq/versions

I found libraries.io through the statistics column on the left-hand side, so it seems version `0.1.0` was released at some point.

Starting at `0.2.0` should fix the problem.

Dcallies:
Thanks for the fix @ianwal , this is resolved, and the new package seems installable!

PR #1514, opened 2024-01-20
[vpdq] Fix vpdq PyPI version collision

ianwal:
Summary
---------

Fix issue #1513

- Bump vpdq Python binding version to 0.2.0 to avoid version collision on PyPI.

- Fix the workflow warnings.

According to https://libraries.io/pypi/vpdq/versions 0.1.* _was_ released, even though it doesn't show on the PyPI version list. It was probably deleted from PyPI.

It's not the build process, because the release version is based on the filename. The filename is `vpdq-0.1.1.tar.gz` which is correct.

This is why yanking is preferred to deleting (unless you leaked nuclear codes :) ):

image

Test Plan
---------

🙏


Dcallies (approved):

Dcallies:
Thanks again for another fix, sorry I haven't had the bandwidth to try things along with you!

I don't have the context for the version changes - the history of the file doesn't show the version: https://github.com/facebook/ThreatExchange/commits/main/vpdq/version.txt

Might have been a mistake by the original author.

Dcallies:
Looks publish worked! https://github.com/facebook/ThreatExchange/actions/runs/7596631693/job/20690555792

I tried running it in my devcontainer environment and needed to make some additional changes: #1515

PR #1515, opened 2024-01-20
[vpdq] Handle pthread error in devcontainer; ignore more build dirs

Dcallies:
Summary
---------

The only environment I was able to get vpdq working before was the docker image container we are using for omm (because I'm on a lamer windows machine).

I attempted to install 0.2.0 there, but wasn't able to get it to install without some more changes.

The problem I had:
```
  INFO:setup.py:Compiling libvpdq with Make...
      CRITICAL:setup.py:/usr/bin/ld: libvpdq.a(filehasher.cpp.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
      /usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
      collect2: error: ld returned 1 exit status
      make[2]: *** [CMakeFiles/vpdq-hash-video.dir/build.make:111: vpdq-hash-video] Error 1
      make[1]: *** [CMakeFiles/Makefile2:101: CMakeFiles/vpdq-hash-video.dir/all] Error 2
      make: *** [Makefile:103: all] Error 2
```

Random googling says -pthread. I didn't do enough research to tell if I only needed this flag in some circumstances.

Test Plan
---------

```
cd vpdq
pip install .
vpdq /host-home-folder/Videos/test.mp4  # See hashes
...
```

### DIFF
--- a/vpdq/.gitignore
+++ b/vpdq/.gitignore
@@ -8,6 +8,9 @@
 *.swp
 *.swo
 
+build/
+cpp/libraries-dirs.txt
 cpp/test
 cpp/build
 output-hashes/*
+python/vpdq.cpp
--- a/vpdq/cpp/CMakeLists.txt
+++ b/vpdq/cpp/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.17)
 
-set(CMAKE_CXX_FLAGS "-O2 -fPIC -pthread -Wall -Wextra -Werror -Wno-unused-function -Wno-deprecated-declarations")
-
+set(CMAKE_CXX_FLAGS "-O2 -fPIC -Wall -Wextra -Werror -Wno-unused-function -Wno-deprecated-declarations")
 project(VPDQ)
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(LIBAV REQUIRED IMPORTED_TARGET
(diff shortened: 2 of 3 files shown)


ianwal:
Argh..

Can you try this attached patch? I think we can threading it with CMake directly. Probably need to link Cython with pthread as well.

Not sure why this error doesn't happen on the CI build or any machine I've tested.

[linkwithpthread.txt](https://github.com/facebook/ThreatExchange/files/13999414/linkwithpthread.txt)

Dcallies:
Not sure either. I don't understand enough about what I'm doing to weigh in on why. I'm on a goofy windows machine so do most of my work in docker devcontainers. https://github.com/facebook/ThreatExchange/blob/main/open-media-match/.devcontainer/Dockerfile

Your patch worked, so replacing the PR contents with that + gitignore.

ianwal:
Glad to hear it works.

I would recommend bumping vpdq PyPI version to v0.2.1 to fix any issues others have installing it as well.

Sorry for the inconvenience! I'm still learning CMake and most of it I've learned from just reading projects and stuff at work, so there is definitely room for improvement.

Dcallies:
No worries, that's basically the same way I've been learning it. Thanks for being responsive!

PR #1542, opened 2024-02-08
[vpdq] Remove unused constant

milesmcc:
Small quick fix. Resolves a compilation error when `-Werror` is used.

Summary
---------
There's an unused variable that causes a compilation error when `-Werror` is used. The constant doesn't appear to be used anywhere else.


Yisu123:
你好，我是瞿文政，你的邮件我已经收到，我会尽快回复。

PR #1565, opened 2024-03-21
HMA: Enable gunicorn task scheduling

### LINKED ISSUE #1566: Flask background task scheduler doesn't run if werkzeug is not the WSGI server
In order to prevent the apscheduler from executing twice in debug mode, hasher-matcher-actioner checks whether WERKZEUG_RUN_MAIN=="true" before starting the scheduler, because in debug mode the Flask service will initialize twice - werkzeug will set that environment variable only for the "outer" process, meaning that if werkzeug is the WSGI server this is a reliable way to only run the scheduler once in debug mode. However, the "production" version of HMA uses gunicorn as the WSGI server. Since gunicorn doesn't set that enviornment variable, the production version of HMA will incorrectly not trigger the scheduler.

jagraff:
Fixes #1566

In order to prevent the apscheduler from executing twice in debug mode, hasher-matcher-actioner checks whether WERKZEUG_RUN_MAIN=="true" before starting the scheduler, because in debug mode the Flask service will initialize twice - werkzeug will set that environment variable only for the "outer" process, meaning that if werkzeug is the WSGI server this is a reliable way to only run the scheduler once in debug mode. However, the "production" version of HMA uses gunicorn as the WSGI server. Since gunicorn doesn't set that enviornment variable, the production version of HMA will incorrectly not trigger the scheduler. This PR fixes this by adding a check for debug mode before checking for WERKZEUG_RUN_MAIN in the environment.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/app.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/app.py
@@ -119,7 +119,9 @@ def create_app() -> flask.Flask:
         # DO NOT RUN multiple workers with TASK_FETCHER=True or TASK_INDEXER=True -
         # running multiple instances of these tasks may cause database conflicts
         # or other undesireable behavior
-        if (_is_werkzeug_reloaded_process() or _is_gunicorn()) and not running_migrations:
+        if (
+            _is_werkzeug_reloaded_process() or _is_gunicorn()
+        ) and not running_migrations:
             now = datetime.datetime.now()
             scheduler = dev_apscheduler.get_apscheduler()
             scheduler.init_app(app)
--- a/hasher-matcher-actioner/src/OpenMediaMatch/app.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/app.py
@@ -48,6 +48,10 @@ def _is_werkzeug_reloaded_process():
     return os.environ.get("WERKZEUG_RUN_MAIN") == "true"
 
 
+def _is_gunicorn():
+    return "gunicorn" in os.environ.get("SERVER_SOFTWARE", "")
+
+
 def _setup_task_logging(app_logger: logging.Logger):
     """Clownily replace module loggers with our own"""
     fetcher.logger = app_logger.getChild("Fetcher")
@@ -106,9 +110,16 @@ def create_app() -> flask.Flask:
     scheduler: APScheduler | None = None
 
     with app.app_context():
+        # For Werkzeug/debug deployments:
         # We only run apscheduler in the "outer" reloader process, else we'll
         # have multiple executions of the the scheduler in debug mode
-        if _is_werkzeug_reloaded_process() and not running_migrations:
+        #
+        # For Gunicorn/production deployments:
+        # There is currently no check for multiple schedulers running.
+        # DO NOT RUN multiple workers with TASK_FETCHER=True or TASK_INDEXER=True -
+        # running multiple instances of these tasks may cause database conflicts
+        # or other undesireable behavior
+        if (_is_werkzeug_reloaded_process() or _is_gunicorn()) and not running_migrations:
             now = datetime.datetime.now()
             scheduler = dev_apscheduler.get_apscheduler()
             scheduler.init_app(app)


hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -97,7 +97,7 @@ def create_app() -> flask.Flask:
     with app.app_context():
         # We only run apscheduler in the "outer" reloader process, else we'll
         # have multiple executions of the the scheduler in debug mode
```
Dcallies, line 99: You may consider adding a note about gunicorn production code, and the dangers/desirability of running multiple schedulers there.

Dcallies (approved):
I think this may start the scheduler in every sub-process, which could be a problem for the curator tasks, but may be desirable for the indexing classes.

Multiple curator tasks running (fetcher, build index, etc) can lead to problems, and I don't think I was sufficiently defensive in those tasks to detect and prevent it. 

Since you are the first person that is likely to run into this, take care to manage your omm_config and production deployment to avoid ending up with multiple curator tasks!

Dcallies (changes requested):
It looks like the CI changes might be related to these changes, can you confirm?

Dcallies:
Saw your updates, let me know when you are ready to review

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -96,12 +101,26 @@ def create_app() -> flask.Flask:
...
+        or app.config.get("TASK_FETCHER", False)
+    ):
+        # Only one Indexer and Fetcher task can be run in gunicorn workers -
+        # running those tasks in multiple workers can cause database conflicts
+        #
+        # Error out if gunicorn is our server and we think we've been misconfigured
+        raise RuntimeError("Cannot run Indexer or Fetcher tasks in gunicorn ROLE worker - must be run in separate deployments")
+
     with app.app_context():
+        scheduler = dev_apscheduler.get_apscheduler()
```
Dcallies, line 119: What is the purpose of this change? If we aren't going to init the scheduler, why get it?

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -96,12 +101,26 @@ def create_app() -> flask.Flask:
...
     scheduler: APScheduler | None = None
 
+    if _is_gunicorn() and (
+        app.config.get("ROLE_HASHER", False)
+        or app.config.get("ROLE_MATCHER", False)
+        or app.config.get("ROLE_CURATOR", False)
+    ) and (
+        app.config.get("TASK_INDEXER", False)
+        or app.config.get("TASK_FETCHER", False)
+    ):
```
Dcallies, line 111: blocking questions: Instead, should the check be against number of workers? These tasks are safe to run as long as long as there is only one thread.

I'm hesitant to add more invariants on the startup that might block reasonable setups.

hasher-matcher-actioner/docker-compose.prod.yaml
```diff
@@ -0,0 +1,71 @@
...
+    restart: unless-stopped
+    ports:
+      - 8081:8080
+    environment:
+      OMM_CONFIG: /build/omm_config.py
+      PRODUCTION: "true"
+      ROLE_HASHER: "false"
+      ROLE_MATCHER: "false"
+      ROLE_CURATOR: "false"
+      TASK_INDEXER: "true"
```
Dcallies, line 23: blocking: This feels like it's heading in the wrong direction for the default setup, and we might be in the territory of something that might be deployment-specific.

Instead, here's what I propose we do:
1. Land #1574 to create more reference configs by base
2. Add support for using environment variables via https://flask.palletsprojects.com/en/2.3.x/config/#configuring-from-environment-variables
3. In the default docker-compose in this file, we can either pick a worker role by default or another prod=True setup
4. In your own deployment in your infra, you should not use this docker file, and instead have your own.

Dcallies (changes requested):
This feels like it's heading a step in the wrong direction. Having the defaults be a combination of file and environment variable inputs makes it more difficult for me to understand the expectations of the docker file, the compose, the omm_config, and what customizations will need to occur inside of each deployment.

I could use some help understanding the expectations we should set of the docker file that is in the base directory for your needs in deployment. Let's chat offline.

Dcallies (approved):
Seems reasonable, we can iterate on detecting misconfiguration or improvement to server bootstrapping as needed.

Issue #1568, opened 2024-03-22
[hma][py-tx] Unclean CI status for HMA and pytx


Dcallies:
Blocking some of the new PRs, throwing onto oncall work.

HMA: https://github.com/facebook/ThreatExchange/actions/runs/8331418249/job/22798281148?pr=1563

Looks like mypy updates picking up small brokenness.

py-tx: https://github.com/facebook/ThreatExchange/actions/runs/8395103331/job/22993645426?pr=1567

VPDQ is failing to build, looks like it might be related to #1331 cc @ianwal

ianwal:
Yes, it looks like the dependencies for vpdq are not installed. They are listed in one of the .md files (will link later).

Is there a proper package manager that could be added for vpdq that is used at Meta, such as Conan? That would allow vpdq to manage its own build dependencies, preferably automatically installing them as needed. vpdq has no runtime dependencies.

Dcallies:
Thanks a ton both @ianwal  and @Jemeljanov for looking into this one.

I've merged #1569, and will do something shortly to confirm it fixed things.

---
In response to other thoughts:

> Is there a proper package manager that could be added for vpdq that is used at Meta, such as Conan

This project is unfortunately limited by what tools I've used personally. So far, that has meant setup.py and pip, but even that is already outdated. I've had limited bandwidth so far this year to make improvements to our code due to competing priorities on my team at Meta, but making the repo more maintainable is a goal I'm tracking.

> Alternatively, we could just publish wheels for vpdq to avoid having to even build in most cases. I think [cibuildwheels](https://github.com/pypa/cibuildwheel) would work and be fairly simple to setup as an action.

I'll also check this out as well, I'm going to open a new issue (#1571) for exploration of both of these.

Dcallies:
Confirmed that #1569 Fixed py-tx build!

aryzle:
Since old runs linked above don't show the logs anymore, here is the latest one
py-tx: https://github.com/facebook/ThreatExchange/actions/runs/9308541738/job/25622160252

HMA looks fine

PR #1569, opened 2024-03-23
Fix threatexchange-ci vpdq build, bump action versions

ianwal:
Summary
---------

- Fix vpdq CI for #1568 by installing dependencies.

- Bump CI action versions
  - `vpdq-ci-python.yaml` runs these new versions so might as well upgrade now that we're touching this file.

Let me know if you think of something better. Not too experienced with Github Actions, so maybe there is a way to avoid duplicating this. Alternatively, we could shove this in a bash file in vpdq directory and run that instead.

Test Plan
---------

Ran it on my fork and vpdq successfully builds. Other format (unrelated to vpdq) format errors still cause the overall workflow to fail, though.

### DIFF
--- a/.github/workflows/python-threatexchange-ci.yaml
+++ b/.github/workflows/python-threatexchange-ci.yaml
@@ -24,10 +24,13 @@ jobs:
     name: Lint and Type-Check python
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Dcallies (approved):
Thanks again @ianwal! Sorry for the delay, I'm looking at both this one and #1570

PR #1574, opened 2024-04-08
[hma] Move reference omm_configs

Dcallies:
Summary
---------

Inspired some of @jagraff's questions, get rid of a single omm_config in the outer directory, and instead force people to select a config with no obvious default. Add many comments to explain. 

Test Plan
---------

pytest && mypy

Let's see what CI says on the docker config, which I was too lazy to test directly.


jagraff:
Overall I agree with this PR, makes sense to force users to think about what should be in their config.

Optionally, for us our omm_config reads most things from environment variables so that we can use a single omm config file for both worker and curator deploys, would an example config like that make sense? See here: https://github.com/facebook/ThreatExchange/pull/1565/files#diff-eb032e1392cdc142041bc9b033dc800fe9d491783c8cb8f6854cdf54c73403a9

hasher-matcher-actioner/reference_omm_configs/scaled_deployment_hasher_omm_config copy.py
jagraff, line 1: Is the word " copy" intended to be in the filename?

jagraff:
I like the reference configs for each type of worker. I do wonder if database config stuff should be in a separate file from role configs? They're sort of two separate concerns, and it's likely that database creds will be shared across an entire deployment, unlike the role configs which are unique to each worker type

hasher-matcher-actioner/reference_omm_configs/scaled_deployment_hasher_omm_config copy.py
Dcallies: Nope, thanks for the look!

Dcallies:
> I like the reference configs for each type of worker. I do wonder if database config stuff should be in a separate file from role configs? They're sort of two separate concerns, and it's likely that database creds will be shared across an entire deployment, unlike the role configs which are unique to each worker type

All of the workers need access to the database to read the persistent settings.

That being said, I'll add env support so you don't need one!

Dcallies:
Discussed offline with @jagraff, will fast follow with some config improvements.

PR #1585, opened 2024-05-08
[HMA] Docker Release GHCR action

juanmrad:
Summary
---------

This PR adds the version.txt file to HMA as well as the github workflow code to release the docker image into the github container registry.

Test Plan
---------

Released image will be available `ghcr.io/facebook/ThreatExchange/hma:___version___`

Was only able to test up to build. 🤞🏻 push works once merged.


hasher-matcher-actioner/version.txt
```diff
@@ -0,0 +1 @@
+1.0.0
```
Dcallies, line 1: Was debating whether we should use a < 1.0.0 version, do you have an opinion?

Dcallies (approved):
Let's see what happens!

hasher-matcher-actioner/version.txt
juanmrad: I was close to do 0.0.1 lol, but also figured you may have a more strong opinion. Considering there seems to be users already with it in prod I think it is on a good state for it to be considered the first release tho.

PR #1592, opened 2024-05-22
Redirect to the homepage.

### LINKED ISSUE #1591: [hma] Bad redirects on dev page
The dev handlers (e.g. `/dev/factory_reset`) redirect to `/dev/` which doesn't exist. It makes it seem like the action failed, when it's just the redirect that is bad. Example here: https://github.com/facebook/ThreatExchange/blob/d4d343c435bb048610de91087e608c6ebcb24923/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/development.py#L104

I think it can be done with `redirect(url_for(ui.bp.home))`? Let me give it a try and send a PR.

sbruens:
Summary
---------

Fix the dev banner redirects. Fixes #1591

Test Plan
---------

Verified by running locally.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/development.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/development.py
@@ -6,7 +6,7 @@ Development only routes for easily testing functionality end-to-end, all running
 
 import itertools
 from flask import Blueprint, abort, request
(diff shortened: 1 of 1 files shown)


Dcallies (approved):
Wrote this on the issue as well, but I likely broke this when I refactored this out of /ui/. Thanks for the fix!

PR #1598, opened 2024-06-12
[HMA] Fix Content Vars missing for template

juanmrad:
Summary
---------

When testing HMA found that we have an issue on the templates missing variables. added it.

Test Plan
---------

Before: 
Screenshot 2024-06-11 at 9 08 39 PM

After:
Screenshot 2024-06-11 at 9 07 17 PM

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/templates/components/add_to_bank_form.html.j2
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/templates/components/add_to_bank_form.html.j2
@@ -14,10 +14,11 @@
         </div>
         <div class="col">
             <select class="form-select" name="content_type" required>
+                <option selected disabled value="">Select content type</option>
                 {% for c in content %}
-                {% if c['enabled'] %}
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/templates/components/add_to_bank_form.html.j2
```diff
@@ -14,10 +14,11 @@
         </div>
         <div class="col">
             <select class="form-select" name="content_type" required>
+                <option selected disabled value="">Select content type</option>
                 {% for c in content %}
-                {% if c['enabled'] %}
-                <option value="{{c['name']}}">{{c['name'].capitalize()}}</option>
-                {% endif %}
+                    {% if c['enabled'] %}
```
Dcallies, line 19: I found a library to auto-format these, I should add it to the devcontainer so we can just let it handle it.

Dcallies (approved):

PR #1600, opened 2024-07-04
[vpdq] Modernize CMake Usage

ianwal:
Summary
---------

Changes are splitting CMake files to their respective directories. This is the preferred way to structure modern CMake projects.

This allows easier modifications to some parts of the project without breaking another.

- Update examples and CI to use modern CLI commands.
  - `cmake -S . -B build` replaces `mkdir build && cmake ..`.
    - Command was added in CMake 3.12, which was released in 2018. [Ubuntu 20.04 ships 3.16](https://packages.ubuntu.com/search?keywords=cmake) so it's old enough to use in the examples to me.

  - [`cmake --build build -j`](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project) replaces `make`

- Move CLI programs to new directory, apps

- Add CMake build types for sanitizers
  - This makes it much easier to run sanitizers locally and through the CI.
  
- Link Python binding with pdq since it is now a separately built library

- Add another stage in the CPP CI for a no sanitizer regtest
  - This only adds 1 minute to the run time and is useful in case there is a problem with the sanitizers.

- Fix benchmark.py
  - This wasn't working before. Updated new CLI apps path. Removed old ffmpeg filepath argument. Removed dead imports. 

I've been learning proper CMake at work and wanted to apply it to a more complex project. This should significantly simplify the work needed for future changes to the project, such as adding unit tests or bindings.

And hopefully it can serve as a reference to others since there is a lack of proper CMake projects out there, especially with Cython.

Test Plan
---------

Tested with container, CI and Arch Linux the Python binding and cpp through regtest.

I building on an M2 Pro, but there are some other unrelated issues in the implementation that causes compilation to fail. I will address these issues in a separate PR. After fixing those problems this CMake setup works on macos + Clang.


.github/workflows/vpdq-ci-cpp.yml
```diff
@@ -30,28 +30,34 @@ jobs:
...
         run: |
           python cpp/regtest.py -i ${GITHUB_WORKSPACE}/tmk/sample-videos
 
-      - name: Make with thread,undefined sanitizer
+      - name: Make with address sanitizer
         run: |
           rm -rf cpp/build
-          mkdir cpp/build
-          cd cpp/build
-          cmake .. -DCMAKE_CXX_FLAGS="-O2 -fPIC -Wall -Wextra -Werror -Wno-unused-function -Wno-deprecated-declarations -fsanitize=thread,undefined" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,leak,undefined"
```
ianwal, line 52: The linker flags in this were wrong. It should be `thread` not `address` Not entirely sure how it was working.

vpdq/benchmark/benchmark.py
```diff
@@ -123,17 +111,15 @@ def run_python(
 
 
 def run_cpp(
-    ffmpegPath: str,
     secondsPerHash: int,
     downsampleFrameDimension: int,
 ):
+    Path.mkdir(OUTPUT, exist_ok=True)
```
ianwal, line 117: Need this directory or the hashes won't be output. This avoids having to make it manually.

vpdq/cpp/CMakeLists.txt
```diff
@@ -1,16 +1,76 @@
...
+    "Flags used by the C++ compiler for Asan build type or configuration." FORCE)
+
+set(CMAKE_EXE_LINKER_FLAGS_TSAN
+    "${CMAKE_EXE_LINKER_FLAGS_DEBUG} -fsanitize=thread,undefined" CACHE STRING
+    "Linker flags to be used to create executables for Tsan build type." FORCE)
+
+set(CMAKE_SHARED_LINKER_FLAGS_TSAN
+    "${CMAKE_SHARED_LINKER_FLAGS_DEBUG} -fsanitize=thread,undefined" CACHE STRING
+    "Linker lags to be used to create shared libraries for Tsan build type." FORCE)
+endif()
```
ianwal, line 65: Not going to pretend that I fully understand this block, but it makes sense at a high level and it does work.

Dcallies (approved):
Thanks for continuing to update the README as you go. 

I have little relevant expertise, but find your overall test plan convincing that it's at least neutral. 

I assume you've run with all the sanitizer builds and they are all clean.

PR #1603, opened 2024-07-11
[HMA] Fix Create Exchange API Selector

### LINKED ISSUE #1601: [hma][ui] Create exchange API selector borked
Yet another tab broken by my refactor. Reported by @jagraff

aryzle:
Summary
---------

Fixes #1601  API Select in the Create Exchange form

also fixes issue with startup.sh script, devcontainer was failing before

Test Plan
---------

before
image

after
image

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/startup.sh
+++ b/hasher-matcher-actioner/.devcontainer/startup.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 export OMM_CONFIG=/workspace/reference_omm_configs/development_omm_config.py
-MIGRATION_COMMAND=1 flask --app OpenMediaMatch.app db upgrade --directory /workspace/src/OpenMediaMatch/migrations
-flask --app OpenMediaMatch.app run --debug
+MIGRATION_COMMAND=1 flask --app src/OpenMediaMatch.app db upgrade --directory /workspace/src/OpenMediaMatch/migrations
+flask --app src/OpenMediaMatch.app run --debug
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/ui.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/ui.py
@@ -135,9 +135,11 @@ def exchanges():
     """
     Exchange management page
     """
-    return render_template(
-        "bootstrap.html.j2", page="exchanges", collabs=_collab_info()
-    )
+    template_vars = {
+        "exchange_apis": _api_cls_info(),
+        "collabs": _collab_info(),
+    }
+    return render_template("bootstrap.html.j2", page="exchanges", **template_vars)
 
 
 @bp.route("/match")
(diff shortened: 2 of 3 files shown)


hasher-matcher-actioner/.devcontainer/startup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 export OMM_CONFIG=/workspace/reference_omm_configs/development_omm_config.py
-MIGRATION_COMMAND=1 flask --app OpenMediaMatch.app db upgrade --directory /workspace/src/OpenMediaMatch/migrations
-flask --app OpenMediaMatch.app run --debug
+MIGRATION_COMMAND=1 flask --app src/OpenMediaMatch.app db upgrade --directory /workspace/src/OpenMediaMatch/migrations
```
Dcallies, line 4: @juanmrad  - can you double check that this is correct? This script should be running with a working directory

Dcallies (approved):
Sorry for the incredibly slow review, I've been out on leave for the last week! 

The only thing I'm unsure about is the change to the startup script. I thought the devcontainer settings would help with that. When you say it was failing, was this failing a clean rebuild in the created terminal window (and not running manually)?

I can ask @juanmrad since I'll be seeing him in person.

Dcallies:
If you are bored while I'm waiting for Juan's answer, a way to make this less fragile in the future would be to add a simple test (probably a new test) to hit the /ui/ blueprints.

hasher-matcher-actioner/.devcontainer/startup.sh
juanmrad: Yes it should be running on the working directory. Where you having issues when running the shell script outside of dev container? or where you running it inside?

Dcallies (changes requested):
@aryzle - Can you include more information about the error you were experiencing so we can make sure this is the right fix?

hasher-matcher-actioner/.devcontainer/startup.sh
aryzle: hey @juanmrad this was failing on the dev container startup
```
Running the postAttachCommand from devcontainer.json...

[4812 ms] Start: Run in container: /bin/sh -c /workspace/.devcontainer/startup.sh
Error: Could not import 'OpenMediaMatch.app'.
```
this is after opening vscode in the `hasher-matcher-actioner` dir and selecting "Reopen in Container"
image

juanmrad: Seems like I was able to repro on a whole new build 🤔. Not sure how this may have changed to cause the problem.

Dcallies: Seems like Juan is convinced, lets land it. I'll re-test on windows just to make sure there's not any evil differences between different vscodes.

Dcallies (approved):

PR #1605, opened 2024-07-17
[vpdq] Fix macOS build

ianwal:
Summary
---------

Clang doesn't like the forward declaration of the templates, so move them to the header file where they should belong.

The current way might actually be ill-formed, but GCC is pretty lax about template stuff in my experience so that's why it may be working.

### Changes

- Fix Clang link error

- Fix deprecated warning for AVCodecContext::frame_number. This warning is present on the version of libav libraries that ship with Ubuntu 24.04.

- Remove unnecessary `std::bind` for spawning hashing threads in VpdqHasher ctor

  - This is not required to fix the build but generally much better practice. Since we're touching this area I made this change now.
 
  - This is the only change to the code while moving from `hasher.cpp` to `hasher.h`
 
  - See Herb Sutter's recommendation for this https://github.com/isocpp/CppCoreGuidelines/issues/1796

Test Plan
---------

Local build and CI passes successfully.

Build and regtest pass on M2 Pro macos Sonoma with Clang with FFmpeg 7.0 installed with brew.


Dcallies (approved):
Apologies for the long delay, I've been traveling.

Thanks again for the fixes!

PR #1618, opened 2024-09-09
[py-tx] Validate hash length

### LINKED ISSUE #1609: [py-tx]Threatexchange CLI PDQ index doesn't validate hash length
There is a corrupt entry in the StopNCII data: 
`('pdq', '00000000000000000000000000000000')`

While it's probably good for StopNCII to remove this data from their upstream as well, it would be nice for threatexchange CLI to ignore corrupt data when rebuilding indexes.

I hit error on this line:
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/pdq/pdq_faiss_matcher.py#L240

but there is similar usage on this line as well:
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/pdq/pdq_faiss_matcher.py#L185

aryzle:
Summary
---------

fixes #1609

validates hash length in SignalExchangeAPIWithSimpleUpdates (used by stopNCII, file, etc)

Test Plan
---------

reproduce error by adding `"00000000000000000000000000000000"` to the list of example PDQ hashes and try to run `tx fetch`
```
...
Building index for pdq with 17 signals...
  File "/home/vscode/.local/bin/tx", line 33, in <module>
    sys.exit(load_entry_point('threatexchange', 'console_scripts', 'tx')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 321, in main
    inner_main()
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 314, in inner_main
    execute_command(settings, namespace)
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/cli/main.py", line 157, in execute_command
    command.execute(settings)
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/cli/fetch_cmd.py", line 163, in execute
    DatasetCommand().execute_generate_indices(settings)
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/cli/dataset_cmd.py", line 292, in execute_generate_indices
    index = index_cls.build(signals.items())
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/signal_type/index.py", line 216, in build
    ret.add_all(entries)
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/signal_type/pdq/pdq_index.py", line 75, in add_all
    self.index.add(
  File "/workspaces/ThreatExchange/python-threatexchange/threatexchange/signal_type/pdq/pdq_faiss_matcher.py", line 240, in add
    self.faiss_index.add_with_ids(numpy.array(vectors), numpy.array(i64_ids))
...
```
after rebuilding `tx` with the fix in `SignalExchangeAPIWithSimpleUpdates`, error goes away as the offending entry is ignored (note the count 17 went down to 16)
```
...
Building index for pdq with 16 signals...
Index for pdq ready
...
```

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
@@ -10,7 +10,7 @@ import typing as t
 
 import dacite
 import requests
-from requests.packages.urllib3.util.retry import Retry
+from urllib3.util.retry import Retry
 from threatexchange.exchanges.clients.utils.common import TimeoutHTTPAdapter
 
 
--- a/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
+++ b/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
@@ -340,7 +340,7 @@ class SignalExchangeAPIWithSimpleUpdates(
                 logging.warning(
                     "Invalid fingerprint (%s): %s",
                     s_type.get_name(),
-                    signal_str,
+                    signal_str if len(signal_str) < 100 else signal_str[:100] + "...",
                 )
                 continue
             inner = ret.get(s_type)
--- a/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
+++ b/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
@@ -6,6 +6,8 @@ The SignalExchangeAPI talks to external APIs to read/write signals
 @see SignalExchangeAPI
 """
 
+import logging
+
 from abc import ABC, abstractmethod
 import typing as t
 
@@ -331,6 +333,15 @@ class SignalExchangeAPIWithSimpleUpdates(
         for (type_str, signal_str), metadata in fetched.items():
             s_type = type_by_name.get(type_str)
             if s_type is None or metadata is None:
+                continue
+            try:
+                s_type.validate_signal_str(signal_str)
+            except ValueError:
+                logging.warning(
+                    "Invalid fingerprint (%s): %s",
+                    s_type.get_name(),
+                    signal_str,
+                )
                 continue
             inner = ret.get(s_type)
             if inner is None:
--- a/python-threatexchange/threatexchange/signal_type/pdq/signal.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/signal.py
@@ -109,4 +109,6 @@ class PdqSignal(
             "c186f9619659658c4b6916daf934496985f4259c72c39676ab9edb631c3c349c",
             # Github octocat - https://github.com/logos
             "62c9789627838f69f23cf98c29e37836d61c87c31c59587c4b49a783f496725c",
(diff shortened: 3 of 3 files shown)


Dcallies:
For test plan, can you confirm if you were able to generate a successful reproduction by adding an invalid length hash in the PDQ samples?

aryzle:
@Dcallies not yet, I'm trying to get a unit test going, but I'll do a manual test too

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
```diff
@@ -10,7 +10,7 @@
 
 import dacite
 import requests
-from requests.packages.urllib3.util.retry import Retry
+from urllib3.util.retry import Retry
```
aryzle, line 13: Pylance was complaining about this, any reason to import from `requests.packages`?

aryzle:
seems like some tests do catch this 😂 will dig into them later today

python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
Dcallies: Probably earlier confusion from us on what was the correct way to import them. Fine with this change.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -332,6 +334,15 @@ def naive_convert_to_signal_type(
             s_type = type_by_name.get(type_str)
             if s_type is None or metadata is None:
                 continue
+            try:
+                s_type.validate_signal_str(signal_str)
+            except ValueError:
+                logging.warning(
+                    "Invalid fingerprint (%s): %s",
+                    s_type.get_name(),
+                    signal_str,
```
Dcallies, line 343: blocking: This string could be quite long, suggest truncating it and doing a "..."

aryzle:
@Dcallies I commented out the bad entry [here](https://github.com/facebook/ThreatExchange/pull/1618/files#diff-fda1271858fb6b256de15f49632e10c464e3a85cd13ef613368b1bfba952775dR113) because a few tests depended on it (`DatasetCommandTest` and `test_utils` in `test_vpdq_faiss.py` because `VPDQSignal.get_examples` calls `PdqSignal.get_examples`) and didn't make sense to include a bad entry in `get_examples` when it's expected to have only valid examples. We can def refactor later to handle this, or make another method that returns mixed/dirty examples to test on.

Dcallies (approved):
Hmm, would like to have a regression test that would exercise this condition, but you've been at it for a while, let's merge this one.

Issue #1610, opened 2024-08-06
[py-tx] Implement Tech Against Terrorism Hash List API


Dcallies:
Looks like TAT add a hash exchange API, we should add a reference client, here's the documentation: https://www.terrorismanalytics.org/docs/hash-list-v1

A good 1:1 comparison for how to do this is from StopNCII:

If you want to give a crack at it, you can follow the same approach we did for StopNCII

1. Add a reference version for a client for the TAT hash api. Here’s StopNCII’s: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/clients/stopncii/api.py
    1. We should add tests
1. Add an “[Exchange](https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py)” interface implementation, which tells the library how to download a copy of all hashes. Here’s StopNCII’s: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/stop_ncii_api.py
    1. You can test to see it works correctly using the `tx` commandline interface to download and then match signals
1. Potentially add implementations for "SHA256", "SHA512", cryptographic types, and a generic "File" content type

Dcallies:
This is thoroughly done :P 

Thanks @Bruce-Pouncey-TAT !

PR #1622, opened 2024-09-10
[pytx] Tech Against Terrorism SignalTypeAPI Implementation

Bruce-Pouncey-TAT:
Summary
---------
Hey @Dcallies 

Here is a draft PR for part 2 of the [issue_1610](https://github.com/facebook/ThreatExchange/issues/1610)

I'm looking for some support on a few key details of my implementation.

**JSON File download** 
Given our approach to delivering the hash list is differently from StopNCII and NCME I have wondered how you want to do this . In the draft I am simply using a `get` to.. 

1. get the tat_api response
2. use the `file_url` in the response to fetch the pre-signed JSON file 
3. write it to a local temp file
4. load the temp file into memory 
5. yield the `FetchDelta` instances. 

I wanted to check this with you before I continued with this approach.

The code for this logic is in `impl/techagainstterrorism_api.py` lines: 148-161

**TATSignalMetadata**
I'm stumped at this time as to what we want to put into `TATSignalMetadata` or my understanding of it is not quite there yet. Outside of the response from the API and along side the JSON file URL we have the fields: `file_name` `created_on` `total_hashes` `ideology` . 

We also have additional information in each hash entry, see below what each hash entry in the JSON file contains.

```
[
   ...
   {
      "hash_digest": "12345abcde",
       "algorithim": "MD5"
       "ideology": "far-right"
       "file_type": "jpg"
   }
   ...
]

```

Am I right to assume we want this information lifting up into the `TATSignalMetadata` ?
or
Is this Metadata that gets populated when a match lookup takes place ? 

Thanks!


python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+    The Hash List is updated nightly and new hashes are appended to the list.
+
+    For more information on our collection process please visit: https://terrorismanalytics.org/about/how-it-works
+    For our Hash List documentation: https://terrorismanalytics.org/docs/hash-list-v1
+    """
+
+    last_fetch_time: int
+
+    def is_stale(self) -> bool:
```
Dcallies, line 45: Does this ever truly stale out though? 

Some API's, like threatexchange have time-limited deltas, which if you don't fetch within that time, the dataset becomes invalid. It sounds like you always get the whole thing in that one file, no?

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+        return time.time() - self.last_fetch_time > 3600 * 24
+
+    @classmethod
+    def from_tat_fetch(cls, response: api.TATHashListResponse) -> "TATCheckpoint":
+        return cls(int(response.created_on.timestamp()))
+
+
+@dataclass
+class TATSignalMetadata(state.FetchedSignalMetadata):
+    """Not sure what should live here"""
```
Dcallies, line 58: This would be where tags would be if your API supports them. I'm guessing this a broadcast-only API from TAT -> consumers, with no feedback mechanisms? If so, there might not be anything here.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+    pass
+
+
+@dataclass
+class TATCredentials(auth.CredentialHelper):
+    ENV_VARIABLE: t.ClassVar[str] = "TX_TAT_CREDENTIALS"
+    FILE_NAME: t.ClassVar[str] = "~/.tx_tat_credentials"
```
Dcallies, line 66: I think the `TX_` in the threatexchange API refers to ThreatExchange the API. To avoid confusion, I'd use `PYTX_` as the full prefix or just leave it at `TAT_CREDENTIALS`

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+    def fetch_iter(
+        self,
+        _supported_signal_types: t.Sequence[t.Type[SignalType]],
+        checkpoint: t.Optional[TATCheckpoint],
+    ) -> t.Iterator[
+        state.FetchDelta[t.Tuple[str, str], TATSignalMetadata, TATCheckpoint]
+    ]:
+        """
+         The TAT Hash List returns a pre-signed URL to a JSON file containing a list of hashes.
+         As well as some metadata about the hash list and information about the hashes themselves including the requested ideology.
```
Dcallies, line 130: nit: line length limit to ~80 characters please.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+        3. Read the JSON file and load into memory
+
+        4. Yield the delta mapping
+        """
+
+        client = self.get_client()
+        result = client.get_hash_list()
+
+        # Download the JSON file from the presigned URL
+        response = requests.get(result.file_url)
```
Dcallies, line 145: confus: Why are we calling requests directly? Your client should hide the file URL details. Suggest instead you have another function on your client that will fetch the hash list, or since here we unconditionally get the file anyways, why doesn't get_hash_list just return the content?

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+        response.raise_for_status()
+
+        with tempfile.NamedTemporaryFile(delete=True) as f:
+            f.write(response.content)
+            f.close()
+
+            # Read the JSON file and load into memory
+            with open(f.name, "r") as file:
+                data = json.load(file)
```
Dcallies, line 154: Lol, you don't need to write it to a file and then read it. Just use `loads` instead: https://docs.python.org/3/library/json.html

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+                translated = (_get_delta_mapping(entry) for entry in data)
+                yield state.FetchDelta(
```
Dcallies, line 157: You don't need to keep your file pointers open all this time, you can yield outside of your with blocks to close the resources.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,168 @@
...
+                    dict(t for t in translated if t[0][0]),
+                    checkpoint=TATCheckpoint.from_tat_fetch(result),
+                )
+
+
+def _get_delta_mapping(
+    record: api.TATHashRecord,
+) -> t.Tuple[t.Tuple[str, str], t.Optional[TATSignalMetadata]]:
+
+    metadata = None  # Unsure at this time
```
Dcallies, line 167: Return `None` instructs the library to delete the record - you just want an empty record here instead. Ref: https://github.com/Bruce-Pouncey-TAT/ThreatExchange/blob/4d3ce4e41d8e2d9184f0fe194e1f82d0bc833a10/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py#L168

Dcallies (changes requested):
> Am I right to assume we want this information lifting up into the TATSignalMetadata ?

Yup, that's where that lives. 

Make sure to rebase your PR on main so it doesn't contain your previous PR in the change contents!

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Dcallies: Thinking about this slightly harder, if you don't record removals to support incremental fetch, the answer is the same here as it is for the "File" signal type API - the fetch is always stale!

Dcallies (commented):
Additionally, for your simplest test plan:
Assume python-threatexchange CLI is aliased to `tx`

1. Add Tech Against Terrorism in the default list of SignalTypeAPIs so that it will show up when you list exchanges on the CLI
2. `tx collab edit <whatever is needed to set up TAT>`
3. `tx fetch`
4. Confirm you have real data by doing `tx dataset`
5. Match a piece of content by doing `tx match ...`
6. Make sure it has sane re-fetch behavior by doing `tx fetch` again

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Bruce-Pouncey-TAT: That's right the hash list is only a broadcast API. Will `pass` be what we want here ? I know `SignalExchangeAPIWithSimpleUpdates` and `SignalExchangeAPI` both expect it to be passed.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Bruce-Pouncey-TAT: Our hash list doesn't support an incremental fetch. So yes I think we can assume the fetch is always stale. In that case I will return `True`

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Bruce-Pouncey-TAT: To follow up, we do have a hash verification service, but this may be a follow up piece of work as my PM specifically mentioned Hash List and this feature is currently time boxed.

I would be happy to return in the future and implement the hash verification.

Bruce-Pouncey-TAT:
Hey @Dcallies 

I have pushed the first working version of the TAT cli!

### Manual test 
Using docker assuming a blank start

0. Build tx
`docker build --tag threatexchange .`

1. Create credentials
`docker run -v $HOME/.threatexchange:/var/lib/threatexchange threatexchange config api tat --credentials '<TCAP_USERNAME>' '<TCAP PASSWORD>'`

2. Create collab config
`docker run -v $HOME/.threatexchange:/var/lib/threatexchange threatexchange config collab edit tat --create "TAT"`

3. Fetch hash list with verbose output
`docker run -v $HOME/.threatexchange:/var/lib/threatexchange threatexchange -v fetch`

4. View dataset information
`docker run -v $HOME/.threatexchange:/var/lib/threatexchange threatexchange dataset`

To test sane fetching run step 3 and 4 again and you will see that output of `dataset` has not changed (no new hashes or new hasn't been appended to old). We don't support incremental fetching.

### Automatic testing
`pytest`

**A couple of thoughts** 🤔 
Our hash list contains the following algorithms for each piece of content 
`MD5` `SH256` `PDQ` `SHA512` `SHA256` for every media type.

I could only find `VideoMD5Signal` and `PdqSignal` and was hesitant to start creating new Signals.

python-threatexchange/threatexchange/exchanges/clients/fb_threatexchange/api.py
```diff
@@ -16,7 +16,7 @@
 
 import requests
 from requests.packages.urllib3.util.retry import Retry
-from threatexchange.exchanges.clients.utils.common import TimeoutHTTPAdapter
+from ..utils.common import TimeoutHTTPAdapter
```
Dcallies, line 19: blocking: I've had too many painful experiences with relative imports, and now always recommend absolute imports.

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/api.py
```diff
@@ -90,14 +90,19 @@ def _get_session(self, auth_token: t.Optional[str] = None):
...
+    ) -> t.Any:
         """
         Perform an HTTP GET request, and return the JSON response payload.
 
         Same timeouts and retry strategy as `_get_session` above.
         """
         with self._get_session(auth_token) as session:
-            url = self.BASE_URL + endpoint
+            url = full_url if full_url else self.BASE_URL + (endpoint or "")
             response = session.get(url=url)
```
Dcallies, line 106: code golf: You might find this more readable

```
if not full_url:
  full_url = self.BASE_URL + (endpoint or "")
with self._get_session(auth_token) as session:
  response = session.get(url=full_url)
```

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/api.py
```diff
@@ -121,7 +126,7 @@ def get_auth_token(self, username: str, password: str) -> t.Optional[str]:
 
         if not isinstance(username, str) or not isinstance(password, str):
```
Dcallies, line 127: nit: This is what typing is for, you don't need to check it at runtime. You might be missing a class-level annotation, but most modern typecheckers intuit from `__init__`

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/api.py
```diff
@@ -133,19 +138,24 @@ def get_auth_token(self, username: str, password: str) -> t.Optional[str]:
 
     def get_hash_list(
         self, ideology: str = TATIdeology._all.value
-    ) -> TATHashListResponse:
+    ) -> t.List[t.Dict[str, str]]:
```
Dcallies, line 141: 👍

python-threatexchange/threatexchange/exchanges/impl/tests/test_techagainstterrorism.py
```diff
@@ -0,0 +1,86 @@
...
+def test_get_credential_cls():
+    assert TATSignalExchangeAPI.get_credential_cls() == TATCredentials
+
+
+def test_get_name(monkeypatch):
+    monkeypatch.setattr(
+        "threatexchange.exchanges.impl.techagainstterrorism_api._API_NAME",
+        "test_api_name",
+    )
+    assert TATSignalExchangeAPI.get_name() == "test_api_name"
```
Dcallies, line 41: blocking: This is a pretty tautological test. Why not just check `get_name()` against the string "tat" ?

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,129 @@
...
+from threatexchange.exchanges import signal_exchange_api
+from threatexchange.exchanges import auth
+from threatexchange.exchanges.collab_config import (
+    CollaborationConfigWithDefaults,
+)
+from threatexchange.signal_type.signal_base import SignalType
+from threatexchange.signal_type.pdq.signal import PdqSignal
+from threatexchange.signal_type.md5 import VideoMD5Signal
+
+_API_NAME: str = "tat"
```
Dcallies, line 20: blocking: It's more confusing to have this outside than inlined. I think in the place where you are copying it from, it is being reused somewhere.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -0,0 +1,129 @@
...
+    record: t.Dict[str, str],
+) -> t.Tuple[t.Tuple[str, str], t.Optional[state.FetchedSignalMetadata]]:
+
+    if not _is_compatible_signal_type(record):
+        return (("", ""), None)
+
+    type_str = _type_mapping().get(record["algorithm"])
+
+    metadata = state.FetchedSignalMetadata()
+    return ((type_str or "", record["hash_digest"]), metadata)
```
Dcallies, line 129: blocking: You are returning photo MD5 hashes as VideoMD5s here.

python-threatexchange/threatexchange/exchanges/impl/tests/test_techagainstterrorism.py
```diff
@@ -0,0 +1,86 @@
...
+        mock_get_delta_mapping,
+    )
+
+    result = list(api_instance.fetch_iter([], None))
+    assert len(result) == 1
+    assert isinstance(result[0], state.FetchDelta)
+    assert result[0].checkpoint == state.NoCheckpointing()
+    assert result[0].updates == {
+        ("signal_type", "signal_value"): {"id": 1, "data": "test_data"}
+    }
```
Dcallies, line 86: blocking: This test is mocking so much that you aren't really testing any of the logic, which is true of much of the tests in this function.

You probably only need to patch exactly one functionality - the return of the client.get_hashes(), which you should be able to copy the shape of a few records from your real API (you can replace the hashes with hashes from test files, or from the example signal types.

Dcallies (changes requested):
> A couple of thoughts 🤔
> Our hash list contains the following algorithms for each piece of content
MD5 SH256 PDQ SHA512 SHA256 for every media type.
> 
> I could only find VideoMD5Signal and PdqSignal and was hesitant to start creating new Signals.

It is fairly straightforward to add more cryptographic hash types. I am of the opinion that cryptographic hashing on image types is a waste of time, but we did have a version of pytx that had it in the past. For video, In general MD5 has been the most common denominator for video hashes.

Some questions for you:
1. You said that you support PDQ for every piece of content. How are you handling that for videos?

My recommendation is to start with just image PDQ and video MD5, and if you have a user who wants support, they are easy to add later.

On your PR:
You are almost there! There is one bug with the wrong signal types being returned in some cases, and I think you should reduce the amount of monkeypatching to just only the client.

I think the NCMEC test does this the best: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/tests/test_ncmec.py

It has a mocked client that is hardcoded to return certain values from it's get() methods, and then the exchange is hardcoded to use the mocked client.

Bruce-Pouncey-TAT:
Hi! @Dcallies 

I have made the changes you requested - thank you for the support and feedback on the tests. I have used the existing pytest api fixture in `clients/techagainstterrorism/tests/test_api.py ` line:43 which will return the right data for testing and has the key methods for this to work already mocked.

And to answer your question. My mistake - we don't support PDQ for videos. We use TMK for videos and PDQ for images, for perceptual hashing. Every piece of content we hash will have 4 hashes, `MD5`, `SHA256`, `SHA512`, `TMK` or `PDQ`

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/tests/test_api.py
```diff
@@ -2,31 +2,37 @@
...
+            "algorithim": "MD5",
+            "ideology": ideology,
+            "file_type": "jpg",
+        },
+        {
+            "hash_digest": "12345abcde",
+            "algorithim": "MD5",
+            "ideology": ideology,
+            "file_type": "jpg",
+        },
```
Dcallies: ignorable: I can't tell whether ot not these are changing between cases, you can could put this in a global to reference multiple times during the test

Dcallies (approved):
All of my blocking comments are addressed, so I accept! If there are more fixes we can do them in future PRs.

Thanks again for all of your hard work @Bruce-Pouncey-TAT !

PR #1625, opened 2024-09-24
Add required copyright headers to files missing it

Dcallies:
Summary
---------

The internal linter is complaining again, so do a fixup. There were a couple of tricky cases
1. jinja template files, which I chose to implement via `{# Comment #}`
2. A section of vpdq declared in the public domain in #1508. I am not a lawyer, so I tried to pick a compromise solution, @Ianwal seems to have left github so can't ask him his opinion.

Test Plan
---------

Let's see if CI is happy, and then I'll just view the HMA UI to make sure it still loads

### DIFF
--- a/Dockerfile.vpdq
+++ b/Dockerfile.vpdq
@@ -1,3 +1,4 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
 # Dockerfile for development of vpdq. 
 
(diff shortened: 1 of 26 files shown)


Dcallies:
Well @ianwal, seems like you can like comments, but your profile and all previous PRs are hidden!

PR #1632, opened 2024-09-30
add endpoint to delete a content_signal from a bank

prenner:
Summary
---------

- add functionality to delete a content_signal from a bank
- there's already a implementation for `bank_remove_content` that I utilized

Test Plan
---------
- tested locally by adding something to a bank and then hitting the new endpoint to delete it

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -212,6 +212,19 @@ def _bank_add_signals(
     }
 
 
+@bp.route("/bank/<bank_name>/content/<content_id>", methods=["DELETE"])
+def bank_delete_content(bank_name: str, content_id: int):
+    """
+    Remove a signal from a bank.
+    """
+    storage = persistence.get_storage()
+    bank = storage.get_bank(bank_name)
+    if not bank:
+        abort(404, f"bank '{bank_name}' not found")
+
+    storage.bank_remove_content(bank.name, content_id)
+    return {"message": "Done"}
+
 @bp.route("/bank/<bank_name>/signal", methods=["POST"])
 def bank_add_as_signals(bank_name: str):
     """
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -222,8 +222,9 @@ def bank_delete_content(bank_name: str, content_id: int):
     if not bank:
         abort(404, f"bank '{bank_name}' not found")
 
-    storage.bank_remove_content(bank.name, content_id)
-    return {"message": "Done"}
+    count = storage.bank_remove_content(bank.name, content_id)
+    return {"deleted": count}
+
 
 @bp.route("/bank/<bank_name>/signal", methods=["POST"])
 def bank_add_as_signals(bank_name: str):
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
@@ -566,11 +566,14 @@ class DefaultOMMStore(interface.IUnifiedStore):
         sesh.commit()
         return content.id
 
-    def bank_remove_content(self, bank_name: str, content_id: int) -> None:
-        database.db.session.execute(
-            delete(database.BankContent).where(database.BankContent.id == content_id)
+    def bank_remove_content(self, bank_name: str, content_id: int) -> int:
+        result = database.db.session.execute(
+            delete(database.BankContent)
+            .where(database.BankContent.id == content_id)
+            .where(database.BankContent.imported_from_id.is_(None))
         )
         database.db.session.commit()
+        return result.rowcount
 
     def get_current_index_build_target(
         self, signal_type: t.Type[SignalType]
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
@@ -567,10 +567,9 @@ class DefaultOMMStore(interface.IUnifiedStore):
         return content.id
 
     def bank_remove_content(self, bank_name: str, content_id: int) -> int:
+        # TODO: throw an exception if deleting imported content
         result = database.db.session.execute(
-            delete(database.BankContent)
-            .where(database.BankContent.id == content_id)
-            .where(database.BankContent.imported_from_id.is_(None))
+            delete(database.BankContent).where(database.BankContent.id == content_id)
         )
         database.db.session.commit()
         return result.rowcount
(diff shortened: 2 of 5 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +212,19 @@ def _bank_add_signals(
...
+    """
+    Remove a signal from a bank.
+    """
+    storage = persistence.get_storage()
+    bank = storage.get_bank(bank_name)
+    if not bank:
+        abort(404, f"bank '{bank_name}' not found")
+
+    storage.bank_remove_content(bank.name, content_id)
+    return {"message": "Done"}
```
prenner, line 226: i took this from another endpoint but happy to make this more specific

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -153,6 +153,19 @@ def test_banks_add_hash(client: FlaskClient):
...
+    image_url = "https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg?raw=true"
+
+    create_bank(client, bank_name)
+    add_hash_to_bank(client, bank_name, image_url, 1)
+
+    post_response = client.delete(
+        f"/c/bank/{bank_name}/content/1"
+    )
+
+    assert post_response.status_code == 200
```
Dcallies, line 167: we should add one more check to make sure the content is actually gone (though I notice this is the case with other tests as well)!

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
Dcallies: I think this is fine, I can't remember why we added it.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +212,19 @@ def _bank_add_signals(
...
 
+@bp.route("/bank/<bank_name>/content/<content_id>", methods=["DELETE"])
+def bank_delete_content(bank_name: str, content_id: int):
+    """
+    Remove a signal from a bank.
+    """
+    storage = persistence.get_storage()
+    bank = storage.get_bank(bank_name)
+    if not bank:
+        abort(404, f"bank '{bank_name}' not found")
```
Dcallies, line 223: blocking: One thing we might need to do here is protect import banks, whose content are managed by the exchange config. If you were planning on using this to move content between exchange configs, then we'll be working on this exact problem for the Child safety hackathon!

Dcallies (changes requested):
Thanks a ton for submitting a PR @prenner ! This looks like an oversight, bank content is expected to have the whole create/update/delete span.

There are two things I think we need still:
1. blocking - We should make sure the unittest has some way to confirm that the content was actually removed
2. Fine as followup - we should protect banks whose contents are managed by an exchange config, since otherwise I think the fetcher/syncer will undo the operation.

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
prenner: Yeah I had started to do this but after rebuilding the index in the test i'm getting a 503. It's also happening with test_banks_add_hash_index:

```
    def test_banks_add_hash_index(app: Flask, client: FlaskClient):
        bank_name = "NEW_BANK"
        bank_name_2 = "NEW_BANK_2"
        image_url = "https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg?raw=true"
        image_url_2 = "https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/c.png?raw=true"

        # Make two banks and add images to each bank
        create_bank(client, bank_name)
        add_hash_to_bank(client, bank_name, image_url, 1)
        create_bank(client, bank_name_2)
        add_hash_to_bank(client, bank_name, image_url_2, 2)

        storage = get_storage()
        # ensure index is empty to start with
        assert storage.get_signal_type_index(PdqSignal) is None

        # Build index
        build_all_indices(storage, storage, storage)

        # Test against first image
        post_response = client.get(
            f"/m/raw_lookup?signal_type=pdq&signal={IMAGE_URL_TO_PDQ[image_url]}"
        )
>       assert post_response.status_code == 200
E       assert 503 == 200
E        +  where 503 = <WrapperTestResponse streamed [503 SERVICE UNAVAILABLE]>.status_code
```

Let me know if you thinkn it's just a set up issue though

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
prenner: ah this is specifically for non imported banks.

I updated the `bank_remove_content` method but let me know if you want me to split this out into a different method

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
Dcallies: I think this is a test issue - it might be related to your build_all_indices, though you can trace (via prints) why it's hitting the 503, but I suspect its because it doesn't think it has the newly built indices ready. 

You don't need to check whether or not a match will work - it's enough that the banking API thinks its gone.

Dcallies (changes requested):

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
```diff
@@ -566,11 +566,14 @@ def bank_add_content(
...
 
-    def bank_remove_content(self, bank_name: str, content_id: int) -> None:
-        database.db.session.execute(
-            delete(database.BankContent).where(database.BankContent.id == content_id)
+    def bank_remove_content(self, bank_name: str, content_id: int) -> int:
+        result = database.db.session.execute(
+            delete(database.BankContent)
+            .where(database.BankContent.id == content_id)
+            .where(database.BankContent.imported_from_id.is_(None))
         )
```
Dcallies, line 574: blocking: Hmm, not sure about this change - would rather it throw a exception rather than silently fail.

I think you can leave this off for now and just add a TODO to throw the exception - the bank as a whole should be marked as import or not so we shouldn't need to check on a row-by-row basis.

prenner: got it, will update!

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
prenner: there's not a get content_id within curation.py so

1) is checking the # of deleted rows in the response sufficient
2) (assuming no, since we want to check if the content is still in the db) do you want me to add a separate GET method that we can use outside of requiring an index?

let me know if im missing something!

Dcallies (approved):

PR #1639, opened 2024-10-02
add similarity to raw_lookup endpoint

prenner:
Summary
---------
- write tests for raw_lookup (some tests are 503'ing?)
- add to lookup
- add to anywhere where we try and find relevant matches
- poke holes in api design

Test Plan
---------
to run api tests:
```
docker exec -it app /bin/bash
pip install .
pip install pytest
cd OpenMediaMatch
pytest tests/
```


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -95,14 +95,14 @@ def raw_lookup():
...
-     * List of matching content items
+     * List of matching with content_id and distance values
     """
     signal = require_request_param("signal")
     signal_type_name = require_request_param("signal_type")
     return lookup_signal(signal, signal_type_name)
 
 
-def lookup_signal(signal: str, signal_type_name: str) -> dict[str, list[int]]:
+def lookup_signal(signal: str, signal_type_name: str) -> dict[str, dict[str, str]]:
```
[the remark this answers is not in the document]
prenner: agreed, probably the right thing to do is version the api change

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -118,7 +119,15 @@ def lookup_signal(signal: str, signal_type_name: str) -> dict[str, list[int]]:
...
-    return {"matches": [m.metadata for m in results]}
+    return {
+        "matches": [
+            {
+                "content_id": m.metadata,
+                "distance": m.similarity_info.pretty_str(),
+            }
+            for m in results
+        ]
+    }
```
prenner, line 130: this is also breaking the /lookup endpoint

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -300,9 +327,17 @@ def index_cache_is_stale() -> bool:
 
 def _get_index(signal_type: t.Type[SignalType]) -> SignalTypeIndex[int] | None:
     entry = _get_index_cache().get(signal_type.get_name())
+
+    if entry is not None and is_in_pytest():
+        entry.reload_if_needed(get_storage())
```
Dcallies, line 332: blocking q: Hmm, I'm very suspicious of any branching in the code that specifically changes behavior in tests.

Is there another solution that you can use in the test itself that can achieve the same results? (e.g. disabling background loading in the fixture for the test? Manually calling a load_all_indices() function?)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -300,9 +327,17 @@ def index_cache_is_stale() -> bool:
...
     if entry is None:
         current_app.logger.debug("[lookup_signal] no cache, loading index")
         return get_storage().get_signal_type_index(signal_type)
     if entry.is_ready:
         return entry.index
     return None
+
+
+def is_in_pytest():
+    return "pytest" in sys.modules
```
Dcallies, line 343: Danger! I don't think this works consistently, and someone could accidentally include a test library in the build that silently changes the behavior of HMA. This is why I am generally suspicious of this kind of branching logic!

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -95,14 +97,19 @@ def raw_lookup():
      * Signal value (the hash)
      * Optional list of banks to restrict search to
     Output:
-     * List of matching content items
+     * List of matching with content_id and, if included, distance values
     """
     signal = require_request_param("signal")
     signal_type_name = require_request_param("signal_type")
-    return lookup_signal(signal, signal_type_name)
+    include_distance = bool(request.args.get("include_distance", False)) == True
```
Dcallies, line 104: This works for me as a solution, though I think this bool parsing is different than what we do for other flags.

blocking: Can you include it in the list of inputs in the docstring?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -95,14 +97,19 @@ def raw_lookup():
...
+    lookup_signal_func = (
+        lookup_signal_with_distance if include_distance else lookup_signal
+    )
 
+    return lookup_signal_func(signal, signal_type_name)
```
Dcallies, line 109: nit: Both of these functions return the same shape, it might be slightly easier to have this return the list and do


```
return {
  "matches": lookup_signal_func()
}
```

Dcallies (changes requested):
Nice job! This ended up being simpler than I had thought it would be.

Tests are not expected to 503! 

I believe the only path that can lead to 503 is if the index is not available when you try and query it. Are you saying they are 503-ing in main or just when you make the changes?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
prenner: ha agreed, glad we found the root cause!

PR #1642, opened 2024-10-03
Fix Broken UI due to lookup change

juanmrad:
Summary
---------

Fixes UI that broke due to return type change on lookup function

Test Plan
---------

Screenshot 2024-10-03 at 12 33 44 PM


Dcallies (changes requested):
run black.

plz test

Dcallies (approved):

PR #1647, opened 2024-10-03
Update Graph API Version

suvamsh:
Summary
---------

Update to graph API version 21

Test Plan
---------
`tx config api fb_threatexchange -L1` produces correct output

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
@@ -11,8 +11,9 @@ import logging
 import time
 import typing as t
 from dataclasses import dataclass, field
-
+from typing import Self
 from threatexchange.exchanges.clients.ncmec import hash_api as api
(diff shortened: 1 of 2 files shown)


python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -11,8 +11,9 @@
 import time
 import typing as t
 from dataclasses import dataclass, field
-
+from typing import Self
```
Dcallies, line 14: blocking: Sadly, as we discovered, we won't be able to support Self until 2026!

Dcallies (changes requested):

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
suvamsh: accidentally committed, reverted.

Dcallies (approved):

PR #1649, opened 2024-10-03
remove dev:query_media function as it's no longer used

prenner:
Summary
---------
remove dev/query_media function as it's no longer used. existing, non dev endpoints have the same functionality

Test Plan
---------

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/development.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/development.py
@@ -4,17 +4,11 @@
 Development only routes for easily testing functionality end-to-end, all running on a single host.
 """
 
-import itertools
-from flask import Blueprint, abort, request
+from flask import Blueprint
 from flask import redirect, url_for
 from werkzeug.exceptions import HTTPException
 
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/development.py
```diff
@@ -14,17 +14,14 @@
...
 )
-from OpenMediaMatch.utils.flask_utils import api_error_handler
+from OpenMediaMatch.utils.flask_utils import api_error_handler, str_to_bool
 
 from OpenMediaMatch.utils import dev_utils
 from OpenMediaMatch import persistence
 from OpenMediaMatch.storage.postgres.flask_utils import reset_tables
 
 
-from threatexchange.exchanges.fetch_state import CollaborationConfigBase
```
prenner, line 24: unused

juanmrad: We don't use the whole `query_media` function I don't see it being used anywhere.

Dcallies (approved):
CI = green; I merge

PR #1651, opened 2024-10-03
Add image hash distance

levi-cloudflare:
Summary
---------

Adds distance between two image hashes on the match debug page.

Screenshot 2024-10-03 at 4 09 07 PM
Screenshot 2024-10-03 at 4 09 18 PM

Test Plan
---------

- Upload two images, verify that the distance is displayed
- Replace one of the images, verify that the distance is re-calculated


Dcallies (approved):
From the file I looked it, it appears to be the right programming language!

/s

PR #1653, opened 2024-10-03
[hma][api] Add option to return banks in matching API

andrew-dillon:
Summary
---------

Include the list of matching banks in the `/m/raw_lookup&include_distance=true` response.

Test Plan
---------

Updated the `src/OpenMediaMatch/tests/test_e2e_workflow.py` test.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_e2e_workflow.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_e2e_workflow.py
@@ -70,4 +70,4 @@ def test_raw_hash_add_to_match_with_distance(app: Flask, client: FlaskClient):
         f"/m/raw_lookup?signal_type=pdq&include_distance=true&signal={hashes[-1]}"
     )
     assert resp.status_code == 200
-    assert resp.json == {"matches": [{"content_id": 16, "distance": "0"}]}
+    assert resp.json == {"matches": [{"content_id": 16, "distance": "0", "banks": ["TEST_BANK"]}]}


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -39,7 +39,7 @@
 class MatchWithDistance(t.TypedDict):
     content_id: int
     distance: str
-
+    banks: list[str]
```
Dcallies, line 42: blocking: A content_id can only ever be in one bank, so I think this should be similar.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -151,11 +151,13 @@ def lookup_signal(signal: str, signal_type_name: str) -> list[int]:
 def lookup_signal_with_distance(
     signal: str, signal_type_name: str
 ) -> list[MatchWithDistance]:
+    banks = lookup(signal, signal_type_name)
     results = query_index(signal, signal_type_name)
```
Dcallies, line 155: blocking: This is essentially causing us to double query to the index, since lookup will query the index, and query_index also queries the index.

Instead, what you want to be doing is taking the returned content_ids and doing a lookup on which (singular) bank they are in, which you can essentially see how to do in the body of lookup.

Dcallies (changes requested):
Hey @andrew-dillon, thanks for contributing to HMA! I think having the ability to lookup content by banks is valuable, though if you explain how you plan on using this in production we can make sure you get the very best version of this functionality.

As written, you will cause a performance regression in the lookup API, because resolving the banks requires another roundtrip to the database. Additionally, the way you are calling it in this PR will duplicate the index lookup! We want to try and maintain a very fast query option for the index since it's essentially the key service of HMA!

Based on my guess of how you might want to use this functionality, here's what I recommend:
1. Reply with the higher level problem you want to solve
2. Do not query the index twice, and keep the `_SignalIndexInMemoryCache` as lightweight as possible - it's used by other functions!
3. At the API where you wish to receive the banks  Only resolve the banks if the API receives an optional parameter, similar to how this is handled on L117.

Dcallies:
No worries! We do need to do some work on the match APIs, @juanmrad and I have been talking about it.

PR #1660, opened 2024-10-11
Update hello world instructions

aokolish:
Summary
---------

Just a small update to the Hello World exercise to make it easier if anyone else tries to do it.

Test Plan
---------

n/a

### DIFF
--- a/hasher-matcher-actioner/CONTRIBUTING.md
+++ b/hasher-matcher-actioner/CONTRIBUTING.md
@@ -92,6 +92,12 @@ def hello_world():
     return "Hello, world!\n"
 ```
 
+If your server is running, stop it by pressing Ctrl+C. Then, start the server and rebuild the image to pick up the changes:
+
+```bash
+docker compose up --build
+```
+
 Retry the `curl localhost:5000/hello`, you should now see your greeting message come back instead of the 404 error.
 
 ## Self-managed Python dev environment


hasher-matcher-actioner/CONTRIBUTING.md
```diff
@@ -92,6 +92,12 @@ def hello_world():
     return "Hello, world!\n"
 ```
 
+If your server is running, stop it by pressing Ctrl+C. Then, start the server and rebuild the image to pick up the changes:
+
+```bash
+docker compose up --build
+```
+
```
Dcallies, line 100: Hmm, this isn't true for the recommended development (dev containers), which will automatically restart without the need for this build. 

We can keep a reminder in here if you aren't using dev containers, but it needs to be caveated e.g. by adding a subheader or note, something like:


---

Let's add the endpoint. In `src/OpenMediaMatch/app.py`, add a route to the Flask app like so:
```python
@app.route("/hello")
def hello_world():
    return "Hello, world!\n"
```

Then save the file! If you are using the recommended development environment with a devcontainer, flask will detect a changed file and automatically restart the server, picking up your changes!

**If you are not using a devcontainer** you might need to stop your server and rebuild the image to pick up changes with `docker compose up --build`. This is a lot slower than using the devcontainer!

Dcallies (changes requested):

hasher-matcher-actioner/CONTRIBUTING.md
aokolish: Oh, I see. I first followed the README instructions which said to use `docker compose up`. Then, found this file and also I don't use VSCode so I skipped that for now. I'll give it a shot today.

Dcallies: The CONTRIBUTING is target at developers, and the hello world is the test to see if you have your development environment set up correctly :P

aokolish: Heh, closing.

PR #1662, opened 2024-10-16
Update contact addresses from @fb -> @meta

### LINKED ISSUE #1615: [py-tx][ez] Update contact addresses from @fb -> @meta
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/setup.py#L46

Still has the @fb.com address, which we should update to @meta.com

16BitNarwhal:
Summary
---------

change all contact info from @fb.com to @meta.com with grep
`grep -rl "@fb.com" . | xargs sed -i 's/@fb.com/@meta.com/g'`

fixes #1615 

Test Plan
---------
n/a

### DIFF
--- a/api-reference-examples/curl/te-tag-query/README.md
+++ b/api-reference-examples/curl/te-tag-query/README.md
@@ -25,7 +25,7 @@ lot.
 
 # Contact
 
-threatexchange@fb.com
+threatexchange@meta.com
 
 # Setup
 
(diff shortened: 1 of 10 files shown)


Dcallies (approved):
Sorry for the long delay in review, I've been totally underwater. 

Nice job on coming up with an approach to comprehensively find all the references!

PR #1669, opened 2024-10-27
[py-tx] Add rotation reference to brute force lookup

### LINKED ISSUE #1665: [py-tx][mlh] Add rotation reference to brute force lookup
This is a small-to-medium project. 

You will learn about:
1. Rotating images
2. the PDQ algorithm
3. The brute force & distance interfaces in SignalType

Images have multiple "primitive" rotations:
1. Rotate 90: counterclockwise 90 degrees
1. Rotate 180: 180 degrees
1. Rotate 270: counterclockwise 270 degrees (i.e. clockwise 90 degrees)
1. FlipX: Left is left and right is right but top and bottom change places
1. FlipY: Top is top and bottom is bottom but left and right change places (mirror image)
1. FlipPlus1: Upper left and lower right stay put; lower left and upper right exchange places
1. FlipMinus: Upper right and lower left stay put; upper left and lower right exchange places

Some algorithms like pdq can generate multiple rotations at once as they hash. You can see the rotations implemented here: https://github.com/facebook/ThreatExchange/blob/main/pdq/cpp/hashing/pdqhashing.cpp#L440-L456

# Goal
The reference brute force approach in `tx match` should also try rotations, and pick the minimum distance. Which rotation it used should be returned in the distance string

### Add a new mixin for hashing rotations
This is similar to what we did for many exceptional cases for SignalType, we can add a new class called `SignalTypeWithRotations` which has a `.hash_rotations(file_ptr)

```
h1 = pdq.hash(image)
queries = PDQSignal.hash_all_rotations(image.flipx())
best = queries[0], PDQSignal.compare(queries[0], h1), Rotation.ORIGINAL
foreach queries as q, rotation:
  d = PDQSignal.compare(q, h1)
  if d < best[0]:
    best = q, d, r 

print(best)
<is match is true, distance is 0[flip_x]>
```

Note that this doesn't (yet) affect the index matching, which we will save for later.

## Code
* You are mostly going to be here: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/pdq/signal.py#L65-L75

## Tests
* Add unittests that test that all rotations are sane, and consider exposing accessors to them in PDQSignalType, which refer to the underlying pdqhash library
* Include outputs from `tx match` or the HMA upload debugger to demonstrate the interface is being used correctly

haianhng31:
Summary
---------

* Fixes #1665 
Test Plan
---------

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_hash_rotations.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_hash_rotations.py
@@ -1,0 +1,102 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import numpy as np
+from enum import Enum, auto
+
+class RotationType(Enum):
+    ORIGINAL = auto()
+    ROTATE90 = auto()
+    ROTATE180 = auto()
+    ROTATE270 = auto()
+    FLIPX = auto()
+    FLIPY = auto()
+    FLIPPLUS1 = auto()
+    FLIPMINUS1 = auto()
+
+class PDQHashRotations:
+    @staticmethod
+    def _hash_to_matrix(hash_str: str) -> np.ndarray:
+        """Convert a 256-bit PDQ hash string to 16x16 float matrix."""
+        # Convert hex string to binary
+        binary = bin(int(hash_str, 16))[2:].zfill(256)
+        # Convert to 16x16 matrix of floats
+        return np.array([float(int(b)) for b in binary]).reshape(16, 16)
+    
+    @staticmethod
+    def _matrix_to_hash(matrix: np.ndarray) -> str:
+        """Convert 16x16 float matrix back to PDQ hash string."""
+        # Flatten matrix and convert to binary string
+        binary = ''.join(['1' if x > 0 else '0' for x in matrix.flatten()])
+        # Convert to hex
+        return hex(int(binary, 2))[2:].zfill(64)
+
+    @classmethod 
+    def rotate90(cls, matrix: np.ndarray) -> np.ndarray: 
+        result = np.zeros((16, 16))
+        for i in range(16):
+            for j in range(16):
+                result[i][j] = -matrix[i, j] if j & 1 else matrix[i, j]
+        return result
+
+    @classmethod
+    def rotate180(cls, matrix: np.ndarray) -> np.ndarray:
+        result = np.zeros((16, 16))
+        for i in range(16):
+            for j in range(16):
+                result[i][j] = -matrix[i, j] if (i + j) & 1 else matrix[i, j]
+        return result
+    
+    @classmethod
+    def rotate270(cls, matrix: np.ndarray) -> np.ndarray:
+        result = np.zeros((16, 16))
+        for i in range(16):
+            for j in range(16):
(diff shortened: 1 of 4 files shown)


Dcallies (changes requested):
Thanks for working on this! I like a lot of the interface proposals, including the new rotation enum.

I spoke to Eric offline, who had look into the PDQ algorithm, and I think there might be a problem with the approach that is hard to fix, though I think you have made some of the right supporting interface.

Namely, my assertion that you can "rotate" a PDQ hash may not actually be true. The rotation code you have copied evaluates rotations on the DCT matrix and not the PDQ hash itself. The output of the DCT -> PDQ looks like it might be destroying information, and so there might not be a valid equivalence from image -> rotate -> hash and image -> hash -> rotate.

We can still get the same effects as rotating the hash, but we may have to do it in more steps (we can discuss here async or over a meeting).
1. Simplest option - move rotating the image as a helper in PhotoContent, then just hash all 8 rotations (we may want to do this anyways), but we will need to go in more steps.
2. Requires research - Using solution #1 or by mathing out the matrix work + torbens in the PDQ code, we can check to see whether there are actually valid rotations of the PDQ hash itself, and then resuming from there.

In either case, I think we should retain your proposed enum, and add some functionality 


Top level changes suggested in current version of code:
1. Let's avoid adding it to the base-level interface, because this doesn't apply to text, and since it's not universal I don't think we should add it just yet - instead add it to PDQ distance
2. For your test plan, look into how we might demonstrate this using either the `hash` or `match` commands.

haianhng31:
> The output of the DCT -> PDQ looks like it might be destroying information, and so there might not be a valid equivalence from image -> rotate -> hash and image -> hash -> rotate.

I've just written and tried to run the test cases with compare_hash using the same pictures (with one rotated) and the match result is false. So I confirm that rotating the hash doesn't work. 

I will now try rotating the image in PhotoContent instead and will update later if this approach works. I will also do some research to see if there's a way to rotate the hash itself

Dcallies:
Thanks for submitting this as a draft @haianhng31  - I often recommend engineers on my team to "have the discussion with code" - it's much easier to talk about concrete things, and I can also give you some pointers on other things you've written.

haianhng31:
@Dcallies Please help me confirm if this is the file I should be working with to rotate the image before hashing & matching and if this approach sounds good: 

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/cli/match_cmd.py#L219-L225
* I'm thinking that in the _match_file function, I will check whether the type is photo. If yes, i will rotate the object then brute force it with index.query

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/content_type/photo.py
* I will add the rotation helpers to this file

Dcallies:
Let's break this up into multiple steps / PRs.

1. PR one should add the helper functions in PhotoContent using pillow. It doesn't need a unittest, and we'll test it in PR 2
2. PR 2, we should add a new command line option (`--rotations`) to turn each photo file added into 8 photos that are all hashed
3. P3, we should update the match function to also take a `--rotations`, that builds on the code we used in PR2 to query all 8 rotations.

PR #1671, opened 2024-10-28
[py-tx] Create rotation helper functions in PhotoContent using Pillow

haianhng31:
Summary
---------
This is a part of the issue #1665
This has been discussed previously in PR #1669 

* Create RotationType enum in content_base.py 
  * Use StrEnum to make the enum’s values independent of the order
* Create 6 helper functions in PhotoContent 
  * 5 functions are to rotate the image in different angles 
  * 1 function is to try all different rotations 

Test plan
---------
@dcallies said not to write a unittest at this stage.

However, I was able to show that the code is correct by writing a simple bit of python:

```
from threatexchange.content_type.photo import PhotoContent

file_path = pathlib.Path("threatexchange/tests/hashing/resources/LA.png")
with open(file_path, "rb") as f:
    image_data = f.read()

all_rotations = PhotoContent.all_simple_rotations(image_data)

for rotation_type, image_data in all_rotations.items():
    with open(f"image_{rotation_type}.png", "wb") as w:
        w.write(image_data)
```

Then I visually inspected all the files to confirm they were rotated as expected.

### DIFF
--- a/python-threatexchange/threatexchange/content_type/content_base.py
+++ b/python-threatexchange/threatexchange/content_type/content_base.py
@@ -34,17 +34,19 @@ class ContentType:
         """
         return []
 
+
 class RotationType(StrEnum):
     """
     Enum for 8 simple rotations of an image.
-    Used to store all generated rotations of an image, 
+    Used to store all generated rotations of an image,
     whose algorithms don't have a native way to generate rotations during hashing.
     """
-    ORIGINAL = auto() # No rotation; the object is in its original orientation
-    ROTATE90 = auto() # Rotates the object 90 degrees 
-    ROTATE180 = auto() # Rotates the object 180 degrees (half-turn)
-    ROTATE270 = auto() # Rotates the object 270 degrees 
-    FLIPX = auto() # Flip the object horizontally along the X-axis 
-    FLIPY = auto() # Flip the object horizontally along the Y-axis 
-    FLIPPLUS1 = auto() # Diagonal flip along the line y = x
-    FLIPMINUS1 = auto() # Diagonal flip along the line y = -x
+
+    ORIGINAL = auto()  # No rotation; the object is in its original orientation
+    ROTATE90 = auto()  # Rotates the object 90 degrees
+    ROTATE180 = auto()  # Rotates the object 180 degrees (half-turn)
+    ROTATE270 = auto()  # Rotates the object 270 degrees
+    FLIPX = auto()  # Flip the object horizontally along the X-axis
+    FLIPY = auto()  # Flip the object horizontally along the Y-axis
+    FLIPPLUS1 = auto()  # Diagonal flip along the line y = x
+    FLIPMINUS1 = auto()  # Diagonal flip along the line y = -x
--- a/python-threatexchange/threatexchange/content_type/content_base.py
+++ b/python-threatexchange/threatexchange/content_type/content_base.py
@@ -7,7 +7,7 @@ Abstraction for different content types.
 This records all the valid signal types for a piece of content.
 """
 
-from enum import StrEnum, auto
+from enum import Enum, auto
 import typing as t
 
 from threatexchange import common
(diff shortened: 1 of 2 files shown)


python-threatexchange/threatexchange/content_type/content_base.py
```diff
@@ -32,3 +33,13 @@ def extract_additional_content(
         * Video => break out photo thumbnail, close caption text, audio
         """
         return []
+
+class RotationType(StrEnum):
```
Dcallies, line 37: ignorable: You can also explain the intention of this enum with a docstring

```
"""
What mysterious purpose could this class represent?
"""
```

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -19,3 +21,75 @@ class PhotoContent(ContentType):
       * frames from videos
       * thumbnails of videos
     """
+    @classmethod
+    def rotate_image(cls, image_data: bytes, angle: float) -> bytes: 
+        """
+        Rotate an image by a given angle
```
Dcallies, line 27: Include a note on why we might want to do this :P 

Namely, this is help for testing hash algorithms.

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -19,3 +21,75 @@ class PhotoContent(ContentType):
...
+        Flip the image diagonally along the line y = -x.
+        """
+        with Image.open(io.BytesIO(image_data)) as img:
+            flipped_img = img.transpose(Image.Transpose.ROTATE_90).transpose(Image.FLIP_LEFT_RIGHT)
+            with io.BytesIO() as buffer:
+                flipped_img.save(buffer, format=img.format)
+                return buffer.getvalue()
+    
+    @classmethod
+    def try_all_rotations(cls, image_data: bytes):
```
Dcallies, line 80: nit: Let's go with `all_simple_rotations`

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -19,3 +21,75 @@ class PhotoContent(ContentType):
...
+        with Image.open(io.BytesIO(image_data)) as img:
+            flipped_img = img.transpose(Image.Transpose.ROTATE_90).transpose(Image.FLIP_LEFT_RIGHT)
+            with io.BytesIO() as buffer:
+                flipped_img.save(buffer, format=img.format)
+                return buffer.getvalue()
+    
+    @classmethod
+    def try_all_rotations(cls, image_data: bytes): 
+        """
+        Try all possible rotations to the image
```
Dcallies, line 82: nit: Something like 
```
"""
Generate the 8 naive rotations of an image.

This can be helpful for testing, and for image algorithms that don't have
a native way to generate rotations during hashing, this can be a way to
brute force rotations.

**Review by Dcallies — CHANGES_REQUESTED** [review]
Looking good!

To test this, on your development environment, create a simple image, and then call, via Ipython the method and write the resultant bytes to files. 

You may find that for some image types, you need to hint pillow on what the format is, which can be done via filename.

You can write the bytes to files and then visually inspect that you have all the rotations! You can include that code in your test plan.

**Comment by haianhng31** [comment]
@Dcallies Thank you for your comments! I have included the docstring where necessary and also tested it on my development environment. If the code is ok, I can start moving to next step which is adding a new command line option (--rotations)

File: python-threatexchange/threatexchange/content_type/photo.py [hunk 5]
```diff
@@ -4,9 +4,11 @@
 """
 Wrapper around the video content type.
 """
+from PIL import Image
+import io
 
 from .content_base import ContentType
-
+from threatexchange.content_type.content_base import RotationType
```
Dcallies, line 11: nit: We shouldn't be using relative imports here, and they are both from the same file.

Dcallies (changes requested):
So close!

Can you:
1. Run `black` so CI passes
2. Confirm that you did something like the test plan?

haianhng31:
I have run it on my development environment and visually inspected all of them to make sure they are right. Can you specify a little bit more about what you meant by test plan? Should I write it in this PR's comment section? 

Here's the exact code I used: 
```
    file_path = pathlib.Path("threatexchange/tests/hashing/resources/LA.png")
    with open(file_path, "rb") as f:
        image_data = f.read()

    all_rotations = PhotoContent.all_simple_rotations(image_data)

    for rotation_type, image_data in all_rotations.items():
        with open(f"image_{rotation_type}.png", "wb") as w:
            w.write(image_data)
```

haianhng31:
@Dcallies it seems like StrEnum cannot be imported with py version before 3.11
Q: Would you like me to remove 3.9 from the workflow? Or would you want me to not use StrEnum and do something else instead?

Dcallies:
Oh no! I totally missed that.

For now, use `Enum`, and then type out the names of the values that would be equivalent to `auto()`. It should just be lowercase

haianhng31:
@Dcallies Fixed! Lmk if you have any other comment

Dcallies (approved):
Nice job!

PR #1672, opened 2024-10-30
[py-tx] Add a new match command line option (--rotations)

haianhng31:
# Summary
---------

* Create a new command line option (--rotation) for matching 
* Create a new dataclass IndexMatchWithRotation
* This is checked out from branch haianhng31:photo-helpers in PR #1671. 
* This is a continuation from PR #1671 (discussed in PR #1669 )

### Files changed: 
* Main changes are in file: match_cmd.py

# Test Plan
* I have included 2 test cases: 
  * [test_non_photo_match_with_rotations](https://github.com/facebook/ThreatExchange/pull/1672/files#diff-c82e2388c19446e2958c3684110f86e1ca872c52f46a2cc5b7ef191390d8c637R42-R48)
  * [test_photo_hash_with_rotations](https://github.com/facebook/ThreatExchange/pull/1672/files#diff-c82e2388c19446e2958c3684110f86e1ca872c52f46a2cc5b7ef191390d8c637R50-R73)
```
$ tx match --rotations text -- "try text"
--rotations flag is only available for Photo content type

$ tx hash --rotations photo pdq/data/bridge-mods/aaa-orig.jpg
$ tx match --rotations photo pdq/data/bridge-mods/aaa-orig.jpg
pdq ORIGINAL 16 (Sample Signals) INVESTIGATION_SEED
$ tx match --rotations photo pdq/data/bridge-mods/aaa-rotated180.jpg
pdq ROTATE180 16 (Sample Signals) INVESTIGATION_SEED
```


python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -51,6 +52,10 @@ class MatchCommand(command_base.Command):
     $ threatexchange match text -- This is my cool text
     ```
 
+    # Additional options: 
+    --rotation: For photo content, generate and match all 8 simple rotations 
+                (0°, 90°, 180°, 270°, flip X, flip Y, flip diagonal +1, flip diagonal -1)
+
```
Dcallies, line 58: blocking: Remove this, we can document the arguments better with argparse features. The rest of this docbloc is meant to explain how the command works in general, not to document specific fields. check out `--help` to see how the options are dispalyed.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -198,31 +217,62 @@ def execute(self, settings: CLISettings) -> None:
...
                     metadatas: t.List[t.Tuple[str, FetchedSignalMetadata]] = r.metadata
                     for collab, fetched_data in metadatas:
                         if not self.all and collab in seen:
                             continue
                         seen.add(collab)
+
+                        # Add rotation information if possible 
+                        rotation_info = ""
+                        if hasattr(r.similarity_info, "rotation"):
+                            rotation_info = f" [{r.similarity_info.rotation.name}]"
```
Dcallies, line 232: While clever, and valid python, this will make typechecking unhappy (try running mypy locally as described in https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/CONTRIBUTING.md).

Instead, change the return type of these functions from IndexMatch to a new object that contains an index match and the rotation info.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -198,31 +217,62 @@ def execute(self, settings: CLISettings) -> None:
...
+    all_matches = []
+
+    for rotation_type, rotated_bytes in rotations.items():
+        # Create a temporary BytesIO object to simulate a file 
+        temp_buffer = io.BytesIO(rotated_bytes)
+        matches = index.query(s_type.hash_from_file(temp_buffer))
+
+        # Add rotation information if any matches were found 
+        for match in matches: 
+            match.similarity_info.rotation = rotation_type
```
Dcallies, line 271: blocking: I think this might be an illegal operation, and I'm unsure whether we should add rotation to the base concept of similarity. 

We have a few options, but the simplest one might be change the return of this function from IndexMatch to instead return a new dataclass that might contains the rotation information. 

I'll accept any solution that:
1. Doesn't change the base interface (since I don't think we yet know how to handle rotations in general - you are pioneering our approach!)
2. Can show which rotation was used in `tx match`.

Dcallies (changes requested):
In the 

For your test plan, showing the ouput of the match function will be critical. Here's what I want to see:

Using https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg

```
# Showing hash of original
$ tx hash photo aaa-orig.jpg
pdq facefacefafce...

# Here we use imageMagick, but any editor will work
$ magick aaa-org.jpg -rotate 90 aaa-90.jpg

# index sample data, which includes aaa-orig
$ tx fetch 

# Now match against rotated
$ tx match photo aaa-90.jpg
pdq sample 0[90º] INGVESTIGATION_SEED

File: python-threatexchange/threatexchange/cli/match_cmd.py [hunk 4]
```diff
@@ -196,33 +214,86 @@ def execute(self, settings: CLISettings) -> None:
             for s_type, index in indices:
                 seen = set()  # TODO - maybe take the highest certainty?
                 if self.as_hashes:
-                    results = _match_hashes(path, s_type, index)
+                    results_from_hashes = _match_hashes(path, s_type, index)
+                    results: t.Sequence[IndexMatchWithRotation] = [
+                        IndexMatchWithRotation(match=match)
+                        for match in results_from_hashes
+                    ]
```
haianhng31, line 221: I created a new dataclass IndexMatchWithRotation which will be the return type of _match_file.

Q: Should I also make it the return type of _match_hashes? Just so the return types of these 2 are similar

Dcallies: To make the python typing pass you may need to be consistent between them (see what mypy says)

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -9,18 +9,23 @@
...
 from threatexchange import common
 from threatexchange.cli.fetch_cmd import FetchCommand
 from threatexchange.cli.helpers import FlexFilesInputAction
 from threatexchange.exchanges.fetch_state import FetchedSignalMetadata
 
-from threatexchange.signal_type.index import IndexMatch, SignalTypeIndex
+from threatexchange.signal_type.index import (
+    IndexMatch,
+    SignalTypeIndex,
+    IndexMatchWithRotation,
```
Dcallies, line 22: blocking: I don't think this should be generalized yet, let's just move it into this file (match_cmd.py)

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -139,6 +140,12 @@ def __eq__(self, other: t.Any) -> bool:
 IndexMatch = IndexMatchUntyped[SignalSimilarityInfo, T]
 
 
+@dataclass
+class IndexMatchWithRotation(t.Generic[T]):
```
Dcallies, line 144: See note about putting this just in match command. It can also be private there (_IndexMatchWithRotation)

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -82,7 +83,7 @@ def flip_minus1(cls, image_data: bytes) -> bytes:
                 return buffer.getvalue()
 
     @classmethod
-    def all_simple_rotations(cls, image_data: bytes):
+    def all_simple_rotations(cls, image_data: bytes) -> t.Dict[RotationType, bytes]:
```
Dcallies, line 86: Good catch

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -139,6 +140,12 @@ def __eq__(self, other: t.Any) -> bool:
...
+    match: IndexMatchUntyped[SignalSimilarityInfo, T]
+    rotation_type: RotationType = RotationType.ORIGINAL
```
Dcallies, line 146: When you move it into the match command, consider making this optional, then adding a `__str__` command which handles the output of the match command. If rotation_type is None, then no rotation is printed. If the rotation is there, then it can format the string you put to string.

Dcallies (changes requested):
Nice progress! Starting to get close.

1. Let's not print the rotation unless --rotations is provided - this is why your tests are failing that check the output
2. Some notes inline about how to handle that.
3. If you haven't done it yet, it will be easier to add --rotations to `tx hash` than it will be to match.

haianhng31:
My current test plan for hash command with rotations: 
'''
$ tx hash --rotations  photo pdq/data/bridge-mods/aaa-orig.jpg >> local_file.txt
$ tx config collab edit local_file file_backed_bank.txt --filename ~/file_backed_bank.txt --create
$ tx fetch
[local_file] file_backed_bank.txt - Up to date
Rebuilding match indices...
$ tx match photo pdq/data/bridge-mods/aaa-orig.jpg 
pdq 16 (Sample Signals) INVESTIGATION_SEED
'''

Dcallies:
Use "Edit" to add your test plan into the summary!

[image](https://github.com/user-attachments/assets/59e043d6-b0e6-4bac-a2e1-7a690d16b98e)

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -99,10 +112,31 @@ def execute(self, settings: CLISettings) -> None:
...
+                                hash_str = hasher.hash_from_file(temp_file_path)
+                                if hash_str:
+                                    print(hasher.get_name(), hash_str)
+
+        else:
+            for file in self.files:
+                for hasher in hashers:
+                    hash_str = hasher.hash_from_file(file)
+                    if hash_str:
+                        print(hasher.get_name(), hash_str)
```
Dcallies, line 142: We can re order these statements to reduce the total indenting, which will make this easier to read.


```
if not rotations:
  <code>
  return

if not issubclass(self.content_type, PhotoContent):
  raise ...

for file in self.files:
   ...
```

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -29,6 +37,21 @@
...
+    rotation_type: t.Optional[RotationType] = field(default=None)
+
+    def __str__(self):
+        # Supposed to be without whitespace, but let's make sure
+        distance_str = "".join(self.match.similarity_info.pretty_str().split())
+        if self.rotation_type is None:
+            return distance_str
+        else:
+            rotation_info = f" [{self.rotation_type.name}]"
+            return distance_str + rotation_info
```
Dcallies, line 52: You can save on indenting by omitting the else.

You can also combinat the format string together since you are making another concat
```suggestion
        if self.rotation_type is None:
            return distance_str
        return f"{distance_str} [{self.rotation_type.name}]"
```

python-threatexchange/threatexchange/tests/hashing/test_pdq_hash_rotations.py
```diff
@@ -0,0 +1,49 @@
...
+from threatexchange.cli.hash_cmd import HashCommand
+from threatexchange.cli.cli_config import CLISettings
+from threatexchange.cli.exceptions import CommandError
+from threatexchange.content_type.photo import PhotoContent, RotationType
+from threatexchange.content_type.text import TextContent
+from threatexchange.content_type.url import URLContent
+from threatexchange.content_type.video import VideoContent
+
+
+class TestHashCommand(unittest.TestCase):
```
Dcallies, line 14: blocking: We are using the py.test framework for this library, not unittest. 

Check out: https://docs.pytest.org/en/stable/

and this reference test: https://github.com/haianhng31/ThreatExchange/blob/8d10599cda78afa49f763c08ca556db94f4ae332/python-threatexchange/threatexchange/cli/tests/match_cmd_test.py

We should probably just use the hash and match_cmd tests rather than creating new tests.

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -23,6 +23,7 @@
 import pickle
 import typing as t
 
+from threatexchange.content_type.content_base import RotationType
```
Dcallies, line 26: remove

python-threatexchange/threatexchange/content_type/content_base.py
```diff
@@ -7,7 +7,7 @@
 This records all the valid signal types for a piece of content.
 """
 
-from enum import Enum, auto
+from enum import Enum
```
Dcallies, line 10: remove

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -196,18 +232,23 @@ def execute(self, settings: CLISettings) -> None:
...
                 else:
-                    results = _match_file(path, s_type, index)
+                    results = _match_file(path, s_type, index, rotations=self.rotations)
 
                 for r in results:
-                    metadatas: t.List[t.Tuple[str, FetchedSignalMetadata]] = r.metadata
+                    metadatas: t.List[t.Tuple[str, FetchedSignalMetadata]] = (
+                        r.match.metadata
+                    )
+                    distance_str = r.__str__()
```
Dcallies, line 245: blocking: Read about how `__str__` works here: https://docs.python.org/3/library/stdtypes.html#str

```
print(s_type.get_name(), result)
```

Dcallies (changes requested):
Getting closer, but maybe let's break down the problem a little more.

It may help to break this down into two PRs - one for the `hash` command, and one for the match command.

Hash will be more straightforward!

In order to do that, you'll need to use some git tricks to "split" the stack, which can be daunting. 

One way we could do that is the following ([ref](https://www.geeksforgeeks.org/how-to-copy-a-version-of-single-file-from-one-git-branch-to-another/))
1. Checkout the `main` branch with `git checkout main`
2. Create a new branch for only the hash command `git checkout -b hash_cmd_rotations`
3. Pull only the contents of the hash command to your new branch `git checkout photo-helpers -- python-threatexchange/threatexchange/cli/hash_cmd.py`
4. Create a new commit on our new branch `git commit -am '[pytx] add --rotations to hash_cmd'
5. Push this branch to start a new PR

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -217,18 +256,54 @@ def execute(self, settings: CLISettings) -> None:
...
+        image_data = f.read()
+
+    rotated_images: t.Dict[RotationType, bytes] = PhotoContent.all_simple_rotations(
+        image_data
+    )
+    all_matches = []
+
+    for rotation_type, rotated_bytes in rotated_images.items():
+        # Create a temporary file to hold the image bytes
+        with tempfile.NamedTemporaryFile() as temp_file:
```
Dcallies, line 285: blocking: This is going to write a lot of files over the course of an execution! We are going to call this method for every single match type.

Another approach you could do is refactor this so that the rotated images are inserted higher up in the stack, and then rather than taking rotations: bool, you could pass in the path of the rotation, and an optional enum representing which enum it can take.

python-threatexchange/threatexchange/cli/match_cmd.py
```diff
@@ -196,18 +230,23 @@ def execute(self, settings: CLISettings) -> None:
             for s_type, index in indices:
```
Dcallies, line 230: Before processing this list of files, if rotations is true here, generate new files that you iterate through.

Here's one way to do that

```
def handle_rotations() -> Iterator[(Path, t.Optional[RotationType)]:
  for file in self.files:
    if not self.rotations:
      yield file, None
      continue
   for rot, dat in PhotoContent.blah():
      with NamedTemporary(...) as f:
         yield Path(f.name), rot

for path in handle_rotations(self.files):
```

Dcallies (changes requested):
Nice job splitting your PR! This is quite difficult to do any many people have a lot more trouble their first time!

python-threatexchange/threatexchange/cli/match_cmd.py
haianhng31: I'm thinking of change s_type to be subclass of BytesHasher so that I can use bytes directly with `hash_from_bytes`. All the tests passed and mypy doesn't complain. What do you think of this approach?

Dcallies: In the last PR, I mentioned the downside - not all photo formats are knowable without the extension. There's likely a workaround, but let's stay with the current course and see if we can fix it in a followup instead.

Dcallies (changes requested):
All that's missing now:

1. We should update the unittest for match to also include a case for --rotations
1. From your test plan it looks like it's not matching hashes that it should, since the rotated image should match the rotated hash.

Here's a e2e test case:
```
$ threatexchange hash --rotations photo https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/b.jpg?raw=true | grep FLIPX | cut -d ' ' -f 2- >> ~/file.txt
$ threatexchange config collab edit local_file --filename ~/file.txt 'file.txt' --create
$ threatexchange fetch
$ threatexchange match photo https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/b.jpg?raw=true
# No match

$ threatexchange match --rotations photo https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/b.jpg?raw=true
pdq - (file.txt) INVESTIGATION_SEED
```

python-threatexchange/threatexchange/cli/tests/match_cmd_test.py
```diff
@@ -31,3 +34,33 @@ def test_invalid_hash(self):
...
+
+        # img = rotated_images[RotationType.ROTATE90] #try with 1 rotated image first
+
+        # with tempfile.NamedTemporaryFile() as tmp_file:
+        #     img = rotated_images[RotationType.ROTATE90]
+        #     tmp_file.write(img)
+        #     self.assert_cli_output(
+        #         ("--rotations", "photo", "--", tmp_file.name),
+        #         "video_md5 - (Sample Signals) INVESTIGATION_SEED",
+        #     )
```
haianhng31, line 66: My current approach for this test: 
1. Hash an image 
2. Fetch 
3. Call `match --rotations` the rotated version of the original image 

Q: @Dcallies 
* I got the hash pdq of the image (line 53) but how do i fetch indexes and call match command on this hash?
* If there is a match found, what should the output look like? When I try it with the CLI it keeps output 
`pdq 16 (Sample Signals) INVESTIGATION_SEED`

haianhng31:
@Dcallies 
Because the hash command output the Rotation_Type at the front, the match command cannot function correctly. Where do you suggest me looking at to fix this? (I've been jumping around files and was confused where the matching implementation is)

python-threatexchange/threatexchange/cli/tests/match_cmd_test.py
Dcallies: If there is a match, then the output is the matching collaboration and distance.

Consider this ouptput:
> pdq 16 (Sample Signals) INVESTIGATION_SEED

| signal_type | distance | collab name | confidence |
|---| --- | --- | --- |
| pdq | 16 | Sample Signals | INVESTIGATION_SEED

Dcallies:
> Because the hash command output the Rotation_Type at the front, the match command cannot function correctly.

Hey @haianhng31, what do you mean? The hash command and the match command are different commands, and so one should not break the other.

My guess is that you mean that the output of the rotation type makes it not read correctly from a file. Take a closer look at the command I gave you before:

```
threatexchange hash --rotations photo https://github.com/facebook/ThreatExchange/blob/main/pdq/data/misc-images/b.jpg?raw=true | grep FLIPX | cut -d ' ' -f 2- >> ~/file.txt
```
* Read about unix piping here: https://en.wikipedia.org/wiki/Pipeline_(Unix)
* Read about what grep does here: https://en.wikipedia.org/wiki/Grep
* Read about what cut does here: https://phoenixnap.com/kb/linux-cut

Let me try patching your code on my machine locally to see if I can see what you mean...

python-threatexchange/threatexchange/cli/tests/match_cmd_test.py
haianhng31: Thank you for your comment! I found the bug. I shouldn't have checked s_type == PhotoContent since it's SignalType not ContentType

python-threatexchange/threatexchange/content_type/content_base.py
haianhng31: Do you mean to keep the `import auto` line?

haianhng31:
@Dcallies thank you for all the comments and suggestions, they are really helpful! I have completed the test cases for match cmd with rotations (updated in the PR's summary). 
The match command now also run correctly with the rotation type printed.

haianhng31:
@Dcallies 
When you have time can you help me review this PR? Also I saw you mentioned about adding -save option in the PR with Mackey, can you tell me more about that?

Dcallies:
Sorry for the delay @haianhng31! You can mark the PR as ready for review by hitting the refresh button in the upper right here: 
image

python-threatexchange/threatexchange/cli/tests/match_cmd_test.py
```diff
@@ -31,3 +38,32 @@ def test_invalid_hash(self):
...
+        rotated_images = PhotoContent.all_simple_rotations(test_file.read_bytes())
+
+        for rotation, image in rotated_images.items():
+            with tempfile.NamedTemporaryFile() as tmp_file:
+                tmp_file.write(image)
+
+                if rotation == RotationType.ROTATE270:
+                    rotation = RotationType.ROTATE90
+                elif rotation == RotationType.ROTATE90:
+                    rotation = RotationType.ROTATE270
```
Dcallies, line 64: blocking q: What is the purpose of this block? It looks like you still end up with all the rotations, so does this just change the order?

If so, why?

If not, consider adding an explanation!

python-threatexchange/threatexchange/cli/tests/match_cmd_test.py
```diff
@@ -31,3 +38,32 @@ def test_invalid_hash(self):
...
+
+                self.assert_cli_output(
+                    ("--rotations", "photo", tmp_file.name),
+                    f"pdq {rotation.name} 16 (Sample Signals) INVESTIGATION_SEED",
```
Dcallies, line 68: I'm not sure if the distance here will be stable (I'm actually surprised this works), but if it passes, I think this is okay for now.

python-threatexchange/threatexchange/content_type/content_base.py
Dcallies: Whoops, mistake on my part, thought this was an unused import rather than a change! Please ignore!

Dcallies (approved):
I had one potentially blocking comment, but I think it's fine to land and merge as is. 

This was a tough stack, nice job @haianhng31 !

PR #1678, opened 2024-11-04
[pytx] Add --rotations to hash_cmd 

haianhng31:
# Summary
---------
* Create a new command line option (--rotation) for hashing

# Test Plan
---------
* I have created 2 test cases in [hash_cmd_test.py](https://github.com/facebook/ThreatExchange/pull/1678/files#diff-80084d07a4b4ea7eb643005894210cc0c62ca6d11a53e0ffce13d55c76c7610e)
  * [test_rotations_with_non_photo_content](https://github.com/facebook/ThreatExchange/pull/1678/files#diff-80084d07a4b4ea7eb643005894210cc0c62ca6d11a53e0ffce13d55c76c7610eR80-R89)
  * [test_rotations_with_photo_content](https://github.com/facebook/ThreatExchange/pull/1678/files#diff-80084d07a4b4ea7eb643005894210cc0c62ca6d11a53e0ffce13d55c76c7610eR91-R107)


python-threatexchange/threatexchange/cli/tests/hash_cmd_test.py
```diff
@@ -75,3 +75,33 @@ def test_mixed(hash_cli: ThreatExchangeCLIE2eHelper, tmp_path: pathlib.Path):
...
+            "pdq accb6d39648035f8125c8ce6ba65007de7b54c67a2d93ef7b8f33b0611306715",
+            "pdq 1f70cbbc77edc5f9524faa1b18f3b76cd0a04a833e20f645d229d0acc8499c56",
+            "pdq 31ddf2513558de0ae56e4a8c8930cadde2ee084df3aed0a75fa512ea0e41e197",
+            "pdq c79931968e880f4b97196df4d5ea0fb489b99e2c10af0dceacf572809b815ea5",
+            "pdq aaca8605440c4a6735dbbf59b947df92e7bf161807e9cd88baf04579523098ab",
+            "pdq 559f78de2530abbdc663b131ed78f10072c4bb13f3acd6017d0e69150e413e5a",
+            "pdq 86a860c1f2bd1a1ec65cf4d10ab55087b4b89f7857da59bbd9200fd5845cc3f9",
+            "pdq 5bb15db9e8a1f03c174a380a55aeaa2985bde9c60abce301bde48df918b5c15b",
+        ],
+    )
```
Dcallies, line 107: I like these tests!

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -99,10 +112,31 @@ def execute(self, settings: CLISettings) -> None:
...
+                image_bytes = f.read()
+                rotated_images = PhotoContent.all_simple_rotations(image_bytes)
+                for _, rotated_bytes in rotated_images.items():
+                    with tempfile.NamedTemporaryFile() as temp_file:  # Create a temporary file to hold the byte data
+                        temp_file.write(rotated_bytes)
+                        temp_file_path = pathlib.Path(temp_file.name)
+                        for hasher in hashers:
+                            hash_str = hasher.hash_from_file(temp_file_path)
+                            if hash_str:
+                                print(hasher.get_name(), hash_str)
```
Dcallies, line 142: blocking: Suggest we should change the output when using --rotations to also include the rotation

Suggest we change it to 

Before:
```
pdq accb6d39648035f8125c8ce6ba65007de7b54c67a2d93ef7b8f33b0611306715 
```
After:
```
original pdq accb6d39648035f8125c8ce6ba65007de7b54c67a2d93ef7b8f33b0611306715
```

This makes it not naively deserializable anymore (without using something like `cut` to manipulate the output, but I think it makes it better for human reading).

What do you think?

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -99,10 +112,31 @@ def execute(self, settings: CLISettings) -> None:
...
         for file in self.files:
-            for hasher in hashers:
-                hash_str = hasher.hash_from_file(file)
-                if hash_str:
-                    print(hasher.get_name(), hash_str)
+            with open(file, "rb") as f:
+                image_bytes = f.read()
+                rotated_images = PhotoContent.all_simple_rotations(image_bytes)
+                for _, rotated_bytes in rotated_images.items():
+                    with tempfile.NamedTemporaryFile() as temp_file:  # Create a temporary file to hold the byte data
```
Dcallies, line 136: ignorable: What are your thoughts on passing in the bytes via `hash_bytes`? Think about it before reading the spoiler.

<details>
  <summary>Spoiler warning</summary>
  `hash_bytes` will work in most cases, but some file formats are only knowable by their extensions. No file name = no extension hint, and pillow will fail to parse the file.
</details>

Dcallies (changes requested):
This overall looks good! One question inline about changing the output of the function when --rotations are provided.

python-threatexchange/threatexchange/cli/hash_cmd.py
haianhng31: That makes sense. I would also prefer it to output the rotation type to know what rotation matches with which pdq

python-threatexchange/threatexchange/cli/hash_cmd.py
Dcallies: Then it would be a BytesHasher!

Dcallies (approved):

Issue #1666, opened 2024-10-22
[py-tx][mlh] Add un-letterboxing reference implementation to Photo Content Type


Dcallies:
Goal: 

Provide a method in [Photo content type](https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/content_type/photo.py) which, given an image, returns either the same image, or a smaller image that has square black boxes removed from the side.

# Example

Take this image:
![Image](https://github.com/user-attachments/assets/7f0703df-e7ba-4d1f-94b3-146e229ae8b4)

Given this "letterboxed" version of the image, the return should be the first image.
![Image](https://github.com/user-attachments/assets/2df01578-6044-47bf-abe6-d4c339a43e07)

# Demonstrating
Add a new flag to `tx hash` that will allow preprocessing steps for images

```
tx hash -s pdq photo original.png
pdq facefacefaceface

tx hash -s pdq photo letterboxed.png
pdq 0000faceface0000

tx hash -s pdq --preprocess=unletterbox letterboxed.png
pdq facefacefaceface 
```

# Implementing
Try to avoid pulling in new dependencies if you can, even though there are a million libraries that might already solve this. Instead, take a look at the image libraries already included, such as Pillow - there is robust documentation.

# Testing
Include a unittest that demonstrates the functionality, including edge cases you might consider. Use the [bridge mod photos](https://github.com/facebook/ThreatExchange/tree/main/pdq/data/bridge-mods), either algorithmically updating them

Dcallies:
We're going to close this as completed for now, there's some cleanup we could do, but that can be a future project.

PR #1684, opened 2024-11-06
[py-tx] embeded tx hash for unletterboxing

Mackay-Fisher:
# Preprocessing and Hashing Enhancements in ThreatExchange CLI
---
This update introduces new preprocessing functionality to remove black letterbox borders and apply image rotations before hashing, improving the flexibility and accuracy of image-based workflows. This update resolves issue #1666 and integrates preprocessing with brute-force rotation workflows for more robust hashing options.
### Enhanced `unletterbox` Method in `photo.py`
The `unletterbox` method preprocesses images by detecting and removing black letterbox borders, with support for flexible configuration:
- **Parameters:**
  - `file_path`: The path to the input image file.
  - `black_threshold`: Sets the brightness threshold for detecting black borders (default: 10). 
- **Returns:** Cropped image bytes, enabling in-memory processing or downstream workflows.

---
### Updated `HashCommand` in `hash_cmd.py`
The `HashCommand` class now supports preprocessing control through additional command-line arguments:
- **`--photo-preprocess`**: Specifies the preprocessing type:
  - `unletterbox`: Removes black borders from images.
  - `rotations`: Generates all simple rotations of the image (e.g., `ROTATE90`, `FLIPX`).
- **`--black-threshold`**: Sets the brightness threshold for detecting black borders. Lower values make the border detection more sensitive, while higher values make it less sensitive.
- **`--save-preprocess`**: If enabled, saves the processed image to disk with a suffix reflecting the preprocessing type (`_unletterboxed`, `_ROTATE90`, etc.).
---

### Improved Handling of Image Formats and Extensions
- The output file format is now inferred from the input file's extension, ensuring consistency between the input and output file types.
- Default fallback to `PNG` for unsupported or missing extensions.
- Output filenames dynamically reflect the preprocessing type (`_unletterboxed`, `_ROTATE90`, etc.) and retain the original format (e.g., `.jpg`, `.png`).
---

### Expanded Testing in `test_pdq_letterboxing.py`
- **`test_unletterbox_image`**: Verifies that unletterboxed image bytes generate the expected PDQ hash.
- **`test_rotations_with_photo_content`**: Validates that all eight rotations produce distinct and accurate hashes.
- **`test_unletterboxfile_creates_output_file`**: Ensures that the processed file is correctly saved with the expected suffix when `--save-preprocess` is enabled.


Dcallies:
Before I look at the code, thank you for the very comprehensive summary and test writeup. You could probably even skimp a little bit in future PRs! 

I also like that you looked for other potential improvements that made sense in the area. I was considering suggesting a --save option during @haianhng31 --rotation diffs as well.

Mackay-Fisher:
Of course, thank you!

I can also update the save output option and file output for the rotation if you would like.

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -79,6 +80,26 @@ def init_argparse(cls, settings: CLISettings, ap: argparse.ArgumentParser) -> No
...
+            "--black-threshold",
+            type=int,
+            default=40,
+            help="Set the black threshold for unletterboxing. Default is 40.",
+        )
+
+        ap.add_argument(
+            "--save-output",
+            type=bool,
+            default=False,
```
Dcallies, line 99: nit: If you make the action `store_true` then the default is false IIRC.

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -118,7 +144,17 @@ def execute(self, settings: CLISettings) -> None:
         if not self.rotations:
             for file in self.files:
                 for hasher in hashers:
-                    hash_str = hasher.hash_from_file(file)
+                    if isinstance(hasher, PdqSignal) and (
```
Dcallies, line 147: Why only PdqSignal? Wouldn't other image perceptual hashing algorithms benefit from this?

We also generally want to avoid places where we do `isinstance(<interface_implementation>)` in preference of it being handled by the interface itself.

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -118,7 +144,17 @@ def execute(self, settings: CLISettings) -> None:
...
+                        self.content_type.get_name() == "photo"
```
Dcallies, line 148: Hmm, more specialization.

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -130,7 +166,15 @@ def execute(self, settings: CLISettings) -> None:
 
         for file in self.files:
             with open(file, "rb") as f:
-                image_bytes = f.read()
+                if (
+                    self.content_type.get_name() == "photo"
+                    and self.preprocess == "unletterbox"
```
Dcallies, line 171: nit: Instead, in the `__init__` validate that unletterbox can only be used with photo, which will save you many future checks like this.

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -101,3 +103,120 @@ def all_simple_rotations(cls, image_data: bytes):
...
+
+        Returns the first column from the right that is not all blacked out in the column.
+        """
+        width, height = grayscale_img.size
+        for x in range(width - 1, -1, -1):
+            col_pixels = list(grayscale_img.crop((x, 0, x + 1, height)).getdata())
+            if all(pixel < black_threshold for pixel in col_pixels):
+                continue
+            return width - x - 1
+        return width
```
Dcallies, line 181: By putting these all at the top level, we are signaling that they are part of the "official" interface for photos.

Instead, let's move this functionality into its own file in a new `/preprocessing` directory. We can add `unletterbox.py` as its own module, with these 4 methods then as standalone.

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -101,3 +103,120 @@ def all_simple_rotations(cls, image_data: bytes):
...
+
+    @classmethod
+    def unletterbox(
+        cls, file_path: Path, save_output: bool = False, black_threshold: int = 40
```
Dcallies, line 185: blocking: Instead of making save_output an argument here, it might be better to compose it from the outside by taking the bytes, which will give the caller more control over the file directory.

blocking q: Can you tell more about how you picked 40? We may want to be very conservative by default (even to only 100% black).

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -101,3 +103,120 @@ def all_simple_rotations(cls, image_data: bytes):
...
+    ) -> bytes:
+        """
+        Remove black letterbox borders from the sides and top of the image.
+
+        Converts the image to grescale then remove the columns and rows that
+        are all completly blacked out.
+
+        Then removing the edges to give back a cleaned image bytes.
+
+        Return the new hash of the cleaned image with an option to create a new output file as well
```
Dcallies, line 195: I don't think this returns the hash, no?

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -101,3 +103,120 @@ def all_simple_rotations(cls, image_data: bytes):
...
+            # Optionally save the unletterboxed image to a new file in the same directory
+            if save_output:
+                path = Path(file_path)
+                output_path = path.parent / f"{path.stem}_unletterboxed{path.suffix}"
+                cropped_img.save(output_path)
+                print(f"Unletterboxed image saved to: {output_path}")
+
+            # Convert the cropped image to bytes for hashing
+            with io.BytesIO() as buffer:
+                cropped_img.save(buffer, format=img.format)
```
Dcallies, line 220: Why .img? Should we use the same format that was passed in?

python-threatexchange/threatexchange/content_type/photo.py
```diff
@@ -101,3 +103,120 @@ def all_simple_rotations(cls, image_data: bytes):
...
+        Converts the image to grescale then remove the columns and rows that
+        are all completly blacked out.
+
+        Then removing the edges to give back a cleaned image bytes.
+
+        Return the new hash of the cleaned image with an option to create a new output file as well
+        """
+        # Open the original image
+        with Image.open(file_path) as img:
+            grayscale_img = img.convert("L")
```
Dcallies, line 199: blocking q: Hmm, why convert to grayscale first? Won't think convert some full colors to black?

Dcallies (changes requested):
As mentioned, thanks for the strong summary writeup.

Toplevel things:
1. I think there's a /data/ directory in python-threatexchange. We've also previously loaded test images from the pdq/data directory - there's a trick you can use to find relative imports from the current file path you can see in a couple of tests. What do you think about synthesizing a new letterboxed bridge photo from the bridge-mods directory and use the meta logo only for the manual testing for this diff? Why - I don't want folks to be concerned about image rights, and the bridge-mods are only for the purpose of this repo.  
2. I think we should simplify the version of the interface we use on PhotoContent for now, and move the specific letterboxing into its own directory under content_type, or at least its own file.
3. Can you explain more how you came to choose the thresholds, .e.g 40?

I found this random library that does something similar - you might want to give their code a read for ideas: https://github.com/Animenosekai/bordercrop/blob/main/bordercrop/bordercrop.py, though as mentioned we are trying to limit the number of external dependencies, otherwise we could even just use the library as is. We don't need all the features they have. 

Overall strong work, thanks for the effort put in!

python-threatexchange/threatexchange/content_type/photo.py
Mackay-Fisher: Ahh no I had it returning the hash at first but then it created a circular dependency so I removed it but did not update the comment. I will update.

python-threatexchange/threatexchange/cli/hash_cmd.py
Mackay-Fisher: Okay, I thought it was specific to PdqSignal and as for why I bypassed the interface it is because they do not all have the method hash from bytes and I did not want always to create the new file with updated images bytes. Is there a way I can go around this or would it be better to always create the new file even if temporarily and then save the output if the user passes the flag to save it?

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -79,6 +80,26 @@ def init_argparse(cls, settings: CLISettings, ap: argparse.ArgumentParser) -> No
...
+
+        ap.add_argument(
+            "--save-output",
+            type=bool,
+            default=False,
+            help="If true, save the processed image as a new file.",
+        )
+
         ap.add_argument(
             "--rotations",
```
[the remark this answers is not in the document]
Mackay-Fisher: Yes, I thought that it made the most sense to combine but didn't want to jump ahead and do it beforehand but I can add it to the list of actions.

Also, would there ever be a workflow where someone may want to both check for rotation and process the image for letterboxing?

python-threatexchange/threatexchange/cli/hash_cmd.py
```diff
@@ -79,6 +80,26 @@ def init_argparse(cls, settings: CLISettings, ap: argparse.ArgumentParser) -> No
             help="only generate these signal types",
         )
 
+        ap.add_argument(
+            "--preprocess",
```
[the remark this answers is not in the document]
Mackay-Fisher: For the photo being a subcommand, how would I make it so that I can accept the arguments without having them processed as multiple files for the photo hash?

I will add the help as well and focus on making it more safe with the targeted subcommands.

Mackay-Fisher:
> 1. What do you think about synthesizing a new letterboxed bridge photo from the bridge-mods directory and use the meta logo only for the manual testing for this diff? Why - I don't want folks to be concerned about image rights, and the bridge-mods are only for the purpose of this repo.

I will update the images.

Mackay-Fisher:
> I found this random library that does something similar - you might want to give their code a read for ideas: https://github.com/Animenosekai/bordercrop/blob/main/bordercrop/bordercrop.py, though as mentioned we are trying to limit the number of external dependencies, otherwise we could even just use the library as is. We don't need all the features they have.

This was super helpful I can go ahead and use the pillow library and implement it in a similar format for simplicity of dependencies and cleaner code. It should work the same way. However, if you would like me to go ahead and use this library I can do that as well.

Mackay-Fisher:
> > I found this random library that does something similar - you might want to give their code a read for ideas: https://github.com/Animenosekai/bordercrop/blob/main/bordercrop/bordercrop.py, though as mentioned we are trying to limit the number of external dependencies, otherwise we could even just use the library as is. We don't need all the features they have.
> 
> This was super helpful I can go ahead and use the pillow library and implement it in a similar format for simplicity of dependencies and cleaner code. It should work the same way. However, if you would like me to go ahead and use this library I can do that as well.

Also, I looked more into it and this focuses on also being able to address image URL types and more than the basic image format that we are using. If that is by design I can add the pass-through to do that.

python-threatexchange/threatexchange/cli/hash_cmd.py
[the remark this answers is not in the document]
Dcallies: For this PR, let's go with what you've written

python-threatexchange/threatexchange/cli/hash_cmd.py
Dcallies: It's not specific to PdqSignal, and the hash_from_bytes method is itself part of a wider interface.

I think I eventually decided it was simpler to write everything to tmpfiles, which is how we ended up with the current implementation.

However, similar to the feedback I gave for --rotations, we can make our life a lot easier by having the preprocessing happen in between the file input and the `for file in self.files`.

I like your idea of providing a way to pass flag to save it, but let's save that for a followup.

Dcallies (changes requested):
Some response to comments - use "request review" in the upper right hand of the summary page to send it back to me for review (little refresh logo).

python-threatexchange/threatexchange/content_type/photo.py
Mackay-Fisher: I revised this and updated to check each individual value of the r g b pixel

python-threatexchange/threatexchange/content_type/preprocess/unletterboxing.py
```diff
@@ -0,0 +1,69 @@
...
+
+def detect_top_border(image: Image.Image, black_threshold: int = 0) -> int:
+    """
+    Detect the top black border by counting rows with only black pixels.
+    Checks each RGB channel of each pixel in each row.
+    Returns the first row that is not all black from the top.
+    """
+    width, height = image.size
+    for y in range(height):
+        row_pixels = list(image.crop((0, y, width, y + 1)).getdata())
```
Dcallies, line 20: I couldn't tell from reading the pillow docs, but can you use the returned core.image object as an iterator without wrapping it in a list?

Dcallies (changes requested):
Getting closer!

blocking: Instead of adding the test file to /threatexchange, can you add it to https://github.com/facebook/ThreatExchange/tree/main/pdq/data/bridge-mods instead?

Because threatexchange is in the same repo as PDQ, you can read the pdq directory from ThreatExchange tests using the trick I noted inline with `__file__`.

At the rate we are going, this feels like it will take a few more passes. You can simplify the PR by breaking out the changes to unletterboxing.py and photo to its own PR (the bottom/simplest part of the stack), though splitting a PR in git is pretty daunting for the first time.  Here's a [random article](https://graphite.dev/guides/how-to-split-an-existing-pull-request-on-github) describing a method using cherry-pick, but there are others.

Mackay-Fisher:
Sorry for the late response, I resolved the blocking comments and tried my best to address the follow-up. Please let me know if there are added things or if you would like more clarification on the testing.

python-threatexchange/threatexchange/content_type/preprocess/unletterboxing.py
Mackay-Fisher: Sorry, I thought I responded to this but I get a mypy error if I do not wrap it in a list because it does not recognize the iterator. It does completely work because there is an iterator to the image object but for the sake of mypy I left it as a list. I can change it to have the ignore if that would be better.

Dcallies (approved):
I didn't see any leftover blockers, so let's merge this version!

Issue #1685, opened 2024-11-08
[pytx][mlh] Convert threatexchange/tests/ to py.test


Dcallies:
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/tests/test_api.py
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/tests/test_common.py

This test is using unittest, we want to use py.test.

py.test docs: https://docs.pytest.org/en/stable/

PR #1690, opened 2024-11-08
[pytx] Convert threatexchange/tests/ to py.test

haianhng31:
Summary
---------
* This resolves issue #1685 
* I updated these 2 tests from unittest to pytest
  * https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/tests/test_api.py
  * https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/tests/test_common.py

Test Plan
---------


Dcallies (approved):
Fast work, good use of fixtures!

PR #1694, opened 2024-11-09
Update README with note on dihedral pdq hashes

### LINKED ISSUE #1676: [pdq] Helper for "lexicographically smallest" PDQ hash
We want to experiment with ways to make PDQ lookup faster, especially as it pertains to looking up rotations.

One idea we had is to compute all of the 8 naive rotations (see #1665), but then only index the "lexicographically smallest" PDQ hash. 

Then, during lookup, you again only lookup the lexicographically smallest hash.

# Step 1: Simple PR
Add a method, which, given an image, returns only the lexicographically smallest hash

# Step 2: Proving This is a Good Idea
1. Find a large public listing of images, e.g. ImageNet - pick ~500-1000 of your favorite diverse set of images
2. First check the sanity of the dataset - find the distance at which you only match the items you are supposed to (default match distance is <32 hamming distance) - hash and brute force compare everything
3. Next make a set of the lexicographically smallest of your reference images, and repeat the experiment from #.2, you should still get the same result
4. Now, using a tool like ImageMagick, actually rotate some number of images (or all of them), and repeat the experiment using rotated images instead - you should still get the same result

# Step 3: Publish your results
We'll add a new section in the README for PDQ about this experiment, which includes your source dataset, the matching distance used, etc. We'll put it under a new section on "querying rotations"

# Step 4: Add a new binding
Head over to https://github.com/faustomorales/pdqhash-python and add an accessor for your new method that only returns the lexicographical smallest hash

# Step 5: Making the binding more functional with a CLI
This will be on https://github.com/faustomorales/pdqhash-python again. We want to add an installable CLI that will make it easier to debug PDQ. Check out #1169 

The cli should be named `pdq` and have the following behavior:
```
# Generates a hash
$ pdq file.png
facefacefacefacefacefaceface0111

$ pdq file.png --rotations
facefacefacefacefacefaceface0111
facefacefacefacefacefaceface0110
facefacefacefacefacefaceface0101
facefacefacefacefacefaceface0100
facefacefacefacefacefaceface0011
facefacefacefacefacefaceface0010
facefacefacefacefacefaceface0001
facefacefacefacefacefaceface0000

$ pdq file.png --smallest
facefacefacefacefacefaceface0000
```

16BitNarwhal:
Summary
---------

Closes #1676  
Updated pdq README with a note on dihedral pdq hashes following this [comment](https://github.com/facebook/ThreatExchange/issues/1676#issuecomment-2466331532)

Test Plan
---------

Script to test this is [here](https://github.com/facebook/ThreatExchange/issues/1676#issuecomment-2466333264)

### DIFF
--- a/pdq/README.md
+++ b/pdq/README.md
@@ -36,6 +36,12 @@ Before evaluating the results on your own to choose the thresholds that work for
 * **Distance Threshold to consider two hashes to be similar/matching**: <=31
 * **Quality Threshold where we recommend discarding hashes**: <=49
 
+## Note on Dihedral PDQ Hashes
+
+The PDQ hashing algorithm is easily capable of producing eight "dihedral" hashes (one for each 90 degree rotation and one for each flip across a horizontal, vertical or diagonal axis). However, PDQ does not guarantee exact rotational invariance. Small variations can occur in the hash values for each rotation due to how PDQ processes the image’s grid alignment in its DCT (Discrete Cosine Transform) phase.
+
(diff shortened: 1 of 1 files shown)


Dcallies (approved):
Looks good! Let's close out the issue!

Issue #1613, opened 2024-08-21
[py-tx] Implement a new cleaner PDQ index solution from scratch


Dcallies:
When we built the PDQ index, it was our first attempt, and we made a lot of strange/bad choices. 

Namely:
* The hash -> id mapping is extremely convoluted and needlessly complicated
* The build and lookup implementations could be simplified

I think we could provide a second implementation that is a lot simpler, which we could then find a way to swap.

They key elements:

# Pass in the index type as an argument during construction
```
    def __init__(
        self,
        threshold: int = DEFAULT_MATCH_DIST,
        faiss_index: t.Optional[faiss.Index] = None,
    ) -> None:
```

# Simplify the stored state of the index implementation
```
# Body of __init__
       super().__init__()
        if faiss_index is None:
            # Brute force
            faiss_index = faiss.IndexFlatL2(DIMENSIONALITY)
        self.faiss_index = _PDQHashIndex(faiss_index)
        self.threshold = threshold
        self._deduper = {}
        self._idx_to_entries: t.List[t.List[T]] = []
```

# Use a simpler inner wrapper to handle some of the PDQ details

class _PDQHashIndex:
    """
    A wrapper around the faiss index for pickle serialization
    """

    def __init__(self, faiss_index: faiss.Index) -> None:
        self.faiss_index = faiss_index

    def search(
        self,
        queries: t.Sequence[str],
        threshhold: int,
    ) -> t.List[t.Tuple[int, float]]:
        """
        Search method that return a mapping from query_str =>  (id, distance)
        """
        qs = convert_pdq_strings_to_ndarray(queries)
        # in Python, the results are returned as a triplet of 1D arrays lims, D, I
        # where result for query i is in I[lims[i]:lims[i+1]] (indices of neighbors)
        # D[lims[i]:lims[i+1]] (distances).
        limits, D, I = self.faiss_index.range_search(qs, threshhold + 1)

        results = []
        for i in range(len(queries)):
            matches = [result.item() for result in I[limits[i] : limits[i + 1]]]
            distances = [dist for dist in D[limits[i] : limits[i + 1]]]
            results.append(list(zip(matches, distances)))
        return results

    def __getstate__(self):
        data = faiss.serialize_index(self.faiss_index)
        return data

    def __setstate__(self, data):
        self.faiss_index = faiss.deserialize_index(data)

# Putting it together with search
```
    def query(self, query: str) -> t.List[IndexMatch[T]]:
[...]

Issue #1318, opened 2023-06-14
[pytx] No match results if creating a local_file with only 1 hash in it


Dcallies:
Repo:
```
$ tx hash photo pdq/data/bridge-mods/aaa-orig.jpg >> local_file.txt
$ tx config collab edit local_file file_backed_bank.txt --filename ~/file_backed_bank.txt  --create
$ tx fetch
$ tx match photo pdq/data/bridge-mods/aaa-orig.jpg 
```
Expected: any matches
However, oddly enough, adding a second hash allows all hashes to to match
```
$ tx hash photo pdq/data/bridge-mods/blur-a-little.jpg >> local_file.txt
$ tx fetch
$ tx match -A photo pdq/data/bridge-mods/aaa-orig.jpg 
pdq 4 (file_backed_bank.txt) INVESTIGATION_SEED
pdq 0 (file_backed_bank.txt) INVESTIGATION_SEED
```

Yisu123:
你好，我是瞿文政，你的邮件我已经收到，我会尽快回复。

jagraff:
This bug shows up in HMA as well, adding repro in case it's helpful.
For this repro to work, HMA (previously OpenMediaMatch) should be running as a docker container and serving `localhost:8080`

Reset the tables:
```
$ docker-compose exec app flask --app OpenMediaMatch.app reset_all_tables
[2024-03-13 18:10:12,303] WARNING in app: No storage class provided, using the default
```

Create a bank:

```
$ curl --location 'localhost:8080/c/banks' \
--header 'Content-Type: application/json' \
--data '{
    "name": "EVIL_CONTENT_BANK"
}'
{"matching_enabled_ratio":1.0,"name":"EVIL_CONTENT_BANK"}'
```

Add a file to the bank:

```
$ curl --location 'localhost:8080/c/bank/EVIL_CONTENT_BANK/content' \
--form 'photo=@"<photo path>"'
{"id":1,"signals":{"pdq":"3517f92351b0e69170c9656ba70c1249d258926d6d65bd2cbcb49cb34bd1c4fb"}}
```

Rebuild indexes:
```
$ docker-compose exec app flask --app OpenMediaMatch.app build_indices
[2024-03-13 18:12:21,582] WARNING in app: No storage class provided, using the default
[2024-03-13 18:12:21,596] INFO in build_index: Running the build_all_indices background task
[2024-03-13 18:12:21,628] INFO in build_index: Building index for pdq (1 signals)
[...]

PR #1695, opened 2024-11-11
[pytx]  Implement a new cleaner PDQ index solution 

haianhng31:
Summary
---------
Resolve issue #1613 . 
This PR introduces a new SignalTypeIndex2 class for managing and querying a PDQ hash-based index using FAISS. Key changes include:
* [**_PDQHashIndex**](https://github.com/facebook/ThreatExchange/pull/1695/files#diff-2173010ed470e9f1a0daa6c696a604eb0937458100d74be1b326ab11ed3224d0R20-R98): A wrapper around the FAISS index to handle the serialization/deserialization of PDQ hashes, along with methods for adding hashes and performing searches using the FAISS library.
* **SignalTypeIndex2**: A class for managing a PDQ index, which includes:
  * The ability to add PDQ hashes and associate them with entries. ([lines](https://github.com/facebook/ThreatExchange/pull/1695/files#diff-2173010ed470e9f1a0daa6c696a604eb0937458100d74be1b326ab11ed3224d0R138-R153))
  * A query method for searching the index for matching entries based on a query hash. ([lines](https://github.com/facebook/ThreatExchange/pull/1695/files#diff-2173010ed470e9f1a0daa6c696a604eb0937458100d74be1b326ab11ed3224d0R128-R137))
  * A serialize and deserialize method for persisting and loading the index from binary streams using pickle. ([lines](https://github.com/facebook/ThreatExchange/pull/1695/files#diff-2173010ed470e9f1a0daa6c696a604eb0937458100d74be1b326ab11ed3224d0R155-R166))

Test Plan
---------
I have included several test cases for this, currently all in one file: 
* test for correct initialization 
* test serialize deserialize (with & without custom index)
* search functionality
  * test empty index query -> return empty array 
  * test sample with 1 exact match 
  * test sample with 1 near exact match 
  * test distance threshold behavior 
  * test duplicate handling

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/index2.py
+++ b/python-threatexchange/threatexchange/signal_type/index2.py
@@ -1,0 +1,166 @@
+import typing as t
+import faiss
+import pickle
+import numpy as np
+
+from threatexchange.signal_type.index import (
+    IndexMatchUntyped,
+    SignalSimilarityInfoWithIntDistance,
+    SignalTypeIndex,
+    T as IndexT,
+    SignalSimilarityInfo,
+    IndexMatch,
+)
+
+T = t.TypeVar("T")
+DEFAULT_MATCH_DIST = 31
+DIMENSIONALITY = 256
+
+
+class _PDQHashIndex:
+    """
+    A wrapper around the faiss index for pickle serialization
+    """
+
+    def __init__(self, faiss_index: faiss.Index) -> None:
+        self.faiss_index = faiss_index
+
+    def add(self, pdq_strings: t.Sequence[str]) -> None:
+        """
+        Add PDQ hashes to the FAISS index.
+
+        Args:
+            pdq_strings (Sequence[str]): PDQ hash strings to add
+        """
+        vectors = self._convert_pdq_strings_to_ndarray(pdq_strings)
+        self.faiss_index.add(vectors)
+
+    def search(
+        self, queries: t.Sequence[str], threshold: int = DEFAULT_MATCH_DIST
+    ) -> t.List[t.List[t.Any]]:
+        """
+        Search the FAISS index for matches to the given PDQ queries.
+
+        Args:
+            queries (Sequence[str]): The PDQ signal strings to search for.
+            threshold (int): The maximum distance threshold for matches.
+
+        Returns:
+
+        """
+        query_array: np.ndarray = self._convert_pdq_strings_to_ndarray(queries)
+        limits, distances, indices = self.faiss_index.range_search(
+            query_array, threshold + 1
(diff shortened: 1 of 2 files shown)


Dcallies:
Sorry I didn't get to this today, will check first thing tomorrow!

Dcallies (changes requested):
blocking: #1613 is about creating a new PDQ index implementation which follows the SignalType interface, but this PR creates a new SignalTypeIndex interface entirely, which is surprising.

Is this an intentional change, and if so, can you walk me through why we should should go this route than a more tailored approach to provide a second, cleaned up PDQ index implementation of the SignalTypeIndex interface which lives in https://github.com/facebook/ThreatExchange/tree/main/python-threatexchange/threatexchange/signal_type/pdq ?

As part of this, can you include tests that show that the bug in #1318 is solved?

haianhng31:
I appreciate your feedback on this PR. To be honest, I was a bit uncareful at the start and didn't fully realize that creating a new SignalTypeIndex was changing the whole interface, rather than a more tailored approach.

But since now that is cleared up, moving forward, I will provide a second, cleaned up PDQ index implementation that adheres to the existing SignalTypeIndex interface. I'll be sure to include needed tests as well!

python-threatexchange/threatexchange/signal_type/index2.py
```diff
@@ -0,0 +1,166 @@
...
+            threshold (int): The maximum distance threshold for matches.
+            faiss_index (faiss.Index): An optional pre-existing FAISS index to use.
+        """
+        super().__init__()
+        if faiss_index is None:
+            # Use a simple brute-force FAISS index by default
+            faiss_index = faiss.IndexFlatL2(DIMENSIONALITY)
+        self.faiss_index = _PDQHashIndex(faiss_index)
+        self.threshold = threshold
+        self._deduper: t.Set[str] = set()
```
Dcallies, line 125: From the meeting:

Should store this as a mapping from hash => faiss_id 

add(h, payload1)
add(h, playload2)

deduper: dict[hash, faiss_id] = dict[]
entries: list[int, payload] = list[]

existing_id = deduper.get(new_hash)
if existing_id is not None: 
   # Don't add to faiss!
   # we add to our internal entry mapping
   entries[existing_id].append(payload2)
else:
  # faiss id is 0 -> size
  next_id = len(deduper)
  faiss.add(h)
  entries.append([payload])
  deduper[h] = next_id

///
lookup(h) -> payloads
  faiss_id = faiss.search(h) 
  entries[faiss_id]

query(h) -> [payload1, payload2]
```

PR #1698, opened 2024-11-14
[pytx] Implement a new cleaner PDQ index solution

haianhng31:
Summary
---------
Resolve issue #1613 . 
This PR introduces a new PDQIndex2 class for managing and querying a PDQ hash-based index using FAISS. Key changes include:
* **_PDQHashIndex**: A wrapper around the FAISS index to handle adding hashes and performing searches using the FAISS library.
* **PDQIndex2**: A class for managing a PDQ index, which includes:
  * The ability to add PDQ hashes and associate them with entries. 
  * A query method for searching the index for matching entries based on a query hash. 
  * A serialize and deserialize method for persisting and loading the index from binary streams using pickle. 

Test Plan
---------
I have included several test cases for this, currently all in one file.

For the bug in issue #1318, I have created 2 test cases relating to it:
* [Test how the index handles when it only has one entry](https://github.com/facebook/ThreatExchange/pull/1698/files#diff-52467a9eead5544de4d6082c507dd5c96bebe25c34a1a5fdfebbdb9f2c223846R182-R195). -> Return correctly 1 hash
* [Test exact matches in sample set](https://github.com/facebook/ThreatExchange/pull/1698/files#diff-52467a9eead5544de4d6082c507dd5c96bebe25c34a1a5fdfebbdb9f2c223846R109-R120) -> Doesn't return all hashes

Also ran `test_pdq_index` with 1,000 banked & 10,000 queries in the unit test. The test passed but it took 4 minutes to run.
And if i tried with 10,000 banked & 100,000 queries, the test took forever.

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
@@ -71,7 +71,7 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
             # Create match objects for each entry
             results.extend(
                 PDQIndexMatch(
-                    SignalSimilarityInfoWithIntDistance(distance=int(distance)),
+                    SignalSimilarityInfoWithIntDistance(distance=distance),
                     entry,
                 )
                 for entry in entries
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
@@ -1,8 +1,7 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 
 """
-Implementation of SignalTypeIndex abstraction for PDQ by wrapping
-hashing.pdq_faiss_matcher.
+Implementation of SignalTypeIndex abstraction for PDQ
 """
 
 import typing as t
@@ -16,9 +15,9 @@ from threatexchange.signal_type.index import (
     SignalTypeIndex,
     T as IndexT,
 )
-from threatexchange.signal_type.pdq.signal import PDQ_CONFIDENT_MATCH_THRESHOLD
 from threatexchange.signal_type.pdq.pdq_utils import (
     BITS_IN_PDQ,
+    PDQ_CONFIDENT_MATCH_THRESHOLD,
     convert_pdq_strings_to_ndarray,
 )
 
@@ -28,6 +27,10 @@ PDQIndexMatch = IndexMatchUntyped[SignalSimilarityInfoWithIntDistance, IndexT]
 class PDQIndex2(SignalTypeIndex[IndexT]):
     """
     Indexing and querying PDQ signals using Faiss for approximate nearest neighbor search.
+
+    This is a redo of the existing PDQ index,
+    designed to be simpler and fix hard-to-squash bugs in the existing implementation.
+    Purpose of this class: to replace the original index in pytx 2.0
     """
 
     def __init__(
@@ -42,7 +45,7 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
 
         if index is None:
             index = faiss.IndexFlatL2(BITS_IN_PDQ)
-        self.index = _PDQFaissIndex(index)
+        self._index = _PDQFaissIndex(index)
 
         # Matches hash to Faiss index
         self._deduper: t.Dict[str, int] = {}
@@ -56,10 +59,10 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
 
     def query(self, hash: str) -> t.Sequence[PDQIndexMatch[IndexT]]:
         """
-        Look up entries against the index, up to the max supported distance.
+        Look up entries against the index, up to the threshold.
         """
         results: t.List[PDQIndexMatch[IndexT]] = []
-        matches_list: t.List[t.Tuple[int, int]] = self.index.search(
+        matches_list: t.List[t.Tuple[int, int]] = self._index.search(
             queries=[hash], threshold=self.threshold
         )
 
@@ -82,7 +85,7 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
         for h, i in entries:
             existing_faiss_id = self._deduper.get(h)
             if existing_faiss_id is None:
-                self.index.add([h])
+                self._index.add([h])
                 self._idx_to_entries.append([i])
                 next_id = len(self._deduper)  # Because faiss index starts from 0 up
                 self._deduper[h] = next_id
@@ -107,7 +110,7 @@ class _PDQFaissIndex:
         self.faiss_index.add(vectors)
 
     def search(
-        self, queries: t.Sequence[str], threshold: int = PDQ_CONFIDENT_MATCH_THRESHOLD
+        self, queries: t.Sequence[str], threshold: int
     ) -> t.List[t.Tuple[int, int]]:
         """
         Search the FAISS index for matches to the given PDQ queries.
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
@@ -8,7 +8,6 @@ hashing.pdq_faiss_matcher.
 import typing as t
 import faiss
 import numpy as np
-import pickle
 
 
 from threatexchange.signal_type.index import (
@@ -17,38 +16,36 @@ from threatexchange.signal_type.index import (
     SignalTypeIndex,
     T as IndexT,
 )
-from threatexchange.signal_type.pdq.pdq_faiss_matcher import (
-    PDQMultiHashIndex,
-    PDQFlatHashIndex,
-    PDQHashIndex,
+from threatexchange.signal_type.pdq.signal import PDQ_CONFIDENT_MATCH_THRESHOLD
+from threatexchange.signal_type.pdq.pdq_utils import (
+    BITS_IN_PDQ,
+    convert_pdq_strings_to_ndarray,
 )
-
-DEFAULT_MATCH_DIST = 31
-DIMENSIONALITY = 256
 
 PDQIndexMatch = IndexMatchUntyped[SignalSimilarityInfoWithIntDistance, IndexT]
 
 
 class PDQIndex2(SignalTypeIndex[IndexT]):
     """
-    Wrapper around the pdq faiss index lib using PDQMultiHashIndex
+    Indexing and querying PDQ signals using Faiss for approximate nearest neighbor search.
     """
 
     def __init__(
         self,
-        threshold: int = DEFAULT_MATCH_DIST,
         index: t.Optional[faiss.Index] = None,
         entries: t.Iterable[t.Tuple[str, IndexT]] = (),
+        *,
+        threshold: int = PDQ_CONFIDENT_MATCH_THRESHOLD,
     ) -> None:
         super().__init__()
         self.threshold = threshold
 
         if index is None:
-            index = faiss.IndexFlatL2(DIMENSIONALITY)
-        self.index = _PDQHashIndex(index)
+            index = faiss.IndexFlatL2(BITS_IN_PDQ)
+        self.index = _PDQFaissIndex(index)
 
         # Matches hash to Faiss index
-        self._deduper: t.Dict[str, faiss.IndexFlatL2] = {}
+        self._deduper: t.Dict[str, int] = {}
         # Entry mapping: Each list[entries]'s index is its hash's index
         self._idx_to_entries: t.List[t.List[IndexT]] = []
 
@@ -62,21 +59,20 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
         Look up entries against the index, up to the max supported distance.
         """
         results: t.List[PDQIndexMatch[IndexT]] = []
-        matches_list: t.List[t.List[t.Any]] = self.index.search(
+        matches_list: t.List[t.Tuple[int, int]] = self.index.search(
             queries=[hash], threshold=self.threshold
         )
 
-        for matches in matches_list:
-            for match_hash, distance in matches:
-                entries = self._idx_to_entries[match_hash]  # Get the Faiss index
-                # Create match objects for each entry
-                results.extend(
-                    PDQIndexMatch(
-                        SignalSimilarityInfoWithIntDistance(distance=int(distance)),
-                        entry,
-                    )
-                    for entry in entries
+        for match, distance in matches_list:
+            entries = self._idx_to_entries[match]
+            # Create match objects for each entry
+            results.extend(
+                PDQIndexMatch(
+                    SignalSimilarityInfoWithIntDistance(distance=int(distance)),
+                    entry,
                 )
+                for entry in entries
+            )
         return results
 
     def add(self, signal_str: str, entry: IndexT) -> None:
@@ -92,23 +88,10 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
                 self._deduper[h] = next_id
             else:
                 # Since this already exists, we don't add it to Faiss because Faiss cannot handle duplication
-                self._idx_to_entries[existing_faiss_id].append(h)
-
-    def serialize(self, fout: t.BinaryIO) -> None:
-        """
-        Serialize the PDQ index to a binary stream.
-        """
-        fout.write(pickle.dumps(self))
-
-    @classmethod
-    def deserialize(cls, fin: t.BinaryIO) -> "PDQIndex2[IndexT]":
-        """
-        Deserialize a PDQ index from a binary stream.
-        """
-        return pickle.loads(fin.read())
+                self._idx_to_entries[existing_faiss_id].append(i)
 
 
-class _PDQHashIndex:
+class _PDQFaissIndex:
     """
     A wrapper around the faiss index for pickle serialization
     """
@@ -119,67 +102,31 @@ class _PDQHashIndex:
     def add(self, pdq_strings: t.Sequence[str]) -> None:
         """
         Add PDQ hashes to the FAISS index.
-        Args:
-            pdq_strings (Sequence[str]): PDQ hash strings to add
         """
-        vectors = self._convert_pdq_strings_to_ndarray(pdq_strings)
+        vectors = convert_pdq_strings_to_ndarray(pdq_strings)
         self.faiss_index.add(vectors)
 
     def search(
-        self, queries: t.Sequence[str], threshold: int = DEFAULT_MATCH_DIST
-    ) -> t.List[t.List[t.Any]]:
+        self, queries: t.Sequence[str], threshold: int = PDQ_CONFIDENT_MATCH_THRESHOLD
+    ) -> t.List[t.Tuple[int, int]]:
         """
         Search the FAISS index for matches to the given PDQ queries.
-        Args:
-            queries (Sequence[str]): The PDQ signal strings to search for.
-            threshold (int): The maximum distance threshold for matches.
-        Returns:
-            2D list of tuples that store (matches, distances) for each query
         """
-        query_array: np.ndarray = self._convert_pdq_strings_to_ndarray(queries)
+        query_array: np.ndarray = convert_pdq_strings_to_ndarray(queries)
         limits, distances, indices = self.faiss_index.range_search(
             query_array, threshold + 1
         )
 
-        results: t.List[t.List[t.Any]] = []
+        results: t.List[t.Tuple[int, int]] = []
         for i in range(len(queries)):
             matches = [idx.item() for idx in indices[limits[i] : limits[i + 1]]]
             dists = [dist for dist in distances[limits[i] : limits[i + 1]]]
-            results.append(list(zip(matches, dists)))
+            for j in range(len(matches)):
+                results.append((matches[j], dists[j]))
         return results
 
     def __getstate__(self):
-        data = faiss.serialize_index(self.faiss_index)
-        return data
+        return faiss.serialize_index(self.faiss_index)
 
     def __setstate__(self, data):
         self.faiss_index = faiss.deserialize_index(data)
-
-    def _convert_pdq_strings_to_ndarray(
-        self, pdq_strings: t.Sequence[str]
-    ) -> np.ndarray:
-        """
-        Convert multiple PDQ hash strings to a numpy array.
-        Args:
-            pdq_strings (Sequence[str]): A sequence of 64-character hexadecimal PDQ hash strings
-        Returns:
-            np.ndarray: A 2D array of shape (n_queries, 256) where each row is the full PDQ hash as a bit array
-        """
-        hash_arrays = []
-        for pdq_str in pdq_strings:
-            print("converting string:", pdq_str)
-            try:
-                # Convert hex string to integer
-                hash_int = int(pdq_str, 16)
-                # Convert to binary string, padding to ensure 256 bits
-                binary_str = format(hash_int, "0256b")
-                # Convert to numpy array
-                hash_array = np.array(
-                    [int(bit) for bit in binary_str], dtype=np.float32
-                )
-                hash_arrays.append(hash_array)
-            except (ValueError, TypeError) as e:
-                raise ValueError(f"Invalid PDQ hash string: {pdq_str}") from e
-
-        # Convert list of arrays to a single 2D array
-        return np.array(hash_arrays, dtype=np.float32)
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
@@ -1,0 +1,185 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+
+"""
+Implementation of SignalTypeIndex abstraction for PDQ by wrapping
+hashing.pdq_faiss_matcher.
+"""
+
+import typing as t
+import faiss
+import numpy as np
+import pickle
+
+
+from threatexchange.signal_type.index import (
+    IndexMatchUntyped,
+    SignalSimilarityInfoWithIntDistance,
+    SignalTypeIndex,
+    T as IndexT,
+)
+from threatexchange.signal_type.pdq.pdq_faiss_matcher import (
+    PDQMultiHashIndex,
+    PDQFlatHashIndex,
+    PDQHashIndex,
+)
+
+DEFAULT_MATCH_DIST = 31
+DIMENSIONALITY = 256
+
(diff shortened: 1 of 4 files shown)


python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        Serialize the PDQ index to a binary stream.
+        """
+        fout.write(pickle.dumps(self))
+
+    @classmethod
+    def deserialize(cls, fin: t.BinaryIO) -> "PDQIndex2[IndexT]":
+        """
+        Deserialize a PDQ index from a binary stream.
+        """
+        return pickle.loads(fin.read())
```
haianhng31, line 108: Q: Since these methods already exist in SignalTypeIndex, do we still need to add them here?

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _convert_pdq_strings_to_ndarray(
+        self, pdq_strings: t.Sequence[str]
+    ) -> np.ndarray:
+        """
+        Convert multiple PDQ hash strings to a numpy array.
+        Args:
+            pdq_strings (Sequence[str]): A sequence of 64-character hexadecimal PDQ hash strings
+        Returns:
+            np.ndarray: A 2D array of shape (n_queries, 256) where each row is the full PDQ hash as a bit array
```
Dcallies, line 166: I don't think this return type is accurate, since I think it needs to be bytes for binary index

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+                hash_int = int(pdq_str, 16)
+                # Convert to binary string, padding to ensure 256 bits
+                binary_str = format(hash_int, "0256b")
+                # Convert to numpy array
+                hash_array = np.array(
+                    [int(bit) for bit in binary_str], dtype=np.float32
+                )
+                hash_arrays.append(hash_array)
+            except (ValueError, TypeError) as e:
+                raise ValueError(f"Invalid PDQ hash string: {pdq_str}") from e
```
Dcallies, line 182: unsure: What's the difference between doing this and letting the exception leak out except coercing TypeError?

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Convert list of arrays to a single 2D array
+        return np.array(hash_arrays, dtype=np.float32)
```
Dcallies, line 185: This is the wrong type, should be `uint8`

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+    T as IndexT,
+)
+from threatexchange.signal_type.pdq.pdq_faiss_matcher import (
+    PDQMultiHashIndex,
+    PDQFlatHashIndex,
+    PDQHashIndex,
+)
+
+DEFAULT_MATCH_DIST = 31
+DIMENSIONALITY = 256
```
Dcallies, line 27: blocking: These look like they are copied from somewhere else in the code - we should only have one set of these constants.

If you are running into a circular import with PDQSignal, then move the constants into their own file

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+class PDQIndex2(SignalTypeIndex[IndexT]):
+    """
+    Wrapper around the pdq faiss index lib using PDQMultiHashIndex
+    """
+
+    def __init__(
+        self,
+        threshold: int = DEFAULT_MATCH_DIST,
+        index: t.Optional[faiss.Index] = None,
+        entries: t.Iterable[t.Tuple[str, IndexT]] = (),
```
Dcallies, line 41: I think we can probably leave this off and rely on `build`

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+
+class PDQIndex2(SignalTypeIndex[IndexT]):
+    """
+    Wrapper around the pdq faiss index lib using PDQMultiHashIndex
+    """
+
+    def __init__(
+        self,
+        threshold: int = DEFAULT_MATCH_DIST,
+        index: t.Optional[faiss.Index] = None,
```
Dcallies, line 40: nit: swap these and turn threshold into a kw arg

```suggestion
        index: t.Optional[faiss.Index] = None,
        *
        threshold: int = DEFAULT_MATCH_DIST,

```

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+
+DEFAULT_MATCH_DIST = 31
+DIMENSIONALITY = 256
+
+PDQIndexMatch = IndexMatchUntyped[SignalSimilarityInfoWithIntDistance, IndexT]
+
+
+class PDQIndex2(SignalTypeIndex[IndexT]):
+    """
+    Wrapper around the pdq faiss index lib using PDQMultiHashIndex
```
Dcallies, line 34: bl;ocking: Comment out of date?

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        self,
+        threshold: int = DEFAULT_MATCH_DIST,
+        index: t.Optional[faiss.Index] = None,
+        entries: t.Iterable[t.Tuple[str, IndexT]] = (),
+    ) -> None:
+        super().__init__()
+        self.threshold = threshold
+
+        if index is None:
+            index = faiss.IndexFlatL2(DIMENSIONALITY)
```
Dcallies, line 47: 👍

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        self.index = _PDQHashIndex(index)
+
+        # Matches hash to Faiss index
+        self._deduper: t.Dict[str, faiss.IndexFlatL2] = {}
```
Dcallies, line 51: blocking: This is the wrong type on the value. I believe it should be `int`

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+    def search(
+        self, queries: t.Sequence[str], threshold: int = DEFAULT_MATCH_DIST
+    ) -> t.List[t.List[t.Any]]:
+        """
+        Search the FAISS index for matches to the given PDQ queries.
+        Args:
+            queries (Sequence[str]): The PDQ signal strings to search for.
+            threshold (int): The maximum distance threshold for matches.
+        Returns:
+            2D list of tuples that store (matches, distances) for each query
```
Dcallies, line 137: blocking: This doesn't match the type annotation for the method. Suggest not mentioning the type and focus on the intent, since the code types can document the types

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        self.faiss_index.add(vectors)
+
+    def search(
+        self, queries: t.Sequence[str], threshold: int = DEFAULT_MATCH_DIST
+    ) -> t.List[t.List[t.Any]]:
+        """
+        Search the FAISS index for matches to the given PDQ queries.
+        Args:
+            queries (Sequence[str]): The PDQ signal strings to search for.
+            threshold (int): The maximum distance threshold for matches.
```
Dcallies, line 135: nit: Recommend removing types from these, since it only creates the possibility they are out of sync with the actual types

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        Add PDQ hashes to the FAISS index.
+        Args:
+            pdq_strings (Sequence[str]): PDQ hash strings to add
+        """
+        vectors = self._convert_pdq_strings_to_ndarray(pdq_strings)
+        self.faiss_index.add(vectors)
+
+    def search(
+        self, queries: t.Sequence[str], threshold: int = DEFAULT_MATCH_DIST
+    ) -> t.List[t.List[t.Any]]:
```
Dcallies, line 130: You can narrow this typing down more I think, tuple(int, int)

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        query_array: np.ndarray = self._convert_pdq_strings_to_ndarray(queries)
+        limits, distances, indices = self.faiss_index.range_search(
+            query_array, threshold + 1
+        )
+
+        results: t.List[t.List[t.Any]] = []
+        for i in range(len(queries)):
+            matches = [idx.item() for idx in indices[limits[i] : limits[i + 1]]]
+            dists = [dist for dist in distances[limits[i] : limits[i + 1]]]
+            results.append(list(zip(matches, dists)))
```
Dcallies, line 148: Do you need to use `tuple` instead of `list` here?

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+        return results
+
+    def __getstate__(self):
+        data = faiss.serialize_index(self.faiss_index)
+        return data
```
Dcallies, line 153: nit: inline
```suggestion
        return faiss.serialize_index(self.faiss_index)
```

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+    def __setstate__(self, data):
+        self.faiss_index = faiss.deserialize_index(data)
+
+    def _convert_pdq_strings_to_ndarray(
+        self, pdq_strings: t.Sequence[str]
+    ) -> np.ndarray:
+        """
+        Convert multiple PDQ hash strings to a numpy array.
+        Args:
+            pdq_strings (Sequence[str]): A sequence of 64-character hexadecimal PDQ hash strings
```
Dcallies, line 164: nit: on leaving off types

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+    return index, pdq_hashes
+
+
+def test_init(empty_index) -> None:
+    assert empty_index.threshold == DEFAULT_MATCH_DIST
+    assert isinstance(empty_index.index, _PDQHashIndex)
+    assert isinstance(empty_index.index.faiss_index, faiss.IndexFlatL2)
+    assert empty_index.index.faiss_index.d == DIMENSIONALITY
+    assert empty_index._deduper == dict()
+    assert empty_index._idx_to_entries == []
```
Dcallies, line 62: Suggest leaving these off - it doesn't really test anything

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+        entries=[
+            ("f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22", 0)
+        ]
+    )
+    return index, pdq_hashes
+
+
+def test_init(empty_index) -> None:
+    assert empty_index.threshold == DEFAULT_MATCH_DIST
+    assert isinstance(empty_index.index, _PDQHashIndex)
```
Dcallies, line 58: suggest leaving this off - doesn't test anything

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+    assert isinstance(empty_index.index.faiss_index, faiss.IndexFlatL2)
+    assert empty_index.index.faiss_index.d == DIMENSIONALITY
+    assert empty_index._deduper == dict()
+    assert empty_index._idx_to_entries == []
+
+
+def test_serialize_deserialize(empty_index) -> None:
```
Dcallies, line 65: blocking: This test doesn't really test much unless the index has some contents.

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+    assert isinstance(deserialized_index, PDQIndex2)
+    assert deserialized_index.threshold == empty_index.threshold
+    assert isinstance(deserialized_index.index, _PDQHashIndex)
+    assert isinstance(deserialized_index.index.faiss_index, faiss.IndexFlatL2)
+    assert deserialized_index.index.faiss_index.d == DIMENSIONALITY
+    assert deserialized_index._deduper == empty_index._deduper
+    assert deserialized_index._idx_to_entries == empty_index._idx_to_entries
+
+
+def test_serialize_deserialize_with_custom_index_threshold(
```
Dcallies, line 80: blocking: This test also doesn't test much, we can probably leave it off

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+    assert len(results) == 0
+
+
+def test_sample_set_near_match(sample_index):
+    """Test near matches in sample set."""
+    index, pdq_hashes = sample_index
+
+    # Create a near-match by flipping a few bits
+    base_hash = pdq_hashes[0]
+    near_hash = hex(int(base_hash, 16) ^ 0xF)[2:].zfill(64)  # Flip 4 bits
```
Dcallies, line 136: blocking: Did you forget to query near_hash?

ignorable: In another similar project, I wrote a helper on the signal type itself to provide a random hash that was X distance away from a given hash by flipping random bits.

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+
+    results = index.query(near_hash)
+    assert len(results) > 0  # Should find near matches
+    assert results[0].similarity_info.distance > 0
+
+
+def test_sample_set_threshold(sample_index):
+    """Test distance threshold behavior."""
```
Dcallies, line 144: nit: Try to avoid docs that are tautological with the function name

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+        "f" * 64,
+        "0" * 64,
+        "a" * 64,
+    ]
+    index = PDQIndex2[str](
+        entries=[
+            ("f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22", 0)
+        ]
+    )
+    return index, pdq_hashes
```
Dcallies, line 53: blocking: Why do you return hashes which aren't in the index?

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+    _PDQHashIndex,
+    DIMENSIONALITY,
+    DEFAULT_MATCH_DIST,
+)
+
+
+@pytest.fixture
+def empty_index():
+    """Fixture for an empty index."""
+    return PDQIndex2[str]()
```
Dcallies, line 15: blocking: Why subscript this?
```suggestion
    return PDQIndex2()
```

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,195 @@
...
+def sample_index():
+    """Fixture for an index with a small sample set."""
+    pdq_hashes = [
+        "f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22",
+        "f" * 64,
+        "0" * 64,
+        "a" * 64,
+    ]
+    index = PDQIndex2[str](entries=[(h, pdq_hashes.index(h)) for h in pdq_hashes])
+    return index, pdq_hashes
```
Dcallies, line 36: blocking: I'm not sure about this pattern of returning an index and the sample hashes. It doesn't seem like you are saving that much time vs something like this:

```
SAMPLE_HASH = "f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22"

SAMPLE_HASHES = [
  SAMPLE_HASH,
  "f" * 64,
  "0" * 64,
  "a" * 64,
]

@dataclass
class IndexRecord:
  original_hash: string
  idx: int 

  def get_expected_distance(matched_hash) -> int:
    return pdq_utils.simple_distance_binary(self.original_hash, matched_hash)

def make_index(hashes):
  return PDQIndex2.build((h, StoredRecord(h, i) for i, h in enumerate(hashes)

def test_something():
  index = make_index(SAMPLE_HASHES)

  query_hash = "..."

  for d, record in index.search([query_hash ]):
    assert d == record.expected_distance(query_hash)
  
```

And you also aren't consistent on the test fixtures about what will be in the return. Fixtures are nice, but they aren't mandatory

Dcallies (changes requested):
Looking good so far!

Toplevel comments:
1. I am not convinced yet that this indexing solution is correct yet. To help convince me, you can write a unittest that generates two large sets of random hashes (PDQSignal.get_random_signal()) b and q (e.g. 1,000 and 10,000), uses the brute force method to check exactly how many of hashes in b should match the hashes in q, then loads the b hashes into your indexing solution, and then queries all of the q hashes and makes sure you exactly get the expected results. 
1. Many of your existing tests are tautological or should check to make sure the matching hashes are actually the ones that should be matching.  You can likely replace many of your existing tests with variations on 2 above - I put some ideas for you in the IndexRecord class, but there are others that might work.
1. I recommend adding a method that can return a random hash exactly X bits different from a given hash, and then using that to help in your tests.

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
haianhng31: so that I can use the other hashes to test for unmatch when needed. 
For example, in the test: `test_one_entry_sample_index`

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
haianhng31: I query the near_hash on the [line](https://github.com/facebook/ThreatExchange/pull/1698/files#diff-52467a9eead5544de4d6082c507dd5c96bebe25c34a1a5fdfebbdb9f2c223846R138) below

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,185 @@
...
+from threatexchange.signal_type.index import (
+    IndexMatchUntyped,
+    SignalSimilarityInfoWithIntDistance,
+    SignalTypeIndex,
+    T as IndexT,
+)
+from threatexchange.signal_type.pdq.pdq_faiss_matcher import (
+    PDQMultiHashIndex,
+    PDQFlatHashIndex,
+    PDQHashIndex,
```
Dcallies, line 23: We don't want to import these because they are the code you are going to be replacing!

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,138 @@
...
+        bits[pos] = "0" if bits[pos] == "1" else "1"  # Flip selected bit positions
+    modified_hash = hex(int("".join(bits), 2))[2:].zfill(64)  # Convert back to hex
+
+    return modified_hash
+
+
+def test_pdq_index():
+    # Make sure base_hashes and query_hashes have at least 100 similar hashes
+    base_hashes = SAMPLE_HASHES + [PdqSignal.get_random_signal() for _ in range(1000)]
+    query_hashes = SAMPLE_HASHES + [PdqSignal.get_random_signal() for _ in range(10000)]
```
Dcallies, line 47: Why add SAMPLE_HASHES here? If you are trying to guarantee matches, you could generate less query hashes and copy the banked hashes in

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,138 @@
...
+
+def test_pdq_index_with_exact_distance():
+    thresholds: t.List[int] = [10, 31, 50]
+    indexes: t.List[PDQIndex2] = []
+    for thres in thresholds:
+        index = PDQIndex2(
+            entries=[(h, SAMPLE_HASHES.index(h)) for h in SAMPLE_HASHES],
+            threshold=thres,
+        )
+        indexes.append(index)
```
Dcallies, line 78: ignorable: Could do this as a comprehension

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,138 @@
...
+    results = index.query(SAMPLE_HASHES[0])
+
+    # Should find all entries associated with the hash
+    assert len(results) == 4
+    for result in results:
+        assert result.similarity_info.distance == 0
+
+
+def test_one_entry_sample_index():
+    """Test how the index handles when it only has one entry."""
```
Dcallies, line 126: nit: Add a reference to the issue

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,138 @@
...
+    base: t.List[str], query: str, threshold: int = 32
+) -> t.Set[int]:
+    matches = set()
+    query_arr = convert_pdq_strings_to_ndarray([query])[0]
+
+    for i, base_hash in enumerate(base):
+        base_arr = convert_pdq_strings_to_ndarray([base_hash])[0]
+        distance = np.count_nonzero(query_arr != base_arr)
+        if distance <= threshold:
+            matches.add(i)
```
Dcallies, line 22: blocking: Why create a new version of this instead of using existing methods like `simple_distance` or PDQSignal.compare_hash`?

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -0,0 +1,132 @@
...
+    convert_pdq_strings_to_ndarray,
+)
+
+PDQIndexMatch = IndexMatchUntyped[SignalSimilarityInfoWithIntDistance, IndexT]
+
+
+class PDQIndex2(SignalTypeIndex[IndexT]):
+    """
+    Indexing and querying PDQ signals using Faiss for approximate nearest neighbor search.
+    """
```
Dcallies, line 31: Nit:

Add a note that the `2` is because it's a redo of the existing PDQ index designed to be simpler and fix hard-to-squash bugs in the existing implementation. The intention is that this class will replace the original index in pytx 2.0

python-threatexchange/threatexchange/signal_type/pdq/pdq_utils.py
```diff
@@ -49,3 +52,19 @@ def pdq_match(pdq_hex_a: str, pdq_hex_b: str, threshold: int) -> bool:
...
+
+def convert_pdq_strings_to_ndarray(pdq_strings: t.Sequence[str]) -> np.ndarray:
+    """
+    Convert multiple PDQ hash strings to a numpy array.
+    """
+    if not all(len(pdq_str) == PDQ_HEX_STR_LEN for pdq_str in pdq_strings):
+        raise ValueError("All PDQ hash strings must be 64 hex characters long")
+
+    binary_arrays = []
+    for pdq_str in pdq_strings:
```
Dcallies, line 65: ignorable: If you move the ValueError check inside the loop you can change the type of pdq_strings to be `t.Iteratble` and accept generators.

Dcallies (changes requested):
So close!

1. The way you are seeding your hashes is causing duplicates in your query set - see inline blocking comment
2. As part of your test plan (doesn't have to be part of the test itself), please try generate a very large number of hashes and queries - e.g. 10,000 banked and 100,000 queries - something that takes 5-10 seconds to check. We want to be certain that the index is correct!
3. Please add a test that serializes and deserializes a non-empty index - you can do something cheeky like:
```
original = ... # do stuff to prepare index
deserialized = PDQIndex2,deserialize(index.serialize(...))
for index in (original, deserialized):
   # The original test goes here, should get same results on original or deserialized
```

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -0,0 +1,144 @@
+import typing as t
+import numpy as np
+import random
+
+from threatexchange.signal_type.pdq.pdq_index2 import PDQIndex2
+from threatexchange.signal_type.pdq.signal import PdqSignal
+from threatexchange.signal_type.pdq.pdq_utils import simple_distance
+
+
+def _generate_sample_hashes(size: int, seed: int = 42):
```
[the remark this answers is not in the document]
haianhng31: I'm sorry what do you mean with the image:P ?

haianhng31:
@Dcallies 
I have tried with 1,000 banked & 10,000 queries in the unit test. The test passed but it took 4 minutes to run. 
And if i tried with 10,000 banked & 100,000 queries, the test took forever.

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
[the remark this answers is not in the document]
Dcallies: 42 is the "Answer to the Ultimate Question of Life, The Universe, and Everything" in the book series "The Hitchhiker's Guide to the Galaxy", of which the image is on the cover. I thought you were a fan :P 

Read more: https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy

PR #1700, opened 2024-11-29
[vpdq] Add data flow diagram, misc docs cleanup

ianwal:
Summary
---------

I was looking at vPDQ again and noticed the initial docs has a lot of words and can be hard to understand.

Tabulating (?) the TMK comparison and adding a simple high level diagram of how vPDQ works can help new readers understand the project better.

- Add simple high level mermaid diagram for vPDQ data flow.
  - [github supports rendering mermaid diagrams](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams). VSCode can render markdown, but currently requires an extension like [Markdown Preview Mermaid Support](https://marketplace.visualstudio.com/items?itemName=bierner.markdown-mermaid) to render it.
  - This diagram doesn't really follow any particular format like UML.
- Put TMK+PDQF / vPDQ comparison in table form for readability
- Other misc cleanup
  - All changes maintain the same amount of info but in a different format that may be easier for new users to understand.

Consideration: mermaid can also be exported to svg, so instead of a codeblock I could add a svg file to avoid readers needing a mermaid renderer. Probably will still need to put the mermaid code somewhere since I think mermaid -> svg is a one way conversion.

Test Plan
---------

Mermaid renders as expected in github:

> image

### DIFF
--- a/vpdq/README.md
+++ b/vpdq/README.md
@@ -8,18 +8,57 @@ See [CPP implementation](#cpp-implementation) for how to install and use vpdq.
 
 ## Compared to TMK+PDQF
 
-Compared to TMK+PDQF (TMK), which also relies on the PDQ image hashing algorithm:
-TMK optimizes for identical videos (same length), vPDQ can match subsequences or clips within videos.
-TMK has a fixed-length hash, which simplifies matching lookup, and can be near constant time with the help of FAISS. vPDQ produces a variable length hash, and requires a linear comparison of candidates. This requires either an O(n*F<sub>c</sub>*F<sub>q</sub>) lookup where n is the number of videos being compared, and F<sub>c</sub> is the average number of frames per compared video and F<sub>q</sub> is the number of frames in the source video, or an initial filtering pass to reduce the candidates, which can potentially discard matching videos.
-Both TMK and vPDQ are backed by PDQ, and so inherit both PDQ’s strengths and weaknesses.
+Compared to TMK+PDQF (TMK):
+
+| Feature                                                   | vPDQ     | TMK+PDQF |
+|-----------------------------------------------------------|:--------:|:--------:|
+| Uses PDQ for hashing frames                               | ✅       | ✅      |
+| Optimized for identical videos (same length)              | ❌       | ✅      |
+| Supports subsequence/clip matching                        | ✅       | ❌      |
+| Fixed-length hash (near constant time search for matches) | ❌       | ✅      |
+
+Both TMK and vPDQ are backed by PDQ for hashing video frames. Therefore, they both inherit PDQ's strengths and weaknesses.
+
+TMK optimizes for identical videos (same length). vPDQ can match subsequences or clips within videos.
+
+TMK has a fixed-length hash, which simplifies matching lookup, and can be near constant time with the help of FAISS. vPDQ produces a variable length hash, and requires a linear comparison of candidates.
+* vPDQ linear search complexity is O(n*F<sub>c</sub>*F<sub>q</sub>), where n is the number of videos being compared, and F<sub>c</sub> is the average number of frames per compared video and F<sub>q</sub> is the number of frames in the source video.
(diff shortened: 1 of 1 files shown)


vpdq/README.md
```diff
@@ -8,18 +8,57 @@ See [CPP implementation](#cpp-implementation) for how to install and use vpdq.
...
+### Producing a Video Hash
+
+The algorithm for producing the video hash is:
+
+1. Given a video, convert it into a sequence of frame images.
+2. For each frame image, use the PDQ hashing algorithm to produce frame perceptual hashes.
+3. Finally, assemble the collection of hashed frames to produce the video perceptual hash.
+
+> **Note:** A subset of all frames can be hashed, such as only the frames at every 1sec interval, to reduce the number of frame
+> perceptual hashes. In general, adjacent frames are very similar, so they are not very useful for finding matches.
```
Dcallies, line 39: I think step 3 and this note can probably be combined in some way - it's easy to glaze over the note. Rather than "assemble", maybe something like

> 3. From the frame perceptual hashes, select hashes to include in the final video perceptual hash.

If we had to name these stages it might be:
1. Convert to image frames
2. Convert images to hashes
3. Select hashes

If there was a word shortage and we had to use only one word it might be:
1. Frames
2. Hash
3. Select

vpdq/README.md
```diff
@@ -8,18 +8,57 @@ See [CPP implementation](#cpp-implementation) for how to install and use vpdq.
...
+  * To speed this up, an initial filtering pass can be used to reduce the candidates. The downside is it could discard matching videos, causing false-negatives.
 
 ## Description of Algorithm
 
-### Producing a Hash
+### Producing a Video Hash
+
+The algorithm for producing the video hash is:
+
+1. Given a video, convert it into a sequence of frame images.
```
Dcallies, line 34: I think you can combine a part of your note here, since it might be easy to glaze over - I think the bolded part is enough, but you can be even more explicit

>  1. Given a video, convert it into a sequence of frame images **using some criteria** (e.g. 1 image / sec).

vpdq/README.md
```diff
@@ -8,18 +8,57 @@ See [CPP implementation](#cpp-implementation) for how to install and use vpdq.
...
+2. For each frame image, use the PDQ hashing algorithm to produce frame perceptual hashes.
+3. Finally, assemble the collection of hashed frames to produce the video perceptual hash.
+
+> **Note:** A subset of all frames can be hashed, such as only the frames at every 1sec interval, to reduce the number of frame
+> perceptual hashes. In general, adjacent frames are very similar, so they are not very useful for finding matches.
+
+The following diagram shows the high level data flow for hashing a video with vPDQ:
+
+```mermaid
```
Dcallies, line 43: fancy fancy, I'll keep this in mind for other projects in this repo

vpdq/README.md
```diff
@@ -99,7 +138,9 @@ is_match = c_pct_matched >= P_c and q_pct_matched >= P_q
 
 > **Note**: The frame number and the timestamp is not used at all in this comparison. The frames are treated as an unordered “bag of hashes”. The frame number and timestamp are included in each feature in the reference implementation in case of future expansion.
 
-### Pruning Candidates
+Beyond pruning frames from candidates, it may be desirable to further prune to just sampled or key frames in candidate videos to control index size, but this may result in videos being incorrectly pruned.
```
Dcallies, line 141:
```suggestion
Beyond pruning frames from candidates, it may be desirable to further prune to just sampled or key frames in candidate videos to control index size, but this may change whether videos are consider matching or not.
```

Dcallies (approved):
Thanks a ton for doing this! I'm going to merge as is, since I don't think anything is blocking, feel free to put up another PR if you feel like doing the suggested changes, otherwise I can do it at some arbitrary future point.

Issue #1688, opened 2024-11-08
[py-tx][mlh] Move IContentTypeConfigStore from HMA to python-threatexchange


Dcallies:
See #1687 for context

https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py#L42-L60

This should be 3 PRs:
1. Copy IContentTypeConfigStore to a new location in python-threatexchange/threatexchange/storage/interfaces.py
2. Bump the minor version of python-threatexchange
3. Remove IContentTypeConfigStore from HMA

b8zhong:
This can be closed I think

PR #1702, opened 2024-11-29
[Pytx] Move ContentTypeConfig to python-threatexchange/threatexchange/storage/interfaces.py

haianhng31:
Summary
---------

* This resolves the 1st part of issue #1688 
* I'm not sure if I should change anything. Right now, I'm merely copying everything from HMA's storage/interface.py to python-threatexchange/threatexchange/storage/interfaces.py (please let me know if i should make any adjustments). 
* Also, for the next step of bumping the minor version of python-threatexchange, I would like to have some guidance as well!


Dcallies (approved):
Looks good!

PR #1704, opened 2024-12-09
TAT API signal checkpointing

Bruce-Pouncey-TAT:
Summary
---------
Hi! @Dcallies

Below is my PR for the updated Tech Against Terrorism API.

We're yet to release the checkpoint compatible API to production but we're expecting to release in the coming days. I have the current changes working against our development environment.

I have a question regarding the checkpointing.

Checkpoints are now saved as part of the `fetch_iter` process and running `tx fetch` a second time does continue where it left off. Though if there are no hashes remaining on a `tx fetch` no checkpoint gets set meaning all the hashes are downloaded again.

example
1st fetch -> gets 10k hashes -> checkpoint set 
2nd fetch -> gets 5k new hashes -> checkpoint set
3rd fetch -> no new hashes -> no checkpoint set 
4th fetch -> gets 10k hashes -> checkpoint set (same as 1st fetch)

I'm wondering if this is expected behaviour ?

Below is the response from our server with our new checkpoint compatible API.

```
{
    "count": 25608,
    "next": "http://dev.terrorismanalytics.org/hash-list/v2/all?limit=1000&offset=1000&order=asc",
    "previous": null,
    "checkpoint": "1724856487.709035,10594",
    "results": [
        {
            "hash_digest": "<hash-digest>",
            "algorithm": "MD5",
            "ideology": "Islamist",
            "file_type": "png",
            "deleted": false,
            "updated_on": 1704901040.222779,
            "id": 2819
        },
        {
            "hash_digest": "<hash-digest>",
            "algorithm": "MD5",
            "ideology": "Islamist",
            "file_type": "png",
            "deleted": false,
            "updated_on": 1704901040.24492,
            "id": 2820
        },
        {
            "hash_digest": "<hash-digest>",
            "algorithm": "MD5",
            "ideology": "Islamist",
            "file_type": "png",
            "deleted": false,
            "updated_on": 1704901040.268179,
            "id": 2821
        },
}
```

Also can you please advise on how we should handle hashes with the `deleted` flag. I did notice `state.FetchedSignalMetadata` but it appears to be for sharing opinions on hashes which we don't support and will unlikely moving forward as our hashes are verified by subject matter experts, unless my understanding is wrong!

Thanks for your patience,

Bruce

Test Plan
---------

Manual
`threatexchange config api tat --credentials 'TCAP_USERNAME' 'TCAP_PASSWORD'`
`threatexchange config collab edit tat --create "TAT"`
`threatexchange -v fetch`

Auto
`pytest`

PS
I had some issues installing vpdq in both dev container and pip there was nothing in the docs that could assist me with this. Below is the output:

```
  error: subprocess-exited-with-error
  
  × Building wheel for vpdq (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [9 lines of output]
      INFO:root:running bdist_wheel
      INFO:root:running build
      INFO:root:running build_ext
      INFO:setup.py:Removing CPP build directory /private/var/folders/t6/gt00fl5j2xd2146shhhnbt7r0000gq/T/pip-install-9w6apfa6/vpdq_7ce79cb2e7944f65955695d1490b98b5/cpp/build...
      INFO:setup.py:Removing compiled Cython files /private/var/folders/t6/gt00fl5j2xd2146shhhnbt7r0000gq/T/pip-install-9w6apfa6/vpdq_7ce79cb2e7944f65955695d1490b98b5/python/vpdq.cpp...
      INFO:setup.py:Removing libraries-dir.txt file /private/var/folders/t6/gt00fl5j2xd2146shhhnbt7r0000gq/T/pip-install-9w6apfa6/vpdq_7ce79cb2e7944f65955695d1490b98b5/cpp/libraries-dirs.txt...
      INFO:setup.py:Creating CPP build directory...
      INFO:setup.py:Generating CMake project...
      error: [Errno 2] No such file or directory: 'cmake'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for vpdq
Failed to build vpdq
ERROR: Could not build wheels for vpdq, which is required to install pyproject.toml-based projects
```


Dcallies:
Hey Bruce, thanks for submitting the updated code. Fetching from the beginning agani not expected behavior if the checkpoint is working, let me see if I can figure out from reading your PR.

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/api.py
```diff
@@ -134,27 +142,52 @@ def get_auth_token(self, username: str, password: str) -> t.Optional[str]:
...
-            hash_list = self._get(full_url=response["file_url"])
-
-            return hash_list
+            return results
 
         except Exception as exception:
             logging.error("Failed to get hash list: %s", exception)
             raise
+
+    def fetch_hashes_iter(self, next_page: str) -> t.Iterator[TATHashListResponse]:
```
Dcallies, line 173: It looks like this was copied from the ThreatExchange version of the API, which I think I mostly regretted. It's fine to keep this as is, but it's likely the source of your bug, since it has slightly different behavior resuming vs not. 

blocking: `next_page` looks like it's actually your checkpoint string, aka "after" parameter

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -10,17 +10,26 @@
...
 from threatexchange.signal_type.md5 import VideoMD5Signal
 
+
+@dataclass
+class TATCheckpoint(state.FetchCheckpointBase):
+    checkpoint: str
+
+    @classmethod
+    def from_tat_fetch(cls, response: api.TATHashListResponse) -> "TATCheckpoint":
+        print("CHECKPOINT:", response["checkpoint"])
```
Dcallies, line 25: blocking: Don't land code with debugging prints in them.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -10,17 +10,26 @@
...
-    CollaborationConfigWithDefaults,
-)
+from threatexchange.exchanges.collab_config import CollaborationConfigWithDefaults
 from threatexchange.signal_type.signal_base import SignalType
 from threatexchange.signal_type.pdq.signal import PdqSignal
 from threatexchange.signal_type.md5 import VideoMD5Signal
 
+
+@dataclass
+class TATCheckpoint(state.FetchCheckpointBase):
```
Dcallies, line 20: Consider also implementing the method for showing the "after" time, which will help your debugging: 

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/fetch_state.py#L41

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -90,27 +99,17 @@ def get_client(self) -> api.TATHashListAPI:
...
-
-
-def _is_compatible_signal_type(record: t.Dict[str, str]) -> bool:
-    compatible_video_types = ["mov", "m4v", "mp4", "webm"]
-
-    if record["algorithm"] == "MD5":
-        return record["file_type"] in compatible_video_types
-
-    return record["algorithm"] == "PDQ"
+        for result in client.fetch_hashes_iter(checkpoint.checkpoint):
```
Dcallies, line 107: blocking: [2/3] Then you are passing in the checkpoint string to fetch_hashes_iter

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/api.py
```diff
@@ -134,27 +142,52 @@ def get_auth_token(self, username: str, password: str) -> t.Optional[str]:
...
+        - results (List[TATHashListEntry]): A list of hash list entries
+
+        if next in null we have no more hashes left to fetch meaning we break out of the loop
+        """
+
+        has_more = True
+        next_page = next_page
+
+        while has_more:
+            response = self.fetch_hashes(after=next_page)
```
Dcallies, line 189: blocking: It looks like this is your "checkpoint" not the next page link.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -90,27 +99,17 @@ def get_client(self) -> api.TATHashListAPI:
...
+            translated = (_get_delta_mapping(r) for r in result["results"])
+            yield state.FetchDelta(
+                dict(t for t in translated if t[0][0]),
+                TATCheckpoint.from_tat_fetch(result),
```
Dcallies, line 111: I suspect your bug is either L107 or L111

blocking question: if your API returns zero hashes:
1. Does it yield a response?
2. If so, what does the response checkpoint show as?

(Expected - if the checkpoint is not advanced, this method should either yield nothing, or should yield zero records with a valid checkpoint of where to call next)

python-threatexchange/threatexchange/exchanges/clients/techagainstterrorism/api.py
```diff
@@ -94,18 +103,17 @@ def _get(
         self,
         endpoint: t.Optional[str] = None,
         auth_token: t.Optional[str] = None,
-        full_url: t.Optional[str] = None,
+        **params,
```
Dcallies, line 106: ignorable: `**params` is slightly dangerous because you can accidentally create slightly misnamed versions of enpoint and auth_tok if you aren't paying attention

Dcallies (changes requested):
Hey Bruce, looking pretty good - congrats on the fast progress on the API update, I think your users will thank you in the long term!

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Bruce-Pouncey-TAT: Thanks David,

Looking at our API it will return the following if there are no hashes left.
```
{
    "count": 19636,
    "next": null,
    "previous": "http://dev.terrorismanalytics.org/hash-list/vafter=1730204429.203484%2C28022&limit=1000&offset=19000&order=asc",
    "checkpoint": null,
    "results": []
}
```

In this case inside `techagainsterroism_api.py` I am checking for this and returning early.

```
for result in client.fetch_hashes_iter(start_time):

    if result.next is None:
        return

    translated = (_get_delta_mapping(r) for r in result.results)
    yield state.FetchDelta(
        dict(t for t in translated if t[0][0]),
        TATCheckpoint.from_tat_fetch(result),
    )
```

This seems to have resolved the issue and now the last checkpoint to get set will remain in place.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Bruce-Pouncey-TAT: Hi David,

Our checkpoint isn't a timestamp, its a string with a timestamp and content id joined by a comma eg, `"123456,123"` when passing said value to `get_progress_timestamp` it throws an error due to the checkpoint not being an integer.

Bruce-Pouncey-TAT:
Hi @Dcallies 

I have made the requested changes and the checkpointing is working as expected now.

I am having issues updating the tests in `threatexchange/exchanges/impl/tests/*` with the following error;

-----------
ERROR tests/test_techagainstterrorism.py - DeprecationWarning: numpy.core._multiarray_umath is deprecated and has been renamed to numpy._core._multiarray_umath. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core._multiarray_umath.__cpu_features__.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-----------

Not sure if this is my setup or numpy needs updating which I am hesitant to do given I'm not too familiar with the package and not sure how the change will propagate across the repo.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -121,13 +140,13 @@ def _type_mapping() -> t.Dict[str, str]:
...
-        return (("", ""), None)
-
-    type_str = _type_mapping().get(record["algorithm"])
+        return (("", ""), state.FetchedSignalMetadata())
 
+    type_str = _type_mapping().get(record.algorithm, "")
     metadata = state.FetchedSignalMetadata()
-    return ((type_str or "", record["hash_digest"]), metadata)
+
+    return ((type_str, record.hash_digest), metadata)
```
Dcallies, line 152: blocking: How to handle deletions:

```suggestion
    return ((type_str, record.hash_digest), None if record.deleted else metadata)
```

Bruce-Pouncey-TAT:
HI David,

I have made the changes you suggested thanks again!

One question on this piece you recommended.
```
if result.checkpoint:
    checkpoint = TATCheckpoint.from_tat_fetch(result)

            translated = (_get_delta_mapping(r) for r in result.results)
            yield state.FetchDelta(
                dict(t for t in translated if t[0][0]),
                checkpoint,
            )
```
 
I added the following else statement:

```
            if result.checkpoint:
                checkpoint = TATCheckpoint(result.checkpoint)
            else:
                checkpoint = TATCheckpoint("")
            
             translated = (_get_delta_mapping(r) for r in result.results)
             ...
```

This seemed to fix the mypy issue and all still works as intended when testing manually just wanted to sanity check.

Thanks!

Dcallies:
> This seemed to fix the mypy issue and all still works as intended when testing manually just wanted to sanity check.

I think this will cause your "rewind" issue that you initially described.

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -112,13 +108,15 @@ def fetch_iter(
 
         for result in client.fetch_hashes_iter(start_time):
 
-            if result.next is None:
-                return
+            if result.checkpoint:
+                checkpoint = TATCheckpoint(result.checkpoint)
+            else:
+                checkpoint = TATCheckpoint("")
```
Dcallies, line 114: What cases can cause the checkpoint on the return to be empty? Is the a valid response?

I'm worried that you can accidentally rewind your checkpoint back to the beginning, which is exactly the behavior you were observing earlier on in this, and so I think we should defend against it

If it's not a valid response, then this is a bit less spooky:
```suggestion
            asset result.checkpoint, "API didn't return a checkpoint?"
            # alt: assert not checkpoint or result.checkpoint, "API rewound checkpoint!"
            checkpoint = TATCheckpoint(result.checkpoint)
```

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
```diff
@@ -90,27 +96,38 @@ def get_client(self) -> api.TATHashListAPI:
...
+        start_time = ""
+
+        if checkpoint is not None:
+            start_time = checkpoint.checkpoint
 
-        yield state.FetchDelta(
-            dict(t for t in translated if t[0][0]),
-            checkpoint=state.NoCheckpointing(),
-        )
+        for result in client.fetch_hashes_iter(start_time):
```
Dcallies, line 109: blocking; This isn't a start time, it's a checkpoint, right?
```suggestion
        checkpoint = "" if checkpoint is None else checkpoint.checkpoint
        for result in client.fetch_hashes_iter(checkpoint):
```

Dcallies (commented):
Some more inline comments

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Bruce-Pouncey-TAT: Okay,

So if we reach the end of the list and there are no hashes remaining the results list will be `[]` and the checkpoint will be 'null'

and yes you were right it does indeed cause a rewind 🤦

python-threatexchange/threatexchange/exchanges/impl/techagainstterrorism_api.py
Dcallies: Suggest that when there are no hashes left, your API should either:
1. [Recommended] Return the checkpoint corresponding to the last record (latest checkpoint)
2. Return the checkpoint passed by the user.

Dcallies (approved):
Let's land and iterate, I'll give a stab at the typing

Dcallies:
I'm working on the vpdq issues on #1711

Issue #1689, opened 2024-11-08
[pytx][mlh] Move ISignalTypeConfigStore from HMA to python-threatexchange


Dcallies:
See https://github.com/facebook/ThreatExchange/issues/1687 for context

https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py#L63-L124

This should be 3 PRs:

1. Copy IContentTypeConfigStore to a new location in python-threatexchange/threatexchange/storage/interfaces.py
2. Bump the minor version of python-threatexchange
3. Remove IContentTypeConfigStore from HMA

There will be followup work after this, will create as a new issue.

b8zhong:
Guessing we can close this, right? @Dcallies

b8zhong:
@Dcallies could you end up posting the followup work (if it wasn't done yet)? Thanks a lot!

Dcallies:
Made #1754 to cover the overall thrust

b8zhong:
Thanks!

PR #1706, opened 2024-12-09
[pytx] Bump the minor version of python-threatexchange

haianhng31:
Summary
---------

* This is the 2nd step from issue #1689 
* Bump the minor version of python-threatexchange


Dcallies:
If we do follow up version changes for other files, let's only do the bug fix number (the last one).

In this case, all the changes that the fellows have made to the the CLI itself constitutes a minor version change I think. So we'll go forward with this one

Dcallies:
If we do follow up version changes for other files, let's only do the bug fix number (the last one).

In this case, all the changes that the fellows have made to the the CLI itself constitutes a minor version change I think. So we'll go forward with this one

Dcallies (approved):

PR #1707, opened 2024-12-11
Remove IContentTypeConfigStore from HMA

haianhng31:
Summary
---------
* This resolves the 3rd part of issue #1688 
* Delete IContentTypeConfigStore in HMA 
* In HMA, import IContentTypeConfigStore from pytx to be included in IUnifiedStore
* Import interfaces from pytx to use in [storage/mocked.py](https://github.com/facebook/ThreatExchange/compare/main...haianhng31:IContentTypeConfigStore3?expand=1#diff-8c3aa92bb914526d0a98d8fdb37914d004b1109fa2e0e4822a6e01a3582d563f) and [storage/postgres/impl.py](https://github.com/facebook/ThreatExchange/compare/main...haianhng31:IContentTypeConfigStore3?expand=1#diff-d75ccee7de8e3db45cc3428fbc79b181b36857278c19f3f24432f78dcc2fb3ed) to make sure that the return types match

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
@@ -22,6 +22,7 @@ import time
 
 import flask
 
+from threatexchange.cli.storage.interfaces import IContentTypeConfigStore
 from threatexchange.utils import dataclass_json
 from threatexchange.content_type.content_base import ContentType
 from threatexchange.signal_type.signal_base import SignalType
@@ -48,16 +49,6 @@ class ContentTypeConfig:
     # Content types that are not enabled should not be used in hashing/matching
     enabled: bool
     content_type: t.Type[ContentType]
-
-
-class IContentTypeConfigStore(metaclass=abc.ABCMeta):
-    """Interface for accessing ContentType configuration"""
-
-    @abc.abstractmethod
-    def get_content_type_configs(self) -> t.Mapping[str, ContentTypeConfig]:
-        """
-        Return all installed content types.
-        """
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -22,6 +22,7 @@
 
 import flask
 
+from threatexchange.cli.storage.interfaces import IContentTypeConfigStore
```
Dcallies, line 25: I think you are getting a mypy failure here because this might need a missing `__init.py__` file in the interfaces directory

Dcallies (changes requested):
From the CI output, it looks like it's importing threatexchange 1.1.1

This is because our 1.2 version publish failed! 

https://github.com/facebook/ThreatExchange/actions/runs/12247017442/job/34164057430

We'll have to fix that before this works

Dcallies:
Okay, I think I fixed the build and package, try re-submitting to see if everything passes again.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
haianhng31: i don't know why mypy keeps complaining about not finding the module. I have checked: 
* threatexchange package is installed 
* `threatexchange.storage.interfaces` is a valid module
* has `__init__.py` in interfaces directory

Dcallies:
You have the same problem I fixed - the `__init__.py` doesn't count unless it's non-empty - you'll need to add the copyright header to the `__init__.py` and then bump the version.

You can confirm that the folder is included by running `setup build` and then inspecting the content of the `build/` directory.

python-threatexchange/threatexchange/storage/__init__.py
```diff
@@ -0,0 +1 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
Dcallies, line 1: You'll need to move this to its own PR, then we'll need to bump the version and release a new bugfix version!

PR #1710, opened 2024-12-12
[pytx] Bump ci+publish version

Dcallies:
Summary
---------

Publish action is failing for mysterious reasons (#1709), maybe blindly increasing the version will help!

Test Plan
---------

Can build and publish locally (tested with pypi-test), but running into a lot of other mysterious problems as well - vpdq won't build on my devcontainer, for example.


Dcallies:
I cannot figure out why vpdq is failing to compile - I am running into different problems locally. The CI links do indeed show that the packages its trying to install are 404s, but as best as I can tell, they are unrelated to the changes I am making here, since they reproduce even untouched.

PR #1713, opened 2024-12-12
[pytx] More twiddling of pytx release

Dcallies:
Just brute forcing the versioning, following https://github.com/facebook/ThreatExchange/actions/runs/12298354254/job/34321583371


python-threatexchange/version.txt
```diff
@@ -1 +1 @@
-1.2.1
```
Dcallies, line 1: Rewinding this will probably fail the test publish, since I already put it there, but will retry assuming build succeeds

PR #1724, opened 2024-12-16
[HMA] Remove ISignalTypeConfigStore from HMA

Mackay-Fisher:
Summary
This resolves the 3rd part of issue #1689 
Delete ISignalTypeConfigStore in HMA
In HMA, import ISignalTypeConfigStore from pytx to be included in IUnifiedStore

Test Plan
I was not entirely sure what to test to verify. There was a module issue, so I added the init py to allow for the module usage in HMA. I am not sure if that was the correct approach however that is the pr referenced in #1725 

Update: Yay the version change worked!

### DIFF
--- a/python-threatexchange/threatexchange/cli/storage/interfaces.py
+++ b/python-threatexchange/threatexchange/cli/storage/interfaces.py
@@ -2,6 +2,7 @@
 
 """
 Common interface for persisting pytx configuration and concepts.
+
 Most of the individual components of pytx are find to use piecemeal, and
 the full interface covers the most complex and complete useage. A usecase
 with one collection of hashes using one algorithm might be better off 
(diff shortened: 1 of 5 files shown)


python-threatexchange/threatexchange/cli/storage/interfaces.py
```diff
@@ -1,60 +1,11 @@
...
-proper as part of a migration to dbm 
-(github.com/facebook/ThreatExchange/issues/1687). The general approach is:
-1. Copy the interface unchanged from HMA to pytx
-2. Release a new version of pytx
-3. Delete the copy in HMA and update all references to the pytx version
-
-In parallel, we intend to slowly migrate the CLI storage components to
-follow the same interface.
-"""
-
```
Dcallies, line 29: blocking: This comment is useful, why delete it?

Dcallies, line 29: blocking: This comment is useful, why delete it?

Dcallies (changes requested):
I think you just had a merge collision that resulted in you deleting the docstring in pytx/.../interfaces.py, I can't easily restore it from here, if you fix that you should be good.

Dcallies (approved):
LGTM!

PR #1725, opened 2024-12-16
[py-tx] update version to include init py to access interfaces storage…

Mackay-Fisher:
Add version change to allow for module access to fix pytx imports in issue #1688 and #1689


Dcallies:
Normally I suggest breaking these up, but this is pretty smol, so let's go for it.

Dcallies (approved):

PR #1729, opened 2024-12-22
Check for hasher role in /lookup endpoint

### LINKED ISSUE #1703: [hma] Cleanup /m/lookup Hash + Match API
Some users have asked for an API that will allow them to combine the hash+match API into one call. It turns out this already exists as [/m/lookup](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py#L188-L222), but this triggers hashing (high CPU cost) even if the hashing role is not enabled, which defeats the point of roles!

There's also a development/UI version of this already here: https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/ui.py#L170 - this probably should be replaced with the new and improved method.

Your mission, should you choose to accept it:
1. Refactor /lookup so that that the URL and POST versions are only enabled when the hashing role is enabled. At the very worst, you can just check for the role in the app config and throw if it's not.
2. Delete the /ui version of lookup and replace it with your new fixed version (can be a separate PR)
3. Make sure everything is tested

aryzle:
Summary
---------

fixes #1703 

TODO:
- [ ]  unit test

Test Plan
---------

1. set `ROLE_HASHER = False` in development_omm_config.py
2. start server
3. try GET/POST to `/lookup` endpoint, should error now with 403 "Hashing is disabled, missing role"

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/postcreate.sh
+++ b/hasher-matcher-actioner/.devcontainer/postcreate.sh
@@ -3,7 +3,7 @@ set -e
 
 pip install --editable .[all]
 
-# Find Python packages in opt and install them
+# Find Python packages in extensions and install them
 for setup_script in "$(find /workspace/extensions -name setup.py)"
 do
     module_dir="$(dirname "$setup_script")"
--- a/hasher-matcher-actioner/src/OpenMediaMatch/app.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/app.py
@@ -7,7 +7,7 @@ import warnings
 with warnings.catch_warnings():
     warnings.simplefilter("ignore")
     from threatexchange.signal_type.pdq import signal as _
-## Resume regularly scheduled imports
+# Resume regularly scheduled imports
 
 import logging
 import os
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -139,7 +139,7 @@ def query_index(
     try:
         signal = signal_type.validate_signal_str(signal)
     except Exception as e:
-        abort(400, f"invalid signal type: {e}")
+        abort(400, f"invalid signal: {e}")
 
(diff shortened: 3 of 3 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -221,6 +223,26 @@ def test_banks_add_hash_index(app: Flask, client: FlaskClient):
     assert post_response.json == {"matches": [2]}
 
 
+def test_lookup_add_hash_without_role(app: Flask, client: FlaskClient):
```
Dcallies, line 226: nit: I don't think we're "adding" a hash in this unittest.  In other tests, this is referring to banking functionality.

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -221,6 +223,26 @@ def test_banks_add_hash_index(app: Flask, client: FlaskClient):
...
+    # role resets to True in the next test
+    client.application.config["ROLE_HASHER"] = False
+
+    # test GET
+    image_url = "https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg?raw=true"
+    get_resp = client.get(f"/m/lookup?url={image_url}")
```
Dcallies, line 232: blocking: I don't see any existing unittests in this file that check against this endpoint at all (the others are using "raw_lookup". Can you also test a lookup is accepted?

This file is getting big enough that it may make sense to break out unittests for the different APIs.

Dcallies (changes requested):
blocking: I think all that's missing is a test that shows the function returns a 200 when it's running normally - it looks uncovered at the moment.

aryzle:
@Dcallies should be GTG, thanks for the review!

Dcallies (approved):
I'm convinced! Thanks for testing improvements!

Issue #1679, opened 2024-11-04
[pytx][ncmec] NCMEC fetch implementation unable to make progress


Dcallies:
There exists some data in some NCMEC environments that have a lot of records all the same second. The current logic will keep fetching until it gets enough data to advance the checkpoint, of which the current smallest granularity is one second. Unfortunately, the amount of data that needs to be fetched in some cases is quite large, and frequently busts storage solutions (especially on HMA).

The fix is to store the "next" URL when the fetch granularity is one second. It's unclear what the behavior of the NCMEC database will be in this circumstance (if it's based on an offset, this may cause records to be skipped in some cases). 

If we want to be double defensive, we can invalidate the next URL if it was stored more than say ~1 day ago.

Dcallies:
I think we've confirmed this is fixed!

PR #1731, opened 2025-01-09
ncmec: store checkpoint on large fetches

prenner:
Summary
---------
sometimes ncmec fails to make progress after hitting a second w/ a large number of results: https://github.com/facebook/ThreatExchange/issues/1679. when that happens (diff of end and start is a second and we have lots of data), store checkpoints occasionally via a next pointer

Test Plan
---------

confirmed that resuming from a checkpoint works around the cursed second

### DIFF
--- a/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
@@ -565,7 +565,7 @@ class NCMECHashAPI:
             )
 
     def get_entries_iter(
-        self, *, start_timestamp: int = 0, end_timestamp: int = 0
+        self, *, start_timestamp: int = 0, end_timestamp: int = 0, next_: str = ""
     ) -> t.Iterator[GetEntriesResponse]:
         """
         A simple wrapper around get_entries to keep fetching until complete.
@@ -574,7 +574,6 @@ class NCMECHashAPI:
         much of the data you have fetched. @see get_entries
         """
         has_more = True
-        next_ = ""
         while has_more:
             result = self.get_entries(
                 start_timestamp=start_timestamp,
--- a/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
@@ -565,7 +565,11 @@ class NCMECHashAPI:
             )
 
     def get_entries_iter(
-        self, *, start_timestamp: int = 0, end_timestamp: int = 0, next_: str = ""
+        self,
+        *,
+        start_timestamp: int = 0,
+        end_timestamp: int = 0,
+        checkpointed_paging_url: str = "",
     ) -> t.Iterator[GetEntriesResponse]:
         """
         A simple wrapper around get_entries to keep fetching until complete.
@@ -574,6 +578,7 @@ class NCMECHashAPI:
         much of the data you have fetched. @see get_entries
         """
         has_more = True
+        next_ = checkpointed_paging_url
         while has_more:
             result = self.get_entries(
                 start_timestamp=start_timestamp,
--- a/python-threatexchange/threatexchange/exchanges/clients/ncmec/tests/data.py
+++ b/python-threatexchange/threatexchange/exchanges/clients/ncmec/tests/data.py
@@ -23,6 +23,14 @@ NEXT_UNESCAPED3 = (
     "&to=2017-10-30T00%3A00%3A00.000Z&start=4001&size=1000&max=5000"
 )
 
+ENTRIES_NO_DATA_XML = """
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<queryResult xmlns="https://hashsharing.ncmec.org/hashsharing/v2">
+    <images count="0"/>
+    <videos count="0"/>
+</queryResult>
+""".strip()
+
 ENTRIES_XML = """
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <queryResult xmlns="https://hashsharing.ncmec.org/hashsharing/v2">
--- a/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
@@ -54,7 +54,7 @@ class NCMECCheckpoint(
     def paging_url(self) -> str:
         PAGING_URL_EXPIRATION = 12 * 60 * 60
         if int(time.time()) - self.last_fetch_time < PAGING_URL_EXPIRATION:
-            return self.paging_url
+            return self._paging_url
         return ""
 
     @classmethod
@@ -298,6 +298,7 @@ class NCMECSignalExchangeAPI(
             duration = max(1, duration)  # Infinite loop defense
             # Don't fetch past our designated end
             current_end = min(end_time, current_start + duration)
+            entry = None
             for i, entry in enumerate(
                 client.get_entries_iter(
                     start_timestamp=current_start,
@@ -346,7 +347,10 @@ class NCMECSignalExchangeAPI(
             else:  # AKA a successful fetch
                 # If we're hovering near the single-fetch limit for a period
                 # of time, we can likely safely expand our range.
-                if len(entry.updates) < api.NCMECHashAPI.ENTRIES_PER_FETCH * 2:
+                if (
+                    entry
+                    and len(entry.updates) < api.NCMECHashAPI.ENTRIES_PER_FETCH * 2
+                ):
                     low_fetch_counter += 1
                     if low_fetch_counter >= self.FETCH_SHRINK_FACTOR:
                         log("multiple low fetches, increasing duration")
@@ -356,7 +360,9 @@ class NCMECSignalExchangeAPI(
                 # If we are not quite at our limit, but getting close to it,
                 # pre-emptively shrink to try and stay under the limit
                 elif (
-                    len(entry.updates) > self.MAX_FETCH_SIZE / self.FETCH_SHRINK_FACTOR
+                    entry
+                    and len(entry.updates)
+                    > self.MAX_FETCH_SIZE / self.FETCH_SHRINK_FACTOR
                 ):
                     log("close to overfetch limit, reducing duration")
                     duration //= self.FETCH_SHRINK_FACTOR
--- a/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
@@ -39,8 +39,8 @@ class NCMECCheckpoint(
 
     # The biggest value of "to", and the next "from"
     get_entries_max_ts: int
-    next_fetch: str
-    last_fetch_time: int
+    next_fetch: t.Optional[str] = ""
+    last_fetch_time: t.Optional[int] = 0
 
     def get_progress_timestamp(self) -> t.Optional[int]:
         return self.get_entries_max_ts
@@ -245,7 +245,7 @@ class NCMECSignalExchangeAPI(
         next_fetch = ""
         if checkpoint is not None:
             start_time = checkpoint.get_entries_max_ts
-            next_fetch = checkpoint.next_fetch
+            next_fetch = checkpoint.next_fetch or ""
         # Avoid being exactly at end time for updates showing up multiple
         # times in the fetch, since entries are not ordered by time
         end_time = int(time.time()) - 5
@@ -324,9 +324,8 @@ class NCMECSignalExchangeAPI(
                         ),
                     )
                     current_next_fetch = entry.next
-                    updates = []
-                else:
-                    updates.extend(entry.updates)
+                    break
+                updates.extend(entry.updates)
             else:  # AKA a successful fetch
                 # If we're hovering near the single-fetch limit for a period
                 # of time, we can likely safely expand our range.
--- a/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
@@ -282,7 +282,9 @@ class NCMECSignalExchangeAPI(
             updates: t.List[api.NCMECEntryUpdate] = []
             for i, entry in enumerate(
                 client.get_entries_iter(
-                    start_timestamp=current_start, end_timestamp=current_end, next_=current_next_fetch,
+                    start_timestamp=current_start,
+                    end_timestamp=current_end,
+                    next_=current_next_fetch,
                 )
             ):
                 if i == 0:  # First batch, check for overfetch
@@ -315,8 +317,11 @@ class NCMECSignalExchangeAPI(
                     log(f"large fetch ({i}), up to {len(updates)}. storing checkpoint")
                     yield state.FetchDelta(
                         {f"{entry.member_id}-{entry.id}": entry for entry in updates},
-                        NCMECCheckpoint(get_entries_max_ts=current_start, next_fetch=entry.next, last_fetch_time=int(time.time())),
-
+                        NCMECCheckpoint(
+                            get_entries_max_ts=current_start,
+                            next_fetch=entry.next,
+                            last_fetch_time=int(time.time()),
+                        ),
                     )
                     current_next_fetch = entry.next
                     updates = []
@@ -342,7 +347,11 @@ class NCMECSignalExchangeAPI(
                     low_fetch_counter = 0
                 yield state.FetchDelta(
                     {f"{entry.member_id}-{entry.id}": entry for entry in updates},
-                    NCMECCheckpoint(get_entries_max_ts=current_end, next_fetch="", last_fetch_time=int(time.time())),
+                    NCMECCheckpoint(
+                        get_entries_max_ts=current_end,
+                        next_fetch="",
+                        last_fetch_time=int(time.time()),
+                    ),
                 )
                 current_start = current_end
                 current_next_fetch = ""
--- a/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
+++ b/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
@@ -39,6 +39,8 @@ class NCMECCheckpoint(
 
     # The biggest value of "to", and the next "from"
     get_entries_max_ts: int
+    next_fetch: str
+    last_fetch_time: int
 
     def get_progress_timestamp(self) -> t.Optional[int]:
         return self.get_entries_max_ts
@@ -46,7 +48,7 @@ class NCMECCheckpoint(
     @classmethod
     def from_ncmec_fetch(cls, response: api.GetEntriesResponse) -> "NCMECCheckpoint":
         """Synthesizes a checkpoint from the API response"""
-        return cls(response.max_timestamp)
+        return cls(response.max_timestamp, response.next, int(time.time()))
 
     def __setstate__(self, d: t.Dict[str, t.Any]) -> None:
         """Implemented for pickle version compatibility."""
@@ -240,15 +242,18 @@ class NCMECSignalExchangeAPI(
                the cursor
         """
         start_time = 0
+        next_fetch = ""
         if checkpoint is not None:
             start_time = checkpoint.get_entries_max_ts
+            next_fetch = checkpoint.next_fetch
         # Avoid being exactly at end time for updates showing up multiple
         # times in the fetch, since entries are not ordered by time
         end_time = int(time.time()) - 5
 
         client = self.get_client(self.collab.environment)
-        # We could probably mutate start time, but new variable for clarity
+        # We could probably mutate start time and next, but new variable for clarity
         current_start = start_time
+        current_next_fetch = next_fetch
         # The range we are fetching
         duration = end_time - current_start
         # A counter for when we want to increase our duration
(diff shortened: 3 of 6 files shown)


python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -39,14 +39,16 @@ class NCMECCheckpoint(
 
     # The biggest value of "to", and the next "from"
     get_entries_max_ts: int
+    next_fetch: str
+    last_fetch_time: int
```
prenner, line 43: this is unsued currently but I know we talked about invalidating checkpoints older than 24hr.. happy to add that! just wanted to get confirmation on the checkpointing piece that this is on the right track

python-threatexchange/threatexchange/exchanges/tests/test_state_compatibility.py
```diff
@@ -147,8 +147,23 @@ def get_NCMECCheckpoint() -> t.Tuple[NCMECCheckpoint, t.Sequence[object]]:
...
+        """
+        0.99.x => 1.0.0
+
+        get_entries_max_ts: int =>
+            get_entries_max_ts: int
+            next_fetch: str
+            last_fetch_time: int
+        """
+
+        get_entries_max_ts: int
```
Dcallies, line 166: Thanks for adding this test case!

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -39,14 +39,16 @@ class NCMECCheckpoint(
 
     # The biggest value of "to", and the next "from"
     get_entries_max_ts: int
+    next_fetch: t.Optional[str] = ""
```
Dcallies, line 42: blocking: Hmm, not sold on this name. NCMEC calls these "[paging urls](https://report.cybertip.org/hashsharing/v2/documentation/?fbclid=IwZXh0bgNhZW0CMTEAAR3z8y_XUaLoQqTw-SlwsYrwMmMpz1xzXoOKgG4vy30xLEIDJxI22qJaiuM_aem_rN1itRmqgz3skMyJ-2q11Q#pagination)" so maybe paging_url?
blocking; Add a comment explaining what this variable represents.

nit: Since the default value is empty string, suggest making this non-`Optional` and just using that.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,12 +310,25 @@ def log(event: str) -> None:
                         # Our entry estimatation (based on the cursor parameters)
                         # occasionally seem to over-estimate
                         log(f"est {entry.estimated_entries_in_range} entries")
-                elif i % 100 == 0:
+
+                updates.extend(entry.updates)
+
+                if i % 100 == 0:
```
Dcallies, line 316: blocking: by change this from `elif` to `if`, I think it will now print the large update warning every update, which is incorrect, no?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -39,14 +39,16 @@ class NCMECCheckpoint(
...
+    last_fetch_time: t.Optional[int] = 0
 
     def get_progress_timestamp(self) -> t.Optional[int]:
         return self.get_entries_max_ts
 
     @classmethod
     def from_ncmec_fetch(cls, response: api.GetEntriesResponse) -> "NCMECCheckpoint":
         """Synthesizes a checkpoint from the API response"""
-        return cls(response.max_timestamp)
+        return cls(response.max_timestamp, response.next, int(time.time()))
```
Dcallies, line 51: blocking: NCMEC's documentation asks us not to store these checkpoints long term, so let's respect that in this implementation. 

One way we could handle this is not to directly access this variable, but wrap it with a helper function that will return empty string if it's too old.

```
def get_resume_url_if_recent(self):
   if time.time() - self.last_fetch_time < SOME_CONSTANT_MAYBE_12_HOURS:
     return self.next_fetch
   return ""
```

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -39,14 +39,16 @@ class NCMECCheckpoint(
 
     # The biggest value of "to", and the next "from"
     get_entries_max_ts: int
+    next_fetch: t.Optional[str] = ""
+    last_fetch_time: t.Optional[int] = 0
```
Dcallies, line 43: blocking: Add a comment on the relationship between the paging URL for why we save this.

ignorable: We could also make this variable explicitly mention the paging_url and not set it if we are not paging

python-threatexchange/threatexchange/exchanges/tests/test_state_compatibility.py
```diff
@@ -147,8 +147,23 @@ def get_NCMECCheckpoint() -> t.Tuple[NCMECCheckpoint, t.Sequence[object]]:
...
+    current = NCMECCheckpoint(
+        get_entries_max_ts=max_ts, next_fetch="", last_fetch_time=0
+    )
+
     # 1.0.x
-    current = NCMECCheckpoint(get_entries_max_ts=max_ts)
+    @dataclass
+    class NCMECCheckpointWithoutNext(FetchCheckpointBase):
+        """
+        0.99.x => 1.0.0
```
Dcallies, line 158: blocking: This version isn't quite correct, the [current version](https://fb.workplace.com/groups/366137801110062) is 1.2.3

python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py
```diff
@@ -565,7 +565,7 @@ def get_entries(
             )
 
     def get_entries_iter(
-        self, *, start_timestamp: int = 0, end_timestamp: int = 0
+        self, *, start_timestamp: int = 0, end_timestamp: int = 0, next_: str = ""
```
Dcallies, line 568: blocking: next_ is a bit cryptic, perhaps we can do `checkpointed_paging_url` and then do

```
next_ = checkpointed_paging_url
```

Dcallies (changes requested):
Overall looking good, thanks for making this change, and I think it will help a lot!

I am slightly suspicious that the paging URLs can go sour (e.g. I have noticed that NCMEC API tends to throw exceptions near the very end of the paging list that make me think that they are invaliding), so I think adding the time-based invalidation logic is a requirement. 

As part of your test plan, can you also attempt fetching past an extremely dense time segment in the NCMEC API and confirm the behavior works as expected?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,12 +310,25 @@ def log(event: str) -> None:
...
+                updates.extend(entry.updates)
+
+                if i % 100 == 0:
                     # If we get down to one second, we can potentially be
                     # fetching an arbitrary large amount of data in one go,
-                    # so log something occasionally
-                    log(f"large fetch ({i}), up to {len(updates)}")
-                updates.extend(entry.updates)
+                    # so store the checkpoint occasionally
+                    log(f"large fetch ({i}), up to {len(updates)}. storing checkpoint")
```
Dcallies, line 320: nit: You don't actually store the checkpoint by yielding, technically the caller can decide whether to keep calling or store.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -277,7 +282,9 @@ def log(event: str) -> None:
             updates: t.List[api.NCMECEntryUpdate] = []
             for i, entry in enumerate(
                 client.get_entries_iter(
-                    start_timestamp=current_start, end_timestamp=current_end
+                    start_timestamp=current_start,
+                    end_timestamp=current_end,
+                    next_=current_next_fetch,
```
Dcallies, line 287: blocking: Danger! It's actually very easy to mess up this argument and accidentally trigger and endless loop. It may be that you have done so in the current code, but it's hard to tell.

The only time `current_next_fetch` should be populated is when you are resuming from checkpoint, and you need to explicitly disable the overfetch check (L290) then.

There might be a refactoring of this code that makes this easier, or now that we are switching over to the next pointer version we can get rid of the probing behavior, which simplifies the implementation quite a bit.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -240,15 +242,18 @@ def fetch_iter(
                the cursor
         """
         start_time = 0
+        next_fetch = ""
```
Dcallies, line 245: nit: just call this variable current_next_fetch - it's unconditionally written at 256.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,12 +310,25 @@ def log(event: str) -> None:
...
+                    log(f"large fetch ({i}), up to {len(updates)}. storing checkpoint")
+                    yield state.FetchDelta(
+                        {f"{entry.member_id}-{entry.id}": entry for entry in updates},
+                        NCMECCheckpoint(
+                            get_entries_max_ts=current_start,
+                            next_fetch=entry.next,
+                            last_fetch_time=int(time.time()),
+                        ),
+                    )
+                    current_next_fetch = entry.next
```
Dcallies, line 329: nit: This value is never read, as it always unconditionally replaced on L359

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: ah so the original `elif` block doesn't need to change? the only real change that's needed is to use the next_url in the for loop on L283?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: Could we use `is_stale` for this?

python-threatexchange/threatexchange/exchanges/tests/test_state_compatibility.py
prenner: fixed!

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: added the func (and not is_stale)

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: it would print for the 0th, which we would not want. I updated this to be (i + 1) % 100 == 0, so it's every 100th iteration

we need to extend `updates` every time and regardless of `i` so this was cleaner than other things I thought of, but please suggest alternatives

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: Yeah as I mentioned in slack looks like we need the probing behavior so I wasn't able to simplify. I added a check to disable the overfetch when resuming from a checkpoint

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: updated the comment 👍

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -37,23 +37,45 @@ class NCMECCheckpoint(
...
         ### field 'max_timestamp' renamed to 'get_entries_max_ts'
         if "max_timestamp" in d:
             d["get_entries_max_ts"] = d.pop("max_timestamp")
+
+        # 1.0.0 => 1.2.3:
+        # Add last_fetch_time
+        # note: the default_factory value was not being set correctly when
+        # reading from pickle
+        if not "last_fetch_time" in d:
+            d["last_fetch_time"] = int(time.time())
```
prenner, line 77: I was getting `AttributeError: 'NCMECCheckpoint' object has no attribute 'last_fetch_time'` without this in the `test_state_compatibility` test

seems sort of related to https://github.com/pydantic/pydantic/issues/7821, since `default` was working (but wouldn't work if we want to set it to the current time)

Dcallies: Wow, tough bug, good find on the fix

Dcallies (changes requested):
Alright, I think I'm only down to questions and non-logical changes! 

This probing behavior is of course quite complicated, and I'm very tempted to try refactoring it to something similar based on the challenge of verifying this PR.

I think a lot of this may depend on the strength of your e2e testing. Were you able to complete a fetch and could tell that you end up with the same hashes at the end in both cases?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,12 +331,22 @@ def log(event: str) -> None:
...
+                    # On large fetches, yield a checkpoint to avoid re-fetching later
+                    log(f"large fetch ({i}), up to {len(updates)}. yielding checkpoint")
+                    yield state.FetchDelta(
+                        {f"{entry.member_id}-{entry.id}": entry for entry in updates},
+                        NCMECCheckpoint(
+                            get_entries_max_ts=current_start,
+                            paging_url=entry.next,
+                            last_fetch_time=int(time.time()),
+                        ),
+                    )
```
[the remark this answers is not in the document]
prenner: are there any downsides to doing so?

prenner:
> Were you able to complete a fetch and could tell that you end up with the same hashes at the end in both cases?

I tested two segments of times with the current implementation on main and another with checkpointing and came away with the same number of hashes. 

Running `threatexchange dataset --signal-summary` should show a distinct breakdown by signal type, right? Is there another approach you would take?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,12 +331,22 @@ def log(event: str) -> None:
...
-                    # If we get down to one second, we can potentially be
-                    # fetching an arbitrary large amount of data in one go,
-                    # so log something occasionally
-                    log(f"large fetch ({i}), up to {len(updates)}")
+
                 updates.extend(entry.updates)
+
+                if (i + 1) % 100 == 0:
+                    # On large fetches, yield a checkpoint to avoid re-fetching later
+                    log(f"large fetch ({i}), up to {len(updates)}. yielding checkpoint")
```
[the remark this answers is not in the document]
prenner: modified but kept the log b/c of what you said [here](https://github.com/facebook/ThreatExchange/pull/1731#discussion_r1929874033)

Dcallies:
> Running threatexchange dataset --signal-summary should show a distinct breakdown by signal type, right? Is there another approach you would take?

We could be even more pedantic by doing some kind of checksum of the actual updates, but I think the counts being the same is good enough

prenner:
👍  looked for hashes between `1605558564` and `1624636980` and had `1332309`

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,12 +331,23 @@ def log(event: str) -> None:
...
+                        f"large fetch ({i}) with {len(updates)} updates. yielding checkpoint"
+                    )
+                    yield state.FetchDelta(
+                        {f"{entry.member_id}-{entry.id}": entry for entry in updates},
+                        NCMECCheckpoint(
+                            get_entries_max_ts=current_start,
+                            _paging_url=entry.next,
+                        ),
+                    )
+                    updates = []
```
Dcallies, line 349: Hmm, What is your intent here? 

It will buffer updates without emitting them until 100 fetches, and then after it will emit every update? 
```suggestion

                if i % 100 == 5:
                    # On large fetches, log notice every once in a while
                    log(
                        f"large fetch ({i}) with {len(updates)} updates"
                    )
                yield state.FetchDelta(
                    {f"{entry.member_id}-{entry.id}": entry for entry in updates},
                    NCMECCheckpoint(
                        get_entries_max_ts=current_start,
                        _paging_url=entry.next,
                    ),
                )
                updates = []
```

Dcallies (approved):
Sorry, had pending review comments! 

If your tests are showing good, then we should proceed. I have some confusion about the emit behavior that I think can be simplified, but I'm happy to make a pass at it it.

Let me know if you prefer to merge as is.

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: how do we know when we have a "large update"? before that was i > 100, but if we're fine considering the 5th iteration to be large, then this makes sense! 

There doesn't seem to be a downside to yielding, so it probably doesn't matter, right?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -39,21 +39,43 @@ class NCMECCheckpoint(
 
     # The biggest value of "to", and the next "from"
     get_entries_max_ts: int
+    # A url to fetch the next page of results
+    # Only reference this value through `paging_url` property
+    _paging_url: str = ""
+    # a timestamp for the last fetch time, specifically used with a paging_url
+    # NCMEC suggests not storing paging_urls long term so we consider them invalid
+    # 12hr after the last_fetch_time
+    last_fetch_time: int = field(hash=True, default_factory=lambda: int(time.time()))
```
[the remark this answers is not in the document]
prenner: yeah I don't think it's needed - thanks for catching that

I don't think the checkpoints are hashable anyway

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
prenner: in the proposed code, there's no need to

`updates.extend(entry.updates)`, right? since we're always resetting it back to an empty array?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -327,10 +366,12 @@ def log(event: str) -> None:
                     low_fetch_counter = 0
                 else:  # Not too small, not too large, just right
                     low_fetch_counter = 0
+
                 yield state.FetchDelta(
                     {f"{entry.member_id}-{entry.id}": entry for entry in updates},
```
prenner, line 371: change to empty array

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -303,16 +330,23 @@ def log(event: str) -> None:
...
+                        get_entries_max_ts=current_start,
+                        _paging_url=entry.next,
+                    ),
+                )
+
             else:  # AKA a successful fetch
                 # If we're hovering near the single-fetch limit for a period
                 # of time, we can likely safely expand our range.
-                if len(updates) < api.NCMECHashAPI.ENTRIES_PER_FETCH * 2:
+                if len(entry.updates) < api.NCMECHashAPI.ENTRIES_PER_FETCH * 2:
```
Dcallies, line 349: blocking: entry will be unset if you if the for loop doesn't yield. I'm surprised typing doesn't catch this. 

Here's a demonstrator:
```
$ python
>>> items = []
>>> for i, val in items:
...   print(i, val)
... else:
...   print(val)
... 
  File "<stdin>", line 4, in <module>
NameError: name 'val' is not defined.
```

To fix, on L300 add
```
entry = None # Alternatively, set it to a empty entry object
```

and then all these checks become 
```
if (entry is not None and len(entry.updates) ...
```

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -321,16 +355,20 @@ def log(event: str) -> None:
                         low_fetch_counter = 0
                 # If we are not quite at our limit, but getting close to it,
                 # pre-emptively shrink to try and stay under the limit
-                elif len(updates) > self.MAX_FETCH_SIZE / self.FETCH_SHRINK_FACTOR:
+                elif (
+                    len(entry.updates) > self.MAX_FETCH_SIZE / self.FETCH_SHRINK_FACTOR
```
Dcallies, line 359: .

Dcallies (changes requested):
sooo close, but new bug introduced (consider adding a test case for it).

python-threatexchange/threatexchange/exchanges/clients/ncmec/tests/data.py
```diff
@@ -23,6 +23,14 @@
     "&to=2017-10-30T00%3A00%3A00.000Z&start=4001&size=1000&max=5000"
 )
 
+ENTRIES_NO_DATA_XML = """
```
Dcallies, line 26: Thanks for adding a test!

Dcallies (approved):
Don't see any more bugs, let's shipit and do a version bump

python-threatexchange/threatexchange/exchanges/clients/ncmec/tests/data.py
prenner: I don't think it's quite what you asked for as we should always get a `GetEntriesResponse` response back or raise an error (meaning it should always enter the for loop)

prenner:
> python

can't merge - can you do so? thanks again 🙏

Dcallies:
Forgot to press the button 😅

Dcallies:
We might have missed something, see #1744

PR #1732, opened 2025-01-11
[HMA] Bank and BankContent disable

### LINKED ISSUE #1682: [hma] Implement content-level disable
https://github.com/haianhng31/ThreatExchange/blob/8d10599cda78afa49f763c08ca556db94f4ae332/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py#L235C1-L258C1

Today we allow disabling entire banks (including fractional coinflip), we need to allow disabling individual hashes.

aryzle:
Summary
---------

stage 1 of #1682 

- [x] Confirm that an API exists (under curator role) that allows setting Bank and BankContent disable states (implement if not)
  - [x] Bank enable_ratio
  - [x] BankContent disable_until_ts

Test Plan
---------

- [x] unittest

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
@@ -72,7 +72,7 @@ def build_index(
     index_cls = for_signal_type.get_index_cls()
     signal_list = []
     last_cs = None
-    for last_cs in bank_store.bank_yield_content(for_signal_type, enabled_only=True):
+    for last_cs in bank_store.bank_yield_content(for_signal_type):
         tuple = (last_cs.signal_val, last_cs.bank_content_id)
         signal_list.append(tuple)
     built_index = index_cls.build(signal_list)
--- a/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
@@ -72,7 +72,7 @@ def build_index(
     index_cls = for_signal_type.get_index_cls()
     signal_list = []
     last_cs = None
-    for last_cs in bank_store.bank_yield_content(for_signal_type):
+    for last_cs in bank_store.bank_yield_content(for_signal_type, enabled_only=True):
         tuple = (last_cs.signal_val, last_cs.bank_content_id)
         signal_list.append(tuple)
     built_index = index_cls.build(signal_list)
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -57,8 +57,8 @@ def bank_create():
     return jsonify(bank_create_impl(name, enabled_ratio)), 201
 
 
-def bank_create_impl(name: str, enabled_ratio: float = 1.0) -> iface.IBank:
-    bank = iface.IBank(name=name, matching_enabled_ratio=enabled_ratio)
+def bank_create_impl(name: str, enabled_ratio: float = 1.0) -> iface.BankConfig:
+    bank = iface.BankConfig(name=name, matching_enabled_ratio=enabled_ratio)
     try:
         persistence.get_storage().bank_update(bank, create=True)
     except ValueError as e:
@@ -133,7 +133,7 @@ def bank_add_content(bank_name: str):
     query parameter), or uploading a file (via multipart/form-data).
 
     @see OpenMediaMatch.blueprints.hashing hash_media()
-    @see OpenMediaMatch.blueprints.hashing hash_file()
+    @see OpenMediaMatch.blueprints.hashing hash_media_from_form_data()
 
     Inputs:
      * The content to be banked, in one of these formats:
@@ -173,14 +173,14 @@ def bank_add_content(bank_name: str):
         hashes = hashing.hash_media()
     # File uploaded via multipart/form-data?
     elif request.files:
-        hashes = hashing.hash_file()
+        hashes = hashing.hash_media_from_form_data()
     else:
         abort(400, "Neither `url` nor multipart file upload was received")
     return _bank_add_signals(bank, hashes, metadata)
 
 
 def _bank_add_signals(
-    bank: iface.IBank,
+    bank: iface.BankConfig,
     signal_type_to_signal_str: dict[str, str],
     metadata: t.Optional[BankedContentMetadata],
 ) -> dict[str, t.Any]:
@@ -200,7 +200,7 @@ def _bank_add_signals(
         except Exception as e:
             abort(400, f"Invalid {name} signal: {str(e)}")
 
-    content_config = iface.IBankContent(
+    content_config = iface.BankContentConfig(
         id=0, disable_until_ts=0, collab_metadata={}, original_media_uri=None, bank=bank
     )
 
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -57,8 +57,8 @@ def bank_create():
     return jsonify(bank_create_impl(name, enabled_ratio)), 201
 
 
-def bank_create_impl(name: str, enabled_ratio: float = 1.0) -> iface.BankConfig:
-    bank = iface.BankConfig(name=name, matching_enabled_ratio=enabled_ratio)
+def bank_create_impl(name: str, enabled_ratio: float = 1.0) -> iface.IBank:
+    bank = iface.IBank(name=name, matching_enabled_ratio=enabled_ratio)
     try:
         persistence.get_storage().bank_update(bank, create=True)
     except ValueError as e:
@@ -127,13 +127,13 @@ def _validate_bank_add_metadata() -> t.Optional[BankedContentMetadata]:
 
 
 @bp.route("/bank/<bank_name>/content", methods=["POST"])
-def bank_add_file(bank_name: str):
+def bank_add_content(bank_name: str):
     """
     Add content to a bank by providing a URI to the content (via the `url`
     query parameter), or uploading a file (via multipart/form-data).
 
     @see OpenMediaMatch.blueprints.hashing hash_media()
-    @see OpenMediaMatch.blueprints.hashing hash_media_post()
+    @see OpenMediaMatch.blueprints.hashing hash_file()
 
     Inputs:
      * The content to be banked, in one of these formats:
@@ -173,14 +173,14 @@ def bank_add_file(bank_name: str):
         hashes = hashing.hash_media()
     # File uploaded via multipart/form-data?
     elif request.files:
-        hashes = hashing.hash_media_post_impl()
+        hashes = hashing.hash_file()
     else:
         abort(400, "Neither `url` nor multipart file upload was received")
     return _bank_add_signals(bank, hashes, metadata)
 
 
 def _bank_add_signals(
-    bank: iface.BankConfig,
+    bank: iface.IBank,
     signal_type_to_signal_str: dict[str, str],
     metadata: t.Optional[BankedContentMetadata],
 ) -> dict[str, t.Any]:
@@ -200,7 +200,7 @@ def _bank_add_signals(
         except Exception as e:
             abort(400, f"Invalid {name} signal: {str(e)}")
 
-    content_config = iface.BankContentConfig(
+    content_config = iface.IBankContent(
         id=0, disable_until_ts=0, collab_metadata={}, original_media_uri=None, bank=bank
     )
 
@@ -210,6 +210,32 @@ def _bank_add_signals(
         "id": content_id,
         "signals": {st.get_name(): val for st, val in signals.items()},
     }
+
+
+@bp.route("/bank/<bank_name>/content/<content_id>", methods=["PUT"])
+def bank_update_content(bank_name: str, content_id: int):
+    """
+    Update the metadata for a banked content item.
+    """
+    storage = persistence.get_storage()
+    bank = storage.get_bank(bank_name)
+    if not bank:
+        abort(404, f"bank '{bank_name}' not found")
+    content = storage.bank_content_get([content_id])
+    if not content:
+        abort(404, f"content '{content_id}' not found")
+    content = content[0]
+    data = request.get_json()
+
+    try:
+        if "disable_until_ts" in data:
+            content.disable_until_ts = flask_utils.str_to_type(
+                data["disable_until_ts"], int
+            )
+        storage.bank_content_update(content)
+    except Exception as e:
+        abort(400, *e.args)
+    return jsonify(content)
 
 
 @bp.route("/bank/<bank_name>/content/<content_id>", methods=["DELETE"])
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -18,7 +18,9 @@ from OpenMediaMatch.utils import flask_utils
 import OpenMediaMatch.storage.interface as iface
 from OpenMediaMatch.blueprints import hashing
 
-MAX_TIME = int(time.mktime((datetime.now() + timedelta(days=365 * 100)).timetuple()))
+
+def five_years_from_now() -> int:
+    return int(time.mktime((datetime.now() + timedelta(days=365 * 5)).timetuple()))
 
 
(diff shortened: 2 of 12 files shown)


aryzle:
@Dcallies this is still a draft but if you get a chance to look at it, would appreciate to know if I'm headed in the right direction 😁🙏

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +212,32 @@ def _bank_add_signals(
...
+    data = request.get_json()
+
+    try:
+        if "disable_until_ts" in data:
+            content.disable_until_ts = flask_utils.str_to_type(
+                data["disable_until_ts"], int
+            )
+        storage.bank_content_update(content)
+    except Exception as e:
+        abort(400, *e.args)
```
Dcallies, line 237: blocking: Some of these exceptions would be the servers fault (500). I don't think we should capture every exception here to try and distinguish. Can you only capture the exceptions you know to be the client's fault?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +212,32 @@ def _bank_add_signals(
...
+    if not bank:
+        abort(404, f"bank '{bank_name}' not found")
+    content = storage.bank_content_get([content_id])
+    if not content:
+        abort(404, f"content '{content_id}' not found")
+    content = content[0]
+    data = request.get_json()
+
+    try:
+        if "disable_until_ts" in data:
```
Dcallies, line 231: We should check to make sure that this is in a sane range.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +212,32 @@ def _bank_add_signals(
     }
 
 
+@bp.route("/bank/<bank_name>/content/<content_id>", methods=["PUT"])
+def bank_update_content(bank_name: str, content_id: int):
+    """
+    Update the metadata for a banked content item.
```
Dcallies, line 218: blocking: Can you document the parameters accepted?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +212,32 @@ def _bank_add_signals(
...
+
+    try:
+        if "disable_until_ts" in data:
+            content.disable_until_ts = flask_utils.str_to_type(
+                data["disable_until_ts"], int
+            )
+        storage.bank_content_update(content)
+    except Exception as e:
+        abort(400, *e.args)
+    return jsonify(content)
```
Dcallies, line 238: not sure/ignorable: I think I wrote something to automatically json-ify the returns, but maybe not

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -72,11 +72,20 @@ def hash_media_post():
...
     Calculate the hash for the provided file.
     """
-    return hash_media_post_impl()
 
+    return hash_file()
 
-def hash_media_post_impl() -> dict[str, str]:
-    """ """
+
+def hash_file() -> dict[str, str]:
```
Dcallies, line 79: blocking: The old name actually was meaningful in that it was referring to the fact that it is one variation of the hash_file implementation - one where the media content is in the form data. There's another version which downloads from a URL. This new name is probably misleading because someone might confuse it with how `tx hash <file>` works.

Maybe something like `hash_media_from_form_data` might be more accurate? (I believe form data can only be included in a POST)

Dcallies (changes requested):
Hey @aryzle, thanks for contributing this PR! This one is in high demand, and I let another user who was asking for it know that there's a PR up.

blocking: Please split this PR into its component parts. It helps during review for PRs to have "one thesis", and the renames adds a lot of additional lines to the review, where the risk profile for renames are a lot smaller than the functional changes, and it can be hard to spot which changes are for which!

Additionally, I would prefer we hold off on the proposed renames. There is an in-progress migration to move these classes from HMA into `python-threatexchange`, and it will be difficult to complete any rename until that is completed. Additionally additionally, in programming languages which have a specific `interface` concept, prefixing with `I` can have some benefits for identifying them. However, in python, the waters are a bit muddy, since you can happily combine implementation in otherwise "abstract" classes. 

HMA itself does mix implementation in some of these "interfaces" - so an author coming from a different language may project expectations onto these classes that they are not expecting as a result of the rename. Thus, I'm not sure it makes sense to rename these classes. As long as you are willing to hold your nose on the current names, I think it's better to leave as is. If you still want to follow through with the rename, let's create a dedicated issue and discuss the tradeoffs of doing so before we follow through - it should not be a part of this functional change.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
aryzle: should we allow updating anything besides `disable_until_ts`? I could see `collab_metadata` or `original_media_uri` as candidates but not sure what they do

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
aryzle: seems like it's needed on our DB entities, like `jsonify(bank)`

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -16,6 +18,8 @@
 import OpenMediaMatch.storage.interface as iface
 from OpenMediaMatch.blueprints import hashing
 
+MAX_TIME = int(time.mktime((datetime.now() + timedelta(days=365 * 100)).timetuple()))
```
aryzle, line 21: I don't love this but wanted to have a max time for `disable_until_ts`

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -201,7 +209,11 @@ def _bank_add_signals(
             abort(400, f"Invalid {name} signal: {str(e)}")
 
     content_config = iface.BankContentConfig(
-        id=0, disable_until_ts=0, collab_metadata={}, original_media_uri=None, bank=bank
```
aryzle, line 204: @Dcallies was this meant to add new content with `disable_until_ts=0` ?

actually it looks like `disable_until_ts` and `collab_metadata` don't get used in `storage.bank_add_content`

Dcallies: > actually it looks like disable_until_ts and collab_metadata don't get used in storage.bank_add_content

Might be needed to complete the issue then!

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -89,6 +93,10 @@ def bank_update(bank_name: str):
         storage.bank_update(bank, rename_from=rename_from)
     except ValueError as e:
         abort(400, *e.args)
+    except IntegrityError:
+        abort(403, "Bank already exists")
```
Dcallies, line 97: blocking: This is an update function, the bank is expected to exist (otherwise it will 404 on L82). In what circumstances are IntegrityError thrown?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -89,6 +93,10 @@ def bank_update(bank_name: str):
...
+    except Exception as e:
+        abort(500, *e.args)
```
Dcallies, line 99: blocking: What happens if you don't catch the exception at all? If you call this abort I think it may "hide" the root exception (by replacement). 

I suspect the default behavior is also to 500, so you may not need this explicit abort.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +224,44 @@ def _bank_add_signals(
...
+            if disable_until_ts > MAX_TIME:
+                raise ValueError(
+                    "disable_until_ts must be less than 100 years in the future"
+                )
+            content.disable_until_ts = disable_until_ts
+        storage.bank_content_update(content)
+    except ValueError as e:
+        abort(400, *e.args)
+    except KeyError as e:
+        abort(404, *e.args)
```
Dcallies, line 259: You are also potentially catching exceptions from `storage.bank_content_update(content)` which may need to be a 500. 

Instead of throwing, catching, and aborting, just `abort(400`) inline.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -458,13 +458,18 @@ def get_current_index_build_target(
 
     @abc.abstractmethod
     def bank_yield_content(
-        self, signal_type: t.Optional[t.Type[SignalType]] = None, batch_size: int = 100
+        self,
+        signal_type: t.Optional[t.Type[SignalType]] = None,
+        batch_size: int = 100,
+        enabled_only: bool = True,
```
Dcallies, line 464: blocking: Why this change?

Dcallies (changes requested):
Hey @aryzle ! Sorry for the slow review, I've been traveling in the last week.

We're getting closer! This looks like a decent scope for the size of this PR, but if you'd like, the only functional changes that are needed is the updates to the API that allows marking a content as disabled.

blocking: I think you are close enough to the right version that you should add some new unittests for you new APIs to make sure they get the right values! This will be easier if you add the GET option for bank content, since you can then just test the full create-update-delete cycle.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
aryzle: I set this to 100 years actually, thinking we def won't hit that 😅 but I can set to lets say 5 years (?) dynamically at runtime

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
aryzle: I believe same as in create, where the passed in `name` already exists (name is set to `unique=True` in db)

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
```diff
@@ -618,12 +625,23 @@ def bank_yield_content(
...
+        if enabled_only:
+            query = query.join(database.BankContent).where(
+                (
+                    database.BankContent.disable_until_ts
+                    == interface.BankContentConfig.ENABLED
+                )
+                or (
+                    database.BankContent.disable_until_ts
+                    < func.extract("epoch", func.now())
+                )
```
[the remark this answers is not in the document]
aryzle: ah yeah I see, check should be `disable_until_ts != 0` and `disable_until_ts < now` (which handles 1s)

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
```diff
@@ -618,12 +625,23 @@ def bank_yield_content(
...
+            )
```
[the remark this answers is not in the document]
aryzle: I guess idk the typical size of these datasets, but by default I tend to do any querying in the DB since it's optimized for that (even without indexes has caching, query optimizer)

aryzle:
@Dcallies Thanks for the reviews! and apologies I've have very little time (1-2 hours) that this thing is dragging...

I can hold on any of the `enabled_only` and skipping logic until the next stages of this issue if it easier for you to review!

For this PR, I'll address your remaining comments + unittest the new API

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -72,7 +72,7 @@ def build_index(
     index_cls = for_signal_type.get_index_cls()
     signal_list = []
     last_cs = None
-    for last_cs in bank_store.bank_yield_content(for_signal_type):
+    for last_cs in bank_store.bank_yield_content(for_signal_type, enabled_only=True):
```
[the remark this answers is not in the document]
aryzle: yeah good point, maybe we could do a bit of both? filter out the permanently disabled and anything more than a few months out? (how often are indexes rebuilt? could help determine the timeframe here)

Also can just filter in python after filtering the permanently disabled like you said, I don't have a sense for how big a typical dataset is yet and if it helps to do it in the DB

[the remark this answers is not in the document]
aryzle: btw any of my comments around this are for the next PR, just planning ahead

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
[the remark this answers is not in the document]
Dcallies: 100% that the backend will probably be better at this. Not sure the tradeoff of that vs it's easier to scale the frontend machines than the backend layer. By sticking to id lookup queries it's easier to show that just inserting a simple cache between will work. 

The other motivator for doing it on server time is we may support multiple backends, and the more requirements we put on the backend, the taller the requirements to make a new one.

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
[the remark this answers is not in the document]
Dcallies: That might be a mistake on our end - we should have probably used an arbitrarily large value for disable instead so that a range query would work. I think we talked about this during the initial implementation so not sure if it was an oversight or a decision...

Dcallies:
I was going to apologize for you to slow reviews as well! I really appreciate any time you can donate @aryzle - every little bit counts and there are real users on the other of your changes that will be helped by your work. I'll give you heads up if something becomes urgent. There's a planning meeting about 1/month where we cover deadlines.

aryzle:
@Dcallies this should be GTG now!

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api_banks.py
```diff
@@ -120,7 +122,42 @@ def test_banks_delete(client: FlaskClient):
     assert post_response.status_code == 200
 
 
-def test_banks_add_hash(client: FlaskClient):
+def test_bank_get_content(client: FlaskClient):
+    bank_name = "TEST_BANK_GET"
+    create_bank(client, bank_name)
+
+    # Add content
+    image_url = "https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg?raw=true"
```
Dcallies, line 130: I'm surprised this doesn't make CI angry

Dcallies (approved):
I'll follow up with some inline tweaks, but this is good enough to get us started. Nice work with the testing!

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -89,6 +95,8 @@ def bank_update(bank_name: str):
         storage.bank_update(bank, rename_from=rename_from)
     except ValueError as e:
         abort(400, *e.args)
+    except IntegrityError:
+        abort(403, "Bank name already exists")
```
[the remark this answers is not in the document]
aryzle: I think I meant to do 400, it's when the passed in `name` already exists - the DB is enforcing unique name on the bank table

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -212,6 +236,38 @@ def _bank_add_signals(
...
+        if "disable_until_ts" in data:
+            disable_until_ts = flask_utils.str_to_type(data["disable_until_ts"], int)
+            if disable_until_ts < 0:
+                abort(400, "disable_until_ts must be a non-negative integer")
+            if disable_until_ts > five_years_from_now():
+                abort(400, "disable_until_ts must be less than 5 years in the future")
+            content.disable_until_ts = disable_until_ts
+        storage.bank_content_update(content)
+    except KeyError as e:
+        abort(404, *e.args)
```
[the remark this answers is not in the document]
aryzle: `bank_content_update` raises KeyError if `sesh.execute(
            select(database.BankContent).where(database.BankContent.id == val.id)
        ).scalar_one_or_none()` returns None

PR #1733, opened 2025-01-23
[HMA] Implement exchange delete API

aokolish:
Summary
---------

This PR implements the exchange delete API which previously would just return a 501 when called.

Test Plan
---------

Unit tests + I manually tested this in my local HMA dev env & UI.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -441,10 +441,6 @@ def exchange_delete(exchange_name: str):
     }
     """
     storage = persistence.get_storage()
-    collab = storage.exchange_get(exchange_name)
-    if collab is None:
-        abort(404, f"exchange '{exchange_name}' not found")
-
     storage.exchange_delete(exchange_name)
     return {"message": "Exchange deleted"}
 
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -431,20 +431,22 @@ def exchange_update(exchange_name: str):
 def exchange_delete(exchange_name: str):
     """
     Delete exchange configuration
+
     Inputs:
-     * Configuration name
-     * (optional) whether to also delete the associated bank (defaults to true)
+     * Exchange name
 
     Returns:
     {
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -431,20 +431,22 @@ def exchange_update(exchange_name: str):
...
-      "message": "success",
+      "message": "Exchange deleted",
     }
     """
     storage = persistence.get_storage()
     collab = storage.exchange_get(exchange_name)
     if collab is None:
-        return {"message": "success"}
-    abort(501, "Not yet implemented")
+        abort(404, f"exchange '{exchange_name}' not found")
```
Dcallies, line 446: blocking: I think we debated this when we originally implemented it, and thought it was less confusing to say that deleting something that doesn't exist is success, so you can naively retry errors.

All the other deletions (e.g. bank delete) follow this pattern as well, so I'm hesitant to add exceptions unless I can understand more about the case.

Did you run into a condition that led you to conclude to change it, or is this just a taste change?

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -163,3 +163,26 @@ def test_compare_hashes(app: Flask, client: FlaskClient):
             json=bad_input,
         )
         assert resp.status_code == 400
+
+
+def test_exchange_delete(app: Flask, client: FlaskClient):
```
Dcallies, line 168: thanks for adding a test!

Dcallies (changes requested):
Let me know if you run into problems internally signing the CLA, I can take over the PR and merge it if you cannot.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
aokolish: Ah, that's ok with me.

I was following what I thought was the established pattern. I see it in `bank_delete_content` for example. However, the rest of the APIs that 404 on things not being found are for updates.

Dcallies:
@aokolish - thanks a ton for contributing to HMA! Let me know if the CLA is a hard blocker - I can merge when it's complete, or else tell me to take over.

Dcallies:
@aokolish - looks like you just need to run the `black` formatter as well

Issue #1734, opened 2025-01-24
[hma][good first issue] Tweak Message For Checkpointing Sanity Check Failure


Dcallies:
https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py#L344

If two fetch processes are running at the same time, one of them will hopefully fail this assert to try and avoid rewinding the checkpoint. 

The message currently cryptically refers to this as a "race condition", but this actually means there are probably multiple instances of the fetch process.

Let's update the message to be less cryptic, and refer to multiple fetch processes or tasks.

anaypurohit0907:
hi i would like to work on this issue, please guide me.

PR #1737, opened 2025-01-30
(doc update): Change Tweak Message For Checkpointing Sanity Check Failure

vincent-4:
See: #1734 https://github.com/facebook/ThreatExchange/issues/1734

Summary
---------
Previously:
`Old checkpoint doesn't match, race condition?`
New message:
`Checkpoint has changed since fetch started - multiple fetch processes may be running simultaneously.`
I think this is a pretty nice message after reviewing the surrounding context.

Test Plan
---------

Only replacing error message, so no updates needed.


vincent-4:
> I agree that no need for any specific testing. Thanks for donating some time!

Thanks for opening this issue. Also, I see there's a backlog of unclosed issues, so I’d be happy to help, especially with ones that aren't blocked by internal (Meta) context. If you mention / add issues where it is convenient for you to do so, I'd be happy to contribute.
Thanks again!

Issue #1681, opened 2024-11-06
[hma] Match Lookup API that provides more metadata


Dcallies:
# Today's Lookup APIs
https://github.com/haianhng31/ThreatExchange/blob/8d10599cda78afa49f763c08ca556db94f4ae332/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py

# Ideal Usecase - Reddit Request
@aokolish representing request:

Input:
1. Required
    1.  Signal type
    1 .hash
2. Optional
    1. Resolve 3P record json

Output:
List of match objects:
```
   Bank: [
       bank_content_id: {
          distance
          // Example
          3P_exchange_type:  exchange_type: NCMEC
          3P_Records: {3p_id: NCMEC_ID        
              exchange_data: {  // Specific to exchange type
                  member_entries: {
                     member_id: [tags]
              }
          } 
       }
   ]
```

Notes:
* Do not return disabled hashes or banks

Dcallies:
This issue was mentioned as a being a priority to an adopting platform in the 7/16/2025 HMA development meeting.

julietshen:
@sreeprasad is working on this via https://github.com/facebook/ThreatExchange/pull/1931. Part of a related issue https://github.com/facebook/ThreatExchange/issues/1925

PR #1738, opened 2025-01-30
[HMA] Change matching API to return more metadata

aokolish:
Summary
---------

This PR changes the API responses for the GET and POST `/m/lookup` APIs based on issue #1681. The motivation being - users of the API need to know the content IDs so that they can turn content IDs off if they're causing false positives. Also, callers may find it useful to log / analyze matching distance as well.

**API response before:** (API just returns list on bank names with matches)
```
["FOO_BANK"]
```

**API response w/this change:** (API returns a dict of bank names with matches + details about each match)
```
{
    "FOO_BANK": [
        {
            "content_id": 16,
            "distance": "13"
        }
    ]
}
```

Note: I modified tests that were calling `/raw_lookup` because there were no tests for the `/lookup` API. I also did that in part because I think we could remove `raw_lookup` once this is merged since the main benefit of that API to me was that it returned the content IDs.

@Dcallies I'm opening this as a draft because I couldn't figure out how to get `3P_exchange_type` and `3P_Records` from the issue. If it's not too hard to add, we can do it now. Otherwise, this is good enough to be valuable to me as-is.

Test Plan
---------

Unit tests + testing in my local HMA dev instance

Here's an example from my dev instance:
```shell
$ curl -s "http://localhost:5000/m/lookup?signal_type=pdq&signal=cccccccc33133333eccecccc33313333cccccccc33313333cce6cccc33933333" | jq
{
  "FOO_BANK": [
    {
      "content_id": 81842,
      "distance": 20
    },
    {
      "content_id": 82244,
      "distance": 10
    },
    {
      "content_id": 82519,
      "distance": 18
    },
    {
      "content_id": 82850,
      "distance": 18
    },
    {
      "content_id": 82870,
      "distance": 18
    },
    {
      "content_id": 82883,
      "distance": 24
    },
    {
      "content_id": 91618,
      "distance": 14
    }
  ]
}
```

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -280,7 +280,7 @@ def lookup(signal, signal_type_name):
         match = results_by_bank_content_id.get(content.id)
         results[content.bank.name].append(
             {
-                "content_id": content.id,
+                "bank_content_id": content.id,
                 "distance": match.similarity_info.distance,
             }
         )
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -278,7 +278,7 @@ def lookup(signal, signal_type_name):
             # is it possible to return the content_signal.signal_val here?
             {
                 "content_id": content.id,
-                "distance": match.similarity_info.pretty_str(),
+                "distance": match.similarity_info.distance,
             }
         )
     return results
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -271,7 +271,7 @@ def lookup(signal, signal_type_name):
     results = defaultdict(list)
     for content in enabled_content:
         if content.bank.name not in enabled_banks:
-            next
+            continue
 
         match = results_by_content_id.get(content.id)
         results[content.bank.name].append(
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -256,7 +256,7 @@ def lookup(signal, signal_type_name):
         r.metadata: r for r in query_index(signal, signal_type_name)
     }
     storage = get_storage()
-    contents = storage.bank_content_get(results_by_content_id.keys())
+    contents = storage.bank_content_get(results_by_content_id)
     enabled_content = [c for c in contents if c.enabled]
     current_app.logger.debug(
         "lookup matches %d content ids (%d enabled_content)",
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -258,8 +258,6 @@ def lookup(signal, signal_type_name):
         "lookup matches %d content ids (%d enabled_content)", len(contents),
         len(enabled_content)
     )
-    if not enabled_content:
-        return []
     banks = {c.bank.name: c.bank for c in enabled_content}
     rand = random.Random(request.args.get("seed"))
     coinflip = rand.random()
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -4,6 +4,7 @@
 Endpoints for matching content and hashes.
 """
 
+from collections import defaultdict
 from dataclasses import dataclass
 import datetime
 import random
@@ -247,11 +248,11 @@ def lookup_post():
 
 def lookup(signal, signal_type_name):
     current_app.logger.debug("performing lookup")
-    raw_results = lookup_signal(signal, signal_type_name)
+    results_by_content_id = {
+        r.metadata: r for r in query_index(signal, signal_type_name)
+    }
     storage = get_storage()
-    current_app.logger.debug("getting bank content")
-    current_app.logger.debug(raw_results)
-    contents = storage.bank_content_get(raw_results)
+    contents = storage.bank_content_get(results_by_content_id.keys())
     enabled = [c for c in contents if c.enabled]
     current_app.logger.debug(
         "lookup matches %d content ids (%d enabled)", len(contents), len(enabled)
@@ -261,13 +262,25 @@ def lookup(signal, signal_type_name):
     banks = {c.bank.name: c.bank for c in enabled}
     rand = random.Random(request.args.get("seed"))
     coinflip = rand.random()
-    enabled_banks = [
+    enabled_banks = {
         b.name for b in banks.values() if b.matching_enabled_ratio >= coinflip
-    ]
+    }
     current_app.logger.debug(
         "lookup matches %d banks (%d enabled)", len(banks), len(enabled_banks)
     )
-    return enabled_banks
+    results = defaultdict(list)
+    for content in enabled:
+        if content.bank.name not in enabled_banks:
+            next
+
+        match = results_by_content_id.get(content.id)
+        results[content.bank.name].append(
+            {
+                "content_id": content.id,
+                "distance": match.similarity_info.pretty_str(),
+            }
+        )
+    return results
 
 
 @bp.route("/index/status")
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +248,40 @@ def lookup_post():
...
     )
-    return enabled_banks
+    results = defaultdict(list)
+    for content in enabled_content:
+        if content.bank.name not in enabled_banks:
+            next
+
+        match = results_by_content_id.get(content.id)
+        results[content.bank.name].append(
+            # is it possible to return the content_signal.signal_val here?
```
aokolish, line 278: @Dcallies I also wanted to return the hash from the bank but couldn't figure that one out. I think each `bank_content` has many `content_signal`s, right? Maybe it's hard to get it here.

Dcallies: It is possible, but it will take more round trips to the DB. I think adding a parameter to request extra information that we can't get "for free", but are useful for debugging makes sense. 

It looks like we didn't expose ContentSignal - it just needs a simple Crud that takes  (bank_content_id), and we'll probably want a multi-lookup so you can use it this context.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +248,40 @@ def lookup_post():
...
+    }
     current_app.logger.debug(
-        "lookup matches %d banks (%d enabled)", len(banks), len(enabled_banks)
+        "lookup matches %d banks (%d enabled_content)", len(banks), len(enabled_banks)
     )
-    return enabled_banks
+    results = defaultdict(list)
+    for content in enabled_content:
+        if content.bank.name not in enabled_banks:
+            next
```
Dcallies, line 274: blocking: I don't think `next` is a keyword in python is it? `continue`?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +248,40 @@ def lookup_post():
 
 def lookup(signal, signal_type_name):
     current_app.logger.debug("performing lookup")
-    raw_results = lookup_signal(signal, signal_type_name)
+    results_by_content_id = {
+        r.metadata: r for r in query_index(signal, signal_type_name)
```
Dcallies, line 252: nit: might want to add `results_by_bank_content_id` to differentiate from a future where we might have on-platform content ids as well

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +248,40 @@ def lookup_post():
...
+    }
     storage = get_storage()
-    current_app.logger.debug("getting bank content")
-    current_app.logger.debug(raw_results)
-    contents = storage.bank_content_get(raw_results)
-    enabled = [c for c in contents if c.enabled]
+    contents = storage.bank_content_get(results_by_content_id.keys())
```
Dcallies, line 255: ignorable: I think this function takes a traversable, and the traverse for dicts is the keys

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +248,40 @@ def lookup_post():
...
     rand = random.Random(request.args.get("seed"))
     coinflip = rand.random()
-    enabled_banks = [
+    enabled_banks = {
         b.name for b in banks.values() if b.matching_enabled_ratio >= coinflip
-    ]
+    }
     current_app.logger.debug(
-        "lookup matches %d banks (%d enabled)", len(banks), len(enabled_banks)
+        "lookup matches %d banks (%d enabled_content)", len(banks), len(enabled_banks)
```
Dcallies, line 269: blocking: this is enabled banks, not enabled content

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_e2e_workflow.py
```diff
@@ -37,37 +37,13 @@ def test_raw_hash_add_to_match_no_distance(app: Flask, client: FlaskClient):
     assert build_info["size"] == len(hashes)
 
     # Now match
-    resp = client.get(f"/m/raw_lookup?signal_type=pdq&signal={hashes[-1]}")
+    resp = client.get(f"/m/lookup?signal_type=pdq&signal={hashes[-1]}")
     assert resp.status_code == 200
-    assert resp.json == {"matches": [16]}
+    assert resp.json == {"TEST_BANK": [{"content_id": 16, "distance": 0}]}
```
Dcallies, line 42: <3

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +248,40 @@ def lookup_post():
 
 def lookup(signal, signal_type_name):
```
Dcallies, line 249: blocking: Make sure to update the documentation on the lookup functions

Dcallies (changes requested):
Overall change looks good, and we can get the content id and distance "for free" (no more additional database calls), so I think it makes sense to return by default.

For additional data (3P Id, hash string, etc), maybe we can add those as optional parameters, since they take extra calls to fetch.

We may need to add a new storage interface to allow getting the signal strings. Seems like an oversight in the Bank interface.

Let me know if you want to tag team this - I can help on the lower parts of the stack

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
aokolish: Yeah, I changed my mind and am thinking we don't need this. It's more of a nice to have and most of the time we won't need it. I can get it via other techniques, if / when I need it.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
aokolish: Did you mean to rename the variable? I didn't totally understand your comment, but that's what I did.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
aokolish: changed

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
aokolish: fixed

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
aokolish: whoops, I'm more fluent in ruby...

aokolish:
@Dcallies I think I addressed all your feedback. This version is adding value for me (enables me to log data regarding the matches we're finding) so I'd like to merge it without any additional data when you're ok with it. Meaning, I don't need the hash values or 3P data...just the content IDs and distance are the main things I'll be needing for now.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
Dcallies: close enough

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -247,27 +252,39 @@ def lookup_post():
...
-    return enabled_banks
+    results = defaultdict(list)
+    for content in enabled_content:
+        if content.bank.name not in enabled_banks:
+            continue
+
+        match = results_by_bank_content_id.get(content.id)
+        results[content.bank.name].append(
+            {
+                "content_id": content.id,
```
Dcallies, line 283: blocking: I missed this version of it, but I forsee the need for a "content id" down the line, and so we should make this clear it's the `bank_content_id`

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_e2e_workflow.py
```diff
@@ -37,37 +37,13 @@ def test_raw_hash_add_to_match_no_distance(app: Flask, client: FlaskClient):
...
-    assert "pdq" in all_build_info  # type: ignore
-    build_info = all_build_info["pdq"]  # type: ignore
-    assert build_info["present"] == True
-    assert build_info["size"] == len(hashes)
-
-    # Now match
+    # Test lookup with no match
     resp = client.get(
-        f"/m/raw_lookup?signal_type=pdq&include_distance=true&signal={hashes[-1]}"
+        "/m/lookup?signal_type=pdq&signal=5555555555555555555555555555555555555555555555555555555555555555"
```
Dcallies, line 46: You can save some space if you want by doing "5" * 64

```
"/m/lookup?signal_type=pdq&signal={}".format("5" * 64)
```

Dcallies (approved):
Looks good, one last rename. Let me know what you want to do about the CLA.

This will be a backwards incompatible change for current users of the API, but I think we should break this for the structured version.

Dcallies:
I'll run the formatter on your code afterwards

PR #1742, opened 2025-01-31
Pytest: `unittest` => `pytest` Conversion

### LINKED ISSUE #1686: [pytx][mlh] Convert some tests in threatexchange/signal_type/ to py.test
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/tests/test_hash_from_x.py
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/tests/test_md5_hash.py
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index.py
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/tests/test_raw_text.py
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/tests/test_url_md5_hash.py

These tests are using unittest, we want to use py.test.

py.test docs: https://docs.pytest.org/en/stable/

b8zhong:
Summary
---------

Previous tests were in unittest, and should now be inline with pytest (resolves #1686)

Previous tests in `signal_type` were converted to pytest

Formatted with `black hasher-matcher-actioner python-threatexchange`.

Test Plan
---------

Added tests.

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
@@ -1,7 +1,7 @@
 import typing as t
 import random
 import io
-import faiss  # type: ignore
+import faiss
 
 from threatexchange.signal_type.pdq.pdq_index2 import PDQIndex2
 from threatexchange.signal_type.pdq.signal import PdqSignal
--- a/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
@@ -1,7 +1,7 @@
 import typing as t
 import random
 import io
-import faiss
+import faiss  # type: ignore
 
 from threatexchange.signal_type.pdq.pdq_index2 import PDQIndex2
 from threatexchange.signal_type.pdq.signal import PdqSignal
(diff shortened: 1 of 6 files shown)


Dcallies (approved):
Thanks a ton @b8zhong for the help! Your conversions look clean, and I appreciate the extra time you took to simplify or improve the tests!

Dcallies (changes requested):
Whoops, went too fast, there are some CI failures on some of the files you touched.

For the typing errors, use your best judgement on whether it's worth the time to fix typing in tests or whether to liberally sprinkle `#type: ignore` . I have soundtimes found typing catches unittest bugs, but more often than not it seems like too much effort.

b8zhong:
@Dcallies Thanks for reviewing! Verified the test files now, passes with `mypy tests --check-untyped-defs`. 

I thought typechecking would trigger whenever py files are touched though lol

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
```diff
@@ -1,7 +1,7 @@
 import typing as t
 import random
 import io
-import faiss
+import faiss  # type: ignore
```
Dcallies, line 4: Hmm, there should be a top-level ignore for this somewhere.

Looks like this turned into an error

```
threatexchange/signal_type/tests/test_pdq_index2.py:4: error: Unused "type: ignore" comment  [unused-ignore]
```

b8zhong:
Thanks for rerunning; put faiss as typeignore but I guess it doesn't like that whoops

Dcallies (changes requested):
It can be tricky to reproduce the exact results as CI - one trick I've done is check the version being installed by the CI, and I think we hardcode the version of mypy to help avoid random failures,

Settings say (from CI output at https://github.com/facebook/ThreatExchange/actions/runs/13074520123/job/36487583072?pr=1742):
```
platform linux -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0

mypy-1.7.1
```

Nothing jumps out to me from the CI failures - you might have some luck breaking the PR into one file at a time, but I'll leave that up to you.

python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
b8zhong: Yeah, gonna remove that

b8zhong:
@Dcallies Could you rerun now? Tried w/ `pytest -W ignore::DeprecationWarning threatexchange/signal_type/tests -vv`, all pass; 

Passes (at least locally), so I think we should be good; caught a few things on my side. The test I marked as skip is because compare_hash isn't supported.. or I could just remove it. But let me know what you think

Dcallies (approved):
Looks like everything is passing now!

b8zhong:
Thanks for merging David!

PR #1743, opened 2025-01-31
[hma] bump version of python-threatexchange to 1.2.4 and version bump…

prenner:
… to 1.0.9

Summary
---------

use the new version of python-threatexchange: https://github.com/facebook/ThreatExchange/pull/1741

Test Plan
---------


hasher-matcher-actioner/version.txt
```diff
@@ -1 +1 @@
-1.0.8
```
prenner, line 1: let me know if it's preferred to split out the version bump

Dcallies: I think it's okay for this case

Dcallies (approved):

PR #1745, opened 2025-02-02
[HMA] Removing IContentTypeConfigStore

b8zhong:
Summary
---------

Removes `IContentTypeConfigStore`, as requested in the last part of [#1688.](https://github.com/facebook/ThreatExchange/issues/1688)

Note: I plan to add copyright boilerplate to the currently empty `init` + bump minor version in a separate PR like you requested here! https://github.com/facebook/ThreatExchange/pull/1707

Test Plan
---------

N/A


b8zhong:
CI fails are expected here I believe..

Dcallies (approved):
Hmm, CI fails are not expected here, and will predict HMA library failures. We need a bumped version which contains the init fix.

This is the same problem that the MLH folks were running into, which you already found the solution to - the missing `__init__.py` file. 

To fix, you'll need to break this up a sequence of PRs:

1. Add `__init__.py` to storage
2. Version bump
3. Now can remove.

That sounds like a pain, so I'll do it now while I'm in the area.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -23,6 +23,7 @@
 import flask
 from threatexchange.cli.storage.interfaces import ISignalTypeConfigStore
```
Dcallies, line 24: Sigh, this should have been threatexchange.storage.interfaces

b8zhong:
Hey! I'm around now; do you still want to take over?

(Actually I just saw your edit, good catch I did not realize that lol)

b8zhong:
(Closing as addressed in #1747 & #1748

Dcallies:
@b8zhong  - didn't quite address! This PR is still needed, I just fixed the underlying stuff for you!

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -23,6 +23,7 @@
...
 from threatexchange.utils import dataclass_json
+from threatexchange.storage.interfaces import IContentTypeConfigStore
```
Dcallies, line 26: Instead of doing a second import on this line, add to L24

b8zhong: Done!

b8zhong:
There we go, mypy is finally happy

Dcallies:
Nice!

Issue #1744, opened 2025-01-31
[hma/py_tx] fetching ncmec hashes ignores all past hashes


prenner:
steps to reproduce
1. set up a new exchange for ncmec (I was only using the TiD hashes)
2. let HMA sync

eventually, I got to a state where it was considered up to date and it was only syncing new hashes moving forward:
```
{"checkpoint_ts":1738363974,"fetched_items":21,"last_fetch_complete_ts":1738363980,"last_fetch_succeeded":true,"running_fetch_start_ts":1738365071,"up_to_date":true}
```

and past hashes were not synced (total_fetched is never > 0 until after it's considered up to date)

Dcallies:
Well, now you know why I was so scared of that block! 1738363974 is essentially now (2025-01-31T17:52:54-0500).

I bet the paging URL is only valid if you exactly reproduce the original query. E.g. it's a naive offset, and if we messed with the end time (current_end), it becomes invalid and speeds you to the end of the list. The probing behavior is constantly changing the end time, and the paging may only work with the current end time. 

We can delete the version in pip to immediately stop distribution of pytx 1.2.4.

My guess is you can get a reproduction any time it switches into your paging behavior.

Dcallies:
Yeah, I'm pretty scared of this bit:

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/clients/ncmec/hash_api.py#L550-L556

We are changing the "to" parameter and trying to reuse the "next" pointer. Experiment with calling it directly and see if it causes it to skip.

prenner:
yeah an extreme example is

https://hashsharing.ncmec.org/npo/v2/entries?from=2021-03-01T17:32:44Z&to=2021-06-28T20:32:45Z&start=537352&size=10&max=980101

https://hashsharing.ncmec.org/npo/v2/entries?from=2021-03-01T17:32:44Z&to=2024-04-28T20:32:45Z&start=537352&size=10&max=980101

which shows hashes being dropped like you said

but i would still expect some of them to be committed

Dcallies:
Okay, we have a repo then, the fix is that we need to add a second field to save the current end for the checkpoint alongside the next pointer. 

TID wasn't launched until the 2023, so there are no old records.

prenner:
> Okay, we have a repo then, the fix is that we need to add a second field to save the current end for the checkpoint alongside the next pointer.

[...]

PR #1750, opened 2025-02-03
[WIP][pytx] Attempt to fix NCMEC paging behavior

Dcallies:
Summary
---------

See #1744

I made a stab at trying for this without redoing the probing behavior.

Test Plan
---------

Pending, ran out of time this morning


prenner (commented):
was able to pick up and resume around the cursed second 👍

Dcallies:
Well, let's land then, I guess it's probably an improvement

PR #1755, opened 2025-02-06
Update benchmarks with PDQ Faiss results

b8zhong:
Summary
---------

Update the benchmarks, plus `benchmark_pdq_faiss_matchers.py` didn't have any results in the markdown file, so I added that as well.

Test Plan
---------

Yes

### DIFF
--- a/python-threatexchange/benchmarks/README.MD
+++ b/python-threatexchange/benchmarks/README.MD
@@ -54,28 +54,28 @@ query: 26.0294s
 Results (PDQ Faiss):
 -------
 ```
-% python3 benchmarks/benchmark_pdq_faiss_matchers.py --dataset-size 10000 --num-queries 1000 --thresholds 255
+% python3 benchmarks/benchmark_pdq_faiss_matchers.py --dataset-size 10000 --num-queries 1000 --thresholds 31
 Benchmark: PDQ Faiss Matcher Comparison
 
 Options:
 	 faiss_threads :  1
 	 dataset_size :  10000
 	 num_queries :  1000
-	 thresholds :  [255]
+	 thresholds :  [31]
 	 seed :  None
 
-using random seed of  1738860222531631000
-use --seed  1738860222531631000  to rerun with same random values
+using random seed of  1739236966565067000
+use --seed  1739236966565067000  to rerun with same random values
 
 Building Stats:
-	PDQFlatHashIndex: time to build (s):  0.01552891731262207
+	PDQFlatHashIndex: time to build (s):  0.015399932861328125
 	PDQFlatHashIndex: approximate size: 390KB
-	PDQMultiHashIndex: time to build (s):  0.03090691566467285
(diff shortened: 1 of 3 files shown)


python-threatexchange/benchmarks/README.MD
```diff
@@ -1,45 +1,81 @@
...
+-------
+```
+% python3 benchmarks/benchmark_pdq_faiss_matchers.py --dataset-size 10000 --num-queries 1000 --thresholds 255
+Benchmark: PDQ Faiss Matcher Comparison
+
+Options:
+	 faiss_threads :  1
+	 dataset_size :  10000
+	 num_queries :  1000
+	 thresholds :  [255]
```
Dcallies, line 64: This is a very unrealistic threshold! We should probably only be testing at 31 - this is why your multihash results are so poor, since you are forcing it search every subtree instead of the ones nearby.

python-threatexchange/benchmarks/README.MD
```diff
@@ -1,45 +1,81 @@
-# pytx-vpdq
-Benchmark vPDQ implementation in threatexchange library
+# pytx-vPDQ
```
Dcallies, line 1: blocking q: I am confused by why this is here instead of python-threatexchange/vpdq/README.md

Any ideas?

python-threatexchange/benchmarks/benchmark_vpdq_index.py
```diff
@@ -71,8 +71,6 @@ def run_benchmark(
         build = lambda: hashes
     elif test_type == IndexType.FLAT:
         build = lambda: build_flat(hashes)
-    else:
-        raise ValueError("Invalid test type")
```
Dcallies, line 75: This seems useful to keep in, since the error will otherwise be on L76 with variable undefined. This is else is also to defend against future developers adding a new test_type and forgetting to update it here.

Dcallies (changes requested):
Threshold of 255 is very unrealistic, and is voiding your benchmark! We could be even faster by just `return not index.is_empty()`, since 255 is the maximum possible distance.

python-threatexchange/benchmarks/benchmark_vpdq_index.py
b8zhong: Kk; I thought argparse would stop it, so that's why I did it

Reverted it back tho

python-threatexchange/benchmarks/README.MD
b8zhong: Lol yep; totally right there. I thought a high threshold enforces strict matching. Changed to 31

python-threatexchange/benchmarks/README.MD
b8zhong: Not sure either... maybe not to clutter the docs? I can move it -- up to you

https://github.com/facebook/ThreatExchange/issues/1122

Don't think there was a specific reasoning originally here either tho

PR #1756, opened 2025-02-06
[py-tx] Implementation of IVF faiss indices in PDQ

b8zhong:
Summary
---------

PDQ now using IVF Faiss for better scalability (as we move away from the old imp)

- Adds PDQSignalTypeIndex2 that automatically switches between flat and IVF indices based on dataset size
- Uses IVF-Faiss for datasets >= 1000 entries (as you said), flat index for less than
- Maintains backward compatibility with existing PDQIndex
- Updates signal.py to use the new index implementation as default (if we're not swapping yet, I'll change it back)

https://github.com/facebook/ThreatExchange/issues/1613

Test Plan
---------

Run the tests....

`python3 -m pytest threatexchange/signal_type/tests/test_pdq_signal_type_index2.py -v -W ignore::DeprecationWarning`

Passes.. ?

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
@@ -8,7 +8,6 @@ import typing as t
 import faiss
 import numpy as np
 
-
 from threatexchange.signal_type.index import (
     IndexMatchUntyped,
     SignalSimilarityInfoWithIntDistance,
@@ -32,6 +31,29 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
     designed to be simpler and fix hard-to-squash bugs in the existing implementation.
     Purpose of this class: to replace the original index in pytx 2.0
     """
+
+    IVF_THRESHOLD = 1000
+
+    @classmethod
+    def build(
+        cls: t.Type["PDQIndex2"], entries: t.Iterable[t.Tuple[str, IndexT]]
+    ) -> "PDQIndex2":
+        """
+        Build an index from a set of entries.
+        Selects between flat and IVF index based on number of entries.
+        """
+        entries_list = list(entries)
+
+        index = faiss.IndexFlatL2(BITS_IN_PDQ)
+        if len(entries_list) >= cls.IVF_THRESHOLD:
+            nlist = len(entries_list) // 2
+            index = faiss.IndexIVFFlat(index, BITS_IN_PDQ, nlist)
+            vectors = convert_pdq_strings_to_ndarray([h for h, _ in entries_list])
+            index.train(vectors)
+        else:
+            index = faiss.IndexFlatL2(BITS_IN_PDQ)
+
+        return cls(index=index, entries=entries_list)
 
     def __init__(
         self,
--- a/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
+++ b/python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py
@@ -178,3 +178,31 @@ def test_one_entry_sample_index():
 
     results = index.query(unmatching_test_hash)
     assert len(results) == 0
+
+
+def test_flat_index_selection():
+    """Test flat index selection for small datasets"""
+    get_random_hashes = _get_hash_generator()
+    base_hashes = get_random_hashes(100)
+    index = PDQIndex2.build([(h, i) for i, h in enumerate(base_hashes)])
+    assert isinstance(index._index.faiss_index, faiss.IndexFlatL2)
+
+
+def test_ivf_index_selection():
+    """Test IVF index selection for large datasets"""
+    get_random_hashes = _get_hash_generator()
+    base_hashes = get_random_hashes(2000)
+    index = PDQIndex2.build([(h, i) for i, h in enumerate(base_hashes)])
+    assert isinstance(index._index.faiss_index, faiss.IndexIVFFlat)
+
+
+def test_build_method():
+    """Test build method creates valid index"""
+    get_random_hashes = _get_hash_generator()
+    base_hashes = get_random_hashes(500)
+    index = PDQIndex2.build([(h, i) for i, h in enumerate(base_hashes)])
+
+    query_hash = base_hashes[0]
+    results = index.query(query_hash)
+    assert len(results) >= 1
+    assert any(r.metadata == 0 for r in results)


python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -5,9 +5,10 @@
 """
 
 import typing as t
+from typing import Self
```
Dcallies, line 8: blocking: I love Self! However, it was only added in 3.11 and we still are planning to support older version of python (back to 3.8), so no self allowed :/ 

https://docs.python.org/3/library/typing.html#typing.Self

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -133,3 +134,78 @@ def __getstate__(self):
 
     def __setstate__(self, data):
         self.faiss_index = faiss.deserialize_index(data)
+
+
+class PDQSignalTypeIndex2(SignalTypeIndex[IndexT]):
```
Dcallies, line 139: blocking: Instead of having this as an entirely new class, instead dynamically swap out which index class you are using in PDQIndex2 based on `build()`.

I believe it should be possible to use one wrapping class, but wrapping a few different indices. Look at L48 in this file.

python-threatexchange/threatexchange/signal_type/pdq/signal.py
```diff
@@ -53,8 +55,8 @@ def get_content_types(cls) -> t.List[t.Type[ContentType]]:
         return [PhotoContent]
 
     @classmethod
-    def get_index_cls(cls) -> t.Type[PDQIndex]:
-        return PDQIndex
+    def get_index_cls(cls) -> t.Type[SignalTypeIndex]:
+        return PDQSignalTypeIndex2
```
Dcallies, line 59: DANGER DANGER DANGER!

We have to very carefully make this change as there are live deployments of this code with stored state that need to load correctly still! (Including every deployment of HMA).

I strongly recommend holding off on this change while we prove the new implementation works perfectly, and then do some dedicated testing to show how we will handle old version capability. At worst we may need to lock this behind a major version change (pytx 2.x.x), but I think we have some options before then.

python-threatexchange/threatexchange/signal_type/tests/test_pdq_signal_type_index2.py
```diff
@@ -0,0 +1,184 @@
+import typing as t
```
Dcallies, line 1: What do you think about these tests?

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -133,3 +134,78 @@ def __getstate__(self):
...
+            quantizer = faiss.IndexFlatL2(BITS_IN_PDQ)
+            index = faiss.IndexIVFFlat(quantizer, BITS_IN_PDQ, nlist)
+
+            vectors = convert_pdq_strings_to_ndarray([h for h, _ in entries_list])
+            index.train(vectors)
+        else:
+
+            index = faiss.IndexFlatL2(BITS_IN_PDQ)
+
+        ret._index = PDQIndex2(index=index, entries=entries_list)
```
Dcallies, line 169: Wait except you are wrapping PDQIndex2 with this class? I am confused!

Dcallies (changes requested):
Hey @b8zhong! Thanks again for contributing, I am very excited about this work, and so glad it's got you on it.

I suspect this change is much simpler than you might think! You have added a new wrapping class to the heirarchy, but I don't think we need it.

Is it possible to make the following changes:
1. Migrate the index selection logic from `PDQSignalTypeIndex2.build` to `PDQIndex2.build`.
2. Take a look at the existing tests for pdqindex2 and see if you can re-use them by just changing the index you pass in

b8zhong:
Didn't forget ab this.. I'll check it out tn; thanks for the review!

python-threatexchange/threatexchange/signal_type/tests/test_pdq_signal_type_index2.py
b8zhong: Did you mean `python-threatexchange/threatexchange/signal_type/tests/test_pdq_index2.py`?

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
b8zhong: Removed rip

python-threatexchange/threatexchange/signal_type/pdq/signal.py
b8zhong: Reverted to original

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
b8zhong: I think I confused myself with this one

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
b8zhong: Yeah; tried to do that now

b8zhong:
Think it's ready to get a second look now?

Dcallies (approved):
Apologies, I had unsubmitted comments!

This looks like it won't break any of the existing code, what I'd like to see is some results from benchmarking to prove that we've gotten any kind of speedup. 

Check out https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/benchmarks/benchmark_pdq_faiss_matchers.py

and 

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/benchmarks/README.md

Let me know if you'd rather take a look at these in a followup instead.

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -33,6 +32,29 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
...
+        cls: t.Type["PDQIndex2"], entries: t.Iterable[t.Tuple[str, IndexT]]
+    ) -> "PDQIndex2":
+        """
+        Build an index from a set of entries.
+        Selects between flat and IVF index based on number of entries.
+        """
+        entries_list = list(entries)
+
+        if len(entries_list) >= cls.IVF_THRESHOLD:
+            nlist = int(len(entries_list) ** 0.5)
```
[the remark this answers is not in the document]
b8zhong: Because setting nlist = number of entries divided by 2 creates clusters averaging 2 items each... ? then the index will have half as many clusters as data points. What do you think? Open to feedback

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
```diff
@@ -33,6 +32,29 @@ class PDQIndex2(SignalTypeIndex[IndexT]):
...
+            quantizer = faiss.IndexFlatL2(BITS_IN_PDQ)
+            index = faiss.IndexIVFFlat(quantizer, BITS_IN_PDQ, nlist)
+            vectors = convert_pdq_strings_to_ndarray([h for h, _ in entries_list])
+            index.train(vectors)
+        else:
+            index = faiss.IndexFlatL2(BITS_IN_PDQ)
```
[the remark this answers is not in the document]
b8zhong: Agree. Done

b8zhong:
> Let me know if you'd rather take a look at these in a followup instead.

I think I would prefer that :)

b8zhong:
@Dcallies Actually I just decided to grind it out, but it doesn't look good so far:

This isn't *broken* so I'd say it's safe to still merge it in, but we definitely need to look at optimization... ?
```
python3 benchmarks/benchmark_pdq_indices.py --dataset-size 10000 --num-queries 1000 --thresholds 31
Benchmark: PDQ Index Implementation Comparison

Options:
	 faiss_threads :  1
	 dataset_size :  10000
	 num_queries :  1000
	 thresholds :  [31]
	 seed :  None
	 implementations :  all
	 serialize_test :  False

using random seed of  1740619317581017000
use --seed  1740619317581017000  to rerun with same random values

Generating dataset...
Generated 10000 hashes in 0.0382s


=== Benchmarking PDQFlatHashIndex ===
Running benchmark with threshold 31...

=== Benchmarking PDQMultiHashIndex ===
Running benchmark with threshold 31...

=== Benchmarking PDQIndex2 (Flat) ===
Running benchmark with threshold 31...

=== Benchmarking PDQIndex2 (IVF) ===
Training IVF index with 10000 vectors of dimension 256
WARNING clustering 10000 points to 1000 centroids: please provide at least 39000 training points
Set nprobe to 100 for searching
Running benchmark with threshold 31...


========== BENCHMARK SUMMARY ==========

===== PDQFlatHashIndex Results =====
Build time: 0.0144s
Size: 390KB

Threshold 31:
	Search time: 0.0116s
	Per query: 0.0116ms
	Targets found: 100.00%
Single item test: SUCCESS

===== PDQMultiHashIndex Results =====
Build time: 0.0274s
Size: 1,208KB

Threshold 31:
	Search time: 0.0164s
	Per query: 0.0164ms
	Targets found: 100.00%
Single item test: SUCCESS

===== PDQIndex2 (Flat) Results =====
Build time: 0.0271s
Size: 10,791KB

Threshold 31:
	Search time: 0.2123s
	Per query: 0.2123ms
	Targets found: 100.00%
Single item test: SUCCESS

===== PDQIndex2 (IVF) Results =====
Build time: 0.3408s
Size: 11,877KB

Threshold 31:
	Search time: 0.0807s
	Per query: 0.0807ms
	Targets found: 100.00%
Single item test: SUCCESS

========== PERFORMANCE COMPARISON ==========
Implementation       | Build Time (s)  | Size (KB)       | Search Time (ms) | Targets Found (%)
----------------------------------------------------------------------------------------
PDQFlatHashIndex     | 0.0144          | 390             | 0.0116          | 100.00         
PDQMultiHashIndex    | 0.0274          | 1,208           | 0.0164          | 100.00         
PDQIndex2 (Flat)     | 0.0271          | 10,791          | 0.2123          | 100.00         
PDQIndex2 (IVF)      | 0.3408          | 11,877          | 0.0807          | 100.00  
```

b8zhong:
Ignore the earlier comment: I think I was doing something dumb.


```
python3 benchmarks/benchmark_pdq_indices.py --dataset-size 10000 --num-queries 1000 --implementations index2_flat flat_hash multi_hash --thresholds 31 --query-mode both
Benchmark: PDQ Index Implementation Comparison

Options:
	 faiss_threads :  1
	 dataset_size :  10000
	 num_queries :  1000
	 thresholds :  [31]
	 seed :  None
	 implementations :  ['index2_flat', 'flat_hash', 'multi_hash']
	 serialize_test :  False
	 query_mode :  both

using random seed of  1740628723234617000
use --seed  1740628723234617000  to rerun with same random values

Generating dataset...
Generated 10000 hashes in 0.0388s


=== Benchmarking PDQFlatHashIndex ===
Running benchmark with threshold 31...
  Running batch query benchmark...
  Running individual query benchmark...

=== Benchmarking PDQMultiHashIndex ===
Running benchmark with threshold 31...
  Running batch query benchmark...
  Running individual query benchmark...

=== Benchmarking PDQIndex2 (Flat) ===
Running benchmark with threshold 31...
  Running batch query benchmark...
  Running individual query benchmark...

========== MEMORY USAGE ESTIMATE ==========
Note: These are serialized sizes, not runtime memory usage
PDQFlatHashIndex: 390KB
PDQMultiHashIndex: 1,208KB
PDQIndex2 (Flat): 10,791KB


========== BENCHMARK SUMMARY ==========

===== PDQFlatHashIndex Results =====
Build time: 0.0156s
Size: 390KB

Threshold 31:
	Batch mode:
		Search time: 0.0120s
		Per query: 0.0120ms
		Targets found: 100.00%
	Individual mode:
		Search time: 0.0200s
		Per query: 0.0200ms
		Targets found: 100.00%
Single item test: SUCCESS

===== PDQMultiHashIndex Results =====
Build time: 0.0281s
Size: 1,208KB

Threshold 31:
	Batch mode:
		Search time: 0.0166s
		Per query: 0.0166ms
		Targets found: 100.00%
	Individual mode:
		Search time: 0.0286s
		Per query: 0.0286ms
		Targets found: 100.00%
Single item test: SUCCESS

===== PDQIndex2 (Flat) Results =====
Build time: 0.0275s
Size: 10,791KB

Threshold 31:
	Batch mode:
		Search time: 0.2088s
		Per query: 0.2088ms
		Targets found: 100.00%
	Individual mode:
		Search time: 0.2089s
		Per query: 0.2089ms
		Targets found: 100.00%
Single item test: SUCCESS

========== PERFORMANCE COMPARISON ==========

--- BATCH QUERY MODE ---
Implementation       | Build Time (s)  | Size (KB)       | Search Time (ms) | Targets Found (%)
----------------------------------------------------------------------------------------
PDQFlatHashIndex     | 0.0156          | 390             | 0.0120          | 100.00         
PDQMultiHashIndex    | 0.0281          | 1,208           | 0.0166          | 100.00         
PDQIndex2 (Flat)     | 0.0275          | 10,791          | 0.2088          | 100.00         

--- INDIVIDUAL QUERY MODE ---
Implementation       | Build Time (s)  | Size (KB)       | Search Time (ms) | Targets Found (%)
----------------------------------------------------------------------------------------
PDQFlatHashIndex     | 0.0156          | 390             | 0.0200          | 100.00         
PDQMultiHashIndex    | 0.0281          | 1,208           | 0.0286          | 100.00         
PDQIndex2 (Flat)     | 0.0275          | 10,791          | 0.2089          | 100.00
```

python-threatexchange/threatexchange/signal_type/pdq/pdq_index2.py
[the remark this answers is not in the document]
Dcallies: The last time I messed around with it, it seemed like there was a warning if there wasn't at least ~40 points per cluster.

There are a bunch of parameters to FAISS that can be tuned, we should pick based on maintaining a very high recall (as we would prefer to find everything indexed), while providing as much speedup as we can.

Dcallies (changes requested):
Let's try and find out why we are getting a performance regression! Ideally this approach should be faster, but maybe we picked the wrong tech and we should be looking more into `IndexBinaryIVF` and `IndexBinaryMultiHash`. I think we should skip `IndexBinaryHash` because the first few bits of a PDQ hash correspond to the upper left hand content of an image, which should have no predictive power towards the rest of the image, and has some bad cases with padding.

We can look at the FAISS Documents: https://github.com/facebookresearch/faiss/wiki/Binary-indexes

Check out also https://github.com/facebookresearch/faiss/wiki/Binary-hashing-index-benchmark, which uses the equivalent of PDQ hashes (256 bit), in roughly our target scale (50M hashes), with radius 32 (our target PDQ dist).

b8zhong:
@Dcallies I agree. Thanks for review -looking into it now.
By the way -- if there's anything else related to dbm I can do concurrently -- let me know?

Dcallies:
There certainly is, which is to start building out the storage interface, backed by DBM.

I can either give you a couple of PR's to serve as the starting point, and you can build the rest out, or you can do a couple of smaller PRs in that direction and hash it out over the review comments. Which is your preference?

b8zhong:
@Dcallies Either works for me! If you want it to get done faster... I don't mind if you do the first few.

Dcallies:
Can get you started then, give me a week to make a few PRs!

b8zhong:
Still working on this: swapping to `index = faiss.IndexBinaryFlat(BITS_IN_PDQ)`, etc. seems to cause a bunch of downstream issues. 

Apparently -

`convert_pdq_strings_to_ndarray function unpacks the binary data to individual bits, creating a 256-element array for a 32-byte PDQ hash`

Gonna try to figure it out something this week

Dcallies:
I made a few small PRs for how you might do DBM and interface migrations, respectively, if you get stuck here and why to try that instead.

b8zhong:
Thanks!

Apologies for the massive delay @Dcallies... to be honest not sure when I'll have time to get around to this. If you would like, feel free to take over the branch if you want 👍

Issue #1757, opened 2025-02-09
CI failure(s) caused by ubuntu-latest 22.04 --> 24.04 update


ianwal:
GitHub finished updating `ubuntu-latest` from Ubuntu 22.04 --> 24.04 on Jan 17 2025. Many of the CI workflow files use `ubuntu-latest`.

This probably broke some of the actions that use `ubuntu-latest`. The actions aren't run periodically, so it's unknown what all workflows are broken. They need to be manually run to check.

The laziest fix right now is would be to replace `ubuntu-latest` to `ubuntu-22.04` everywhere.

At the least, pdq build CI is currently broken due to some change in `imagemagick`, which is a runtime dependency for CImg for decoding images files and is required for tests. I don't think `imagemagick` installed by default anymore. Installing `imagemagick` in the CI workflow should fix this (#1758)

Dcallies:
@ianwal - should we close this after your fix in #1758?

ThisIsMissEm:
You probably would want to pin the distro versions, e.g., `ubuntu-22.04, ubuntu-24.04` instead of whatever "latest" is.

It may also make sense to test on a system such as alpine or debian, which can produce smaller docker images, but have more up to date or varying system packages.

ianwal:
> [@ianwal](https://github.com/ianwal) - should we close this after your fix in [#1758](https://github.com/facebook/ThreatExchange/pull/1758)?

pdq was fixed, not sure if there's any issues with any of the other workflows. Feel free to close this one. If there's issues with the other projects they can be fixed as they're found.

PR #1758, opened 2025-02-09
[pdq] Fix pdq CI failure caused by ubuntu-latest 22.04 --> 24.04 update

ianwal:
Summary
---------

Fix pdq CI test failure on Ubuntu 24.04 (#1757 for pdq only)

`imagemagick` isn't installed by Ubuntu 24.04 by default anymore, and it's a runtime dependency of CImg for decoding images. Installing it fixes the issue.

### Changes

- Install `imagemagick` before building pdq
  - This is the ONLY required change to get the build working.

- Added multiple OS profiles to the CI to make it easier to spot a failure caused by OS differences in the future.
  - The only thing I had to debug this issue with was this runtime message: ` [CImg] *** CImgIOException *** [instance(0,0,0,0,(nil),non-shared)] CImg<unsigned char>::load(): Failed to recognize format of file './reg_test/../../data/reg-test-input/dih/bridge-1-original.jpg'.` It would have been very helpful to be able to see the failure was only occurring on one profile but not the other.

- Updated `checkout` action from v2 to v4

Test Plan
---------

CI should fail without the changes (first commit).

After the changes, CI should pass on both 22.04 and 24.04.

### DIFF
--- a/.github/workflows/pdq-ci-cpp.yml
+++ b/.github/workflows/pdq-ci-cpp.yml
@@ -21,8 +21,15 @@ defaults:
 
 jobs:
   build-and-test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: ["ubuntu-22.04", "ubuntu-latest"]
+    runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


Dcallies (approved):
Many thanks!

Issue #1761, opened 2025-02-12
[hma] Add the ability to GET the signals from a content_id


Dcallies:
Problem to solve:
* After getting a match hit, we want to inspect the actual hash that we matched.
* However, you can't get, we only have the content_id
* Getting signal_type requires an extra query on the database
* Therefore, add it as an optional parameter on the GET for bank_content

Extensions:
* We made need to add something on the match API to say what algorithm we matched with

PR #1763, opened 2025-02-13
[hma][api] bank_get_content can optionally return signals

vincent-4:
Summary
---------

An attempt:
Not 100% on the DB stuff. I'm having issues running tests locally, so I'm mainly putting this up for now to see the CI workflow output. If there are failures, feel free to come back later, when it would make more sense.

Test Plan
---------

Yes, new tests in PR: tests are _pretty_ self-explanatory, in this case at least.
Edit: This is woards #1761

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -150,6 +150,26 @@ def bank_get_content(bank_name: str, content_id: int):
     Query Parameters:
         signal_type (optional): If specified, includes the signal value for this signal type
 
+    Returns: JSON representation of the bank content
+
+    Without signal_type parameter:
+    {
+      'id': 1234,
+      'bank': 'TEST_BANK',
+      'enabled': true,
+      'original_media_uri': 'file:///data/media/uploaded_content_123.jpg'
+    }
+
+    With signal_type parameter:
+    {
+      'id': 1234,
+      'bank': 'TEST_BANK',
+      'enabled': true,
+      'original_media_uri': 'file:///data/media/uploaded_content_123.jpg',
+      'signals': {
+         'pdq': 'f8f8f0cee0f4a84f06370a22038f63f0b36e2ed596621e1d33e6b39c4e9c9b22'
+      }
+    }
     """
     storage = persistence.get_storage()
     bank = storage.get_bank(bank_name)
@@ -177,7 +197,7 @@ def bank_get_content(bank_name: str, content_id: int):
     if signal_type and content_obj.signals:
         response["signals"] = content_obj.signals
 
-    return jsonify(response)
+    return response
 
 
 @bp.route("/bank/<bank_name>/content", methods=["POST"])
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -34,7 +34,7 @@ class BankContentResponse(t.TypedDict):
     bank: str
     enabled: bool
     original_media_uri: t.Optional[str]
-    signals: t.Dict[str, str]
+    signals: t.NotRequired[dict[str, str]]
 
 
 bp = Blueprint("curation", __name__)
@@ -149,6 +149,7 @@ def bank_get_content(bank_name: str, content_id: int):
 
     Query Parameters:
         signal_type (optional): If specified, includes the signal value for this signal type
+
     """
     storage = persistence.get_storage()
     bank = storage.get_bank(bank_name)
@@ -171,8 +172,10 @@ def bank_get_content(bank_name: str, content_id: int):
         "bank": content_obj.bank.name,
         "enabled": content_obj.enabled,
         "original_media_uri": content_obj.original_media_uri,
-        "signals": content_obj.signals,
     }
+
+    if signal_type and content_obj.signals:
+        response["signals"] = content_obj.signals
 
     return jsonify(response)
 
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -29,6 +29,14 @@ class BankedContentMetadata(t.TypedDict):
     json: t.NotRequired[dict[t.Any, t.Any]]
 
 
+class BankContentResponse(t.TypedDict):
+    id: int
+    bank: str
+    enabled: bool
+    original_media_uri: t.Optional[str]
+    signals: t.Dict[str, str]
+
+
 bp = Blueprint("curation", __name__)
 bp.register_error_handler(HTTPException, flask_utils.api_error_handler)
 
@@ -136,14 +144,37 @@ def _validate_bank_add_metadata() -> t.Optional[BankedContentMetadata]:
 
 @bp.route("/bank/<bank_name>/content/<content_id>", methods=["GET"])
 def bank_get_content(bank_name: str, content_id: int):
+    """
+    Get content from a bank by ID.
+
+    Query Parameters:
+        signal_type (optional): If specified, includes the signal value for this signal type
+    """
     storage = persistence.get_storage()
     bank = storage.get_bank(bank_name)
     if not bank:
         abort(404, f"bank '{bank_name}' not found")
-    content = storage.bank_content_get([content_id])
+
+    signal_type = request.args.get("signal_type")
+    if signal_type:
+        signal_type_cfgs = storage.get_signal_type_configs()
+        if signal_type not in signal_type_cfgs:
+            abort(400, f"No such signal type '{signal_type}'")
+
+    content = storage.bank_content_get([content_id], signal_type)
     if not content:
         abort(404, f"content '{content_id}' not found")
-    return jsonify(content[0])
+
+    content_obj = content[0]
+    response: BankContentResponse = {
+        "id": content_obj.id,
+        "bank": content_obj.bank.name,
+        "enabled": content_obj.enabled,
+        "original_media_uri": content_obj.original_media_uri,
+        "signals": content_obj.signals,
+    }
+
+    return jsonify(response)
 
 
 @bp.route("/bank/<bank_name>/content", methods=["POST"])
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -178,8 +178,8 @@ def lookup_signal_with_distance(
         {
             "bank_content_id": m.metadata,
             "distance": m.similarity_info.pretty_str(),
-            "signal_type": signal_type_name,
-            "signal_value": signal,
+            "signal_type": m.similarity_info.signal_type.get_name(),
+            "signal_value": m.similarity_info.signal_value,
         }
         for m in results
     ]
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -178,8 +178,8 @@ def lookup_signal_with_distance(
         {
             "bank_content_id": m.metadata,
             "distance": m.similarity_info.pretty_str(),
-            "signal_type": m.similarity_info.signal_type.get_name(),
-            "signal_value": m.similarity_info.signal_value,
+            "signal_type": signal_type_name,
+            "signal_value": signal,
         }
         for m in results
     ]
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -43,6 +43,8 @@ bp.register_error_handler(HTTPException, api_error_handler)
 class MatchWithDistance(t.TypedDict):
     bank_content_id: int
     distance: str
+    signal_type: str
+    signal_value: str
 
 
 TMatchByBank = t.Mapping[str, t.Sequence[MatchWithDistance]]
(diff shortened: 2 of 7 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -29,6 +29,14 @@ class BankedContentMetadata(t.TypedDict):
     json: t.NotRequired[dict[t.Any, t.Any]]
 
 
+class BankContentResponse(t.TypedDict):
```
Dcallies, line 32: ignorable: I think we've found that dataclass also translates, but I'm fine with this

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -165,6 +167,8 @@ def lookup_signal_with_distance(
         {
             "content_id": m.metadata,
             "distance": m.similarity_info.pretty_str(),
+            "signal_type": signal_type_name,
+            "signal_value": signal,
```
Dcallies, line 171: blocking: They want to the value of the matched signal (which would need another fetch on the db for content_id), not the value they passed in. They already know the value they passed in :P 

Propose we leave this off this PR and do it after we figure out the content GET first.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -279,12 +283,15 @@ def lookup(signal, signal_type_name):
...
-                "distance": match.similarity_info.distance,
-            }
-        )
+        if match:
+            results[content.bank.name].append(
+                {
+                    "bank_content_id": content.id,
+                    "distance": match.similarity_info.distance,
+                    "signal_type": signal_type_name,
+                    "signal_value": signal,
```
Dcallies, line 292: ditto about the matched signal on this one

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -329,6 +329,7 @@ class BankContentConfig:
     original_media_uri: t.Optional[str]
 
     bank: BankConfig
+    signals: t.Dict[str, str] = field(default_factory=dict)
```
Dcallies, line 332: blocking: I'm not sure about this, because we know that we'll need to store it in a separate database table, and making the interface promise to return it every time will add another join to the database that we don't need most of the time, and relying on it only being populated some of the time (depending on the get method) will be confusing.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -409,8 +410,19 @@ def bank_delete(self, name: str) -> None:
 
     # Bank content
     @abc.abstractmethod
-    def bank_content_get(self, id: t.Iterable[int]) -> t.Sequence[BankContentConfig]:
-        """Get the content config for a bank"""
+    def bank_content_get(
+        self, id: t.Iterable[int], signal_type: t.Optional[str] = None
```
Dcallies, line 414: I'm trying to convince myself between adding this as the query here vs having a separate interface which takes the bank_content_id and only returns the signal types. 

Normally, we don't expect the users to be getting the signals in the same pass.

The key question is whether we think there is a usecase for fetching the signals without content_config, and I think there might be, but maybe not one worth optimizing for.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -409,8 +410,19 @@ def bank_delete(self, name: str) -> None:
...
+    ) -> t.Sequence[BankContentConfig]:
```
Dcallies, line 415: unsure: As an alternative to including it in the bank content config, another solution would be to return this as `Mapping[int, Ret]` where ret could include the config and the requested signals separately.

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -166,6 +166,11 @@ def as_storage_iface_cls(self) -> BankContentConfig:
             collab_metadata={},
             original_media_uri=None,
             bank=self.bank.as_storage_iface_cls(),
+            signals=(
+                {s.signal_type: s.signal_val for s in self.signals}
+                if self.signals
+                else {}
+            ),
```
Dcallies, line 173: blocking: self.signals is hiding a database call here I believe (though rusty on the dark magic that flask uses on these classes), by including it in this method you are always fetching it (the thing I was concerned about for performance :P )

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -166,6 +166,11 @@ def as_storage_iface_cls(self) -> BankContentConfig:
             collab_metadata={},
             original_media_uri=None,
             bank=self.bank.as_storage_iface_cls(),
+            signals=(
+                {s.signal_type: s.signal_val for s in self.signals}
+                if self.signals
+                else {}
```
Dcallies, line 172: ignorable/code golf: I believe this produces the same output
```suggestion
```

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
```diff
@@ -523,14 +523,35 @@ def bank_delete(self, name: str) -> None:
...
-            .filter(database.BankContent.id.in_(ids))
-            .all()
-        ]
+        query = database.db.session.query(database.BankContent)
+
+        if signal_type is not None:
+            query = query.outerjoin(
+                database.ContentSignal,
+                database.ContentSignal.content_id == database.BankContent.id,
+            )
```
Dcallies, line 534: ignorable; I think if we are going to fetch this information, doing it as part of the original query is the most efficient way to do it. I just haven't done the labor to try and prove what the access patterns are and whether it's worth optimizing for this one.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -409,8 +410,19 @@ def bank_delete(self, name: str) -> None:
...
+        """
+        Get the content config for a bank.
+
+        Args:
+            id: The IDs of the bank content to retrieve
+            signal_type: Optional signal type to include in the response
+
+        Returns:
+            List of bank content configs, optionally including signal values
+        """
```
Dcallies, line 426: unsure: As an alternative, we could provide a method that allows fetching the signals one at a time directly. The upsides would be that it's simpler to implement multiple interfaces, and would allow fetching strictly only the signals.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -136,14 +144,37 @@ def _validate_bank_add_metadata() -> t.Optional[BankedContentMetadata]:
...
+    content_obj = content[0]
+    response: BankContentResponse = {
+        "id": content_obj.id,
+        "bank": content_obj.bank.name,
+        "enabled": content_obj.enabled,
+        "original_media_uri": content_obj.original_media_uri,
+        "signals": content_obj.signals,
+    }
+
+    return jsonify(response)
```
Dcallies, line 177: ignorable; I believe jsonify is optional if you are returning a dict, but might be misremembering

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -136,14 +144,37 @@ def _validate_bank_add_metadata() -> t.Optional[BankedContentMetadata]:
 
 @bp.route("/bank/<bank_name>/content/<content_id>", methods=["GET"])
 def bank_get_content(bank_name: str, content_id: int):
+    """
+    Get content from a bank by ID.
+
+    Query Parameters:
+        signal_type (optional): If specified, includes the signal value for this signal type
+    """
```
Dcallies, line 152: nit: Include a sample output on the docstring (see bank_add_content)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -29,6 +29,14 @@ class BankedContentMetadata(t.TypedDict):
     json: t.NotRequired[dict[t.Any, t.Any]]
 
 
+class BankContentResponse(t.TypedDict):
+    id: int
+    bank: str
+    enabled: bool
+    original_media_uri: t.Optional[str]
+    signals: t.Dict[str, str]
```
Dcallies, line 37: You can denote this key is optional with NotRequired (https://typing.readthedocs.io/en/latest/spec/typeddict.html#required-and-notrequired), which I think is fine since our min version for HMA is high enough (3.11)

```suggestion
    signals: dict[str, str]
```

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -40,6 +40,8 @@
 class MatchWithDistance(t.TypedDict):
     content_id: int
     distance: str
+    signal_type: str
+    signal_value: str
```
Dcallies, line 44: blocking: Let's leave this off in this pass, since this is trickier than it looks

Dcallies (changes requested):
Whoops, forgot to hit request changes.

I think your base is overall solid, and since your unittest is on the API interface you don't have to change much on that side.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
vincent-4: I ditched it for now

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
vincent-4: Do you have an updated opinion on this? Tbh, w/o feedback from users, I'll just ditch it for now.

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
vincent-4: Removed this

vincent-4:
I haven't hit resolve on a few blocking comments that I don't have a particular opinion on. I think this is ready for a second look, if you're able to.

I think, maybe I could make the interface changes in another PR, either separate or combined with the second needed one about more of the nitty gritty implementation.

Dcallies:
Hey @vincent-4! I lost track of this one, let me self-put it up for review, and I might do some work to split out some of the changes to help land it, sorry for the delay!

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
Dcallies: Let's instead pass a bool to include all the signals
```suggestion
        self, id: t.Iterable[int], *, include_signals: bool = false
```

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -159,15 +159,33 @@ def set_typed_config(self, cfg: BankContentConfig) -> t.Self:
         self.disable_until_ts = cfg.disable_until_ts
         return self
 
-    def as_storage_iface_cls(self) -> BankContentConfig:
-        return BankContentConfig(
-            self.id,
+    def as_storage_iface_cls(self, include_signals: bool = False) -> BankContentConfig:
```
Dcallies, line 162: nit: For boolean arguments, we can force them to be self-documenting in the caller like this
```suggestion
    def as_storage_iface_cls(self, *, include_signals: bool = False) -> BankContentConfig:
```

ThisIsMissEm:
Out of curiosity, would it be better to have an endpoint: `//c/bank/{bank_name}/content/{content_id}/signals` which returns all signals for a given `content_id` and then allow filtering that result set by signal type?

That'd reduce the amount of data returned / queried when just checking "does this content_id exist?"

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -159,15 +159,33 @@ def set_typed_config(self, cfg: BankContentConfig) -> t.Self:
         self.disable_until_ts = cfg.disable_until_ts
         return self
 
-    def as_storage_iface_cls(self) -> BankContentConfig:
-        return BankContentConfig(
-            self.id,
+    def as_storage_iface_cls(self, *, include_signals: bool = False) -> BankContentConfig:
```
ThisIsMissEm, line 162: This fixes the code style issue reported by `black`:

```suggestion
-    def as_storage_iface_cls(self, *, include_signals: bool = False) -> BankContentConfig:
+    def as_storage_iface_cls(
+        self, *, include_signals: bool = False
+    ) -> BankContentConfig:
```

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -409,8 +417,25 @@ def bank_delete(self, name: str) -> None:
 
     # Bank content
     @abc.abstractmethod
-    def bank_content_get(self, id: t.Iterable[int]) -> t.Sequence[BankContentConfig]:
-        """Get the content config for a bank"""
+    def bank_content_get(
+        self, id: t.Iterable[int], *, include_signals: bool = false
```
ThisIsMissEm, line 421:
```suggestion
        self, id: t.Iterable[int], *, include_signals: bool = False
```

Fixes one mypy error

ThisIsMissEm: Though looking at this method signature in impl.py and mocked.py, I think you want the interface here to be:

```suggestion
        self, id: t.Iterable[int], *, signal_type: t.Optional[str] = None
```

Which fixes the errors in impl.py and mocked.py

PR #1762, opened 2025-02-12
[hma] Fix typos in HMA contributing document

vincent-4:
Summary
---------

:p Fix some typos while I was onboarding to work on the latest HMA issue...
Btw, this onboarding doc is very good.

Test Plan
---------

Only doc changes– N/A

PR #1764, opened 2025-02-13
Trim formatting

vincent-4:
Summary
---------

It's kinda silly, but I was wondering how to do it _without_ removing spaces for sentences that had a line break.
https://github.com/facebook/ThreatExchange/pull/1762#discussion_r1954813278
```
This is a sentence[SPACE]
and it breaks back there.
```
We don't overwrite this type of formatting, which is expected, right?.
This would be kiinda hard to do with sed so I ended up trying a python script which works pretty well

Test Plan
---------

Only doc changes- not needed.


Dcallies:
Seems fine - put it up for review if you'd like!

Dcallies (approved):
ltgm

PR #1767, opened 2025-02-22
[hma] Fix lookup with dist

Dcallies:
Summary
---------

A prior PR accidentally used the wrong distance API, and the type checker couldn't tell, likely due to the number of generics used along the way.

The real villain here is the lack of comprehensive testing on the lookup APIs, which I made the start of here.

Fixes #1766

Test Plan
---------

Added regression test, breaks before with the stack trace from #1766, works after.

PR #1770, opened 2025-02-24
[hma] Align lookup APIs, fix more bugs

### LINKED ISSUE #1766: [hma] unable to lookup a video_md5 hashes after distance was added to m/lookup
to reproduce:
- get a video_md5 hash
- try to look it up (m/lookup?signal=<hash>&signal_type=video_md5)
- error:
```
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/OpenMediaMatch/blueprints/matching.py", line 221, in lookup_get
    resp = lookup(signal, signal_type)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/OpenMediaMatch/blueprints/matching.py", line 284, in lookup
    "distance": match.similarity_info.distance,
AttributeError: 'SignalSimilarityInfo' object has no attribute 'distance'
```

relevant PR: https://github.com/facebook/ThreatExchange/pull/1738

Dcallies:
Summary
---------

In #1767, I attempted to fix #1766, but missed the actual APIs in the stacktrace because I am a hack and a fraud.

Add regression tests for /lookup as well.

As part of fixing this, I added some missing type annotations, which caught a few other bugs. I also chose to align all the APIs to the same shapes, which of course will be backwards incompatible, but I don't think current users are relying on the ones I changed being 100% stable at the moment. 

Test Plan
---------

Regression tests, will ask @prenner if it really works this time.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -253,7 +253,6 @@ def lookup_get() -> t.Union[TMatchByBank, TBankMatchBySignalType]:
     else:
         signal = require_request_param("signal")
         signal_type = require_request_param("signal_type")
-        # Don't
         return lookup(signal, signal_type)
 
     selected_st = request.args.get("signal_type")
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -37,9 +37,16 @@ bp = Blueprint("matching", __name__)
 bp.register_error_handler(HTTPException, api_error_handler)
 
 
+# Type helpers
+
+
 class MatchWithDistance(t.TypedDict):
-    content_id: int
+    bank_content_id: int
     distance: str
+
+
+TMatchByBank = t.Mapping[str, t.Sequence[MatchWithDistance]]
+TBankMatchBySignalType = t.Mapping[str, TMatchByBank]
 
 
 @dataclass
@@ -167,7 +174,7 @@ def lookup_signal_with_distance(
     results = query_index(signal, signal_type_name)
     return [
         {
-            "content_id": m.metadata,
+            "bank_content_id": m.metadata,
             "distance": m.similarity_info.pretty_str(),
         }
         for m in results
@@ -190,7 +197,7 @@ def _validate_and_transform_signal_type(
 
 
 @bp.route("/lookup", methods=["GET"])
-def lookup_get():
+def lookup_get() -> t.Union[TMatchByBank, TBankMatchBySignalType]:
     """
     Look up a hash in the similarity index. The hash can either be specified via
     `signal_type` and `signal` query params, or a file url can be provided in the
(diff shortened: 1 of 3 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -167,7 +174,7 @@ def lookup_signal_with_distance(
     results = query_index(signal, signal_type_name)
     return [
         {
-            "content_id": m.metadata,
+            "bank_content_id": m.metadata,
```
Dcallies, line 177: @prenner - will this break you? I can temporarily write both if that helps, but I think the extra bytes for "bank" are worth it.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -206,29 +213,57 @@ def lookup_get():
...
         hashes = hashing.hash_media()
-        resp = {}
         for signal_type in hashes.keys():
             signal = hashes[signal_type]
             resp[signal_type] = lookup(signal, signal_type)
     else:
         signal = require_request_param("signal")
         signal_type = require_request_param("signal_type")
-        resp = lookup(signal, signal_type)
+        # Don't
```
Dcallies, line 256: Don't! (to fix)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -206,29 +213,57 @@ def lookup_get():
      Also (applies to both cases):
      * Optional seed (content id) for consistent coinflip
     Output:
-     * JSON object with a key of each bank name with matches.
-       For each bank, there is list of objects containing details
-       about each match.
+     * JSON object keyed by signal type, to a JSON object of bank
+       matches. If Signal Type is given, the outer JSON object is
+       elided.
```
Dcallies, line 218: This formalizes what had been sort of happening already, but I'm generally skeptical of these kinds of behaviors.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -206,29 +213,57 @@ def lookup_get():
...
+
+    Example output:
+    {
+        "pdq": {
+            "BANK_A": [
+                {"bank_content_id": 1001, "distance": 4},
+                {"bank_content_id": 1002, "distance": 0},
+            ],
+            "BANK_B": [
```
Dcallies, line 227: I personally find this nesting a bit strange, I'd prefer banks be on the "outside".

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_e2e_workflow.py
```diff
@@ -39,7 +39,7 @@ def test_add_hashes_and_find_match(app: Flask, client: FlaskClient):
     # Now match
     resp = client.get(f"/m/lookup?signal_type=pdq&signal={hashes[-1]}")
     assert resp.status_code == 200
-    assert resp.json == {"TEST_BANK": [{"bank_content_id": 16, "distance": 0}]}
+    assert resp.json == {"TEST_BANK": [{"bank_content_id": 16, "distance": "0"}]}
```
Dcallies, line 42: This is the net result of this change - I can try and do something to detect when there's a primitive type in the distance slot and return the correctly typed JSON if that would be preferable.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -282,13 +316,12 @@ def lookup(signal, signal_type_name):
...
-        results[content.bank.name].append(
-            {
-                "bank_content_id": content.id,
-                "distance": match.similarity_info.distance,
-            }
-        )
+        matched_content = results_by_bank_content_id[content.id]
+        match: MatchWithDistance = {
+            "bank_content_id": content.id,
+            "distance": matched_content.similarity_info.pretty_str(),
```
prenner, line 321: this is what I need 👍

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
prenner: no preferences on my side (though we are using the outer-level bank response already)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
aokolish: Not using it.

PR #1773, opened 2025-02-26
[py-tx] Deprecate SignalExchange API method

### LINKED ISSUE #1218: [py-tx] Deprecate SignalExchangeAPI.fetch_value_merge
This API was intended to support a fancy data storage model that we don't actually need if we split up downloading from the API and converting to signals, like we eventually did.

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/signal_exchange_api.py#L127

Next steps:

1. Should be marked as deprecated in the comments and marked for removal in 2.x.x
2. Existing callsites should be removed

b8zhong:
Summary
---------

Added deprecation notices to both methods in SignalExchangeAPI class, indicating they will be removed in version 2.x.x

Remove the few references to it there was.

Closes #1218

Test Plan
---------
Yep


python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -407,6 +410,10 @@ def fetch_value_merge(
         old: t.Optional[api.NCMECEntryUpdate],
         new: t.Optional[api.NCMECEntryUpdate],
     ) -> t.Optional[api.NCMECEntryUpdate]:
+        """
+        DEPRECATED: This method will be removed in version 2.x.x.
+        The deletion logic has been moved directly to fetch_iter().
+        """
         assert new is not None, "fetch shouldn't do this"
         if new.deleted:
             return None
```
Dcallies, line 419: Hmm, this is a real functionality we need to store somewhere. The task mentions changing the behavior of fetch(). See L434 on the new code for where this is handled in convert

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -189,6 +193,8 @@ def naive_fetch_merge(
         """
         Merge the results of a fetch in-memory.
 
+        DEPRECATED: This method will be removed in version 2.x.x.
+
```
Dcallies, line 197: We should keep this version around - it's marked final, and the argument to keep it around is on L198.

Instead, we should should try and tell if it's safe to change this method not to call fetch_value_merge. 

```suggestion
```

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -167,6 +167,10 @@ def fetch_value_merge(
         """
         Merge a new update produced by fetch.
 
+        DEPRECATED: This method will be removed in version 2.x.x.
```
Dcallies, line 170: In addition, we should use the `@deprecated` annotation - you can include the removal message there.

https://typing.readthedocs.io/en/latest/spec/directives.html#deprecated

python-threatexchange/threatexchange/exchanges/helpers.py
```diff
@@ -56,7 +56,13 @@ def delta(self, value: fetch_state.FetchDeltaTyped) -> None:
             self._delta = value
         else:
             old = self._delta.updates
-        self.api_cls.naive_fetch_merge(old, value.updates)
```
Dcallies, line 59: Let's keep this version around, it's useful to formally describe the update logic somewhere.

Dcallies (changes requested):
Thanks for taking a look!

Let's use the `@deprecated` decorator, and leave naive_fetch_merge around as a documentation of the correct way to merge this data.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
b8zhong: @Dcallies `py-tx` goes back to Python 3.8 right? `@deprecated` I think is only introduced in 3.13

What do you think of using `warnings.warn`?

python-threatexchange/threatexchange/exchanges/helpers.py
b8zhong: I agree

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -197,11 +210,10 @@ def naive_fetch_merge(
         together keyed by ID will eventually get you an entire copy of the database.
         """
         for k, v in new.items():
-            new_v = cls.fetch_value_merge(old.get(k), v)
-            if new_v is None:
+            if v is None:
```
b8zhong, line 213: I think... this might be right? If something is deleted -- it would already be `None`... ?

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -407,6 +411,16 @@ def fetch_value_merge(
         old: t.Optional[api.NCMECEntryUpdate],
         new: t.Optional[api.NCMECEntryUpdate],
     ) -> t.Optional[api.NCMECEntryUpdate]:
+        """
+        DEPRECATED: This method will be removed in version 2.x.x.
+        The deletion logic has been moved directly to fetch_iter().
+        """
+        warnings.warn(
+            "fetch_value_merge() is deprecated and will be removed in version 2.x.x. "
+            "The deletion logic has been moved directly to fetch_iter().",
```
Dcallies, line 420: Not quite - you need to add this logic in fetch_iter!

python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py
```diff
@@ -373,7 +374,10 @@ def log(event: str) -> None:
                     log(f"large fetch ({i}) with {total_fetched} updates.")
 
                 yield state.FetchDelta(
-                    {f"{entry.member_id}-{entry.id}": entry for entry in entry.updates},
+                    {
+                        f"{update.member_id}-{update.id}": update
+                        for update in entry.updates
+                    },
```
Dcallies, line 380: Instead of yielding `: update`, you need to yield None if update is deleted - you could inline it here.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
Dcallies: D'oh! It goes back to 3.9 (3.8 is deprecated). 

Warnings is a fine workaround.

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
```diff
@@ -167,13 +168,25 @@ def fetch_value_merge(
...
 
         Most implementations will probably prefer the default, which is to
         replace the record entirely.
 
         It is safe to mutate `new` inline and return it, if needed.
         """
+        warnings.warn(
+            "fetch_value_merge() is deprecated and will be removed in version 2.x.x. "
+            "Any implementations that convert records into None to signal a deletion "
+            "should instead move that logic into the fetch() method.",
```
Dcallies, line 185: I think we figured out it's `fetch_iter`

python-threatexchange/threatexchange/exchanges/signal_exchange_api.py
Dcallies: If we were outright deleting fetch_value_merge, then yes!

However, we might need to keep the old code around until 2.x.x. See toplevel comment about what to do.

Dcallies (changes requested):
Hmm, as written I think this might break the NCMEC fetch code! We might be missing a unittest that tests to make sure NCMEC deletions are handled properly (or else it might be a bug that is compensated for later).

Let's break this up into stages:
1. Review the current API integrations and see which ones are relying on an override of `fetch_value_merge`
2. Fix the implementations relying on `fetch_value_merge` to not do so, which allows for the deletion of the override on that exchange class
3. In order to defend against the existence of any third party extensions that rely on overrides, you could try detecting whether the class has an override which is not equal to the default. I think you might be able to determine that via reflection. If you detect an override, emit a warning.

PR #1775, opened 2025-03-02
[CI] Fix labelling actions

b8zhong:
Summary
---------

I'm pretty sure labelling actions doesn't work. Trying to fix it and see if it works, that's why there's random typo fixes lol

Test Plan
---------

Let CI do it's thing


b8zhong:
I believe this will only work when it's merged into main... can we give this a shot?

Dcallies (approved):
I made an attempt to fix this last year and couldn't quite get it to work. Thanks for looking into it!

Let's see if it works!

PR #1797, opened 2025-03-11
DESCRIPTION.rst => into `pyproject.toml` file

b8zhong:
Summary
---------
Screenshot 2025-03-11 at 2 05 04 PM

Apparently, we can't handle `DESCRIPTION.rst`... so I think it's easier just to put it in the toml file directly.

Test Plan
---------

Yeah


Dcallies (approved):
Noticed this as well, thanks for doing it

PR #1805, opened 2025-03-27
Fix bug in HMA redirect logic

aokolish:
Summary
---------

Fixes a bug I introduced in #1803 where the redirect logic is inverted.

Test Plan
---------

n/a


Dcallies (approved):
Oops - we should add a regression test

PR #1809, opened 2025-04-09
[tmk] Fix build on Linux with latest FAISS

IvanPizhenko:
Summary
---------

Fixes broken build of tmk-query-with-faiss on Linux systems with the latest version of FAISS.

Test Plan
---------

Tested on Ubuntu 22.04 with GCC 12. Should work with default GCC 11 too.

1. Built FAISS from sources (took latest commit in the FAISS repo).
2. Built "parallel" TMK version `make parallel`.


Dcallies:
Thanks a ton @IvanPizhenko! If the CI passes I'll merge!

Dcallies (approved):
Thanks for the PR!

PR #1814, opened 2025-04-24
Update TAT API base URL

aokolish:
Summary
---------

We received an update from Tech Against Terrorism that they are changing the subdomain of their API:

image

Test Plan
---------

I verified their server responds when I `curl` it using the new subdomain, so I'm pretty confident this will work.


Dcallies (approved):
Thanks for helping us keep this up to date!

PR #1817, opened 2025-04-26
[pdq] optimize PDQ linear scan implementation

### LINKED ISSUE #1810: Proposal: A faster strategy to do PDQ matching in practice with AVX512
Hi, 

I’ve been working on a solution for performing PDQ-based screening in high-throughput or resource-constrained environments. My approach involves 10x+ faster PDQ-compatible variant and an specialized exact matching solution tailored for PDQ-specific tasks; the latter is the focus here (though I’d be happy to share more about my variant once I gather more concrete validation data). While reviewing your benchmark and tutorial on MIH for PDQ hash matching, **I noticed an opportunity for specialization of the simpler linear scan method that could perform significantly faster than MIH at 10 million-vector haystack without the need for big MIH indices that is invariant on dynamic changes in radii, particularly when processing all 8 dihedral variants required for reliable matching.**

I was wondering would you like some documentation (or C++/Rust code) contributed on this approach which may actually be more scalable and fruitful than MIH in high-throughput content-moderation systems.

## Algorithm Description

The key insight is that since PDQ is not invariant to dihedrals, to reliably match images one must always process all 8 dihedral variants. While MIH is faster for single-hash queries, a flat scan can amortize the cost by processing all 8 matches simultaneously, whereas MIH requires 8 separate lookups.

Looking at the characteristics of this problem (256-bit hashes, 8 dihedral variants), we can observe that 512/8=64, which enables highly efficient matching on AVX512VPOPVNTDQ CPUs (most AVX512 CPUs since 2021).

The basic setup is similar to [Faiss](https://github.com/facebookresearch/faiss/blob/117aafd90b06b70d81bc7cad442a3fc56b7ac866/faiss/utils/hamming_distance/avx512-inl.h#L12), however we can specialize to the constraint that the query is always batched by 8, and eliminate the expensive unaligned loads and horizontal reduction step entirely (which Faiss cannot avoid due to its generic nature, but with my strategy 256*8 happened to line up perfectly to the architecture). 

At the beginning of each query, we do a one-time shuffle of the 64-bit words of all 8 hashes (A-H) like this:

ZMM1: [A0, B0, C0, D0, E0, F0, G0, H0]
ZMM2: [A1, B1, C1, D1, E1, F1, G1, H1]
ZMM3: ... the 3rd quad words..
ZMM4:  ... the 4th quad words ..

In the inner matching loop, for a database entry [Q0, Q1, Q2, Q3], we can broadcast each qword to a register (avoiding unaligned zmmword loads):

ZMM5: [Q0, Q0, Q0, Q0, Q0, Q0, Q0, Q0]
ZMM6: .. the 2nd quad word of the DB entry ..
ZMM7: ..
ZMM8: ..

The operation `(POPCNT(ZMM1 ^ ZMM5) + POPCNT(ZMM2 ^ ZMM6) + ... + ...)` yields the hamming distances for all 8 queries in the 64-bit lanes of the final register (ZMM9), without requiring horizontal shuffling.

Furthermore, on micro-architectures where vectorized comparison may have lower throughput (not common and I don't have concrete examples but just in case), if we only need to detect matches (rather than exact indices or the matching distance), we can optimize by first complementing all queries (or the database). This allows thresholding to be done with ZMM10 |= (ZMM9 + threshold), followed by checking if any lane in ZMM10 & 256 is non-zero.

## Implementation and Benchmark

All time/throughput data except from ThreatExchange official benchmark code considers matching __8 hashes against the haystack as an atomic unit of work__, because that is an universal requirement for almost all "serious" usage of PDQ..

[My solution](https://github.com/eternal-flame-AD/yume-pdq/blob/d033fb2d166e2dde9d3cbb810a10ed0c06dd7085/src/matching/mod.rs#L252) When compiled down to assembly with LLVM optimizations this is nearing the theoretical minimum number of operations for a linear scan, about 6.1 cycles per haystack on my machine (i.e. an amortized cost of less than 1 cycle per pairwise comparison):

```asm
<yume_pdq::matching::CpuMatcher as yume_pdq::matching::PDQMatcher>::scan:
        .cfi_startproc
        vpbroadcastq zmm0, qword ptr [rdi + 256]
        vmovdqa64 zmm1, zmmword ptr [rdi]
        vmovdqa64 zmm2, zmmword ptr [rdi + 64]
        vmovdqa64 zmm3, zmmword ptr [rdi + 128]
        vmovdqa64 zmm5, zmmword ptr [rdi + 192]
        vpxor xmm4, xmm4, xmm4
        xor eax, eax
        .p2align        4
.LBB36_1:
        vpxorq zmm6, zmm1, qword ptr [rsi + rax]{1to8}
        vpxorq zmm7, zmm2, qword ptr [rsi + rax + 8]{1to8}
        vpxorq zmm8, zmm3, qword ptr [rsi + rax + 16]{1to8}
        vpxorq zmm9, zmm5, qword ptr [rsi + rax + 24]{1to8}
        add rax, 32
        vpopcntq zmm6, zmm6
        vpopcntq zmm7, zmm7
        vpopcntq zmm8, zmm8
        vpaddq zmm7, zmm7, zmm8
        vpaddq zmm6, zmm6, zmm0
        vpopcntq zmm8, zmm9
        vpaddq zmm6, zmm6, zmm7
        vpaddq zmm6, zmm6, zmm8
        vpord zmm4, zmm4, zmm6
        cmp rax, 65536
        jne .LBB36_1
        vptestmd k0, zmm4, zmmword ptr [rip + .LCPI36_0]
        kortestw k0, k0
        setne al
        vzeroupper
        ret
```

This is the benchmark on my machine (I modified your code to show the number divided by 8 but otherwise identical), and 10 million was not chosen arbitrarily (doesn't even fit in my L3 cache), but to my knowledge the current size of NCMEC database, rounded up to power of 10:

```
CFLAGS="-march=native -funroll-loops -ftree-vectorize -O3" make benchmark-query 
g++  -g -std=c++11 -march=native -funroll-loops -ftree-vectorize -O3 -I../.. -I. -Wall -Wno-char-subscripts -Wno-stringop-truncation -Wno-class-memaccess -Wno-misleading-indentation -Wno-unknown-warning-option -Werror -O3 bin/benchmark-query.cpp common/pdqhashtypes.cpp io/hashio.cpp common/pdqhamming.cpp hashing/pdqhashing.cpp downscaling/downscaling.cpp io/pdqio.cpp hashing/torben.cpp common/pdqutils.cpp -o benchmark-query -lm -lpthread
 ./benchmark-query -b (math 10_000_000) -m mih 

METHOD: mutually-indexed hashing query
QUERY COUNT:             1000
INDEX COUNT:             10000000
TOTAL MATCH COUNT:       1000
TOTAL QUERY SECONDS:     8.774241
QUERIES PER SECOND:     113.97
8 DIHEDRAL QUERIES PER SECOND (computed):     14.25
```

If we get a CPU baseline without vectorization (this was still slower than MIH, but not by much):
```
RUSTFLAGS="-Ctarget-cpu=core2" cargo bench -pyume-pdq  popcnt
NUM_BATCHES: 4883 (10000384 vectors, 305 MiB)
8needles/scan10mil/cpu popcnt
                        time:   [115.21 ms 115.30 ms 115.39 ms]
                        thrpt:  [86.664 Melem/s 86.735 Melem/s 86.803 Melem/s]
```

And if we try it with AVX512 we can be faster than MIH by a big margin (~6x from this benchmark), I also have a Vulkan based matcher that is even faster:

```
RUSTFLAGS="-Ctarget-feature=+avx512vpopcntdq,+avx2" cargo bench -pyume-pdq  --features "vulkan" 8needles
8needles/scan10mil/cpu popcnt
                        time:   [11.421 ms 11.440 ms 11.460 ms]
                        thrpt:  [872.62 Melem/s 874.19 Melem/s 875.62 Melem/s]
                 change:
                        time:   [-90.096% -90.078% -90.056%] (p = 0.00 < 0.05)
                        thrpt:  [+905.63% +907.89% +909.65%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
adapter: AdapterInfo { name: "NVIDIA GeForce RTX 4070", vendor: 4318, device: 10118, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "570.133.07", backend: Vulkan }
8needles/scan10mil/vulkan
                        time:   [773.95 µs 779.13 µs 785.39 µs]
                        thrpt:  [12.733 Gelem/s 12.835 Gelem/s 12.921 Gelem/s]
Found 11 outliers among 100 measurements (11.00%)
  3 (3.00%) high mild
  8 (8.00%) high severe
```

## Notes on Security

Another benefit of doing linear scan is it is constant time, you will not have certain inputs consistently take significantly longer than others, which can become primitives for information disclosure or even denial of service. This is an inherent property of MIH and AFAIK no way around it.

### Notes on Memory Bandwidth

On high-end server CPUs some can indeed fit the whole 10M database in L3, making memory bandwidth mostly irrelevant, but even if the database is 100M (~3G), each query only does a single scan, a production-grade quad-channel DDR5 can comfortably fulfill 5+ busy looping threads not counting any shared working set caches).

### Why MIH is so slow despite pruning

The constant factor in MIH is actually very high, as seen in this profiler output, just loading the hit candidates by index is enough to stall the CPU to spend around 1 trillion cycles (and another one trillion not shown here spent on hashmap probing, likely stalled for branch prediction) for 1000 singular queries loading data to confirm the distance to candiates, if you count by 8 that is 125 effective queries, while some happened to get hit in L3, since the index cannot fit in L3 a significant portion will incur the full memory load latency. As index size grows the probabilistic L3 hit rate will continue to drop.

Image

eternal-flame-AD:
Closes #1810 

Summary
---------

This is a **breaking** API change and need some notice in release notes, as the alignment of Hash256 is now 64-bit, although it should only affect users' who reuse old artifacts (static/dynamic libs) while building new code using the new header. No downstream code change should be needed.

I also refactored the pdq benchmark data structure to not interleave strings with actual hash data, it is bad for caching and increases TLB pressure that would hurt both the existing MIH implementation and linear scan. (Also fixed a logic bug that can cause arithmetic wraparound/infinite looping in the benchmark execution, not relevant to results.

I can probably shave off some constant factor in MIH too but that won't change the big picture, so I will defer it to maybe a later PR if needed.

Performance as of now:

```bash
> CFLAGS="-march=znver4 -funroll-loops -ftree-vectorize -O3" make benchmark-query
>  ./benchmark-query -b (math 10_000_000)  -m $method

MIH index build time: 74.216144

METHOD: mutually-indexed hashing query
QUERY COUNT:             1000 * 8 = 8000
INDEX COUNT:             10000000
TOTAL MATCH COUNT:       8000
TOTAL QUERY SECONDS:     59.697593
QUERIES PER SECOND:     16.75
THROUGHPUT (millions of amortized tests/sec): 1340.09

METHOD: linear query (SIMD accelerated)
QUERY COUNT:             1000 * 8 = 8000
INDEX COUNT:             10000000
TOTAL MATCH COUNT:       8000
TOTAL QUERY SECONDS:     14.524117
QUERIES PER SECOND:     68.85
THROUGHPUT (millions of amortized tests/sec): 5508.08
```

<details>
<summary> Assembly for queryAll hot path</summary>

```asm
   151d0:	62 71 fd 48 6f cf    	vmovdqa64 zmm9,zmm7
   151d6:	62 71 fd 48 6f de    	vmovdqa64 zmm11,zmm6
   151dc:	62 71 fd 48 6f ed    	vmovdqa64 zmm13,zmm5
   151e2:	62 71 fd 48 6f fc    	vmovdqa64 zmm15,zmm4
   151e8:	62 53 b5 58 25 0f 99 	vpternlogq zmm9,zmm9,QWORD BCST [r15],0x99
   151ef:	62 53 a5 58 25 5f 01 	vpternlogq zmm11,zmm11,QWORD BCST [r15+0x8],0x99
   151f6:	99 
   151f7:	62 52 fd 48 55 d1    	vpopcntq zmm10,zmm9
   151fd:	62 53 95 58 25 6f 02 	vpternlogq zmm13,zmm13,QWORD BCST [r15+0x10],0x99
   15204:	99 
   15205:	62 53 85 58 25 7f 03 	vpternlogq zmm15,zmm15,QWORD BCST [r15+0x18],0x99
   1520c:	99 
   1520d:	62 52 fd 48 55 e3    	vpopcntq zmm12,zmm11
   15213:	62 52 fd 48 55 f5    	vpopcntq zmm14,zmm13
   15219:	62 d1 ad 48 d4 c4    	vpaddq zmm0,zmm10,zmm12
   1521f:	62 d2 fd 48 55 d7    	vpopcntq zmm2,zmm15
   15225:	62 d1 fd 48 d4 ce    	vpaddq zmm1,zmm0,zmm14
   1522b:	62 f1 f5 48 d4 da    	vpaddq zmm3,zmm1,zmm2
   15231:	62 d3 e5 48 1f c0 05 	vpcmpnltq k0,zmm3,zmm8
   15238:	c5 f9 98 c0          	kortestb k0,k0
   1523c:	0f 85 7d 01 00 00    	jne    153bf <_ZL5queryPciPS_+0x165f>
   15242:	48 ff c2             	inc    rdx
   15245:	49 83 c7 20          	add    r15,0x20
   15249:	4c 39 ca             	cmp    rdx,r9
   1524c:	72 82                	jb     151d0 <_ZL5queryPciPS_+0x1470>
```
</details>

"Barcelona" is AMD K10, a 2007 CPU with almost no SIMD support, it is placed here for comparison for difference achievable only using scalar optimizations (alignment, scalar POPCNT r/64 encouragement that at least LLVM seems to fail to do automatically, etc). 

```bash
> CFLAGS="-march-barcelona -funroll-loops -ftree-vectorize -O3" make benchmark-query
> ./benchmark-query -b (math 10_000_000) -m $method
MIH index build time: 63.761978

METHOD: mutually-indexed hashing query
QUERY COUNT:             1000 * 8 = 8000
INDEX COUNT:             10000000
TOTAL MATCH COUNT:       8000
TOTAL QUERY SECONDS:     58.247728
QUERIES PER SECOND:     17.17
THROUGHPUT (millions of amortized tests/sec): 1373.44

METHOD: linear query
QUERY COUNT:             1000 * 8 = 8000
INDEX COUNT:             10000000
TOTAL MATCH COUNT:       8000
TOTAL QUERY SECONDS:     62.094976
QUERIES PER SECOND:     16.10
THROUGHPUT (millions of amortized tests/sec): 1288.35

```

Test Plan
---------

The matching is mirrored from my Rust implementation with no logical changes, which has robust Unit testing and edge case testing, plus end-to-end image matching tests, the benchmark results are also comparable to my statistics-based Criterion benchmark in Rust implementation.

Unit test and Rust implementation: https://github.com/eternal-flame-AD/yume-pdq/blob/main/src/matching/mod.rs
End to end test using synthetic DB: https://github.com/eternal-flame-AD/yume-pdq/blob/main/examples/end_to_end.rs

A comprehensive Unit test is written in C++ too using similar methodology (exhaustively check every possible combination of needle/haystack index), Test results:

```
Using scalar flat index / Using SIMD accelerated flat index
Testing maxDistance = 0
        PASS: initially no hits
        PASS: 1 hit
        PASS: 2 hits
        PASS: no edge-case false positives
        PASS: no hits at the end
Testing maxDistance = 31
        PASS: initially no hits
        PASS: 1 hit
        PASS: 2 hits
        PASS: no edge-case false positives
        PASS: no hits at the end
Testing maxDistance = 64
        PASS: initially no hits
        PASS: 1 hit
        PASS: 2 hits
        PASS: no edge-case false positives
        PASS: no hits at the end
Testing correctness under misuse
        PASS: correctness under misuse
```

Since MIH index used `int` for specifying threshold I followed suit, but I added a test to make sure even under nonsensical input (not within 0-255) the answer is still technically correct.

Load/Store alignment is tested fully using RFLAGS AC bit in both Rust and C++ implementation.

Although we won't be able to test the AVX512 kernel on Github Actions, AFAIK it does not have AVX512, and QEMU also does not support it, for now we have to test locally.


eternal-flame-AD:
As a disclosure, you might notice that the performance of MIH increase dramatically at really big query counts, this artifact of hashmap micro-architecture behavior and is not achievable in production environments IMO.

The benchmark is inherently O(n) with regards to "query count", thus this speedup is almost certainly branch prediction or caching (cache learned where the matches artificially introduced in the indices are, thus boosting the cache hit rate), in a noisy, multi-threaded environment with a lot of irrelevant code in between there is very little opportunity for a CPU core to learn data access behavior that cannot be amortized in 10k queries under ideal back-to-back query conditions..

eternal-flame-AD:
I think this is ready for review.

The ASM is a little more convoluted than the one generated from LLVM Rust for some reason, I "agree" with the rust output more on how it should look like.

I am not too familiar with C++ codegen optimization, I tried writing exactly what the ASM I want says GCC still rewrote it, so it isn't some syntax/alias thing but the optimizer is actively saying they think this is better. I don't think I can fight back on that except write asm directly. 

I asked AI it says it is GCC's heuristics prefer to first do a move then do ternary when LLVM prefer to do broadcast then xor with broadcast operand directly as I prescribed, but the speedup is there and the answer is correct, so I guess it doesn't really hurt.

pdq/cpp/common/pdqhashtypes.h
```diff
@@ -78,14 +86,24 @@ struct Hash256 {
     }
     return n;
   }
+#ifdef __POPCNT__
+  int hammingDistance(const Hash256& that) const {
+    const uint64_t* quadwords = reinterpret_cast<const uint64_t*>(this->w);
+    const uint64_t* that_quadwords = reinterpret_cast<const uint64_t*>(that.w);
+    return _mm_popcnt_u64(quadwords[0] ^ that_quadwords[0]) +
```
ianwal, line 93: Hi!

When I changed hamming distance calculation from __builtin_popcount to use std::bitset::count() in (#1333) (see `pdqhamming.h`), I inspected the assembly on x86-64 and I was convinced that `std::bitset<>(x ^ y).count()` generated nearly the same, if not identical, code to using any popcount intrinsic function I threw at it.

See this test: https://godbolt.org/z/dKMG5hn18
image


Can you try running some benchmarks with this:

```cpp
  int hammingDistance(const Hash256& that) const {
    const uint64_t* quadwords = reinterpret_cast<const uint64_t*>(this->w);
    const uint64_t* that_quadwords = reinterpret_cast<const uint64_t*>(that.w);
    return std::bitset<64>(quadwords[0] ^ that_quadwords[0]).count() +
        std::bitset<64>(quadwords[1] ^ that_quadwords[1]).count() +
        std::bitset<64>(quadwords[2] ^ that_quadwords[2]).count() +
        std::bitset<64>(quadwords[3] ^ that_quadwords[3]).count();
  }
```

If the above is just as fast, then it would be preferable because bitset is portable, and we only need one implementation for this function.

We can probably add a `hammingDistance64()` either way, which may be a bit cleaner than implementing this function twice.

eternal-flame-AD: Yes if you explicitly write it as 64-bit numbers probably compilers can vectorize I can get rid of intrinstic, but the original code is very difficult to optimize out (if you look at the original issue the profile output clearly shows the compiler doing peeling instead of clean 64-bit popcnt).

The AVX512 implementation it is not possible for optimizer to do this, it might be able to use vpopcntq if the vectorizer is aggressive enough but it is not possible for it to do data layout optimizations.

eternal-flame-AD: Done! thanks for catching that

ianwal: >  but the original code without alignment is very difficult to optimize out (if you look at the original issue the profile output clearly shows the compiler doing peeling instead of clean 64-bit popcnt).

Absolutely, makes sense.

ianwal:
Please copy any modified pdq files to `vpdq/cpp/pdq/cpp` as well (this sucks...)

eternal-flame-AD:
Done, I copied the core files there that should have 100% the same behavior but I am not familiar with vpdq so not sure how to wire it up.

I think it is best handled in another PR if needed as much more things need to be validated and the diff is already not small.

pdq/cpp/common/pdqhashtypes.h
eternal-flame-AD: I saw the hammingDistance*() functions, it says backwards compatibility only, so I am confused about whether that function is intended to be just there or?

And it obviously isn't viable to have a LUT for hammingDistance64.

I can certainly put the bitset impl there too for unification.

ianwal:
> Done, I copied the core files there that should have 100% the same behavior but I am not familiar with vpdq so not sure how to wire it up.
> 
> I think it is best handled in another PR if needed as much more things need to be validated and the diff is already not small.

No need to copy the new `index/flat.h` file to vpdq. Actually, the whole index folder could be removed from vpdq because it's not used, but you don't need to do that in this PR. CI runs tests for vpdq if there's changes to pdq, so no need to worry about testing anything manually there.

eternal-flame-AD:
Since this were brought up I will share some history on development and how I came to this instruction sequence,

If I recall correctly my initial solution was pretty similar to Faiss (straightforward vpxor[d|q] then vpopcnt[d|q] then do a horizontal add), I benchmarked it was not faster than a properly aligned scalar solution.

That is intuitive too, there are 4 qwords in a hash so even if the POPCNT is free, at least 2 permutations and then 2 additions (which has at least 1 chain of dependency) would be required for each POPCNT result to be reduced into a single distance metric that could be obtained using cheaper scalar sequences, but to get further performance we need to spend on average <3 cycles per hash so obviously that won't cut it.

So I thought about with the natural dihedral batching we can do this one-time transpose approach which saved all the expensive, ILP-hard horizontal operations with broadcasts and vertical additions, and vectorized tests. And indeed with no unaligned loads or horizontal operations we can do 8 things at once with minimal overhead and thus we can get this throughput.

Dcallies:
Sorry, I'm way underwater. Thanks @ianwal  for the help here!

> This is a breaking API change and need some notice in release notes

We aren't maintaining release notes at the moment, but I can start a document. The biggest source of downstream usage that I know of is https://github.com/faustomorales/pdqhash-python (pdq bindings). 

Huge huge thanks for the high quality PR here with unittests. I haven't coded C++ in like a decade, but I'll do my best look.

Dcallies (approved):
I don't fully understand the details and potential downstream issues we may see, but I think I see the crux of the optimization. I'm personally fine always going with FAISS as the index of choice, but can also understand why that might not be the preferred case. 

I'm okay with landing this as is as the CI is passing (the format issue you are fine to do yourself or we can just do it as a followup PR, since it's unrelated to the actual changes).

@ianwal - I'll give you a few days in case you have any other review feedback, since you appear to understand the nuts and bolts more than I do. If not, I'll go ahead and merge Thursday or later, and hopefully we don't have to do a backout later.

eternal-flame-AD:
It seems like clang isn't capturing .clang-format file, I looked at the CI logs and ran that container, if this doesn't fix the formatting complaint I probably don't know how to fix it .

vpdq/cpp/pdq/cpp/common/pdqhashtypes.h
```diff
@@ -12,7 +12,8 @@
 #include <pdq/cpp/common/pdqbasetypes.h>
 #include <pdq/cpp/common/pdqhamming.h>
 
-#include <stdio.h>
+#include <bitset>
+#include <cinttypes>
```
ianwal, line 16: Replace this <cstdint>. It's smaller than <cinttypes> for the fixed-sized types.

Is `<bitset>` used in this file anymore or can it be removed?

pdq/cpp/bin/test-flat.cpp
```diff
@@ -0,0 +1,232 @@
+// ================================================================
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+// ================================================================
+
+#include <stdio.h>
```
ianwal, line 5: Use the C++ equivalent headers for the C-standard library headers, e.g., `stdio.h` --> `<cstdio>`. I know pdq doesn't do this everywhere, but it should.

pdq/cpp/common/pdqhashtypes.h
```diff
@@ -31,7 +32,10 @@ using Hash256Text = char[HASH256_TEXT_LENGTH];
 
 // ================================================================
 struct Hash256 {
-  Hash16 w[HASH256_NUM_WORDS];
+  // align to 64-bit boundary for faster scalar popcnt
+  // (compiler can optimize without peeling to natural word size)
+  // index/flat.h also relies on this assumption, do not remove.
```
ianwal, line 37: Consider adding a static assert for `alignof(Hash256) == 8` and/or `alignof(Hash256::w) == 8` wherever code requires this alignment assumption, i.e., at least in `flat.h`.

pdq/cpp/index/flat.h
```diff
@@ -0,0 +1,265 @@
...
+#else
+  Flat(const std::array<facebook::pdq::hashing::Hash256, 8>& needles)
+      : _needles(needles) {}
+#endif
+
+  ~Flat() {}
+
+ private:
+  // Disallow copying
+  Flat(const Flat& /*that*/) {}
```
ianwal, line 92: C++11 supports deleted functions which provides a much more friendly compile error if it's used: `Flat(const Flat&) = delete`.

No need to add `/*that*/`. Omitted parameter is fine.

Probably should delete the move constructor and move assignment operators as well (rule of five).

pdq/cpp/index/flat.h
```diff
@@ -0,0 +1,265 @@
...
+  void operator=(const facebook::pdq::hashing::Hash256& /*that*/) {}
```
ianwal, line 93: Not really sure why this has a Hash256& as a parameter. I'm assuming this is supposed to be a plain-old copy-assignment operator? If so, the copy-assignment operator should have the following signature:

`Flat& operator=(const Flat&);`

then give it a definition as well to follow the rule-of-five:

`Flat& operator=(const Flat&) = delete;`

And move to public.

pdq/cpp/index/flat.h
```diff
@@ -0,0 +1,265 @@
...
+      }
+      _packedNeedles[regi] = _mm512_xor_epi64(ones, _mm512_load_si512(tmp));
+    }
+  }
+#else
+  Flat(const std::array<facebook::pdq::hashing::Hash256, 8>& needles)
+      : _needles(needles) {}
+#endif
+
+  ~Flat() {}
```
ianwal, line 88: `~Flat() {}` --> `~Flat() = default;`, see https://github.com/facebook/ThreatExchange/pull/1817/files#r2065413659

vpdq/cpp/pdq/cpp/common/pdqhamming.h
```diff
@@ -9,35 +9,56 @@
...
 #ifdef USE_BUILTIN_POPCOUNT
-// Inlined right here
+
+/**
+ * @brief Returns the number of bits set.
+ *
+ * @param T An unsigned integer type.
+ */
+template <typename T>
+int popcount(T const x) {
```
ianwal, line 32:
```suggestion
int popcount(const T x) {
```

Argh... sorry!

vpdq/cpp/pdq/cpp/common/pdqhashtypes.h
ianwal, line 16:
```suggestion
#include <cstdint>
```

Replace `<cinttypes>` with `<cstdint>`. It's smaller than <cinttypes> for the fixed-sized types.

Is `<bitset>` used in this file anymore or can it be removed?

ianwal (commented):
The changes in `pdq/cpp/common/` look good. I don't see any risks there.

I'm not very familiar with the new flat.h addition, and I'm pretty busy this week at work to have much energy afterwards to review that material. So if it's not breaking anything I'm okay with it. Seems well documented enough.

Left a few suggestions that are mostly style stuff. Let me know if you want any clarification for them them :)

Please make sure any changes from my suggestions are synced between the vpdq files and pdq files!

After that, I'm good with merging.

pdq/cpp/index/flat.h
ianwal, line 88:
```suggestion
   ~Flat() = default;
```

`~Flat() {}` --> ``,

See https://github.com/facebook/ThreatExchange/pull/1817/files#r2065413659

pdq/cpp/index/flat.h
ianwal, line 93:
```suggestion
  `Flat& operator=(const Flat&);`
```

Not really sure why this has a Hash256& as a parameter. I'm assuming this is supposed to be a plain-old copy-assignment operator? If so, the copy-assignment operator should have the suggested signature.

This suggestion is included in https://github.com/facebook/ThreatExchange/pull/1817/files/d0f9fb0f75da3c587d6b2f60bcd8f4372ada6bb5#r2065413659 as well.

ianwal:
Do you need to approve workflow run @Dcallies? Let's double check for any new compiler warnings that show up there before merging.

pdq/cpp/index/flat.h
eternal-flame-AD: I applied your suggestion, I am not too familiar with C++ coding standards, thanks for the help.

pdq/cpp/bin/test-flat.cpp
eternal-flame-AD: I changed all C headers to C++ ones in files that this PR involved.

Dcallies:
Yeah, it needs me to manually run the CI for non-contributors (security step). 

Looks like we are aligned on landing, @eternal-flame-AD , we just need to get the CI green again and we can merge!

eternal-flame-AD:
I fixed the CMake complaint but It seems there is some weirdness going on with the Makefile warning flags, it first sets a WFLAGS that is different for each compiler than resets to another constant value?

https://github.com/facebook/ThreatExchange/blob/5eae856ab2e8ed6984469e97beec6958c463507d/pdq/cpp/Makefile#L2-L17

I tried `make wasm` it doesn't seem to work for me, but unrelated to this PR:

```c++
 make wasm                                                                                                                                                                                                                                                  10:08:14
emcc -g -std=c++11 -D BUILD_WASM -s ERROR_ON_UNDEFINED_SYMBOLS=0 -s EXPORTED_RUNTIME_METHODS=['ccall','UTF8ToString ']  -s  EXPORTED_FUNCTIONS=['_getHash'] -s FORCE_FILESYSTEM=1 -s LLD_REPORT_UNDEFINED -s ERROR_ON_UNDEFINED_SYMBOLS=0 -s DISABLE_EXCEPTION_CATCHING=0 -s ALLOW_MEMORY_GROWTH=1  -I../.. -I. -Wall -Wno-char-subscripts -Wno-stringop-truncation -Wno-class-memaccess -Wno-misleading-indentation -Wno-unknown-warning-option -Werror -O3 bin/pdq-photo-hasher.cpp common/pdqhashtypes.cpp io/hashio.cpp common/pdqhamming.cpp hashing/pdqhashing.cpp downscaling/downscaling.cpp io/pdqio.cpp hashing/torben.cpp common/pdqutils.cpp -o pdq-photo-hasher.js -lm -lpthread
In file included from bin/pdq-photo-hasher.cpp:7:
In file included from ../../pdq/cpp/io/pdqio.h:14:
./CImg.h:81800:23: error: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'CImg<char>' [-Werror,-Wnontrivial-memcall]
 81800 |           std::memcpy(new_data, _data, sizeof(CImg<T>) * npos);
       |                       ^
./CImg.h:81867:7: note: in instantiation of member function 'cimg_library::CImgList<char>::insert' requested here
 81867 |       insert(empty, npos + i);
       |       ^
./CImg.h:16345:18: note: in instantiation of member function 'cimg_library::CImgList<char>::insert' requested here
 16345 |     move_to(list.insert(1, npos)[npos]);
       |                  ^
./CImg.h:87439:27: note: in instantiation of function template specialization 'cimg_library::CImg<char>::move_to<char>' requested here
 87439 |             full_filename.move_to(res);
       |                           ^
./CImg.h:81800:23: note: explicitly cast the pointer to silence this warning
 81800 |           std::memcpy(new_data, _data, sizeof(CImg<T>) * npos);
       |                       ^
       |                       (void*)
./CImg.h:81803:31: error: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'CImg<char>' [-Werror,-Wnontrivial-memcall]
 81803 |               new_data + npos + 1,
       |                               ^
./CImg.h:81803:31: note: explicitly cast the pointer to silence this warning
 81803 |               new_data + npos + 1,
       |                               ^
       |               (void*)
./CImg.h:81819:21: error: first argument in call to 'memset' is a pointer to non-trivially copyable type 'CImg<char>' [-Werror,-Wnontrivial-memcall]
 81819 |         std::memset(_data, 0, sizeof(CImg<T>) * (_width - 1));
       |                     ^
./CImg.h:81819:21: note: explicitly cast the pointer to silence this warning
 81819 |         std::memset(_data, 0, sizeof(CImg<T>) * (_width - 1));
       |                     ^
       |                     (void*)

<...>
```

eternal-flame-AD:
@Dcallies any updates? Can you approve the CI so I know it passes?

The WASM bug seems to be unrelated to this PR.

Dcallies:
Apologies, swamped on my end, I'll give notice to @faustomorales that there's an updated PDQ version.

faustomorales:
> Apologies, swamped on my end, I'll give notice to @faustomorales that there's an updated PDQ version.

Here to report that the Python bindings include this new version on [v0.2.8](https://pypi.org/project/pdqhash/0.2.8/) 🫡 🎉 😂

PR #1821, opened 2025-05-12
OMM: Update hashing to support FileHashers

ThisIsMissEm:
@Dcallies this is the change I needed to make VPDQ usable through the Curator Web UI, since it uses `hash_media_from_form_data` internally.

This is the same thing that the `hash_media` endpoint does when receiving a `url`, as that uses FileHasher, not BytesHasher.

We'd probably also need something in Hashing to have support for `URLContent` and `TextContent`, but that'd require different UI completely (text input instead of file input)

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
@@ -121,9 +121,9 @@ def hash_media_from_form_data() -> dict[str, str]:
                 elif issubclass(st, FileHasher):
                     with tempfile.NamedTemporaryFile("wb") as tmp:
                         current_app.logger.debug(
-                           "Writing to %s for hashing, signal_type=%s",
-                           tmp.name,
-                           st.get_name(),
+                            "Writing to %s for hashing, signal_type=%s",
+                            tmp.name,
+                            st.get_name(),
                         )
                         with tmp.file as temp_file:  # this ensures that bytes are flushed before hashing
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
(Also, just updated my CLA, because I don't remember signing the CLA recently, so it may have been before my identity changed)

Dcallies (approved):
My one long-term concern here is I want to try and avoid writing anything to disk for high risk harms. I think there is probably an interface tweak to allow this for vPDQ, e.g. passing the metadata need to identify the file type without necessarily using a tmpfile.

However, I'd rather have HMA work consistently, especially at this stage. Thanks for the fix!

Dcallies:
Looks like you have just a code format lint, let me know if you can't run it on your side, I can merge and run it on mine.

ThisIsMissEm:
@Dcallies my understanding is that vPDQ needs the file on disk, because it passes it to ffmpeg? I'm not sure if you can stream the file into ffmpeg or something?

Dcallies:
Not sure! It might be a much larger refactor if we'd have to replace ffmpeg

ThisIsMissEm:
@Dcallies yeah, so perhaps just make it known that anything that extends `FileHasher` may temporarily write bytes to disk.

ThisIsMissEm:
Screenshot 2025-05-13 at 21 23 24

As an aside, this isn't particularly helpful. Adding `--diff` to `black` makes more sense:
Screenshot 2025-05-13 at 21 24 08

PR #1822, opened 2025-05-14
[HMA] Add URL Validation for hashing.

juanmrad:
Summary
---------

Add URL validation to the service as a way to prevent big files from download and SSRF prevention.
Also added possibility for whitelist of host to help with abuse prevention. 

Test Plan
---------

added test

### DIFF
--- a/hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
+++ b/hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
@@ -40,6 +40,7 @@ UI_ENABLED = True
 TASK_FETCHER = True
 TASK_INDEXER = True
 TASK_INDEX_CACHE = True
+MAX_CONTENT_LENGTH = 100 * 1024 * 1024  # 100MB max file size
 
 # Core functionality configuration
 STORAGE_IFACE_INSTANCE = DefaultOMMStore(
--- a/hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
+++ b/hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
@@ -56,3 +56,11 @@ STORAGE_IFACE_INSTANCE = DefaultOMMStore(
 
 # Debugging stuff
 # SQLALCHEMY_ENGINE_LOG_LEVEL = logging.INFO
+
+# URL validation configuration
+# ALLOWED_HOSTNAMES = {
+#     'example.com',
+#     'cdn.example.com',
+#     'media.example.com'
+# }
+# Note: When ALLOWED_HOSTNAMES is not set or empty, all hostnames are allowed
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
@@ -90,34 +90,17 @@ def hash_media():
     if not is_valid_url(media_url):
         abort(400, "Invalid or unsafe URL provided")
 
-    # First make a HEAD request to check content length
     try:
-        head_resp = requests.head(media_url, allow_redirects=True, timeout=5)
-        head_resp.raise_for_status()
-
-        content_length = head_resp.headers.get("content-length")
-        if content_length and int(content_length) > MAX_CONTENT_LENGTH:
-            abort(
-                413,
-                f"Content too large. Maximum size is {MAX_CONTENT_LENGTH/1024/1024}MB",
-            )
-    except requests.exceptions.RequestException as e:
-        abort(400, f"Failed to fetch URL: {str(e)}")
-
-    # Now make the actual GET request with streaming to handle large files safely
-    try:
+        # Use a single GET request with body limiting
         download_resp = requests.get(
-            media_url, allow_redirects=True, timeout=30, stream=True
+            media_url,
+            allow_redirects=True,
+            timeout=30,
+            stream=True,
+            # Limit the response body to MAX_CONTENT_LENGTH
+            max_content_length=MAX_CONTENT_LENGTH
         )
         download_resp.raise_for_status()
-
-        # Check content length again from the actual response
-        content_length = download_resp.headers.get("content-length")
-        if content_length and int(content_length) > MAX_CONTENT_LENGTH:
-            abort(
-                413,
-                f"Content too large. Maximum size is {MAX_CONTENT_LENGTH/1024/1024}MB",
-            )
 
         url_content_type = download_resp.headers["content-type"]
 
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
@@ -124,8 +124,12 @@ def hash_media() -> dict[str, str]:
 
     try:
         # Get response with content length tracking
-        max_content_length = current_app.config.get("MAX_CONTENT_LENGTH", DEFAULT_MAX_CONTENT_LENGTH)
-        with _check_content_length_stream_response(media_url, max_content_length) as download_resp:
+        max_content_length = int(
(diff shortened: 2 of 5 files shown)


hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
```diff
@@ -56,3 +56,11 @@
 
 # Debugging stuff
 # SQLALCHEMY_ENGINE_LOG_LEVEL = logging.INFO
+
+# URL validation configuration
+# ALLOWED_HOSTNAMES = {
```
ThisIsMissEm, line 61: Perhaps we should have a toggle for `ALLOW_REMOTE_FILES` (and maybe `ALLOW_FILE_UPLOADS`) to completely disable

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -38,33 +85,64 @@ def hash_media() -> dict[str, str]:
...
     if media_url is None:
-        abort(400, "url is required")
+        abort(400, "Missing required parameter: url")
+
+    if not is_valid_url(media_url):
+        abort(400, "Invalid or unsafe URL provided")
 
-    download_resp = requests.get(media_url, allow_redirects=True, timeout=30 * 1000)
-    download_resp.raise_for_status()
+    # First make a HEAD request to check content length
```
ThisIsMissEm, line 93: The way this is written would allow a malicious server to return a passing response from the HEAD request and a malicious response for the GET; see the requests body limiting docs

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
juanmrad: Not sure why we would need to disable both. Disabling URLs. Also this can be something enabled/disabled by the consumer of the API. If no hashing needs to be happening we already have disabled the whole hashing path as part of the role for the container.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
juanmrad: used `max_content_length` instead.

juanmrad:
@Dcallies Whenever you have time :)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -246,6 +246,9 @@ def lookup_get() -> t.Union[TMatchByBank, TBankMatchBySignalType]:
             abort(403, "Hashing is disabled, missing role")
 
         hashes = hashing.hash_media()
+        if not hashes:
+            abort(400, "Failed to hash media from URL")
```
Dcallies, line 250: blocking: Hmm, in some cases this might be the expected result - if only using photo signal type, if the image is low quality, you'll get no hashes back. Hashing happened successfully and the correct result is no hashes. It's not a failure.

Can you help me understand what this is intended to catch?

Dcallies (changes requested):
Thanks for the PR @juanmrad! Only one blocking inline question about throwing an exception for no hashes generated, which is the expected behavior in some cases. 

I think we may want to return 204 (no content)  instead of 400 (user error), since in some cases everything has operated as expected.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
juanmrad: I was having some issues with hashes and figured returning a bad error over empty hashes 🤔 I guess the consumer can also handle the empty case over handling the error.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -38,33 +85,47 @@ def hash_media() -> dict[str, str]:
...
-    url_content_type = download_resp.headers["content-type"]
+    try:
+        # Use a single GET request with body limiting
+        download_resp = requests.get(
+            media_url,
+            allow_redirects=True,
+            timeout=30,
+            stream=True,
+            # Limit the response body to MAX_CONTENT_LENGTH
+            max_content_length=MAX_CONTENT_LENGTH
```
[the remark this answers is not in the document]
juanmrad: We did, but realized doing this here actually broke the other change I made to chunk download and write the file. So I instead did the content size check as we chunk download. That way we can raise the a content length error if we are over length and also take advantage of stream download.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -72,6 +77,33 @@ def is_valid_url(url: str) -> bool:
...
+    if content_length is not None and int(content_length) > max_length:
+        abort(413, "Content too large")
+
+    # If content length is acceptable, proceed with GET request
+    response = requests.get(url, stream=True, timeout=30, allow_redirects=True)
+    response.raise_for_status()
+
+    # Double check content length from GET response
+    content_length = response.headers.get("content-length")
+    if content_length is not None and int(content_length) > max_length:
```
Dcallies, line 101: Hmm, I think you can still get longer than the content-length if place you are calling doesn't return you a content-length header, which is the server discretion (it could also lie to you). I think if you want to truly defend against an attack, you need to keep track of the size yourself as you are streaming.... (see comment below)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -72,6 +77,33 @@ def is_valid_url(url: str) -> bool:
...
+        abort(413, "Content too large")
```
Dcallies, line 102: Do you need to do something to explicitly close the connection here?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -20,13 +26,86 @@
...
+    except Exception as e:
+        logger.warning(f"URL validation error: {str(e)}")
+        return False
+
+
+def _get_and_check_content_length(
+    url: str, max_length: int = MAX_CONTENT_LENGTH
+) -> requests.Response:
+    """
+    Get URL content with explicit content length tracking.
```
Dcallies, line 84: nit: Note that this won't get the content, but instead returns a stream to the content

Dcallies (approved):
Let me know if you want to merge as is or want to touch it more

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
ThisIsMissEm: Yeah, usually you're better setting an explicit hard upper limit 200mb 500mb etc.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -20,13 +26,86 @@
...
+    content_length = response.headers.get("content-length")
+    if content_length is not None and int(content_length) > max_length:
+        abort(413, "Content too large")
+
+    return response
+
 
 @bp.route("/hash", methods=["GET"])
-def hash_media() -> dict[str, str]:
+def hash_media():
```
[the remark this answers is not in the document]
juanmrad: reverted lol.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
juanmrad: changed comment

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
juanmrad: used context manager instead to ensure we clean-up in case of any exception.

juanmrad:
@Dcallies we should be good to merge :)

PR #1823, opened 2025-05-17
[CI] Update github workflows with black --diff

b8zhong:
Summary
---------

I agree with https://github.com/facebook/ThreatExchange/pull/1821#issuecomment-2877695616

I think we should do this?

Test Plan
---------


Dcallies (approved):
Thanks for doing this!

PR #1824, opened 2025-05-21
Process PNG files by ignoring the alpha channel

momo286:
Summary
---------

To process PNG images, they often have an alpha channel, thus the number of channels is >3. By changing to >3, it changes nothing to jpeg processing, but allows PNG processing by ignoring the alpha channel

with this change i can hash jpeg files and png files with no problem

Test Plan
---------
I modified a copy of this repo and compiled with required libraries for PNG processing and now i can process png files
image


momo286:
David, thanks for your answer and your detailed instructions. I will take a look and try to write a regression test that justifies this change. thank you

Issue #1830, opened 2025-06-12
[hma] Docker deployment improvements


ThisIsMissEm:
This may be a documentation issue, or perhaps it's just how the containers are built, but when trying to work with HMA using the container image, I found that I was having problems with the bash executing things part. I wasn't using the docker-compose file, as I already had postgresql running within docker, and wanted to connect to that.

I ended up with something like:
```sh
docker run --net development_default -p 5001:5000 -e OMM_DATABASE_URI="postgresql://postgres:@postgresql/hma_media_match" -e OMM_CONFIG="/build/reference_omm_configs/development_omm_config.py" -e OMM_SQLALCHEMY_ENGINE_LOG_LEVEL="DEBUG" -it ghcr.io/facebook/threatexchange/hma:latest '/bin/bash -c "MIGRATION_COMMAND=0 OMM_CONFIG=/build/reference_omm_configs/development_omm_config.py flask --app OpenMediaMatch.app run --host=0.0.0.0 --debug"'
```

Which felt rather complex. I also needed to run with migrations 0 instead of 1 after the initial migration for some reason (but I can't recall why)

I think the [dockerfile](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/Dockerfile) can definitely be improved and use a script to do multiple things on startup, for example, bluesky's ozone uses [a small node.js script](https://github.com/bluesky-social/ozone/blob/main/service/index.js) as it's command in its [dockerfile](https://github.com/bluesky-social/ozone/blob/main/Dockerfile)

So we could probably do something similar for HMA. Usually you want to use dumb-init or tini or something similar as PID 0 inside docker containers too.

Perhaps it'd also be worth having a python tool that could be a wizard to generate the configuration required?

julietshen:
fixed in https://github.com/facebook/ThreatExchange/pull/1832

PR #1832, opened 2025-06-12
Improve developer experience when using docker

ThisIsMissEm:
Summary
---------

This addresses #1830, we may want to consider writing a helm chart for easier production deployments, with the correct configuration files for high-availability.

This PR does the following things:
- Changes the port for the container from 5000 to 5100, as port 5000 on MacOS is now a reserved port (a system process runs on it)
- Always boots the application using gunicorn in the docker container
- Adds a database migration script
- Upgrades the base image version to python 3.13 (security)
- Uses environment variables in the development config for the database.
- Adds logging for the development config (so you can see the container is actually doing things)
- Places the reference configuration files at `/reference_configs/*` instead of `/build/reference_omm_configs/*` (short path to type / remember)
- Changes the docker-compose to use service dependencies to first start postgresql, then run the database migrations, and then start the application.

Test Plan
---------

- [x] Verify docker commands in README.md work as expected
- [x] I have verified the `docker compose up` command works as expected (by using `docker compose up --build` since the published image doesn't have these changes in it)


hasher-matcher-actioner/docker-compose.yaml
```diff
@@ -1,40 +1,55 @@
...
+    networks:
+      - backend
+
+  migrations:
+    build: .
+    restart: no
     environment:
       <<: *postgres-variables
-      OMM_CONFIG: /build/reference_omm_configs/development_omm_config.py
+      OMM_CONFIG: /reference_configs/development_omm_config.py
```
juanmrad:
```suggestion
      OMM_CONFIG: /reference_omm_configs/development_omm_config.py
```

hasher-matcher-actioner/reference_omm_configs/scaled_deployment_cron_omm_config.py
```diff
@@ -16,8 +16,8 @@
 # Database configuration
 DBUSER = os.environ.get("POSTGRES_USER", "media_match")
 DBPASS = os.environ.get("POSTGRES_PASSWORD", "hunter2")
-DBHOST = os.environ.get("POSTGRESS_HOST", "db")
-DBNAME = os.environ.get("POSTGRESS_DBNAME", "media_match")
+DBHOST = os.environ.get("POSTGRES_HOST", "db")
+DBNAME = os.environ.get("POSTGRES_DB", "media_match")
```
juanmrad: Thanks lol I kept forgetting to fix the spelling here lol!

juanmrad (approved):

hasher-matcher-actioner/docker-compose.yaml
ThisIsMissEm: This is actually already correct, see the copy in the dockerfile

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
```diff
@@ -24,10 +25,10 @@
...
 
 # Database configuration
-DBUSER = "media_match"
-DBPASS = "hunter2"
-DBHOST = "db"
-DBNAME = "media_match"
+DBUSER = os.environ.get("POSTGRES_USER", "media_match")
+DBPASS = os.environ.get("POSTGRES_PASSWORD", "hunter2")
+DBHOST = os.environ.get("POSTGRES_HOST", "db")
+DBNAME = os.environ.get("POSTGRES_DB", "media_match")
```
Dcallies: This may not be needed, I think we may already have the omm config to prefer environment variables over the file. 

If not, that is probably easy to add, and it might be confusing to try and read environment in the file. 

See more: https://flask.palletsprojects.com/en/stable/config/#configuring-from-environment-variables

Here's the corresponding code: https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/app.py#L83

If you are intentionally adding this so that DBUSER can be used instead of OMM_DBUSER, then just add a comment in this file that refers to the OMM_ prefix and explain why you are hardcoding it here.

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
```diff
@@ -65,3 +66,23 @@
 #     'media.example.com'
 # }
 # Note: When ALLOWED_HOSTNAMES is not set or empty, all hostnames are allowed
+
+# We always want some logging by default, otherwise it's hard to tell whats happening inside the container:
```
Dcallies, line 75: I think there is logging by default, no? This is only changing the default log level and format of the output?

hasher-matcher-actioner/reference_omm_configs/json_log_omm_config.py
```diff
@@ -26,10 +26,11 @@
...
 # Database configuration
-DBUSER = "media_match"
-DBPASS = "hunter2"
-DBHOST = "db"
-DBNAME = "media_match"
+DBUSER = os.environ.get("POSTGRES_USER", "media_match")
+DBPASS = os.environ.get("POSTGRES_PASSWORD", "hunter2")
+DBHOST = os.environ.get("POSTGRES_HOST", "db")
+DBNAME = os.environ.get("POSTGRES_DB", "media_match")
+
```
Dcallies: Same comment on this - I question this version more because I think this is supposed to be more for reference and not direct usage

Dcallies (changes requested):
Publish the PR when you are ready for review!

Did you know it's possible to configure with environment variables already? 

> I'm not sure what a test plan would look like for these changes

Perhaps we could run the docker files and make sure we could still use the system, such as by doing operations that require database writes!

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
ThisIsMissEm: Yeah, it's just in the docker-compose we're reusing these variables, without the `OMM_` prefix, because these are the environment variables needed to configure postgresql.

We could also change the docker-compose to do `OMM_DATABASE_URI` and construct it from those variables.

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
ThisIsMissEm: Not under gunicorn, only under the direct flask app with `--debug` — I'm not sure if there's a way to pass that through gunicorn

hasher-matcher-actioner/reference_omm_configs/json_log_omm_config.py
ThisIsMissEm: Would it be better to move that to an examples folder then?

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
ThisIsMissEm: Yeah, this is necessary for gunicorn, which is really what we want the docker image to use.

hasher-matcher-actioner/README.md
```diff
@@ -49,13 +49,20 @@ HMA requires a configuration file passed as an environment variable, `OMM_CONFIG
...
-$ docker run -e OMM_CONFIG='/build/reference_omm_configs/development_omm_config.py' -p 5000:5000 ghcr.io/facebook/threatexchange/hma flask --app OpenMediaMatch.app run --host=0.0.0.0
+$ docker run -e OMM_CONFIG='/reference_configs/development_omm_config.py' ghcr.io/facebook/threatexchange/hma /app/scripts/db-migrate.sh
 ```
 
-This command sets the necessary environment variable and exposes the app on port 5000 of your host machine, making the API accessible locally.
+
+Then we can start the server with:
+
+```bash
+$ docker run -e OMM_CONFIG='/reference_configs/development_omm_config.py' -p 5100:5100 ghcr.io/facebook/threatexchange/hma
```
ThisIsMissEm: This wouldn't be quite right for development, since we're relying on a built image, we may be better suggesting devcontainers here?

ThisIsMissEm (commented):
I've updated the changes based on code review, rather than doing changes in the config files, I've used the `OMM_DATABASE_URI` environment variable instead of setting individual parts of that database URI as environment variables.

ThisIsMissEm:
Have updated this after I noticed that the existing `docker compose up` just doesn't boot locally for me.

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
ThisIsMissEm, line 1: The `DATABASE_URI` and `DB*` variables in this file aren't used when running under `docker compose`, as we're setting `OMM_DATABASE_URI` there.

hasher-matcher-actioner/README.md
Dcallies: +1 to that change, up to you if you want to take a stab at it now or merge-and-iterate

hasher-matcher-actioner/README.md
```diff
@@ -64,7 +71,7 @@ For production environments, it is recommended to use a more robust server like
 Here is an example command to run the application with Gunicorn:
 
 ```bash
-$ docker run -e OMM_CONFIG='/build/reference_omm_configs/production_omm_config.py' -p 5000:5000 ghcr.io/facebook/threatexchange/hma gunicorn --bind 0.0.0.0:5000 "OpenMediaMatch.app:create_app()"
+$ docker run -v 'config:/config:ro' -e OMM_CONFIG='/config/production.py' -p 5100:5100 ghcr.io/facebook/threatexchange/hma"
```
Dcallies: blocking: Is this the right name? I don't see a production.py in this PR or [in the code](https://github.com/facebook/ThreatExchange/tree/main/hasher-matcher-actioner/reference_omm_configs). 

I wonder if we want to downplay the name a bit more - the longer "reference" name was to try and give the impression that the deployer should be tweaking it to their environment, and by shedding that name

hasher-matcher-actioner/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
-COPY ./src /build
+COPY ./src /app
+COPY ./scripts/ /app/scripts
 
 # Here is where you need to install your config and role settings
 # You will probably use your own rather than copying one of the examples
 # Be careful using the development config with gunicorn due to strange
 # interactions with apscheduler and forking.
-COPY reference_omm_configs/ ./reference_omm_configs
+COPY reference_omm_configs/ /reference_configs
```
Dcallies, line 14: Recommend against renaming it from the github version, since the purpose of this is to make it easier to draw the connection from the code. 

We could rename the folder in the repo /reference_configs so they lined up again

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
Dcallies: Hmm, gunicorn is what we want the production image to use, the development one we expect to be in debug/development mode. 

What's the behavior if running in with the default devcontainer setup? Any changes to output?

hasher-matcher-actioner/README.md
juanmrad: You can change this command by mounting the volume. 
```
--mount type=bind
```

This will make it such that changes on local will also change the container. I would also recommend instead using Docker compose as it will have both the app container and the postgresql container.

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
juanmrad: Agree with above for development it is better to use flask directly with debug flag to have all logs there.

for production it would be gunicorn and each consumer will define their logging interface.

hasher-matcher-actioner/Dockerfile
ThisIsMissEm: I went for renaming here simply because the `docker run -e OMM_CONFIG='/reference_omm_configs/development_omm_config.py'` was kinda a lot of OMM's, but yeah, if you're up for us renaming `reference_omm_configs/` to `reference_configs/` I'd be all for it. We could also rename the development config to just `development.py` since the other parts of the file name are redundant via the path / environment variable?

hasher-matcher-actioner/README.md
ThisIsMissEm: Would we be happy merging as is as it'll give you a local server you can *develop against* even though it doesn't give you a server you can *develop on* (i.e., differentiating development for integrators and development for us building HMA)

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
ThisIsMissEm: As mentioned in my other comment, I think we probably need to differentiate between development for integraters that want to run a server locally to test their code against, and development for the HMA contributors — for the latter, I think we can work out how to run the code, if not, improve a `contributing.md` file with better instructions?

ThisIsMissEm:
> blocking: As part of your test plan, can you run the commands from the readme as they are written and confirm the server starts successfully?

@Dcallies I can't since it relies on the built image for many of them. So I've only been able to test against a locally built image.

hasher-matcher-actioner/README.md
ThisIsMissEm: Yes, this is correct, this is showing how to mount a directory `config/`  into the container, where that directory would contain a `production.py` file, which would depend on the deployer's requirements. I'm doubtful anyone is deploying in production via just `docker run` though.

There's also no `reference_omm_configs/production_omm_config.py` file as is currently referenced.

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
ThisIsMissEm: This doesn't change working in the devcontainer either, just the final built image, as the devcontainer uses it's own docker file: https://github.com/facebook/ThreatExchange/blob/9e8c36efc11fbd338356c9838b070764560a17eb/hasher-matcher-actioner/.devcontainer/Dockerfile

So for people developing HMA, the devcontainer remains the same, and it'd use the flask configuration from here: https://github.com/facebook/ThreatExchange/blob/44276483382d4112a47b7f33c8778a39d8047440/hasher-matcher-actioner/.devcontainer/startup.sh

Maybe we can check if we're running under gunicorn and only add the logging in development config if that's the case? I'm not sure how well that'd work though.

Or we need a `contributing.py` configuration for using within the devcontainer / flask directly?

hasher-matcher-actioner/README.md
ThisIsMissEm: We've already covered "developing HMA" in the `CONTRIBUTING.md` file, so I don't think the `README.md` needs any further changes here.

hasher-matcher-actioner/README.md
ThisIsMissEm: I've clarified the wording of this section to emphasize that we're mounting a custom configuration file in a directory `config` into the container to use.

hasher-matcher-actioner/Dockerfile
ThisIsMissEm: @Dcallies I can do a follow up PR to rename `reference_omm_configs` to `reference_configs` in the repo.

hasher-matcher-actioner/README.md
```diff
@@ -47,24 +47,39 @@ You can see a complete example using Docker Compose in the provided [docker-comp
...
 
+#### Database Connection
+
+HMA requires configuration for the database connection to postgresql, if you're not using the `docker-compose` approach described above, you will need to set the database URI for HMA to connect to. For the docker commands below, you may need to specify the network on which the postgresql database is available, in which case the commands may look something like the following:
+
+```sh
+docker run --rm --net development_default -p 5100:5100 -e OMM_CONFIG="/reference_configs/development_omm_config.py" -e OMM_DATABASE_URI="postgresql://postgres:@postgresql/hma" ghcr.io/facebook/threatexchange/hma /app/scripts/db-migrate.sh
+```
+
 #### Running the Application in Development
```
Dcallies, line 58: ignorable: We can move some of these sections into contributing, in the future

hasher-matcher-actioner/README.md
```diff
@@ -47,24 +47,39 @@ You can see a complete example using Docker Compose in the provided [docker-comp
...
+```bash
+$ docker run --rm -e OMM_CONFIG='/reference_configs/development_omm_config.py' ghcr.io/facebook/threatexchange/hma /app/scripts/db-migrate.sh
+```
+
+
+Then we can start the server with:
 
 ```bash
-$ docker run -e OMM_CONFIG='/build/reference_omm_configs/development_omm_config.py' -p 5000:5000 ghcr.io/facebook/threatexchange/hma flask --app OpenMediaMatch.app run --host=0.0.0.0
+$ docker run --rm -e OMM_CONFIG='/reference_configs/development_omm_config.py' -p 5100:5100 ghcr.io/facebook/threatexchange/hma
```
Dcallies, line 70: ignorable: I think we should be recommending using the devcontainer for development in general

hasher-matcher-actioner/reference_omm_configs/development_omm_config.py
Dcallies: ignorable: Hmm, didn't realize we had a second copy (which we should bump the base versions of in the future to match your changes here), I wonder if we should delete this one, or point the devcontainer to this one. We might be giving too many options to developers, and just standardize on devcontainer.

Dcallies (approved):
Thank you for updating your test plan! 

CI looks like it's passing, so let's merge this version. In the longer term, I think we might benefit from standardizing on devcontainer for development, which might make things simpler for potential contributors.

hasher-matcher-actioner/README.md
ThisIsMissEm: Yeah, I think the heading here is wrong too, it's really "Running a local version of HMA for developing integrations"

PR #1833, opened 2025-06-24
[HMA] Adding HMA UI documentation and detailed walkthrough of demo functions

julietshen:
Summary
---------

Wrote up detailed documentation of Hasher-Matcher-Actioner UI, including definitions for each demo button, more details on what hash exchanges are supported, and how to navigate the web UI.

Test Plan
---------

These are manually written docs, so did not test besides previewing locally.


doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+There are 4 pages in the app:
+
+## **Dashboard (`/ui/`)**
+
+* View system status  
+* See signal types and content types  
+* Check index statistics
+
+There are 3 demo buttons and an option to do a factory reset:
```
ThisIsMissEm: Might be worth noting that these are only available if the environment is "development", otherwise these do not exist on the page.

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+2. **After Setup:** Button shows "✅ Sample API created" and is disabled  
+3. **Exchange Status:** Shows SEED\_SAMPLE exchange in the exchanges page  
+4. **Index Building:** Triggers automatic background indexing of the sample data  
+5. **Matching Ready:** System now has data to match against
+
+### Technical Implementation:
+
+The StaticSampleSignalExchangeAPI is essentially a hardcoded data source that:
+
+* Returns a fixed set of sample hashes (PDQ and MD5)
```
ThisIsMissEm, line 82: As someone reading this, my question would be: Where do these sample hashes come from? How do I find the images that would trigger matches on these sample hashes?

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+* Demo Workflows: Enable full end-to-end testing without real uploads
+
+The seeded data is stored in a PostgreSQL database that runs inside the dev container:
+
+#### **Database Structure**
+
+* Database: media\_match  
+* Tables: bank, bank\_content, content\_signal  
+* Host: localhost:5432 (inside the dev container)  
+* Credentials: media\_match/hunter2
```
ThisIsMissEm:
```suggestion
* Database: `media_match`  
* Tables: `bank`, `bank_content`, `content_signal`  
* Host: `localhost:5432` (inside the dev container)  
* Credentials: `media_match/hunter2`
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+#### **Data Organization**
+
+* 2 Banks: SEED\_BANK\_0 and SEED\_BANK\_1
```
ThisIsMissEm:
```suggestion
* 2 Banks: `SEED_BANK_0` and `SEED_BANK_1`  
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+* Each Content Item: Has a unique ID and associated signal/hash values
+
+
+### “Setup Tx API"
+
+ The “Setup Tx API” button creates a demo integration with [ThreatExchange](https://developers.facebook.com/docs/threat-exchange/). 
+
+#### **Integration Details**
+
+* **Exchange Name:** TX\_EXAMPLE\_COLLAB
```
ThisIsMissEm:
```suggestion
* **Exchange Name:** `TX_EXAMPLE_COLLAB`
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+#### **Important Caveats**
+
+* **Demo Mode:** Uses test privacy group that may not have live data  
+* **Credentials Needed:** Won't fetch real data without valid API credentials, which users can apply for access for [here](https://developers.facebook.com/docs/threat-exchange/getting-access/).  
+* **Production Setup:** Real deployments need proper app registration with Facebook for Developers.
+
+## **Banks (`/ui/banks`)**
+
+* See your test banks: `SEED_BANK_0` and `SEED_BANK_1`
```
ThisIsMissEm:
```suggestion
* See your content banks, e.g., the `SEED_BANK_0` and `SEED_BANK_1` that are created using the demo buttons.
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+This shows counts of banks, contents, signals, etc.
+
+#### **Direct Database Queries**
+
+You can also connect to PostgreSQL directly to browse the data:
+
+```shell
+# Connect to database  
+psql -h localhost -p 5432 -U media_match -d media_match
```
ThisIsMissEm, line 200:
```suggestion
psql -h localhost -p 5432 -U media_match -d media_match
# Enter password: hunter2
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+The Exchanges page lets users configure external hash exchanges that are API based.
+
+
+### Supported Hash Exchanges
+
+**NCMEC (National Center for Missing & Exploited Children)**
+
+* Child safety hash sharing program  
+* Requires authentication credentials
```
ThisIsMissEm:
```suggestion
* Requires authentication credentials and additional configuration
```

The NCMEC one needs to know which hashsets and stuff too, not just a credential like the ThreatExchange one.

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+* Requires no external network calls or authentication
+
+
+## **"Seed Banks"**
+
+Based on the code analysis and logs, the Seed Banks button:
+
+### Creates Test Banks
+
+* 2 Banks: SEED\_BANK\_0 and SEED\_BANK\_1
```
ThisIsMissEm:
```suggestion
* Creates two Banks: `SEED_BANK_0` and `SEED_BANK_1`  
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+* **Example Implementation:** Shows how signal exchanges are structured  
+* **API Pattern**: Demonstrates the exchange API interface without complexity  
+* **No External Dependencies:** Works completely offline
+
+#### **What You See in the UI:**
+
+1. **Before Setup:** Button shows "Setup Sample API"  
+2. **After Setup:** Button shows "✅ Sample API created" and is disabled  
+3. **Exchange Status:** Shows SEED\_SAMPLE exchange in the exchanges page
```
ThisIsMissEm:
```suggestion
3. **Exchange Status:** Shows `SEED_SAMPLE` exchange in the exchanges page  
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+* Purpose: Provides a mock signal exchange that simulates real threat intelligence feeds without requiring external API credentials  
+* No Authentication Required: Unlike real exchanges (NCMEC, GIFCT, ThreatExchange), it doesn't need API keys or credentials  
+* Instant Setup: Creates sample data immediately for testing the system
+
+##### Sample Data Provider
+
+When you click "Setup Sample API" it:
+
+1. Creates the Exchange: Sets up SEED\_SAMPLE exchange configuration
```
ThisIsMissEm:
```suggestion
1. Creates the Exchange: Sets up `SEED_SAMPLE` exchange configuration  
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+* No Authentication Required: Unlike real exchanges (NCMEC, GIFCT, ThreatExchange), it doesn't need API keys or credentials  
+* Instant Setup: Creates sample data immediately for testing the system
+
+##### Sample Data Provider
+
+When you click "Setup Sample API" it:
+
+1. Creates the Exchange: Sets up SEED\_SAMPLE exchange configuration  
+1. Provides Sample Hashes: Returns a fixed set of example hashes for both PDQ (images) and MD5 (videos)  
+1. Populates the System: Gives you 18 sample records initially (as seen in logs: "fetch\_iter() with 18 new records")
```
ThisIsMissEm:
```suggestion
1. Populates the System: Gives you 18 sample records initially (as seen in logs: `fetch_iter() with 18 new records`)  
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
+# Introduction to Hasher-Matcher-Actioner UI
+
+Open Media Match (aka Hasher-Matcher-Actioner) includes a web app where users can configure their hash banks, connect to hash exchanges like NCMEC, Stop NCII, GIFCT, and ThreatExchange, and debug matches. These features are available as demos to show the features and functionality of HMA.  
+
+There are 4 pages in the app:
+
+## **Dashboard (`/ui/`)**
```
ThisIsMissEm, line 7: Might be worth noting that the UI isn't just for demo, but also for curation, if desired — e.g., you can do curation via the API as well, but the UI will let you do it too.

ThisIsMissEm (commented):
A few comments, mostly formatting related (removing escaping in favour of using inline code formatting.

doc/hma-ui.md
Dcallies: They should be configured in a roughly equivalent manner. ThreatExchange also needs to know which hashsets and stuff.

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
+# Introduction to Hasher-Matcher-Actioner UI
+
+Open Media Match (aka Hasher-Matcher-Actioner) includes a web app where users can configure their hash banks, connect to hash exchanges like NCMEC, Stop NCII, GIFCT, and ThreatExchange, and debug matches. These features are available as demos to show the features and functionality of HMA.
```
Dcallies: ignorable/fine as followup: Consider linking the third party pages for these.

I think we should probably continue to exclusively refer to HMA as HMA. Open Media Match is the name of the library (the name of which I was partial for, but we got pushback on renaming HMA).

```suggestion
Hasher-Matcher-Actioner (aka HMA) includes a web app where users can configure their signal banks, connect to exchanges like NCMEC, StopNCII, GIFCT, ThreatExchange, and others, and debug matches. These features are available as demos to show the features and functionality of HMA.  
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
+# Introduction to Hasher-Matcher-Actioner UI
+
+Open Media Match (aka Hasher-Matcher-Actioner) includes a web app where users can configure their hash banks, connect to hash exchanges like NCMEC, Stop NCII, GIFCT, and ThreatExchange, and debug matches. These features are available as demos to show the features and functionality of HMA.  
+
+There are 4 pages in the app:
```
Dcallies, line 5: ignorable: This page would benefit a lot from images, but I'd prefer not to embed them in repo itself.

The workaround we used in the past was to add images to an issue or the PR, and then deep link to them in the READMEs, but I'm not sure if that's the official way, @ThisIsMissEm  might have some ideas.

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+## **Dashboard (`/ui/`)**
+
```
Dcallies, line 8: Ignorable/Idea: Include a link to the default localhost port hosted version for each of these pages.

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+* **Industry Scale:** Access to millions of real threat content hashes  
+* **Collaborative Defense:** Shared threat intelligence across major platforms  
+* **Live Updates:** Periodic sync with updated threat databases  
+* **Multiple Signal Types:** PDQ, MD5, and other content signatures
+
+#### **Technical Details**
+
+* **Authentication Required:** Needs valid Facebook App credentials for production (more info [here](https://developers.facebook.com/docs/threat-exchange/getting-access/))  
+* **Background Fetching:** Automatic periodic synchronization  
+* **Bank Integration:** Imported data goes into dedicated bank
```
Dcallies: This should be true for all exchanges that implement the interface

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+* **Collaborative Defense:** Shared threat intelligence across major platforms  
+* **Live Updates:** Periodic sync with updated threat databases  
+* **Multiple Signal Types:** PDQ, MD5, and other content signatures
+
+#### **Technical Details**
+
+* **Authentication Required:** Needs valid Facebook App credentials for production (more info [here](https://developers.facebook.com/docs/threat-exchange/getting-access/))  
+* **Background Fetching:** Automatic periodic synchronization  
+* **Bank Integration:** Imported data goes into dedicated bank  
+* **API Limits:** Subject to ThreatExchange rate limiting
```
Dcallies: (Spoilers: The rate limits are high enough that no one will hit them)

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Exchange Configuration
+
+Several of the hash exchanges supported require credentials, provided in the form of a JSON blob:
+
+1. Configure authentication credentials (API keys, certificates)  
+2. Select signal types to synchronize  
+3. Set fetch frequency and retry policies  
+4. Monitor connection health and data flow
+
```
Dcallies, line 294: ignorable: You could add a mention that exchanges can be added to HMA by adding an extension: https://github.com/facebook/ThreatExchange/tree/main/python-threatexchange/threatexchange/extensions

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+**StopNCII**
+
+* Non-consensual intimate image hashes  
+* Privacy-focused hash sharing  
+* Victim-centered approach
+
+### Exchange Configuration
+
+Several of the hash exchanges supported require credentials, provided in the form of a JSON blob:
```
Dcallies: blocking: Link to the pytx repo on further instruction: 

The JSON represents the [CollabConfig](https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/collab_config.py#L12) for the exchange. The keys and data types must match the structure data. For example, here's the exchange config for NMCEC: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/ncmec_api.py#L101-L119

From that, we can see it has two fields:
1. [Required] A field named environment which is a string enum for the API endpoint
2. An optional field only_esp_ids, which accepts a container of ints

So one valid json blob would be
```
{
  'environment': 'https://hashsharing.ncmec.org/npo'
}
```

And one that filled out the optional field would be:
```
{
  'environment': 'https://hashsharing.ncmec.org/npo',
  'only_esp_ids': [1]
}
```

@ThisIsMissEm - Here's the automatic type-correct parsing and documentation reflection attempt I told you about - it could be adapted for a UI form:
https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/cli/config_cmd.py#L95-L101
```suggestion
Several of the exchanges supported require configuration that match their [CollabConfig](https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/collab_config.py). For now, these must be provided in JSON form - consider using the [pytx CLI](https://github.com/facebook/ThreatExchange/tree/main/python-threatexchange#threatexchange-cli) to help with this. Set up a collaboration on the CLI with `threatexchange collab edit` and then print the JSON to paste in with `threatexchange config collab print`
```

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+
+1. Configure authentication credentials (API keys, certificates)
```
Dcallies, line 289: We might need to give a hint on the authentication configuration. This might also be a documentation need for us on the exchanges themselves - we might want to include a future README.md in https://github.com/facebook/ThreatExchange/tree/main/python-threatexchange/threatexchange/exchanges which includes the collaboration config explainer and which credentials are accepted, which hopefully is just copying docs from the code.

doc/hma-ui.md
```diff
@@ -0,0 +1,280 @@
...
+```shell
+curl http://localhost:5000/c/bank/SEED_BANK_0/content/1
+```
+
+## **Exchanges (`/ui/exchanges`)**
+
+The Exchanges page lets users configure external hash exchanges that are API based.
+
+
+### Supported Hash Exchanges
```
Dcallies: blocking: We should link to https://github.com/facebook/ThreatExchange/tree/main/python-threatexchange/threatexchange/extensions

or 

https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges
```suggestion
### Default Signal Exchanges
```

Dcallies (changes requested):
On review - I will approve/merge with just blocking comments addressed and if @ThisIsMissEm gives a thumbs up

Top level thoughts:

# README or Wiki?

We also have the wiki as a place to store information, but I'm fine with this, but check out what is in the wiki to see what you think the source of truth should be: https://github.com/facebook/ThreatExchange/wiki

# Terminology
1. We generally prefer `signal` over `hash`, since there are more than just hashes on several of these databases
1. Similarly `hash exchange` should be either `signal exchange` or just `exchange`
1. Stop NCII -> StopNCII

# Adding images
Even though imagery tends to get out of date, I think it would be worth adding images in a followup - see the inline comment on how to do that without requiring the images be checked into the repo, which I recommend against (as it tends to make fetching the repo much slower). 

# Refer to existing documentation in python-threatexchange
Things like which exchanges are supported, how exactly to credential and configure exchanges are based on the underlying libraries. We should link to them, though we may want to consider additionally improving the documentation there to help explain the credential string, configuration json, etc.

doc/hma-ui.md
ThisIsMissEm: @Dcallies yeah, that's the similar approach I took for the UI that I demoed in the discord, before I abandoned it because it got very complex very quick. Would rather have a json schema to be honest for these.

ThisIsMissEm: I also found in testing that `only_esp_ids` couldn't be missing, but had to be present and contain at least one item. But maybe I was doing something wrong.

doc/hma-ui.md
ThisIsMissEm: Usually it's best to have images for documentation within the repo, since then they can be versioned when the UI or features change, and you can always refer to a specific release's documentation with the correct images.

doc/hma-ui.md
julietshen: Asked David if this is the same test dataset referenced in https://github.com/facebook/ThreatExchange/tree/main/pdq/data

doc/hma-ui.md
julietshen: This is embarrassing but I forgot how to add images. @ThisIsMissEm if you think having them in the repo is best, I can figure it out!

doc/hma-ui.md
julietshen: ooooh. Worth mentioning in the documentation?

doc/hma-ui.md
julietshen: ack! I will add this in.

doc/hma-ui.md
julietshen: fixed in latest commit

julietshen:
sorry in advance for chaotic commiits, but i've addressed most of the feedback (especially the blocking issues)

doc/hma-ui.md
ThisIsMissEm: Yeah, so it's like add an image to `doc/images/name.png` then `alt text` iirc.

You can also just use html `img` tags: `NLNet`

Dcallies: Hmm, @ThisIsMissEm - just for clarity - I think it's possible to version which images are being pointed to without adding them to repo, which increases the size.

Please see this video for this trick: https://www.youtube.com/watch?v=2ajWUp8F694

I think this allows us to version the images without *checking in* the images.

doc/hma-ui.md
Dcallies: I think removing the reference to rate limiting is probably the easiest.

doc/hma-ui.md
Dcallies: More information is in StaticSample: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/exchanges/impl/static_sample.py#L3-L9

And inline in the various signal types: https://github.com/facebook/ThreatExchange/blob/main/python-threatexchange/threatexchange/signal_type/pdq/signal.py#L93-L114

Different SignalTypes might use different sources of samples.

doc/hma-ui.md
ThisIsMissEm: The repo is already really large, a few PNGs isn't gonna make a difference:

```
$ du -sh ThreatExchange
435M	ThreatExchange
```

And that way the images are with the repo when you check it out, and you don't need to rely on a third-party host being available.

ThisIsMissEm: Oh, and that "image in a issue" thing, that's been a security issue in the past, I can't find the report now, but basically it was a vulnerability to have github store files forever that weren't actually referenced anywhere — essentially upload the file, copy the URL and never use the file in anything.

Dcallies: Hey @ThisIsMissEm, make sure you don't include gitignored files when calculating the size. I think you might find that the majority of the size is mypy cache, which are not checked in. The comment about keeping the repo smaller was about initial fetch size. 

We've spent a bit of time on this, and I believe this may be a style discussion, with the functional tradeoff of whether we trust github to not replace uploaded image links with malware or optimize for first fetch download. 

Again, my preference is we optimize for first fetch download, which is to refer to github-hosted images. I understand your preference is different.

Dcallies (approved):
Lets land and iterate

PR #1838, opened 2025-07-31
Get signal from contentid

miriyalamani:
This code is for issue 1761
 I took ideas and suggestions from this closed pr https://github.com/facebook/ThreatExchange/pull/1763.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -143,7 +143,7 @@ def _validate_bank_add_metadata() -> t.Optional[BankedContentMetadata]:
     return t.cast(BankedContentMetadata, metadata)
 
 
-@bp.route("/bank/<bank_name>/content/<content_id>", methods=["GET"])
+@bp.route("/bank/<bank_name>/content/<int:content_id>", methods=["GET"])
 def bank_get_content(bank_name: str, content_id: int):
     storage = persistence.get_storage()
     bank = storage.get_bank(bank_name)
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -19,6 +19,15 @@ import OpenMediaMatch.storage.interface as iface
 from OpenMediaMatch.blueprints import hashing
 
 
+class BankContentResponse(t.TypedDict):
+    id: int
+    disable_until_ts: int
+    collab_metadata: t.Dict[str, t.List[str]]
+    original_media_uri: t.Optional[str]
+    bank: iface.BankConfig
+    signals: t.NotRequired[t.Dict[str, str]]
+
+
 def five_years_from_now() -> int:
     return int(time.mktime((datetime.now() + timedelta(days=365 * 5)).timetuple()))
 
@@ -141,18 +150,34 @@ def bank_get_content(bank_name: str, content_id: int):
     if not bank:
         abort(404, f"bank '{bank_name}' not found")
     include_signals = request.args.get("include_signals", "false").lower() == "true"
-    content = storage.bank_content_get([content_id], include_signals=include_signals)
+    content = storage.bank_content_get([content_id])
     if not content:
         abort(404, f"content '{content_id}' not found")
-    resp = content[0]
-    # If signals were requested and present, add them to the response dict
-    if include_signals and hasattr(resp, "_signals"):
-        resp_dict = resp.__dict__.copy()
-        resp_dict["signals"] = getattr(resp, "_signals")
-        # Remove private/protected keys
-        resp_dict = {k: v for k, v in resp_dict.items() if not k.startswith("_")}
-        return jsonify(resp_dict)
-    return jsonify(resp)
+    content_config = content[0]
+    
+    # If signals were requested, fetch them separately and include in response
+    if include_signals:
+        signals = storage.bank_content_get_signals([content_id])
+        if content_id in signals:
+            response: BankContentResponse = {
+                "id": content_config.id,
+                "disable_until_ts": content_config.disable_until_ts,
+                "collab_metadata": content_config.collab_metadata,
+                "original_media_uri": content_config.original_media_uri,
+                "bank": content_config.bank,
+                "signals": signals[content_id],
+            }
+            return jsonify(response)
+    
+    # Return content without signals
+    response: BankContentResponse = {
+        "id": content_config.id,
+        "disable_until_ts": content_config.disable_until_ts,
+        "collab_metadata": content_config.collab_metadata,
+        "original_media_uri": content_config.original_media_uri,
+        "bank": content_config.bank,
+    }
+    return jsonify(response)
 
 
 @bp.route("/bank/<bank_name>/content", methods=["POST"])
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -155,21 +155,7 @@ def bank_get_content(bank_name: str, content_id: int):
         abort(404, f"content '{content_id}' not found")
     content_config = content[0]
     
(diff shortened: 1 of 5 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -305,10 +305,10 @@ def enabled(self) -> bool:
 @dataclass
 class BankContentConfig:
     """
-    Represents all the signals (hashes) for one piece of content.
+    Represents the configuration for one piece of banked content.
```
Dcallies: I think some combination of both this version and the previous version comment is accurate.

Conceptually, this object represents the collection of signals associated with the content, as well as the configuration.

hasher-matcher-actioner/src/OpenMediaMatch/storage/interface.py
```diff
@@ -305,10 +305,10 @@ def enabled(self) -> bool:
...
 
-    When signals come from external sources, or the original content
-    has been lost
+    Signals are only included in this object when explicitly requested via
+    include_signals parameter in bank_content_get().
```
Dcallies: nit: Misplaced comment, this no longer includes the signals on it, must be fetched separately (which I think is fine)

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
```diff
@@ -525,12 +525,28 @@ def bank_delete(self, name: str) -> None:
...
-        return [
-            b.as_storage_iface_cls()
-            for b in database.db.session.query(database.BankContent)
+        contents = list(
+            database.db.session.query(database.BankContent)
             .filter(database.BankContent.id.in_(ids))
             .all()
-        ]
+        )
+        return [b.as_storage_iface_cls() for b in contents]
```
Dcallies: blocking: This change causes an extra O(n) of work, because you are copying it into a list, and then copying it again. 

Suggest leaving the prior version of the code

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/impl.py
```diff
@@ -525,12 +525,28 @@ def bank_delete(self, name: str) -> None:
...
+            .filter(database.BankContent.id.in_(ids))
+            .options(joinedload(database.BankContent.signals))
+            .all()
+        )
+        
+        signals_dict = {}
+        for b in contents:
+            signals_dict[b.id] = {s.signal_type: s.signal_val for s in b.signals}
+        
+        return signals_dict
```
Dcallies: This looks good to me

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api_banks.py
```diff
@@ -288,3 +288,28 @@ def test_banks_add_hash_index(app: Flask, client: FlaskClient):
     )
     assert post_response.status_code == 200
     assert post_response.json == {"matches": [2]}
+
+
+def test_bank_get_content_with_signals(client: FlaskClient):
+    bank_name = "TEST_BANK_GET_SIGS"
+    create_bank(client, bank_name)
+
+    image_url = "https://github.com/facebook/ThreatExchange/blob/main/pdq/data/bridge-mods/aaa-orig.jpg?raw=true"
```
Dcallies, line 297: nit: This unittest will fail when run without internet access, but it looks like that is what all the other tests are doing, so /shrug

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api_banks.py
```diff
@@ -288,3 +288,28 @@ def test_banks_add_hash_index(app: Flask, client: FlaskClient):
...
+    get_response = client.get(f"/c/bank/{bank_name}/content/{content_id}?include_signals=true")
+    assert get_response.status_code == 200, str(get_response.get_json())
+    assert get_response.json
+    assert get_response.json.get("id") == content_id
+    # The API may not include signals field if no signals are found
+    # This is the current behavior - signals field is only included when signals exist
+    if "signals" in get_response.json:
+        assert isinstance(get_response.json["signals"], dict)
+        # If signals exist, they should not be empty
+        assert len(get_response.json["signals"]) > 0
```
Dcallies: Can you call the method twice, one without `include_signals` and one without, expecting that they are included when called.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -19,6 +19,15 @@
 from OpenMediaMatch.blueprints import hashing
 
 
+class BankContentResponse(t.TypedDict):
+    id: int
+    disable_until_ts: int
+    collab_metadata: t.Dict[str, t.List[str]]
+    original_media_uri: t.Optional[str]
+    bank: iface.BankConfig
+    signals: t.NotRequired[t.Dict[str, str]]
```
Dcallies, line 28: I like your use of NotRequired here!

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -19,6 +19,15 @@
 from OpenMediaMatch.blueprints import hashing
 
 
+class BankContentResponse(t.TypedDict):
+    id: int
+    disable_until_ts: int
+    collab_metadata: t.Dict[str, t.List[str]]
```
Dcallies: nit: you don't have to use the typing version of these
```suggestion
    collab_metadata: dict[str, list[str]]
```

Dcallies (approved):
Thanks @miriyalamani for the help! 

I made some minor suggestions inline, but the logic all looks correct. Do you want to give a shot at fixing the lint errors? (Running formatter, some typing)

Dcallies:
Hey @miriyalamani, do you want to try a stab at fixing the lint failures?

miriyalamani:
@Dcallies  I have addressed the comments and fixed the lint and type errors locally.  Can you please review?

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api_banks.py
ThisIsMissEm: Would it be worth us opening an issue to refactor tests to not use files from URLs but instead files from fixtures, and if we do need to test from URL we start up a local http file server (e.g., `python -m http.server -b 127.0.0.1 4000`) for that file URL?

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api_banks.py
```diff
@@ -288,3 +288,38 @@ def test_banks_add_hash_index(app: Flask, client: FlaskClient):
...
+    )
+    assert get_response.status_code == 200, str(get_response.get_json())
+    assert get_response.json
+    assert get_response.json.get("id") == content_id
+    # The API may not include signals field if no signals are found
+    # This is the current behavior - signals field is only included when signals exist
+    if "signals" in get_response.json:
+        assert isinstance(get_response.json["signals"], dict)
+        # If signals exist, they should not be empty
+        assert len(get_response.json["signals"]) > 0
```
ThisIsMissEm, line 325: Does the `add_response` request add a signal to the bank? What happens if we run the tests multiple times?

miriyalamani: yes it adds a signal to the bank. And the tests reset tables (client fixture depends on app fixture in utils.py ) each time they run, so tests pass on multiple runs.

ThisIsMissEm: So perhaps here the check should be `== 1`?

ThisIsMissEm: And perhaps worth checking that the signal added for this test is the one returning?

Dcallies: @ThisIsMissEm - all the tests use a fixture to isolate their environments (client is a fixture wrapping a server setup)

Issue #1813, opened 2025-04-23
[hma] HMA large memory leaks based on index size


Dcallies:
Reported by an adopter with ~10M hashes in an index - doing suggested prod deployment with curator, matcher, and hasher in different indices. Running with docker deployment, gunicorn with work threads = 1

Seeing memory leaks fairly consistently across all roles:

![Image](https://github.com/user-attachments/assets/e1ecd063-9f84-47cb-a4ff-a7fb14c2e6f0)

Some ideas:
1. Since it's consistent across curator and matchers, it might be that faiss is not truly deallocating when we finalize the old index version
2. Don't know what debugging tools we have available in python
3. If it's a pure python issue, it should be easy to reproduce by seeding a large sample dataset with 10M hashes and adding a small amount of hashes every once in a while to force rebuild
4. If it's a flask issue, should be possible to reproduce by querying locally in a loop

Dcallies:
@juanmrad  - I think you were actively working on this, so marking it to you, let me know if that's not sure.

Dcallies:
New image is released, waiting for feedback from production deployments on whether this fixes the observed issues. Thanks @juanmrad  for all the hard work here!

Dcallies:
@prenner - any news on whether @juanmrad's changes solved the issue?

DiggidyDave:
FYI I'm seeing this also. I am looking into standing up an instance of this, currently I have curator role and worker (hasher+matcher) roles separated,  with 2 decent sized indices for pdq (7.7M) and photodna (2.8M). If I run it under load for an extended period of time I see worker baseline memory climbing. (1 curator, 10 workers)

I ran overnight with about 400 RPS on the match endpoint and, in about a 10 hour window, I see the worker memory increase from 5.4G on all pods, to a range of 8.8-10.5G 10 hours later.  Note the variability, which is likely explained by the fact that I  implemented a refresh jitter on the workers to avoid having them all simultaneous refresh and kill the system periodically, so its workers will refresh at different rates and over 10 hours likely have very different numbers of refreshes.

juanmrad:
@DiggidyDave Thanks for the context on what you are experiencing.

I'm actually taking another look into this right now as I've been doing some local testing with some load :). I think I have some more thoughts on what could be the cause and will create a PR if that is the case. 🤞🏻

juanmrad:
[...]

PR #1839, opened 2025-08-08
[HMA] #1813 Clear out memory on index building to prevent bloating.

juanmrad:
Summary
---------

This is for #1813 I added memory logging around index operations ( Building, Download, Swap ) to monitor usage. I generated 150K hashes to have a big enough DB.

Without the changes here I saw a jump between reloads of 50-80MB while only freeing 5-15MB in between the process. 

Start: 
102MB → 150MB → 265MB → 285MB → 346MB → 378MB → 450MB
In a span of 10 minutes. 

PDQ loads: 48MB → 77MB → 76MB

```
CachedIndex[pdq] Index load: +48.21 MB
CachedIndex[pdq] Index load: +77.24 MB  
CachedIndex[pdq] Index load: +75.99 MB

[2025-08-08 03:57:04,926] INFO in matching: CachedIndex[video_md5] Periodic task memory: start=285.62 MB, storage=285.62 MB, end=298.36 MB (change: +12.74 MB)
[2025-08-08 03:57:05,040] INFO in matching: CachedIndex[pdq] Index load: 285.62 MB -> 362.86 MB (+77.24 MB)
[2025-08-08 03:57:35,041] INFO in matching: CachedIndex[pdq] Index load: 377.89 MB -> 453.88 MB (+75.99 MB)
```

After adding explicit GC I saw improvements, but the memory wasn't being freed between index swaps as expected. I added malloc_trim as a way to force C to free up pages. 

> malloc_trim(0) asks the C allocator to return currently unused heap pages back to the OS.

After the changes I saw logging as follows:
```
[2025-08-08 04:05:39,166] INFO in matching: CachedIndex[pdq] Index load: 216.17 MB -> 332.91 MB (+116.74 MB)
[2025-08-08 04:05:39,166] INFO in matching: CachedIndex[pdq] Replacing index type: PDQIndex
[2025-08-08 04:05:39,210] INFO in matching: CachedIndex[pdq] Called malloc_trim(0) to release freed memory
[2025-08-08 04:05:39,210] INFO in matching: CachedIndex[pdq] Cleanup complete: 215.97 MB (freed: 116.94 MB)

[2025-08-08 04:06:39,020] INFO in matching: CachedIndex[video_md5] Index load: 249.70 MB -> 261.15 MB (+11.45 MB)
[2025-08-08 04:06:39,020] INFO in matching: CachedIndex[video_md5] Replacing index type: TrivialSignalTypeIndex
[2025-08-08 04:06:39,042] INFO in matching: CachedIndex[video_md5] Called malloc_trim(0) to release freed memory
[2025-08-08 04:06:39,043] INFO in matching: CachedIndex[video_md5] Cleanup complete: 246.47 MB (freed: 14.68 MB)
```

Memory jumps are expected as we do the swap, but it being freed right after.

Test Plan
---------

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
@@ -2,7 +2,9 @@
 
 import logging
 import time
+import gc
 import typing as t
+from typing import Optional
 
 from threatexchange.signal_type.signal_base import SignalType
 
@@ -58,6 +60,8 @@ def build_index(
     Build one index from scratch with the current bank contents and persist it.
     """
     start = time.time()
+    logger.info(f"Starting index build for {for_signal_type.get_name()}")
+
     # First check to see if new signals have appeared since the last build
     idx_checkpoint = index_store.get_last_index_build_checkpoint(for_signal_type)
     bank_checkpoint = bank_store.get_current_index_build_target(for_signal_type)
@@ -69,24 +73,57 @@ def build_index(
         for_signal_type.get_name(),
         0 if bank_checkpoint is None else bank_checkpoint.total_hash_count,
     )
-    index_cls = for_signal_type.get_index_cls()
+    
+    # Use try/finally to ensure cleanup happens even on exceptions
     signal_list = []
+    built_index: Optional[t.Any] = None
     last_cs = None
-    for last_cs in bank_store.bank_yield_content(for_signal_type):
-        tuple = (last_cs.signal_val, last_cs.bank_content_id)
-        signal_list.append(tuple)
-    built_index = index_cls.build(signal_list)
-    checkpoint = SignalTypeIndexBuildCheckpoint.get_empty()
-    if last_cs is not None:
-        checkpoint = SignalTypeIndexBuildCheckpoint(
-            last_item_timestamp=last_cs.bank_content_timestamp,
-            last_item_id=last_cs.bank_content_id,
-            total_hash_count=len(signal_list),
-        )
+    signal_count = 0
+    
+    try:
+        # Collect signals
+        for last_cs in bank_store.bank_yield_content(for_signal_type):
+            signal_list.append((last_cs.signal_val, last_cs.bank_content_id))
+            signal_count += 1
+        
+        # Build index
+        index_cls = for_signal_type.get_index_cls()
+        built_index = index_cls.build(signal_list)
+        
+        # Clear signal_list early to reduce memory peak during storage
+        signal_list.clear()
+        
+        # Create checkpoint
+        checkpoint = SignalTypeIndexBuildCheckpoint.get_empty()
+        if last_cs is not None:
+            checkpoint = SignalTypeIndexBuildCheckpoint(
+                last_item_timestamp=last_cs.bank_content_timestamp,
+                last_item_id=last_cs.bank_content_id,
+                total_hash_count=signal_count,
+            )
+            
+        # Store the index
+        if built_index is not None:
+            index_store.store_signal_type_index(for_signal_type, built_index, checkpoint)
+        
+    finally:
+        # Guaranteed cleanup even if exceptions occur
+        # Explicitly clear large objects to help with memory management  
+        if 'signal_list' in locals():
+            signal_list.clear()
+            del signal_list
+        
+        # Clear reference to built_index after storage
+        if 'built_index' in locals() and built_index is not None:
+            built_index = None
+        
+        # Force garbage collection to reclaim memory
+        gc.collect()
+    
     logger.info(
         "Indexed %d signals for %s - %s",
-        len(signal_list),
+        signal_count,
         for_signal_type.get_name(),
         duration_to_human_str(int(time.time() - start)),
     )
-    index_store.store_signal_type_index(for_signal_type, built_index, checkpoint)
+
--- a/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
@@ -17,6 +17,7 @@ from OpenMediaMatch.storage.interface import (
     SignalTypeIndexBuildCheckpoint,
 )
 from OpenMediaMatch.utils.time_utils import duration_to_human_str
+from OpenMediaMatch.utils.memory_utils import trim_process_memory
 
 logger = logging.getLogger(__name__)
 
@@ -117,8 +118,8 @@ def build_index(
         if 'built_index' in locals() and built_index is not None:
             built_index = None
         
-        # Force garbage collection to reclaim memory
-        gc.collect()
+        # Force garbage collection to reclaim memory and attempt to free pages
+        trim_process_memory(logger, "Indexer")
     
     logger.info(
         "Indexed %d signals for %s - %s",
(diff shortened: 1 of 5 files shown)


ThisIsMissEm:
Would it make sense to add a reset method to SignalIndexType? So then we just call that which does the right thing to reset the index, as implemented by the index type? That'd avoid leaking implementation details about faiss into HMA, I think?

ThisIsMissEm:
@juanmrad based on my comment on friday morning, I think this might give a cleaner abstraction for resetting indexes and cleaning up memory used by them: https://github.com/facebook/ThreatExchange/pull/1840

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -148,7 +148,8 @@ class BankContent(db.Model):  # type: ignore[name-defined]
```
Dcallies: Did we change black settings?

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -456,6 +462,11 @@ def commit_signal_index(
         self.serialized_index_large_object_oid = l_obj.oid
         db.session.add(self)
         raw_conn.commit()
+        # Ensure raw connection is closed to avoid resource leaks
+        try:
+            raw_conn.close()
+        except Exception:
+            pass
```
Dcallies: Hmm, my impression is that flask was handling this, and it was still in use here.

If you want to make this less error prone, is there a place we can add `with` or the `closing` helper?

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -69,24 +74,58 @@ def build_index(
...
+        # Store the index
+        if built_index is not None:
+            index_store.store_signal_type_index(
+                for_signal_type, built_index, checkpoint
+            )
+
+    finally:
+        # Guaranteed cleanup even if exceptions occur
+        # Explicitly clear large objects to help with memory management
+        if "signal_list" in locals():
```
Dcallies: Wont this always be true?

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -69,24 +74,58 @@ def build_index(
...
+            signal_list.clear()
+            del signal_list
+
+        # Clear reference to built_index after storage
+        if "built_index" in locals() and built_index is not None:
```
Dcallies: nit: built_index should always be in locals

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -90,8 +91,28 @@ def reload_if_needed(self, store: interface.IUnifiedStore) -> None:
...
+                ):
+                    try:
+                        faiss_idx = old_index._index.faiss_index
+                        if hasattr(faiss_idx, "reset"):
+                            faiss_idx.reset()
+                    except Exception:
+                        pass
+
+                del old_index
+                trim_process_memory()
```
Dcallies: Agree with @ThisIsMissEm that we should standardize this on the interface with either a `reset` or `dispose` functionality. See the note here about delete: https://docs.python.org/3/reference/datamodel.html#object.__del__

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -90,8 +91,28 @@ def reload_if_needed(self, store: interface.IUnifiedStore) -> None:
...
+                    old_index._index, "faiss_index"
+                ):
+                    try:
+                        faiss_idx = old_index._index.faiss_index
+                        if hasattr(faiss_idx, "reset"):
+                            faiss_idx.reset()
+                    except Exception:
+                        pass
+
+                del old_index
```
Dcallies: It might be possible other threads are finishing a request that still holds a reference to this, in which case this might have strange behavior

hasher-matcher-actioner/src/OpenMediaMatch/utils/memory_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+        libc = ctypes.CDLL("libc.so.6")
+        if hasattr(libc, "malloc_trim"):
+            libc.malloc_trim(0)
+            if logger:
+                logger.info("%s Called malloc_trim(0) to release freed memory", label)
+            return True
+        if logger:
+            logger.debug("%s trim_process_memory: libc has no malloc_trim", label)
+    except Exception as exc:  # pragma: no cover — platform specific
```
Dcallies: Catch the specific exception instead of all

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -69,24 +74,58 @@ def build_index(
...
+            index_store.store_signal_type_index(
+                for_signal_type, built_index, checkpoint
+            )
+
+    finally:
+        # Guaranteed cleanup even if exceptions occur
+        # Explicitly clear large objects to help with memory management
+        if "signal_list" in locals():
+            signal_list.clear()
+            del signal_list
```
Dcallies: After clearing it, there's only a single empty lists worth of memory left, probably safe to leave

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
ThisIsMissEm: This could be a case of black changing under us, as iirc the versions of tools we use aren't pinned, this already caused test failures in python-threatexchange

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -69,24 +74,58 @@ def build_index(
...
+
+        # Clear reference to built_index after storage
+        if "built_index" in locals() and built_index is not None:
+            built_index = None
+
+        # Force garbage collection to reclaim memory and attempt to free pages
```
Dcallies: Ignorable: We can restructure this code to make it easier for normal python refcounting to naturally decrement to zero - if we move the body of the try into its own method, then wrap that with this try finally, I believe we can just call your trim memory method without manually manipulating the collections.

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -69,24 +74,58 @@ def build_index(
...
+        for last_cs in bank_store.bank_yield_content(for_signal_type):
+            signal_list.append((last_cs.signal_val, last_cs.bank_content_id))
+            signal_count += 1
+
+        # Build index
+        index_cls = for_signal_type.get_index_cls()
+        built_index = index_cls.build(signal_list)
+
+        # Clear signal_list early to reduce memory peak during storage
+        signal_list.clear()
```
Dcallies: Will it, without a garbage collect call?

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -69,24 +72,56 @@ def build_index(
         for_signal_type.get_name(),
         0 if bank_checkpoint is None else bank_checkpoint.total_hash_count,
     )
-    index_cls = for_signal_type.get_index_cls()
-    signal_list = []
+
+    # Use try/finally to ensure aggressive memory trim after build
+    signal_count = 0
+    built_index: t.Any | None = None  # keep in locals per review nit
```
Dcallies, line 78: touche comment!

PR #1840, opened 2025-08-09
Add reset method to SignalTypeIndex and all downstream index classes

ThisIsMissEm:
Summary
---------

This would hopefully provide a cleaner way to "reset" an index when we don't need it anymore, based on the ideas in https://github.com/facebook/ThreatExchange/pull/1839/files#diff-1322c39b70910bd2d549a2b05510a0502cd01be14a17be592bede1f48795dc66R101

So instead of:
```python
            old_index = self.index

            self.index = new_index
            self.checkpoint = curr_checkpoint

            if old_index is not None:
                # Try to explicitly cleanup FAISS internals
                if hasattr(old_index, "_index") and hasattr(
                    old_index._index, "faiss_index"
                ):
                    try:
                        faiss_idx = old_index._index.faiss_index
                        if hasattr(faiss_idx, "reset"):
                            faiss_idx.reset()
                    except Exception:
                        pass

                del old_index
                trim_process_memory()
```

We would have:

```python
            old_index = self.index

            self.index = new_index
            self.checkpoint = curr_checkpoint

            if old_index is not None:
                old_index.reset()
                del old_index
                trim_process_memory()
```

Test Plan
---------

I've added various tests for the reset method in each modified index.

### DIFF
--- a/python-threatexchange/pyproject.toml
+++ b/python-threatexchange/pyproject.toml
@@ -42,7 +42,7 @@ extensions = [
     "pytesseract",
 ]
 dev = [
-    "pytest",
+    "pytest == 8.3.5",
     "black",
     "mypy",
     "types-python-dateutil",
--- a/python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
+++ b/python-threatexchange/threatexchange/extensions/vpdq/tests/test_vpdq_faiss.py
@@ -96,6 +96,19 @@ def test_simple():
     assert res[0] == IndexMatch(VPDQSimilarityInfo(100.0, 100.0), EXAMPLE_META_DATA)
 
 
+def test_reset():
+    index = VPDQIndex.build([[HASH, EXAMPLE_META_DATA]])
+    assert index._entry_idx_to_features_and_entries[0][0] == FEATURES
+    assert len(index._index_idx_to_vpdqHex_and_entry) == len(FEATURES)
+
+    index.reset()
+
+    assert index.index.faiss_index is not None
+    assert index.index.faiss_index.ntotal == 0
+    assert len(index._entry_idx_to_features_and_entries) == 0
+    assert len(index._index_idx_to_vpdqHex_and_entry) == 0
+
+
 def test_half_match():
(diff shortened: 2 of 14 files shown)


ThisIsMissEm:
@Dcallies are the tests for python-threatexchange working on main currently? I'm getting failures in files unrelated to my changes both in CI and locally.

ThisIsMissEm:
@Dcallies worked it out, pytest 8.3.5 was last used on python-threatexchange CI, since then, pytest 8.4.x was released, and this introduced errors for: `pytest.PytestWarning: matching against an empty string will *always* pass. [...]`

I would guess we'd want to pin pytest to 8.3.5 until we can do the work to upgrade to 8.4.x, I don't really understand the changes in detail, but the changelog for 8.4.0 is here: https://docs.pytest.org/en/stable/changelog.html#pytest-8-4-0-2025-06-02

Dcallies (commented):
Are we certain that this is what's causing the problem, or is this an attempt to speed up effective normal garbage collection?

Reading #1839, it looks like Juan's test plan shows that gc and malloc free appear to solve the memory issues. Why should we add this treatment in addition, given that clearing an index is not an operation that will be used in production?

You mention it is helpful in the summary here, and propose this solution, but I'm missing the problem we're looking to solve.

Dcallies:
> we also wouldn't have this problem if reloading/fetching updates to the index didn't involve pickle creating an entirely new index structure in memory

We have had this discussion offline already, and I still believe you are ignoring some of the reasons why we have chosen to do so, but I you've made clear your position.

ThisIsMissEm:
> > we also wouldn't have this problem if reloading/fetching updates to the index didn't involve pickle creating an entirely new index structure in memory
> 
> We have had this discussion offline already, and I still believe you are ignoring some of the reasons why we have chosen to do so, but I you've made clear your position.

@Dcallies yes, so I think we need a method to explicitly free up and dispose of an index, given that. Unless we were to change that behavior.

ThisIsMissEm:
@juanmrad would your latest PR (#1847 ) remove the need for this PR?

juanmrad:
@ThisIsMissEm I believe so.

PR #1843, opened 2025-08-13
Fix crash due to missing MAX_CONTENT_LENGTH configuration value

### LINKED ISSUE #1842: h/hash fails if MAX_CONTENT_LENGTH is not specified in config
breaking change when upgrading to 1.0.18

there was a new config value added here: https://github.com/facebook/ThreatExchange/pull/1822/

if you do not set that config value before deploying, 

```
max_content_length = int(
  current_app.config.get("MAX_CONTENT_LENGTH", DEFAULT_MAX_CONTENT_LENGTH)
)
```

will raise an error. MAX_CONTENT_LENGTH is defined as a default flask config value so it uses None instead of the fallback used when calling `get` 

```
current_app.config.get("MAX_CONTENT_LENGTH", 20) # current code - returns None unless you update the omm config first
current_app.config.get("MAX_HMA_CONTENT_LENGTH", 20) # this returns 20 because MAX_HMA_CONTENT_LENGTH is not part of the default flask config
```

to reproduce:
- remove MAX_CONTENT_LENGTH from the development omm config settings
- try to hash an image by hitting the h/hash endpoint
- should see an error

ThisIsMissEm:
Summary
---------

This fixes #1842, which was potentially the case if you were using existing configuration files from before #1822 was merged, where a new configuration variable of `MAX_CONTENT_LENGTH` was introduced.

Test Plan
---------

Have added additional pytest test cases to catch this misconfiguration.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
@@ -124,9 +124,18 @@ def hash_media() -> dict[str, str]:
 
     try:
         # Get response with content length tracking
-        max_content_length = int(
-            current_app.config.get("MAX_CONTENT_LENGTH", DEFAULT_MAX_CONTENT_LENGTH)
-        )
+        max_content_length = current_app.config.get("MAX_CONTENT_LENGTH")
+        if max_content_length is None:
+            max_content_length = DEFAULT_MAX_CONTENT_LENGTH
+        
+        # cast to integer if necessary (the value could have come from an environment variable)
+        if isinstance(max_content_length, str):
+            if not max_content_length.isdigit():
+                logger.error(f"MAX_CONTENT_LENGTH misconfigured, expected integer, received: {max_content_length}")
+                abort(500, "Service misconfigured, see logs for details")
+
+            max_content_length = int(max_content_length)
+
         with _check_content_length_stream_response(
             media_url, max_content_length
         ) as download_resp:
--- a/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_hashing.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_hashing.py
@@ -155,3 +155,35 @@ def test_content_length_validation(mock_get, mock_head, client):
     # Test with acceptable content length
     response = client.get("/h/hash?url=https://example.com/image.jpg")
     assert response.status_code != 413
+
+@patch("requests.head")
+def test_content_length_validation_default(mock_head, client, app):
+    with app.app_context():
+        # Override MAX_CONTENT_LENGTH to None, simulating it not 
+        # being present in an OMM_CONFIG file:
+        app.config["MAX_CONTENT_LENGTH"] = None
+
+        # Mock successful HEAD response with large content length
+        mock_head_resp = Mock()
+        mock_head_resp.headers = {"content-length": str(DEFAULT_MAX_CONTENT_LENGTH + 1)}
+        mock_head_resp.raise_for_status = (
+            Mock()
+        )  # Add this to prevent the raise_for_status() call from failing
+        mock_head.return_value = mock_head_resp
+
+        # Test with large content length
+        response = client.get("/h/hash?url=https://example.com/image.jpg")
+        assert response.status_code == 413
+        assert "Content too large" in response.get_data(as_text=True)
+
+@patch("requests.head")
+def test_content_length_validation_misconfiguration(client, app):
+    with app.app_context():
+        # Override MAX_CONTENT_LENGTH to a non-integer value, simulating it 
+        # being set from the environment variables incorrectly:
+        app.config["MAX_CONTENT_LENGTH"] = "not-an-integer"
+
+        # Test with large content length
+        response = client.get("/h/hash?url=https://example.com/image.jpg")
+        assert response.status_code == 500
+        assert "Service misconfigured, see logs for details" in response.get_data(as_text=True)


ThisIsMissEm:
@juanmrad could you give this a review? thanks!

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -124,9 +124,18 @@ def hash_media() -> dict[str, str]:
 
     try:
         # Get response with content length tracking
-        max_content_length = int(
-            current_app.config.get("MAX_CONTENT_LENGTH", DEFAULT_MAX_CONTENT_LENGTH)
-        )
+        max_content_length = current_app.config.get("MAX_CONTENT_LENGTH")
```
prenner: only other option would be to change the setting const - that would simplify a lot of this since we'd fall back to the `get` 

that might be a breaking change for anyone who is using it, though :(

ThisIsMissEm: @prenner hmm, I'm not sure I understand? (I'm still getting familiar with pythonisms)

ThisIsMissEm: Or is `MAX_CONTENT_LENGTH` used by Flask somewhere, and that's why it's defaulting to None?

ThisIsMissEm: oooh, I see it now: https://flask.palletsprojects.com/en/stable/patterns/fileuploads/#improving-uploads

Yeah, we should probably change the setting then.

prenner: yeah, MAX_CONTENT_LENGTH is already a config used by Flask - https://flask.palletsprojects.com/en/stable/config/#MAX_CONTENT_LENGTH

if we use something else, like MAX_CONTENT_LENGTH_HMA, the `.get` default would work as expected

```
default_flask_config = {"MAX_CONTENT_LENGTH": None}
default_flask_config.get("MAX_CONTENT_LENGTH", 5)  # returns None
```

ThisIsMissEm: @juanmrad maybe we could change the setting to `MAX_REMOTE_FILE_SIZE` or something, since `MAX_CONTENT_LENGTH` already is used by Flask and modifying it may have unintended side-effects? (unless it was deliberately chosen because it also would control POST content-length on all requests?)

juanmrad: Let's update the name to be HMA specific. It is the easiest forward fix here and the intended effect is that we use the default 100MB or whatever is set on the config file and not mess with flask internals.

ThisIsMissEm: Okay, I'll update to reflect that. Wasn't sure if you'd chosen `MAX_CONTENT_LENGTH` to give parity with the POST multipart endpoint or something.

Will be a "breaking" change, but should be fine.

ThisIsMissEm:
@juanmrad okay, kept most of the changes I had, but renamed `MAX_CONTENT_LENGTH` to `MAX_REMOTE_FILE_SIZE`

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -124,11 +124,22 @@ def hash_media() -> dict[str, str]:
 
     try:
         # Get response with content length tracking
-        max_content_length = int(
-            current_app.config.get("MAX_CONTENT_LENGTH", DEFAULT_MAX_CONTENT_LENGTH)
-        )
+        max_file_size = current_app.config.get("MAX_REMOTE_FILE_SIZE")
+        if max_file_size is None:
+            max_file_size = DEFAULT_MAX_REMOTE_FILE_SIZE
```
ThisIsMissEm: We could maybe write this as:
```python
         max_file_size = current_app.config.get("MAX_REMOTE_FILE_SIZE", DEFAULT_MAX_REMOTE_FILE_SIZE)
```

But I think I like the explicit None check in case the configuration you actually have is: `MAX_REMOTE_FILE_SIZE=None`?

juanmrad (approved):

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
juanmrad: Looking back I do wonder if we should just keep using the `MAX_CONTENT_LENGTH` over a new variable. I would guess expected behavior is that flasks variable should persist to all bodies, and if there is a function that behaves different from that behavior it would be odd. 

We could add the explicit None check tho. as seems like the get doesn't cover that. I'm open to keep it with this new variable name tho.

juanmrad:
@ThisIsMissEm we are missing changing the name on the reference omm.

ThisIsMissEm:
> @ThisIsMissEm we are missing changing the name on the reference omm. `hasher-matcher-actioner/reference_omm_configs/development_omm_config.py`

So we are! I could've sworn I'd changed it there but mustn't have, will fix tomorrow

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
ThisIsMissEm: I think it's debatable, I think I like the explicit new variable, most.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -124,11 +124,22 @@ def hash_media() -> dict[str, str]:
...
+
+        # cast to integer if necessary (the value could have come from an environment variable)
+        if isinstance(max_file_size, str):
+            if not max_file_size.isdigit():
+                logger.error(
+                    f"MAX_REMOTE_FILE_SIZE misconfigured, expected integer, received: {max_file_size}"
+                )
+                abort(500, "Service misconfigured, see logs for details")
+
+            max_file_size = int(max_file_size)
```
ThisIsMissEm, line 140: Interestingly, this wouldn't be necessary with fastapi, where you have a "configuration class" that does all the parsing and validation of configuration before starting the app.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
Dcallies: ignorable, alt:
```suggestion
        max_file_size = current_app.config.get("MAX_REMOTE_FILE_SIZE") or DEFAULT_MAX_REMOTE_FILE_SIZE
```

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -124,11 +124,22 @@ def hash_media() -> dict[str, str]:
...
+        if max_file_size is None:
+            max_file_size = DEFAULT_MAX_REMOTE_FILE_SIZE
+
+        # cast to integer if necessary (the value could have come from an environment variable)
+        if isinstance(max_file_size, str):
+            if not max_file_size.isdigit():
+                logger.error(
+                    f"MAX_REMOTE_FILE_SIZE misconfigured, expected integer, received: {max_file_size}"
+                )
+                abort(500, "Service misconfigured, see logs for details")
```
Dcallies, line 138:
```
try:
  `int(max_file_size)`
catch ValueError
```

Is possibly more durable. I'd recommend sending a more explicit log to the abort - it's only creating an indirection for the user, best to present them the answer

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
Dcallies: Unless you are personally signing up for a month of work to migrate, I think we should discard that as a viable route for now.

Dcallies:
@ThisIsMissEm - following up, would you like me to merge?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
ThisIsMissEm: Yeah, the comment was more meant to say that "maybe there's a way to do similar configuration loading with Flask, where environment variables / config are validated as they're loaded to be a specific type.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
ThisIsMissEm: So I don't think we should expose to the API what the misconfiguration is. Ideally we should probably do this check on application startup, rather than in the request/response lifecycle.

ThisIsMissEm:
@Dcallies yes, we can merge.

PR #1847, opened 2025-08-21
[PyTx] Optimization to Serialize/Deserialize and interface to clean data.

juanmrad:
Summary
---------

This PR attempt to help reduction of peak memory by making some changes on how we serialize/deserialize data and makes index cleanup explicit and reliable:

- Switches index serialization/deserialization to streamed pickle APIs.
- Introduces a `dispose()` hook on indices and implements it for FAISS-backed PDQ indexes.
- Adds a weakref-based finalizer in the FAISS wrapper as a safety net so resources are cleaned up on GC/exit.

Why pickle changes:

Before: `pickle.dumps()`/`pickle.loads(fin.read())` created a duplicate of the entire index in memory, temporarily inflating RSS by the size of the serialized blob.

Now: `pickle.dump()`/`pickle.load()` stream directly to/from the file, lowering peak memory and reducing allocator fragmentation during load/store.

Test Plan
---------

### DIFF
--- a/python-threatexchange/threatexchange/signal_type/index.py
+++ b/python-threatexchange/threatexchange/signal_type/index.py
@@ -239,9 +239,17 @@ class SignalTypeIndex(t.Generic[T]):
 
         Could also be premature optimization, you decide!
         """
-        fout.write(pickle.dumps(self))
+        pickle.dump(self, fout, protocol=pickle.HIGHEST_PROTOCOL)
 
     @classmethod
     def deserialize(cls: t.Type[Self], fin: t.BinaryIO) -> Self:
         """Instantiate an index from a previous call to serialize"""
-        return pickle.loads(fin.read())
+        return pickle.load(fin)
+
+    def dispose(self) -> None:
+        """
+        Release any native resources or large auxiliary structures held by the
+        index implementation. Default is a no-op; implementations that wrap
+        native libraries (e.g., FAISS) should override.
+        """
+        return None
--- a/python-threatexchange/threatexchange/signal_type/pdq/pdq_faiss_matcher.py
+++ b/python-threatexchange/threatexchange/signal_type/pdq/pdq_faiss_matcher.py
@@ -1,6 +1,7 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 
 import typing as t
+import weakref
 import faiss
 import binascii
 import numpy
@@ -32,6 +33,11 @@ class PDQHashIndex(ABC):
     def __init__(self, faiss_index: faiss.IndexBinary) -> None:
         self.faiss_index = faiss_index
         super().__init__()
+        # Ensure native resources are released when the wrapper is GC'd, even if
+        # callers forget to dispose explicitly. Do not capture `self`.
+        self._finalizer = weakref.finalize(
+            self, PDQHashIndex._finalize_faiss, self.faiss_index
+        )
 
     @abstractmethod
     def hash_at(self, idx: int) -> str:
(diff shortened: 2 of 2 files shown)


python-threatexchange/threatexchange/extensions/vpdq/vpdq_faiss.py
```diff
@@ -83,3 +87,24 @@ def __getstate__(self):
...
+            pass
+
+    @staticmethod
+    def _finalize_faiss(faiss_index: faiss.Index) -> None:
+        try:
+            reset_fn = getattr(faiss_index, "reset", None)
+            if callable(reset_fn):
+                reset_fn()
+        except Exception:
+            pass
```
Dcallies, line 110: We talked about this in the last PR, but are we sure this is doing anything?

This logic looks fairly fragile, and my previous impression is that faiss memory cleanup after garbage collection is working without the need to call it directly.

python-threatexchange/threatexchange/signal_type/index.py
```diff
@@ -239,9 +239,17 @@ def serialize(self, fout: t.BinaryIO) -> None:
 
         Could also be premature optimization, you decide!
         """
-        fout.write(pickle.dumps(self))
+        pickle.dump(self, fout, protocol=pickle.HIGHEST_PROTOCOL)
```
Dcallies, line 242: 🤦 not sure why I wrote it that way in the first place

Dcallies (changes requested):
Changes to default serialize/deserialize seem uncontroversial, and I think are fine to merge as long as tests pass. 

However, I still have the same questions about whether we need to manually manage faiss data - I had thought in your previous PR we had shown that even with large indices, garbage collection and malloc were enough. Here we're adding manual management back in, are we certain that those are doing anything?

blocking q: How did you test this?

python-threatexchange/threatexchange/extensions/vpdq/vpdq_faiss.py
juanmrad: @Dcallies I created this Gist to provide more context as there is a lot to explain for one pr: https://gist.github.com/juanmrad/f60f5cdb4da838945d20ca5dbe60cb43

Dcallies: Perfect, add that to the test plan. I'd love to bother the FAISS people about this later, but let's get a version of this merged.

python-threatexchange/threatexchange/signal_type/pdq/pdq_faiss_matcher.py
```diff
@@ -148,6 +154,27 @@ def __getstate__(self):
...
+            self, PDQHashIndex._finalize_faiss, self.faiss_index
+        )
+
+    def dispose(self) -> None:
+        try:
+            reset_fn = getattr(self.faiss_index, "reset", None)
+            if callable(reset_fn):
+                reset_fn()
+        except Exception:
+            pass
```
Dcallies, line 168: Catching all exceptions unconditionally is a sure way to hide bugs, both here and in other sites. 

Is there a specific exception you are guarding against? If not, not catching the exception, or catching and logging with `logging` might be preferable.

python-threatexchange/threatexchange/cli/tests/e2e_test_helper.py
```diff
@@ -71,7 +71,7 @@ def assert_cli_output(
             assert lines[line] == expected_line_output
 
     def assert_cli_usage_error(
-        self, args: t.Iterable[str], msg_regex: str = ""
+        self, args: t.Iterable[str], msg_regex: t.Optional[str] = None
```
Dcallies, line 74: Orphaned change?

PR #1848, opened 2025-08-21
[HMA] Make Schedules configurable

juanmrad:
Summary
---------

Make schedules configurable for the background task as to help adopters adjust how often to refresh the index and fetch it. 

Test Plan
---------

Screenshot 2025-08-21 at 6 55 34 PM

Added a log line to check :)


hasher-matcher-actioner/reference_omm_configs/json_log_omm_config.py
```diff
@@ -41,6 +41,11 @@
 TASK_FETCHER = True
 TASK_INDEXER = True
 TASK_INDEX_CACHE = True
+# Optional: configure background task intervals (in seconds)
+# Defaults: fetcher=240, indexer=60, index cache refresh=30
+TASK_FETCHER_INTERVAL_SECONDS = 60 * 4
+TASK_INDEXER_INTERVAL_SECONDS = 60
+TASK_INDEX_CACHE_INTERVAL_SECONDS = 30
```
Dcallies, line 48: ignorable: Should we consider merging these two sets of configs?

Dcallies (approved):
Apologies, thought I approved!

Can you confirm that you ran the server and changed the settings, and saw they were respected?

hasher-matcher-actioner/reference_omm_configs/json_log_omm_config.py
juanmrad: I wasn't fully sure I thought of it maybe being the same, but also given how each env may be different there may be a case where you want to space out the start of tasks?

PR #1855, opened 2025-08-27
[HMA] Bypass Enabled Ratio for UI

### LINKED ISSUE #1844: HMA: Find Content in Banks in UI doesn't always find content
I was trying out the `Matching Enabled Ratio` value on a given bank, and when using the form at the top of the page "Find Content in Banks", this `Matching Enabled Ratio` ends up affecting results, so they're not deterministic in the UI.

I think we should probably bypass `Matching Enabled Ratio` in the UI, such that you can always find which bank includes which content, even though the API for matching may not always return a match.

juanmrad:
Summary
---------

Fixes #1844

This PR Allows for the front-end to bypass the enabled ratio on queries to match content as well as add the functionality to the matching lookup as an optional arg. 

This is useful to check what the content may match on different banks regardless of the enabled_ratio while also giving a visual of what is the current enabled value for the banks. This is an optional setting on the UI as to keep the same behavior.

Test Plan
---------
Tested locally adding the same image to multiple banks and then changing their enabled ratio. I then requested the match bypassing enabled ratio and all got returned and then did a couple without the bypass to see the coinflip still working.

Before:

Screenshot 2025-08-26 at 9 22 00 PM

After:

Screenshot 2025-08-26 at 9 37 08 PM
Screenshot 2025-08-26 at 9 37 01 PM
Screenshot 2025-08-26 at 9 36 39 PM

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -79,8 +79,7 @@ class _SignalIndexInMemoryCache:
     def reload_if_needed(self, store: interface.IUnifiedStore) -> None:
         now = time.time()
         # There's a race condition here, but it's unclear if we should solve it
-        curr_checkpoint = store.get_last_index_build_checkpoint(
-            self.signal_type)
+        curr_checkpoint = store.get_last_index_build_checkpoint(self.signal_type)
         if curr_checkpoint is not None and self.checkpoint != curr_checkpoint:
             new_index = store.get_signal_type_index(self.signal_type)
             if new_index is None:
@@ -140,8 +139,7 @@ def raw_lookup():
     """
     signal = require_request_param("signal")
     signal_type_name = require_request_param("signal_type")
-    include_distance = str_to_bool(
-        request.args.get("include_distance", "false"))
+    include_distance = str_to_bool(request.args.get("include_distance", "false"))
     lookup_signal_func = (
         lookup_signal_with_distance if include_distance else lookup_signal
     )
@@ -153,8 +151,7 @@ def query_index(
     signal: str, signal_type_name: str
 ) -> t.Sequence[IndexMatchUntyped[SignalSimilarityInfo, int]]:
     storage = get_storage()
-    signal_type = _validate_and_transform_signal_type(
-        signal_type_name, storage)
+    signal_type = _validate_and_transform_signal_type(signal_type_name, storage)
 
     try:
         signal = signal_type.validate_signal_str(signal)
@@ -297,7 +294,9 @@ def lookup_post() -> TBankMatchBySignalType:
     return resp
 
 
-def lookup(signal: str, signal_type_name: str, bypass_coinflip: bool = False) -> TMatchByBank:
+def lookup(
+    signal: str, signal_type_name: str, bypass_coinflip: bool = False
+) -> TMatchByBank:
     current_app.logger.debug("performing lookup")
     results_by_bank_content_id = {
         r.metadata: r for r in query_index(signal, signal_type_name)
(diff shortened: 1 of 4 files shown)


ThisIsMissEm:
Cursory look at the screenshots seems nice

hasher-matcher-actioner/src/OpenMediaMatch/templates/components/banks/match_form.html.j2
```diff
@@ -64,27 +99,74 @@
...
+    const renderMatchResult = async (result) => {
+        // Fetch bank data with enabled ratios
+        const bankData = await fetchBankData(result.banks);
+        
+        // Render matched banks with enabled ratios
+        const banksList = result.banks.map(bankName => {
+            const bank = bankData[bankName];
+            const enabledRatio = bank ? (bank.matching_enabled_ratio * 100).toFixed(1) : 'Unknown';
+            const badgeClass = bank && bank.matching_enabled_ratio > 0.8 ? 'bg-success' : 
+                              bank && bank.matching_enabled_ratio > 0.5 ? 'bg-warning' : 'bg-danger';
```
Dcallies: unsure: bg-danger usually means that something is wrong or you need to address, but the user put it into that state and might be fine to maintain based on the state of their rollout. What do you think about standardizing on either:
1. 100% = green; <100% = warn
2. 100% = show nothing; <100% = show badge with warn
3. 0% = grey instead of warn

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -308,11 +311,15 @@ def lookup(signal: str, signal_type_name: str) -> TMatchByBank:
         len(enabled_content),
     )
     banks = {c.bank.name: c.bank for c in enabled_content}
+
+    # Always allow all banks, whether matching is enabled or not if bypass_coinflip is True
     rand = random.Random(request.args.get("seed"))
-    coinflip = rand.random()
+    coinflip = rand.random() if not bypass_coinflip else 0
+    current_app.logger.debug("coinflip: %s", coinflip)
```
Dcallies, line 318: I was considering asking if we should also match against disabled banks if we disabled coinflip, but I think we don't even load hashes into the index for disabled banks (updating the bank to be disabled is faster than rebuilding the index, so it would match *some* of the time if queried right after status change).

Therefore, I think the current logic is correct.

Dcallies (approved):
Thanks for doing this! A nice usability change. Some suggestions inline that I think will help with usability, hit me up on Discord with your preferences on merging.

Since the match ratio is a coinflip, is there any reason not to always bypass on the UI? I can't think of a reason to not always check the box.

hasher-matcher-actioner/src/OpenMediaMatch/templates/components/banks/match_form.html.j2
```diff
@@ -64,27 +99,74 @@
...
+            
+            return `
+                <li class="list-group-item">
+                    <span class="fw-bold">${bankName}</span>
+                    <span class="badge ${badgeClass} rounded-pill ms-2" title="Matching Enabled Ratio">
```
Dcallies: Should we add a tooltip to the badge that explains that for <100%, there's only a chance it will match in production?

juanmrad: Yeah I'll update the messaging a bit then.

hasher-matcher-actioner/src/OpenMediaMatch/templates/components/banks/match_form.html.j2
juanmrad: Screenshot 2025-08-27 at 1 01 49 PM


Something like this?

hasher-matcher-actioner/src/OpenMediaMatch/templates/components/banks/match_form.html.j2
```diff
@@ -130,13 +130,14 @@
...
-                              bank && bank.matching_enabled_ratio > 0.5 ? 'bg-warning' : 'bg-danger';
+            // Only tooltip if not 100% enabled, use warning color for partial rollouts
+            const badgeClass = bank && bank.matching_enabled_ratio < 1.0 ? 'bg-warning' : 'bg-success';
+            const showBadge = bank && bank.matching_enabled_ratio < 1.0;
             
             return `
                 <li class="list-group-item">
                     <span class="fw-bold">${bankName}</span>
-                    <span class="badge ${badgeClass} rounded-pill ms-2" title="Matching Enabled Ratio">
+                    <span class="badge ${badgeClass} rounded-pill ms-2" title="${showBadge ? `Partial rollout: Only ${enabledRatio}% chance this bank will match in production` : ''}">
```
Dcallies: I'm good to merge as is, if we wanted to get more specific, it might be something like 

> "This bank is only partially enabled, and may not count as matching in production based on coinflip. Enable the bank at 100% to ensure it matches consistently."

I like to make sure there is a consistent call to action if we are drawing attention to it.

juanmrad: adjusted.

Dcallies (approved):

PR #1857, opened 2025-08-27
[HMA] Split up README.md into individual documentation files

ThisIsMissEm:
Summary
---------

This is the most minimal change to start work on documentation for HMA, where I'm just taking what is already in the `README.md` and splitting it up into separate files to keep the main `README.md` more focused.

Test Plan
---------

- [x] n/a, documentation change without rewording anything (just moving content)


hasher-matcher-actioner/docs/architecture.md
```diff
@@ -0,0 +1,148 @@
...
+- Nginx
+- Træfik
+- Kubernetes ingress controller
+- Apache HTTPD
+- Amazon Application Load Balancer
+- Azure Application Gateway
+
+If you're operating at the scale where this becomes necessary, you will more than likely already have your own preferred solution here.
+
+## Deployment scenarios and scaling strategies
```
ThisIsMissEm, line 46: Maybe this should be a separate section? `scaling.md` or `deployment.md` ?

I think we'd probably want information about the "roles" in the architecture document, but maybe move deployment scenarios / scaling strategies to a separate piece of documentation and expand upon that?

hasher-matcher-actioner/docs/history.md
```diff
@@ -0,0 +1,46 @@
...
+| HMA 1.0 (AWS Deployed)                                          | HMA 2.0 (Docker image only)                                                |
+| --------------------------------------------------------------- | -------------------------------------------------------------------------- |
+| Heavy AWS dependencies: VPC, DynamoDB, Lambda, API Gateway, ... | OSS dependencies only                                                      |
+| Opinionated: only supported by included Terraform scripts       | Runs in any Docker based infra                                             |
+| Service capabilities provided by AWS Lambda                     | Service capabilities provided by composable and repackageable components   |
+| AWS only, needs its own S3 buckets and DynamoDB tables          | Integrates anywhere Docker can be used, or under self-managed Python/Flask |
+| Includes a GUI and API                                          | API only                                                                   |
+| Scales automatically                                            | Doesn't auto-scale but can be auto-scaled                                  |
+
+## Motivations
```
ThisIsMissEm, line 22: I'm not sure if this belongs in the history or if it's maybe better in the goals documentation?

hasher-matcher-actioner/docs/api.md
```diff
@@ -0,0 +1,174 @@
+# API
```
ThisIsMissEm, line 1: This is verbatim from the existing README.md, and isn't the `Getting Started with the API" document that we've proposed else where.

hasher-matcher-actioner/docs/history.md
Dcallies: I think it mattered more when this was new, we can probably delete this entirely once HMA1.0 falls all the way in the memory hole.

Issue #1808, opened 2025-04-09
[hma] Provide a UI to remove content from the bank


Dcallies:
See #1682 

One adopter is using the UI to manager their banks, but the ability to remove content from a bank is a key feature.

PR #1859, opened 2025-08-28
[HMA] Banks View UI Improvements

juanmrad:
Summary
---------

This PR Creates Bank UI Improvements. Takes care of some of the requests on #1808 By creating the ability to match against URLs or hashes directly. It also now includes the distance of those hashes and the content id it matched with on the UI.

Added also a remove content modal to be able to remove content based on a file, url or hash. Similarly the add content got the ability to add based on URL, and hash. 

Test Plan
---------

Tested locally video:

https://github.com/user-attachments/assets/a6af284b-27a3-4be3-a28d-5be999dc598e

Screenshot 2025-08-27 at 10 58 35 PM

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
@@ -119,6 +119,25 @@ def hash_media() -> dict[str, str]:
     if media_url is None:
         abort(400, "Missing required parameter: url")
 
+    try:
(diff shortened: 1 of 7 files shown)


juanmrad:
@Dcallies I think the biggest JS was moving it out of the j2 files as to have better type hinting/colors on my code editor as it was getting harder to read the JS there ( may be just me/skill issue or missing extension ). 

Also tried my best on the video to make sure it showed all the functionality.

Issue #1852, opened 2025-08-26
ThreatExchange CLI file management issue on Windows


hexydec:
When using ThreatExchange to import hashes from StopNCII on Windows, I encountered the following issue:

```
2025-08-13 14:58:12,698 I] fetch() with 800 new records @ 2025-06-26T09:41:00
[stop_ncii] test-stop-ncii - Checkpointing at 2025-06-26T09:41:00
[stop_ncii] test-stop-ncii - failed to fetch!
2025-08-13 14:58:14,853 E] Failed to fetch test-stop-ncii
Traceback (most recent call last):
  File "C:\tools\miniconda3\envs\threat\lib\site-packages\threatexchange\cli\fetch_cmd.py", line 231, in execute_for_collab
    store.flush()
  File "C:\tools\miniconda3\envs\threat\lib\site-packages\threatexchange\exchanges\helpers.py", line 132, in flush
    self._write_state(collab_name, state.delta)
  File "C:\tools\miniconda3\envs\threat\lib\site-packages\threatexchange\cli\cli_state.py", line 183, in _write_state
    tmpfile.rename(file)
  File "C:\tools\miniconda3\envs\threat\lib\pathlib.py", line 1310, in rename
    self._accessor.rename(self, target)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\Username\\.threatexchange\\fetched\\stop_ncii\\.test-stop-ncii.state.pickle' -> 'C:\\Users\\Username\\.threatexchange\\fetched\\stop_ncii\\test-stop-ncii.state.pickle'
Traceback (most recent call last):
  File "C:\tools\miniconda3\envs\threat\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\tools\miniconda3\envs\threat\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\tools\miniconda3\envs\threat\Scripts\threatexchange.exe\__main__.py", line 7, in <module>
  File "C:\tools\miniconda3\envs\threat\lib\site-packages\threatexchange\cli\main.py", line 340, in main
    inner_main()
  File "C:\tools\miniconda3\envs\threat\lib\site-packages\threatexchange\cli\main.py", line 332, in inner_main
    execute_command(settings, namespace)
  File "C:\tools\miniconda3\envs\threat\lib\site-packages\threatexchange\cli\main.py", line 161, in execute_command
[...]

PR #1860, opened 2025-08-28
[pytx] Fix state file rename logic on windows #1852

hexydec:
Summary
---------

Renaming a temp file to a file that already exists doesn't work on Windows, this adds code to delete the file first.

Test Plan
---------

Tested locally.

### DIFF
--- a/python-threatexchange/threatexchange/cli/cli_state.py
+++ b/python-threatexchange/threatexchange/cli/cli_state.py
@@ -14,7 +14,6 @@ import pickle
 import pathlib
 import typing as t
 import logging
-import os
 
 from threatexchange.signal_type.index import SignalTypeIndex
 from threatexchange.signal_type.signal_base import SignalType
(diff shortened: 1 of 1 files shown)


python-threatexchange/threatexchange/cli/cli_state.py
```diff
@@ -180,4 +181,10 @@ def _write_state(
         tmpfile = file.with_name(f".{file.name}")
         with tmpfile.open("wb") as f:
             pickle.dump(delta, f)
+
+        # delete the target file first if it exists, otherwise windows doesn't like it
+        if os.path.exists(file):
+            os.remove(file)
+
+        # move temp file to target
         tmpfile.rename(file)
```
Dcallies: Hey @hexydec! Thanks for contributing! I was reading through the Pathlib docs, and found https://docs.python.org/3/library/pathlib.html#pathlib.Path.replace which I think does exactly what we need, and preserves an attempt at atomic replace (by using a [special flag on windows systems](https://stackoverflow.com/questions/69363867/difference-between-os-replace-and-os-rename)). Can you give this version a try?
```suggestion
        # move temp file to target
        tmpfile.replace(file)
```

hexydec: Sure no probs

hexydec: Have updated and retested. All good.

Dcallies (approved):
Thanks a ton for the help!

Issue #1825, opened 2025-05-23
[hma] Add documentation for API


julietshen:
Creating this issue to track discussion and planning around a documentation tool for HMA, like OpenAPI

julietshen:
cc @ThisIsMissEm @juanmrad @Dcallies

julietshen:
I did a first attempt, with some (a lot of) help from Cursor:

https://github.com/facebook/ThreatExchange/pull/1826

julietshen:
Based on some feedback from @juanmrad, splitting the HTML into its own file. I added a pyyaml dependency and can remove it, but it creates some other risks

- openapi.py will become very long: Embedding this much data will make the blueprint harder to read and edit.,
- Spec updates become cumbersome: Any change to the OpenAPI spec will require regenerating this large Python dictionary and updating openapi.py.,
- Potential for syntax errors: Manually converting or pasting such a large structure into Python code is error-prone.

Dcallies:
Comments on the PR, but we should also consider using tooling integrated with the technology itself like https://luolingchun.github.io/flask-openapi3/v3.x/ - manually maintaining a second documentation file might be faster in the short term, but burdensome in the long term.

ThisIsMissEm:
From the HMA meeting today, we had two options to move forwards:
1. Write basic documentation in markdown for some common APIs to help people integrating get started with the HMA endpoints that they will most likely will need, this could also include some information on scaling rollouts (e.g., match %)
2. Using flash-openapi3 to do code-driven OpenAPI specification for the existing APIs.

turt2live:
I'm hoping to contribute this, likely aiming to land option 1 with a start on option 2 ahead of the API rework, though very much welcome others getting to it faster 😅

PR #1866, opened 2025-09-05
[hma] Rudimentary docs for status, hashing, and matching functions

turt2live:
In pursuit of https://github.com/facebook/ThreatExchange/issues/1825, some *very* barebones docs on a few of the endpoints. 

This is not intended to be comprehensive or cover the entire API surface. I've essentially picked the ones I wish I knew about when I started writing code to use HMA, with the (undocumented) assumption that the reader has used the dev UI to get some banks going.

I've also fixed the heading levels while here.

### DIFF
--- a/hasher-matcher-actioner/docs/api.md
+++ b/hasher-matcher-actioner/docs/api.md
@@ -2,7 +2,48 @@
 
 As HMA is not planning to provide a production UI, the API is the key point of integration. We hope that users are able to treat it as a “black box” with the API defining the inputs and outputs.
 
-### Fetching / Exchanges
+## Accessing the API
+
+Some endpoints may only be available to specific roles. All endpoints are based upon the HMA process's base URL.
+In the monolithic docker compose setup from this repository, that will be `http://localhost:5100`.
+
+**Note**: The API docs are a work in progress and may be missing some (useful) endpoints.
+
+### Status/Information
+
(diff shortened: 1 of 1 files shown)


hasher-matcher-actioner/docs/api.md
```diff
@@ -2,7 +2,48 @@
 
 As HMA is not planning to provide a production UI, the API is the key point of integration. We hope that users are able to treat it as a “black box” with the API defining the inputs and outputs.
```
Dcallies, line 3: (ignorable): The latter does not follow the former - a UI is not the correct place to do production scale matching on platform :P

hasher-matcher-actioner/docs/api.md
```diff
@@ -2,7 +2,48 @@
...
+## Accessing the API
+
+Some endpoints may only be available to specific roles. All endpoints are based upon the HMA process's base URL.
+In the monolithic docker compose setup from this repository, that will be `http://localhost:5100`.
+
+**Note**: The API docs are a work in progress and may be missing some (useful) endpoints.
+
+### Status/Information
+
+When HMA has started up, it will respond 200 OK to `GET /status`.
```
Dcallies, line 14: As a followup, we could explain in what circumstances it will not return 200
* On startup, it will return a different code if it is trying to get a new index
* If the index becomes stale, it will return a different code

PR #1867, opened 2025-09-06
Fix build on macOS Apple silicon

lxcode:
Add -Wno-deprecated-declarations to fix build with system clang. All builds and tests complete successfully on my M4.


Dcallies (approved):
Hey @lxcode! Big fan of your work, it's great to have you here! 

I'm a C++ dummy, is there a way we can limit the use of this flag only if we can detect it's an apple device? My thought is that if there is a future deprecation that affects the ubuntu build we can detect (and maybe fix it).

Dcallies:
Hey @lxcode  - just following up, do you know the answer to the question above? I can also ask around internally to see if I can find anyone.

Dcallies:
I figured it can't get any more broken, so let's merge as is.

PR #1879, opened 2025-10-10
Add bank filtering to lookup endpoints

### LINKED ISSUE #1875: [hma] Filter lookup by bank or banks
**Request: Add filtering by bank to `/m/lookup` and `/h/raw_lookup`**

We’ve started splitting our content into different banks, and want to perform searches limited to specific banks.  
Currently, this isn’t possible with the existing API endpoints.

Right now:

- `/m/lookup` returns matches grouped by bank, but does not support filtering by bank input.  
- `/h/raw_lookup` doesn’t return any bank information at all ([see #1827](https://github.com/facebook/ThreatExchange/issues/1827)).

This means we’re inefficiently re-hashing images we already have the content hash for, then filtering results from HMA in our own code to exclude banks we don’t care about.

We’d prefer to make a single request to `/h/lookup`with the hash we already have, and specify which banks to include in the search.

---

### Problem

We want to run separate lookups for each bank.
But the current API design forces us to:

1. Re-hash images we already have hashes for, since we must use `/m/lookup` for bank information.  
2. Retrieve all matches, then post-process and discard unwanted banks.  

This is redundant and wasteful, and likely slows down matching because the API searches across banks we don’t need.

There’s even a comment in `matching.py` that suggests filtering by bank was planned, but it does not appear implemented:

https://github.com/facebook/ThreatExchange/blob/5a1d53a8d05676add176381f9ff354d2f42d63fc/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py#L125-L147

---

### Feature Request

Extend `/m/lookup` and `/h/raw_lookup` to support **optional bank filtering**.

**Proposed API design example:**

```json5
POST /h/raw_lookup
{
  "signal": "facd8bcb2a49bcebdec1985298d5fe84bcd006c187c598c720c3c087b3fdb318",
  "signal_type": "PDQ",
  "banks": ["EXAMPLE_BANK_A", "EXAMPLE_BANK_B"] // this is new, by ID would also be acceptable, name is just more convenient
}

### PR DESCRIPTION
**Pull request by cnaples79**
## Summary
- Implements optional bank filtering for `/m/lookup` and `/h/raw_lookup` endpoints
- Allows clients to restrict search results to specific banks using a comma-separated list
- Addresses the feature request in issue #1875

## Changes Made
- Modified `raw_lookup()` endpoint to accept optional `banks` parameter
- Modified `lookup_get()` endpoint to accept optional `banks` parameter  
- Modified `lookup_post()` endpoint to accept optional `banks` parameter
- Updated `lookup_signal()` helper function to filter by banks
- Updated `lookup_signal_with_distance()` helper function to filter by banks
- Updated `lookup()` function to filter enabled banks by requested banks

## API Usage
```bash
# Filter lookup by banks
GET /m/lookup?signal_type=pdq&signal=hash&banks=BANK_A,BANK_B

# Filter raw lookup by banks
GET /h/raw_lookup?signal_type=pdq&signal=hash&banks=BANK_A
```

## Implementation Details
- Banks parameter is optional and accepts comma-separated bank names
- When provided, results are filtered to only include content from the specified banks
- When not provided, behavior is unchanged (all enabled banks are searched)
- Filtering is applied after existing enable/disable and coinflip logic

## Test Plan
- Manually tested syntax validation with Python
- CI will run full test suite including type checking and linting
- Integration testing should verify that:
  - Requests without banks parameter work as before
  - Requests with banks parameter only return results from specified banks
  - Invalid bank names are handled appropriately

Fixes #1875

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -143,7 +143,7 @@ def raw_lookup():
 
     # Parse optional banks parameter
     banks_param = request.args.get("banks")
-    requested_banks = banks_param.split(",") if banks_param else None
+    requested_banks = set(banks_param.split(",")) if banks_param else None
 
     lookup_signal_func = (
         lookup_signal_with_distance if include_distance else lookup_signal
@@ -183,16 +183,13 @@ def lookup_signal(
     if banks is not None:
         storage = get_storage()
         contents = storage.bank_content_get(content_ids)
-        requested_banks_set = set(banks)
-        content_ids = [
-            c.id for c in contents if c.bank.name in requested_banks_set
-        ]
+        content_ids = [c.id for c in contents if c.bank.name in banks]
 
     return content_ids
 
 
 def lookup_signal_with_distance(
-    signal: str, signal_type_name: str, banks: t.Optional[t.Sequence[str]] = None
+    signal: str, signal_type_name: str, banks: t.Optional[t.Set[str]] = None
 ) -> list[MatchWithDistance]:
     results = query_index(signal, signal_type_name)
     matches = [
@@ -208,11 +205,8 @@ def lookup_signal_with_distance(
         storage = get_storage()
         content_ids = [m["bank_content_id"] for m in matches]
         contents = storage.bank_content_get(content_ids)
-        requested_banks_set = set(banks)
         # Create a set of valid content IDs
-        valid_content_ids = {
-            c.id for c in contents if c.bank.name in requested_banks_set
-        }
+        valid_content_ids = {c.id for c in contents if c.bank.name in banks}
         # Filter matches to only include valid content IDs
         matches = [m for m in matches if m["bank_content_id"] in valid_content_ids]
 
@@ -281,7 +275,7 @@ def lookup_get() -> t.Union[TMatchByBank, TBankMatchBySignalType]:
     """
     # Parse optional banks parameter
     banks_param = request.args.get("banks")
-    requested_banks = banks_param.split(",") if banks_param else None
+    requested_banks = set(banks_param.split(",")) if banks_param else None
 
     resp: dict[str, TMatchByBank] = {}
     if request.args.get("url", None):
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -140,11 +140,16 @@ def raw_lookup():
     signal = require_request_param("signal")
     signal_type_name = require_request_param("signal_type")
     include_distance = str_to_bool(request.args.get("include_distance", "false"))
+
+    # Parse optional banks parameter
+    banks_param = request.args.get("banks")
+    requested_banks = banks_param.split(",") if banks_param else None
```
Dcallies: unsure: Should we validate that these are real banks?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -168,23 +173,51 @@ def query_index(
     return results
 
 
-def lookup_signal(signal: str, signal_type_name: str) -> list[int]:
+def lookup_signal(
+    signal: str, signal_type_name: str, banks: t.Optional[t.Sequence[str]] = None
```
Dcallies: Looks like we turn into a set when we use it, we could go straight to set at the type here
```suggestion
    signal: str, signal_type_name: str, banks: t.Optional[t.Set[str]] = None
```

cnaples79: Yep I agree with you here, committing this.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
cnaples79: Current behavior is that incorrect bank names return no results. This design is consistent with how other filters typically work. We could add validation and return 400 Bad Request, but I think current design is appropriate.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -168,23 +173,45 @@ def query_index(
...
             "bank_content_id": m.metadata,
             "distance": m.similarity_info.pretty_str(),
         }
         for m in results
     ]
 
+    # Filter by banks if specified
+    if banks is not None:
+        storage = get_storage()
+        content_ids = [m["bank_content_id"] for m in matches]
```
Dcallies, line 206: typing is showing that this isn't metdata isn't an int, and so will throw/not find anything when passed into L207

This looks to be a previous bug, not sure why it wasn't picked up before.

Issue #1856, opened 2025-08-27
Hasher-Matcher-Actioner (HMA) Office Hours!


Dcallies:
The Hasher Matcher Matcher sub-project has its own biweekly meeting hosted by [ROOST](https://roost.tools/)!

You can see what issues are in active discussion and actively being worked at the ROOST [office hours project](Project: https://github.com/orgs/roostorg/projects/3/views/1)!

Details for how to get an invite are forthcoming!

ghostinhershell:
Hey! Just wanted to throw my hat in the ring and say I would be interested in attending 👀

PR #1881, opened 2025-10-15
[hma] Topk threshold tests clip

miriyalamani:
I needed to add some new end points for clip index, for evaluation. Please review

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
@@ -146,11 +146,16 @@ def raw_lookup():
     signal = require_request_param("signal")
     signal_type_name = require_request_param("signal_type")
     include_distance = str_to_bool(request.args.get("include_distance", "false"))
+
+    # Parse optional banks parameter
+    banks_param = request.args.get("banks")
+    requested_banks = set(banks_param.split(",")) if banks_param else None
+
     lookup_signal_func = (
         lookup_signal_with_distance if include_distance else lookup_signal
     )
 
-    return {"matches": lookup_signal_func(signal, signal_type_name)}
+    return {"matches": lookup_signal_func(signal, signal_type_name, requested_banks)}
 
 
 @bp.route("/lookup_threshold")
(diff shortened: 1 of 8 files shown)


Dcallies (changes requested):
Hey @miriyalamani, thanks for submitting a PR!

I like the idea for this feature, but you may run into some challenges, as threshold is not numeric for all SignalTypes (e.g. check out VideoPDQ's threshold: https://github.com/facebook/ThreatExchange/tree/main/vpdq / - which is a pair of overlap percentages).

Can you tell me more about the tradeoffs between adding threshold and N as an optional parameter on lookup vs adding another API route?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -173,6 +251,58 @@ def query_index(
...
+        abort(503, "index not yet ready")
+    if not hasattr(index, "query_top_k"):
+        abort(
+            501,
+            f"Signal type '{signal_type_name}' does not support query_top_k method",
+        )
+    current_app.logger.debug("[lookup_signal_topk] querying index")
+    results = index.query_top_k(signal, k)  # type: ignore[attr-defined]
+    current_app.logger.debug("[lookup_signal_topk] query complete")
+    return results
```
juanmrad, line 303: Is there a reason why we need to filter by top k items? shouldn't the signal have a threshold instead? I don't understand the implementation here. One thing would be to have a defined threshold to pass down to the signal, but here you are just returning the top k items that matched a given hash.

Why would this live in HMA over the implementation from the adopter?

miriyalamani: For our evaluation, we want threshold to be configurable. @Dcallies I can make threshold an optional parameter for lookup which is Comparable (float or int) , as this would work for pdq or any other single threshold signals. If a threshold is passed and it's not numeric I can throw an error. I defined new end points to make the functionality more explicit and also only one of k or threshold exclusively at once. 

@Juan currently hma signals have a default threshold which do not guarantee a minimum number of matches. The idea was to run these tests on a series of images which have threshold configurable or adjusted internally for top k so we find exact k matches.

miriyalamani: @Dcallies, should I make changes to lookup and remove new methods?

Dcallies: @miriyalamani - do you want to come to the HMA office hours to talk through it (details: #1856) ? Might be faster than going back and forth on the PR, especially since we have the PR to reference now. 

What I think we'll need to figure out:
1. I think we should not make a new lookup to take in these parameters (we might already have too many lookups with `lookup` and `raw_lookup`
2. We need to come up with a strategy for how topk and thresholding works for signal types other than PDQ, or otherwise come up with a thesis for why it makes sense to specializes parts of HMA to PDQ vs long term

miriyalamani: okay, I will be there for HMA office hours

PR #1884, opened 2025-10-21
[hma] Add hook for setting up deployment-specific details

turt2live:
Summary
---------

In our deployment, we'd like to have API keys set up so we can limit access to HMA within our system. API keys as the sole (preferred) authorization model is opinionated though, so instead of adding that particular functionality this PR adds a generic hook so consumers can choose to do whatever they want. For example, while we use API keys, another deployment might want to support cookies, complete RBAC, or something else entirely. 

The approach to specify a function through the config was picked because there doesn't appear to be a great way (if any?) to acquire the Flask app itself - it appears as though the `app` instance that can be acquired by the config isn't yet set up enough to support HTTP middleware-like functions.

Test Plan
---------

The PR includes a rough example of how one might use this in development by accessing the Flask app's logger. If things are working, the log line should be present. 

An alternative test would be to add API (Bearer) auth to the endpoints, open the UI, and expect to see "Authentication required" instead of the UI itself. An example which does that is:

```python
def on_flask_ready(app):
    api_key = os.environ.get("API_KEY", "")
    if not api_key:
        app.logger.warning("No API authentication enabled")
        return

    app.logger.info("Adding API authentication")

    @app.before_request
    def require_auth():
        app.logger.info("%s %s", request.method, request.path)
        auth = request.headers.get("Authorization", "")
        if not auth.startswith("Bearer ") or auth[7:] != api_key:
            return Response("Authentication required", 401)


APP_HOOK = on_flask_ready
```

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/app.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/app.py
@@ -175,7 +175,7 @@ def create_app() -> flask.Flask:
         # doesn't supply such functionality as individual deployments may have
         # different, competing, requirements.
         hook = app.config.get("APP_HOOK", None)
-        if hook != None:
+        if hook != None and callable(hook):
             hook(app)
 
     @app.route("/")


hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -170,6 +170,14 @@ def create_app() -> flask.Flask:
         if app.config.get("ROLE_CURATOR", False):
             app.register_blueprint(curation.bp, url_prefix="/c")
 
+        # Allow the config to hook into the Flask app to add things like auth,
+        # new endpoints, etc as may be required by their environments. HMA itself
+        # doesn't supply such functionality as individual deployments may have
+        # different, competing, requirements.
+        hook = app.config.get("APP_HOOK", None)
+        if hook != None and callable(hook):
+            hook(app)
```
Dcallies, line 179: Should we not do the callable check to deliberately cause a failure if the configuration is incorrect?
```suggestion
        app.config.get("APP_HOOK", lambda _: None)(app)
```

turt2live: Probably, yea. Especially if this gets used for things like auth

turt2live: https://github.com/facebook/ThreatExchange/pull/1885

PR #1891, opened 2025-11-10
[hma] Add OpenAPI docs using flask-openapi3 (#1825)

rossdm:
Summary
---------

Introduces code-driven OpenAPI generation using [flask-openapi3](https://luolingchun.github.io/flask-openapi3/v3.x/), as suggested in https://github.com/facebook/ThreatExchange/issues/1825#issuecomment-2925313878. Added a "Documentation" tab to the app.

Screenshot 2025-11-09 at 11 52 01 PM

Test Plan
---------

- verified that OpenAPI spec is generated manually
- added [minimal API test](https://github.com/rossdm/ThreatExchange/blob/4128c08a4eaf5fc812546b4e1604315ad35121e2/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py#L32) to assert that `openapi.json` is generated

### DIFF
--- a/hasher-matcher-actioner/pyproject.toml
+++ b/hasher-matcher-actioner/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
     "flask_apscheduler",
     "python-dateutil",
     "python-json-logger",
+    "flask-openapi3[swagger]"
 ]
 
 [project.optional-dependencies]
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -1,11 +1,14 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 
-from datetime import datetime, timedelta
 import re
 import time
 import typing as t
+from datetime import datetime, timedelta
 
-from flask import Blueprint, Response, request, jsonify, abort
+from flask_openapi3 import APIBlueprint
+from flask_openapi3.models import Tag
+from flask import Response, abort, jsonify, request
+from pydantic import BaseModel, ValidationError
 from sqlalchemy.exc import IntegrityError
 from werkzeug.exceptions import HTTPException
 
@@ -17,41 +20,61 @@ from OpenMediaMatch import persistence
 from OpenMediaMatch.utils import flask_utils
 import OpenMediaMatch.storage.interface as iface
 from OpenMediaMatch.blueprints import hashing
-
-
-class BankContentResponse(t.TypedDict):
-    id: int
-    disable_until_ts: int
-    collab_metadata: t.Mapping[str, t.Sequence[str]]
-    original_media_uri: t.Optional[str]
-    bank: iface.BankConfig
-    signals: t.NotRequired[t.Mapping[str, str]]
+from OpenMediaMatch.schemas.curation import (
+    BankConfig,
+    BankContentResponse,
+    BankCreateRequest,
+    BankUpdateRequest,
+    BankedContentMetadata,
+    ExchangeConfig,
+    ExchangeUpdateRequest,
+)
+from OpenMediaMatch.schemas.shared import ErrorResponse, SuccessResponse
 
 
 def five_years_from_now() -> int:
     return int(time.mktime((datetime.now() + timedelta(days=365 * 5)).timetuple()))
 
 
-class BankedContentMetadata(t.TypedDict):
-    content_id: t.NotRequired[str]
-    content_uri: t.NotRequired[str]
-    json: t.NotRequired[dict[t.Any, t.Any]]
-
-
-bp = Blueprint("curation", __name__)
+bp = APIBlueprint("curation", __name__, url_prefix="/c")
 bp.register_error_handler(HTTPException, flask_utils.api_error_handler)
 
 # Banking
 
 
-@bp.route("/banks", methods=["GET"])
+class BankPathParams(BaseModel):
+    bank_name: str
+
+
+class BankContentPathParams(BankPathParams):
+    content_id: int
+
+
+class ExchangePathParams(BaseModel):
+    exchange_name: str
+
+
+@bp.get(
+    "/banks",
+    tags=[Tag(name="Banks")],
+    responses={"200": {"description": "List of banks"}},
+    summary="List banks",
+    description="Get list of all banks",
+)
 def banks_index():
     storage = persistence.get_storage()
     return list(storage.get_banks().values())
 
 
-@bp.route("/bank/<bank_name>", methods=["GET"])
-def bank_show_by_name(bank_name: str):
+@bp.get(
+    "/bank/<bank_name>",
+    tags=[Tag(name="Banks")],
+    responses={"200": BankConfig, "404": ErrorResponse},
+    summary="Get bank",
+    description="Get bank configuration by name",
+)
+def bank_show_by_name(path: BankPathParams):
+    bank_name = path.bank_name
     storage = persistence.get_storage()
 
     bank = storage.get_bank(bank_name)
@@ -60,16 +83,21 @@ def bank_show_by_name(bank_name: str):
     return jsonify(bank)
 
 
-@bp.route("/banks", methods=["POST"])
-def bank_create():
-    name = flask_utils.require_json_param("name")
-    data = request.get_json()
-    enabled_ratio = 1.0
-    if "enabled_ratio" in data:
-        enabled_ratio = flask_utils.str_to_type(data["enabled_ratio"], float)
-    elif "enabled" in data:
-        enabled_ratio = 1.0 if flask_utils.str_to_bool(data["enabled"]) else 0.0
-    return jsonify(bank_create_impl(name, enabled_ratio)), 201
+@bp.post(
+    "/banks",
+    tags=[Tag(name="Banks")],
+    responses={"201": BankConfig, "400": ErrorResponse, "403": ErrorResponse},
+    summary="Create bank",
+    description="Create a new bank",
+)
+def bank_create(body: BankCreateRequest):
+    if body.enabled_ratio is not None:
+        enabled_ratio = body.enabled_ratio
+    elif body.enabled is not None:
+        enabled_ratio = 1.0 if body.enabled else 0.0
+    else:
+        enabled_ratio = 1.0
+    return jsonify(bank_create_impl(body.name, enabled_ratio)), 201
 
 
 def bank_create_impl(name: str, enabled_ratio: float = 1.0) -> iface.BankConfig:
@@ -83,23 +111,34 @@ def bank_create_impl(name: str, enabled_ratio: float = 1.0) -> iface.BankConfig:
     return bank
 
 
-@bp.route("/bank/<bank_name>", methods=["PUT"])
-def bank_update(bank_name: str):
+@bp.put(
+    "/bank/<bank_name>",
+    tags=[Tag(name="Banks")],
+    responses={
+        "200": BankConfig,
+        "400": ErrorResponse,
+        "403": ErrorResponse,
+        "404": ErrorResponse,
+    },
+    summary="Update bank",
+    description="Update bank configuration",
+)
+def bank_update(path: BankPathParams, body: BankUpdateRequest):
+    bank_name = path.bank_name
     storage = persistence.get_storage()
-    data = request.get_json()
     bank = storage.get_bank(bank_name)
     rename_from: t.Optional[str] = None
     if bank is None:
         abort(404, "Bank not found")
 
     try:
-        if "name" in data:
+        if body.name is not None:
             rename_from = bank.name
-            bank.name = data["name"]
-        if "enabled" in data:
-            bank.matching_enabled_ratio = 1 if bool(data["enabled"]) else 0
-        if "enabled_ratio" in data:
-            bank.matching_enabled_ratio = data["enabled_ratio"]
+            bank.name = body.name
+        if body.enabled is not None:
+            bank.matching_enabled_ratio = 1 if bool(body.enabled) else 0
+        if body.enabled_ratio is not None:
+            bank.matching_enabled_ratio = body.enabled_ratio
 
         storage.bank_update(bank, rename_from=rename_from)
     except ValueError as e:
@@ -109,8 +148,15 @@ def bank_update(bank_name: str):
     return jsonify(bank)
 
 
-@bp.route("/bank/<bank_name>", methods=["DELETE"])
-def bank_delete(bank_name: str):
+@bp.delete(
+    "/bank/<bank_name>",
+    tags=[Tag(name="Banks")],
+    responses={"200": SuccessResponse, "404": ErrorResponse},
+    summary="Delete bank",
+    description="Delete a bank",
+)
+def bank_delete(path: BankPathParams):
+    bank_name = path.bank_name
     storage = persistence.get_storage()
     storage.bank_delete(bank_name)
     return {"message": "Done"}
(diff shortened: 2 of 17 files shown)


hasher-matcher-actioner/pyproject.toml
```diff
@@ -25,6 +25,9 @@ dependencies = [
     "flask_apscheduler",
     "python-dateutil",
     "python-json-logger",
+    "flask-openapi3[swagger]",
+    "pydantic>=2.0",
```
Dcallies: unsure: I believe you are safe to omit this, since it's a requirement of flask-openapi

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -105,22 +113,30 @@ def hash_media() -> dict[str, str]:
...
     try:
-        return hash_url_content(media_url)
+        result = hash_url_content(
+            query.url,
+            content_type_hint=query.content_type,
+            signal_type_names=query.types,
+        )
+        return HashResponse(**result)
     except ValueError as e:
         abort(400, str(e))
```
Dcallies: nit: this hash response should probably be outside the try
```suggestion
    except ValueError as e:
        abort(400, str(e))
   return HashResponse(**result)
```

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -32,21 +38,32 @@
...
 
-class MatchWithDistance(t.TypedDict):
+class MatchWithDistancePayload(t.TypedDict):
     bank_content_id: int
     distance: str
 
 
-TMatchByBank = t.Mapping[str, t.Sequence[MatchWithDistance]]
-TBankMatchBySignalType = t.Mapping[str, TMatchByBank]
+TMatchByBank = dict[str, list[MatchWithDistancePayload]]
```
Dcallies, line 63: One reason why this might be `Mapping` is to suggest to the caller that they shouldn't edit these objects. We should make sure that we are wrapping any internal state dicts to make sure callers can't break anything by editing the dicts

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -456,26 +514,32 @@ def compare():
         }
     }
     """
-    request_data = request.get_json()
+    request_data = request.get_json(silent=True)
```
Dcallies: Why this change?

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
Dcallies: blocking q: I've been conditioned against adding things in the `__init__.py` by many hard to debug issues around imports. Can we instead move this to a dedicated file and leave the init blank, or is there a compelling reason to include this here?

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+"""
+Pydantic schema models for OpenAPI documentation and validation.
+"""
+
+from typing import Optional, Dict, List, Any, Union
+from pydantic import BaseModel, Field, ConfigDict
+
+
+# Common/Shared Schemas
+class ErrorResponse(BaseModel):
```
Dcallies: What are your thoughts about the pros/cons of inlining the models with the API that defines them, as long as they aren't shared by multiple APIs?

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+    )
+
+
+class HashResponse(BaseModel):
+    """Response schema for hash generation."""
+
+    model_config = ConfigDict(extra="allow")
+
+    def __init__(self, **data):
+        super().__init__(**data)
```
Dcallies: blocking q: Is this the same as not defining the init?

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+    )
+
+
+class LookupResponse(BaseModel):
+    """Response schema for content lookup."""
+
+    model_config = ConfigDict(extra="allow")
+
+    def __init__(self, **data):
+        super().__init__(**data)
```
Dcallies: ditto

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+class CompareRequest(BaseModel):
+    """Request schema for hash comparison."""
+
+    model_config = ConfigDict(extra="allow")
+
+    def __init__(self, **data):
+        super().__init__(**data)
```
Dcallies: ditto

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+class CompareResponse(BaseModel):
+    """Response schema for hash comparison."""
+
+    model_config = ConfigDict(extra="allow")
+
+    def __init__(self, **data):
+        super().__init__(**data)
```
Dcallies: Just imagine I marked the rest too

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -31,6 +28,15 @@ def test_status_response(client: FlaskClient):
     assert response.data == b"I-AM-ALIVE"
 
 
+def test_openapi_documentation_available(client: FlaskClient):
```
Dcallies, line 31: Thanks for adding a test to cover the new docs!

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
```diff
@@ -0,0 +1,369 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+
+"""
+Pydantic schema models for OpenAPI documentation and validation.
+"""
+
+from typing import Optional, Dict, List, Any, Union
```
Dcallies: blocking q: Elsewhere in the codebase we are using `dict` and `list` instead of the typing equivalents. Is there any reason to do those here?

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -132,19 +184,23 @@ def _validate_bank_add_metadata() -> t.Optional[BankedContentMetadata]:
...
-            400, f"metadata contains unexpected keys: {' ,'.join(sorted(unexpected))}"
-        )
-    return t.cast(BankedContentMetadata, metadata)
 
-
-@bp.route("/bank/<bank_name>/content/<int:content_id>", methods=["GET"])
-def bank_get_content(bank_name: str, content_id: int):
+    try:
+        return BankedContentMetadata(**metadata)
+    except Exception as e:
```
Dcallies: blocking q: Can we narrow this exception type to only what we might expect to see for invalid metadata? I'm worried this will pick up unrelated exceptions

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -155,26 +211,40 @@ def bank_get_content(bank_name: str, content_id: int):
...
 
-    # Create base response
-    response: BankContentResponse = {
-        "id": content_config.id,
-        "disable_until_ts": content_config.disable_until_ts,
-        "collab_metadata": content_config.collab_metadata,
-        "original_media_uri": content_config.original_media_uri,
-        "bank": content_config.bank,
+    collab_metadata = {
+        key: list(values) for key, values in content_config.collab_metadata.items()
```
Dcallies, line 209: Is this a behavior change?

hasher-matcher-actioner/pyproject.toml
rossdm: ty, [done](https://github.com/facebook/ThreatExchange/compare/4128c08a4eaf5fc812546b4e1604315ad35121e2..92963b8ca31b0b102e9891a4664f245b236e32ea#diff-db58e24776840f3be640c42e055bdba603464e345ae2fc3e70a792e0b8c35543L28).

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
rossdm: good tip, [done](https://github.com/facebook/ThreatExchange/compare/4128c08a4eaf5fc812546b4e1604315ad35121e2..92963b8ca31b0b102e9891a4664f245b236e32ea#diff-ee2f7e851a87f239a8ac18c9c4adf1ece622adac0e86e17cd36391f3e79f8254R125).

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
rossdm: good catch; extracted to [separate modules ](https://github.com/facebook/ThreatExchange/compare/4128c08a4eaf5fc812546b4e1604315ad35121e2..92963b8ca31b0b102e9891a4664f245b236e32ea#diff-a143095dd7b194e907040392e7ed3c9b714600187265fdeecf5149fc6c25faefR9)by category.

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
rossdm: valid point, and I could see this being refactored further if this supports the project's trajectory.

pros
- context is readily accessible to reader
- fewer files

cons
- patterns may be less obvious
- not all readers may care about the details of schemas upfront

for now, I elected to add separate modules to `schemas`,

```
├── blueprints
│   ├── __init__.py
│   ├── curation.py
│   ├── development.py
│   ├── hashing.py
│   ├── matching.py
│   └── ui.py
├── schemas
│   ├── __init__.py
│   ├── curation.py
│   ├── development.py
│   ├── hashing.py
│   ├── matching.py
│   ├── shared.py
│   ├── status.py
│   └── ui.py
```

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
rossdm: doh! removed.

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
rossdm: ty, good catch. [Done](https://github.com/facebook/ThreatExchange/compare/4128c08a4eaf5fc812546b4e1604315ad35121e2..92963b8ca31b0b102e9891a4664f245b236e32ea#diff-34cf3043a672e5fdd63f398d452138d2398c632108e3bf0430cdc63865a3d917R1).

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
rossdm: well spotted. [Done](https://github.com/facebook/ThreatExchange/compare/4128c08a4eaf5fc812546b4e1604315ad35121e2..92963b8ca31b0b102e9891a4664f245b236e32ea#diff-542c3a60901c2e911e63b483ba90a19c97a58f6168e3f88daa56ed19855155c0R189).

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -155,26 +210,40 @@ def bank_get_content(bank_name: str, content_id: int):
...
+        id=content_config.id,
+        disable_until_ts=content_config.disable_until_ts,
+        collab_metadata=collab_metadata,
+        original_media_uri=content_config.original_media_uri,
+        bank=bank_schema,
+        signals=signals_payload,
+    )
 
-    return jsonify(response)
+    return jsonify(response.model_dump(exclude_none=True))
```
juanmrad: Could this break some expectations if a key is None to still be present as null?

rossdm: good catch; [refactored](https://github.com/facebook/ThreatExchange/compare/92963b8ca31b0b102e9891a4664f245b236e32ea..82fdca35850b494c2595505f5cf3631541b3d1a1#diff-542c3a60901c2e911e63b483ba90a19c97a58f6168e3f88daa56ed19855155c0R237) to only exclude the original [intended](https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api_banks.py#L311) field.

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
Dcallies: You said that you extracted them to separate modules, but I expect the content of the `__init__` file to be empty, because otherwise it changes the import behavior in a way that I generally find to be hard to debug. 

Did you mean to leave this file populated?

hasher-matcher-actioner/src/OpenMediaMatch/schemas/curation.py
```diff
@@ -0,0 +1,162 @@
...
+    "ExchangeCreateRequest",
+    "ExchangeConfig",
+    "ExchangeUpdateRequest",
+    "ExchangeApiConfigResponse",
+    "ExchangeFetchStatus",
+    "SignalTypeConfig",
+    "SignalTypeUpdateRequest",
+    "ContentTypeConfig",
+    "SignalTypeIndexStatus",
+]
```
Dcallies: nit: This looks like it is the same content as the file - I generally recommend against all imports and `*` imports because it makes it harder to manage the namespace, so I don't think we need to optimize for them

hasher-matcher-actioner/src/OpenMediaMatch/schemas/development.py
```diff
@@ -0,0 +1,18 @@
+from pydantic import BaseModel, Field
+
+
+__all__ = ["SeedSampleResponse", "SetupTxExampleRequest"]
```
Dcallies: ditto

hasher-matcher-actioner/src/OpenMediaMatch/schemas/hashing.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+__all__ = ["HashRequest", "HashResponse", "HashPostRequest"]
```
Dcallies: ditto

hasher-matcher-actioner/src/OpenMediaMatch/schemas/matching.py
```diff
@@ -0,0 +1,87 @@
+from typing import Optional, Union
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+__all__ = [
```
Dcallies: And all the rest

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
```diff
@@ -137,19 +160,25 @@ def raw_lookup():
...
+
+    if query.include_distance:
+        distance_matches = t.cast(list[MatchWithDistancePayload], matches)
+        response = RawLookupResponse(
+            matches=[MatchWithDistanceModel(**match) for match in distance_matches]
+        )
+        return response.model_dump()
+
+    int_matches = t.cast(list[int], matches)
+    matches_union: list[int | MatchWithDistanceModel] = [m for m in int_matches]
```
Dcallies: This typing appears incorrect - you are manually casting (usually a sign that something is wrong) to int, but then saying that it's not int on the next line.

juanmrad (approved):
Thanks for the time you spent working on this. Highly appreciated :shipit:

hasher-matcher-actioner/src/OpenMediaMatch/schemas/curation.py
rossdm: ah, doh! Sorry for missing your intent the first time. [Removed](https://github.com/facebook/ThreatExchange/compare/69d54abfca39aaafd0a2161697d18e2793ae9d43..38573556ee171b3e100ee46732bba3e0493db460).

hasher-matcher-actioner/src/OpenMediaMatch/schemas/__init__.py
rossdm: Ah, thank you for your patience with my Python naivety. [Fixed](https://github.com/facebook/ThreatExchange/pull/1891#discussion_r2525874139).

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
rossdm: Ah, ty. Another [attempt](https://github.com/facebook/ThreatExchange/compare/38573556ee171b3e100ee46732bba3e0493db460..8be1350a95e283229b035946af591ae5d38cdb94).

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/matching.py
rossdm: I think I was trying to do something clever & misguided w/validation; reverted.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
rossdm: didn't find a good alternative for type-safety and it didn't seem to have any undesirable JSON response implications, but I'll welcome future scrutiny here & in any other places since I may have easily overlooked something.

rossdm:
@Dcallies @juanmrad thank you so much for your guidance & patience w/getting this ready. There may be a few idioms that you might refine later, but I think this might be ready to merge. LMK if you find any other blockers, otherwise I'm comfortable w/letting this proceed.

PR #1895, opened 2025-12-14
[hma] Provide more detailed error when exchange fields are missing

turt2live:
Summary
---------

image

This is very much not intended to be a complete solution, but rather a step towards "better".

Test Plan
---------

1. Attempt to create an exchange (in the UI) which requires config values, but don't specify them.
2. Receive error stating there's missing fields.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
@@ -513,7 +513,10 @@ def exchange_create():
     try:
         cfg = dataclass_json.dataclass_load_dict(api_json, api_type.get_config_cls())
     except Exception as e:
-        abort(400, f"Failed to parse `api_json` - verify all fields are supplied. {str(e)}")
+        abort(
+            400,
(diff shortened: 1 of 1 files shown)


turt2live:
The type check error appears unrelated to my changes.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/curation.py
```diff
@@ -512,10 +512,10 @@ def exchange_create():
 
     try:
         cfg = dataclass_json.dataclass_load_dict(api_json, api_type.get_config_cls())
-    except Exception:
+    except Exception as e:
         abort(
             400,
-            "Failed to parse `api_json` - did you provide all the fields needed for your api cls?",
+            f"Failed to parse `api_json` - verify all fields are supplied. {str(e)}",
```
Dcallies, line 518: You aren't guaranteed that missing key is the reason you got an exception, recommend not including that bit
```suggestion
            f"Failed to parse `api_json`: {str(e)}",
```

Dcallies (approved):

PR #1901, opened 2026-01-18
[HMA] Add Database replication configurations

juanmrad:
Summary
---------

This PR adds configuration to sqlalchemy to create a new binding for postgres using the read UI config if available. This then is used on the database file to return either the read binding or the default session when requesting a read session and always returning default session  for write queries. 

Created a simple reference file to show this and updated architecture to explain read/write configs. 

Also had a small fight with CI and local Black so looked into this and changes:

Changed from hardcoded paths to "importStrategy": "fromEnvironment" for both Black and mypy
This ensures VS Code uses the versions installed by `pip install -e .[all]` (from postcreate.sh), matching what CI uses.

Test Plan
---------

Tested locally creating a new dockerfile with both read and write postgreSQL and added some banks and tested read queries where going to read replica. 

relevant docs: 
- https://flask-sqlalchemy.readthedocs.io/en/stable/binds/

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
@@ -19,6 +19,7 @@ import time
 import typing as t
 import os
 
+import flask
 from flask import current_app
 import flask_sqlalchemy
 from sqlalchemy import (
@@ -41,6 +42,8 @@ from sqlalchemy.orm import (
     relationship,
     validates,
     Session,
+    scoped_session,
+    sessionmaker,
 )
 from sqlalchemy.types import DateTime
 from sqlalchemy.sql import func
@@ -75,7 +78,28 @@ class Base(DeclarativeBase):
 # Standard Flask-SQLAlchemy initialization
 db = flask_sqlalchemy.SQLAlchemy(model_class=Base)
 
-_last_read_session_warning_time: float = 0
+_read_scoped_session: t.Optional[scoped_session] = None
+
+
+def init_read_replica(app: flask.Flask) -> None:
+    """
+    Initialize the read replica scoped session. Call after db.init_app().
+
+    Creates a scoped session bound to the "read" engine with proper
+    per-request lifecycle management.
+    """
+    global _read_scoped_session
+
+    read_engine = db.engines["read"]
+    _read_scoped_session = scoped_session(
+        sessionmaker(bind=read_engine),
+        scopefunc=db.session.registry.scopefunc,
+    )
+
+    @app.teardown_appcontext
+    def cleanup_read_session(exc: t.Optional[BaseException] = None) -> None:
+        if _read_scoped_session is not None:
+            _read_scoped_session.remove()
 
 
 def get_read_session():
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -71,11 +72,33 @@ class Base(DeclarativeBase):
...
+    """
+    Get a session bound to the read replica database.
+
+    Use for read-only queries that can tolerate slight staleness.
+    Falls back to primary if no replica is configured.
+    """
+    try:
+        read_engine = db.get_engine(bind_key="read")
+        return Session(read_engine)
+    except (KeyError, RuntimeError):
```
Dcallies: blocking: Slightly concerned about catching this without logging anything. 

What's the intent of this code block?

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -133,10 +133,30 @@ def create_app() -> OpenAPI:
...
+    sqlalchemy_config = {
+        "SQLALCHEMY_DATABASE_URI": database_uri,
+        "SQLALCHEMY_TRACK_MODIFICATIONS": False,
+    }
+
+    # If a separate read database is configured, set up binds
+    if database_read_uri:
+        sqlalchemy_config["SQLALCHEMY_BINDS"] = {
+            "read": database_read_uri,
+        }
```
Dcallies: Why not instead undconditionally add this bind, and for read/write database it's the same URL? 

That will let you consistently use the key.

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
juanmrad: was mostly trying to avoid crashing altogether. Would something like a global variable for storing the last warning? 

```python

_last_read_session_warning_time: float = 0

.....
    except RuntimeError as e:
        current_time = time.time()
        if current_time - _last_read_session_warning_time > 10:
            _last_read_session_warning_time = current_time
            logging.warning(
                "Failed to get read session, falling back to primary: %s", e
            )
        return db.session
```

Like this?

aokolish:
Thanks for working on this. I was just thinking about requesting this feature because it's been an issue for us when we need HMA to scale up quickly - loading the (large) index from Postgres many times concurrently is causing scaling issues for our DB. I'm thinking separating the reads and writes may help (and adding more read replicas).

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
Dcallies: Yes, something like that

juanmrad: Done

hasher-matcher-actioner/src/OpenMediaMatch/app.py
juanmrad: Done

Issue #1902, opened 2026-01-23
[vpdq] Crashes on videos with specific characteristics


marcodimakis:
## Bug Description
VPDQ crashes when processing videos with certain characteristics. Such as the one generated with the command below. I think it is to do with the display vs storage dimensions and padding. Interestingly this seems to not occur on my aarch64 laptop but on a VPS and my main machine I get this output:
`corrupted size vs. prev_size
Aborted (core dumped)`

I've already made a fix at https://github.com/marcodimakis/ThreatExchange/tree/vpdq-avframe-allocation-crash
I made it as minimal as I could. I moved away from manually allocating the AVFrame data using av_image_alloc to using av_frame_get_buffer which also allows the destructor to be smaller.

I'm happy to open a pr although I'm inexperienced and thought opening up and issue first would be best.

## Reproduction Steps
1. Create a test video and run vpdq-hash-video on said video:
`ffmpeg -f lavfi -i "testsrc=duration=5:size=608x1080:rate=60" -vf "pad=1080:1080:236:0" -c:v libx264 -pix_fmt yuv420p output.mp4`

2. Run vpdq-hash-video on the output:
`vpdq-hash-video -r 1 -i output.mp4 -o hash`

PR #1915, opened 2026-01-28
[vPDQ] AVFrame crash fix

marcodimakis:
Summary
---------

Fixes issue #1902. VPDQ crashes when processing videos with certain characteristics, this is due to improper allocation when using the function `av_image_alloc` to fix this I replaced it's use with `av_frame_get_buffer` which allows libav to handle AVFrame allocation and freeing. 

Test Plan
---------
For testing I added another test file  `tmk/sample-videos/test-padded-dimensions.mp4` which when `vpdq/cpp/regtest.py` is ran will try hashing it and if the fix is not in place it will get caught by the try except and exit early. Additionally, I also added it's sample hash `vpdq/sample-hashes/test-padded-dimensions.txt`.
##
As asked in the issue I did attach the output of `ffmpeg` on some systems I tested. The one that passed I do not really know why it does not crash, likely to do with it being on a different architecture and using 16k pages.

[ffmpeg.txt](https://github.com/user-attachments/files/24922525/ffmpeg.txt)

### DIFF
--- a/vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
+++ b/vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
@@ -56,10 +56,7 @@ AVFramePtr createRGB24Frame(size_t const width, size_t const height) {
   frame->width = width;
   frame->height = height;
 
-  // TODO: Alignment is forced to 16, but the alignment should be variable (0).
-  // This requires updating PDQFrameBufferHasher and VpdqHasher to pass the
-  // linesize as the row stride to PDQ (see GH #1918)
-  int ret = av_frame_get_buffer(frame.get(), 16);
+  int ret = av_frame_get_buffer(frame.get(), 0);
   if (ret < 0) {
     char errbuf[AV_ERROR_MAX_STRING_SIZE];
     int strerr = av_strerror(ret, errbuf, sizeof(errbuf));
--- a/vpdq/cpp/vpdq/cpp/hashing/bufferhasher.cpp
+++ b/vpdq/cpp/vpdq/cpp/hashing/bufferhasher.cpp
@@ -21,10 +21,12 @@ const int MIN_HASHABLE_DIM = 5;
 // ----------------------------------------------------------------
 bool PDQFrameBufferHasher::hashFrame(
     unsigned char* buffer,
+    int linesize,
     pdq::hashing::Hash256& hash, // The result pdq hash
     int& quality // Hashing Quality
 ) {
-  if (_frameHeight < MIN_HASHABLE_DIM || _frameWidth < MIN_HASHABLE_DIM) {
+  if (_frameHeight < MIN_HASHABLE_DIM || _frameWidth < MIN_HASHABLE_DIM ||
+      linesize < _frameWidth * 3) {
     hash.clear();
     quality = 0;
     return false;
@@ -35,7 +37,7 @@ bool PDQFrameBufferHasher::hashFrame(
       &buffer[2], // pBbase
       _frameHeight,
       _frameWidth,
-      3 * _frameWidth, // rowStride
+      linesize, // rowStride
       3, // colStride
       _fullLumaImageBuffer1.data());
 
--- a/vpdq/cpp/vpdq/cpp/hashing/bufferhasher.h
+++ b/vpdq/cpp/vpdq/cpp/hashing/bufferhasher.h
@@ -41,7 +41,10 @@ class AbstractFrameBufferHasher {
   virtual int getFeatureDimension() = 0;
 
   virtual bool hashFrame(
-      unsigned char* buffer, pdq::hashing::Hash256& hash, int& quality) = 0;
+      unsigned char* buffer,
+      int linesize,
+      pdq::hashing::Hash256& hash,
+      int& quality) = 0;
 };
 
 /**
@@ -76,6 +79,7 @@ class PDQFrameBufferHasher : public AbstractFrameBufferHasher {
   // Get PDQ Hash in Hash256 format
   bool hashFrame(
       unsigned char* buffer,
+      int linesize,
       pdq::hashing::Hash256& hash,
       int& quality) override;
 };
--- a/vpdq/cpp/vpdq/cpp/hashing/ffmpegwrapper.cpp
+++ b/vpdq/cpp/vpdq/cpp/hashing/ffmpegwrapper.cpp
@@ -126,8 +126,10 @@ bool FFmpegVideo::createSwsContext() {
   return (swsContext.get() != nullptr);
 }
 
-FFmpegFrame::FFmpegFrame(AVFramePtr frame, uint64_t frameNumber)
-    : m_frame(std::move(frame)), m_frameNumber(frameNumber) {}
+FFmpegFrame::FFmpegFrame(AVFramePtr frame, uint64_t frameNumber, int linesize)
+    : m_frame(std::move(frame)),
+      m_frameNumber(frameNumber),
+      m_linesize(linesize) {}
 
 uint64_t FFmpegFrame::get_frame_number() const {
   return m_frameNumber;
@@ -135,6 +137,10 @@ uint64_t FFmpegFrame::get_frame_number() const {
 
 unsigned char* FFmpegFrame::get_buffer_ptr() {
   return m_frame->data[0];
+}
+
+int FFmpegFrame::get_linesize() const {
+  return m_linesize;
 }
 
 } // namespace ffmpeg
--- a/vpdq/cpp/vpdq/cpp/hashing/ffmpegwrapper.h
+++ b/vpdq/cpp/vpdq/cpp/hashing/ffmpegwrapper.h
@@ -99,8 +99,10 @@ class FFmpegFrame {
    *
    *  @param frame The AVFrame.
    *  @param frameNumber The frame number in the video.
+   *  @param linesize The number of bytes per row in the buffer including
+   *padding.
    **/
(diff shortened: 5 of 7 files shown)


Dcallies (commented):
Hey @marcodimakis , thanks for contributing!

Thanks also for generating a novel test image to check against in the future.

Since I'm a C++ dummy, I'll wait to see if @ianwal has time to take a brief look, but as best as I can tell this looks okay.

EDIT: Just need to clang-format your files

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
```diff
@@ -47,17 +47,20 @@ void saveFrameToFile(AVFramePtr frame, const std::string& filename) {
...
   }
 
   frame->format = PIXEL_FORMAT;
   frame->width = width;
   frame->height = height;
 
-  if (av_image_alloc(
-          frame->data, frame->linesize, width, height, PIXEL_FORMAT, 1) < 0) {
-    throw std::bad_alloc();
+  int ret = av_frame_get_buffer(frame.get(), 1);
```
ianwal:
```suggestion
  int ret = av_frame_get_buffer(frame.get(), 0);
```

ffmpeg docs recommends alignment to be chosen automatically (0). Does this need to be 1 or can it be automatically chosen?

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
```diff
@@ -47,17 +47,20 @@ void saveFrameToFile(AVFramePtr frame, const std::string& filename) {
...
+  if (ret < 0) {
+    char errbuf[AV_ERROR_MAX_STRING_SIZE];
+    av_strerror(ret, errbuf, sizeof(errbuf));
```
ianwal:
```suggestion
    int strerr = av_strerror(ret, errbuf, sizeof(errbuf));
    if (strerr != 0) {
        throw std::runtime_error("av_frame_get_buffer failed.");
    }
    ...
```

Handle `av_strerror()` return error to avoid any potential UB on converting it to a string. I don't see any guarantee it will be null-terminated if it fails.

No need to report the status in the exception msg on this error. av_strerror prints some message even if it fails, according to FFmpeg docs.

ianwal:
Reviewed, if CI workflows pass after the suggestions are applied then it should be good to merge.

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
marcodimakis: The issue with using alignment 0 is that it pads the buffers making it so the computed hashes are different to the ones generated by previous versions of pdq, if you run regtest.py with alignment 0 you'll see that a lot of the hashes do not match. I tried avoiding using 1 but I did some benchmarking with a fix I came up with and it seems to only be slower. It should be the same performance wise.

ianwal: I'll look more into this alignment issue before we merge. I want to avoid unaligned memory access faults for architectures/SIMD where that may occur.

I'm not sure why the buffer alignment would affect the actual frame data in the buffer unless there is a bug somewhere else in the code. I think a bug elsewhere is more likely.

ianwal: Just a pure guess, we're probably missing the frame linesize when iterating over the frame data and assuming all the lines are adjacent in memory. I think that would be in PDQ so it might require a small PDQ update to add this as a parameter somewhere. Alternatively, make a copy of the frame and smush them adjacent before passing to PDQ (probably bad performance).

marcodimakis: > I'll look more into this alignment issue before we merge. I want to avoid unaligned memory access faults for architectures/SIMD where that may occur.
> 
> I'm not sure why the buffer alignment would affect the actual frame data in the buffer unless there is a bug somewhere else in the code. I think a bug elsewhere is more likely.

Yeah you were mostly right it had to do with the linesize. The alignment pads for SIMD optimization, importantly it does that by appending each row with padding. The issue is that in `bufferhasher` in the function `fillFloatLumaFromRGB` row stride is set to `3 * _frameWidth, // rowStride` since though that is not the case it was reading the padding as frame data. 

It is pretty late for me I can have a fix done tomorrow, I think all that needs to get done is to pass the m_frame->linesize[0] down to fillFloatLumaFromRGB and use it as the rowStride.

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
```diff
@@ -47,17 +47,21 @@ void saveFrameToFile(AVFramePtr frame, const std::string& filename) {
...
 
-  if (av_image_alloc(
-          frame->data, frame->linesize, width, height, PIXEL_FORMAT, 1) < 0) {
-    throw std::bad_alloc();
+  int ret = av_frame_get_buffer(frame.get(), 1);
+  if (ret < 0) {
+    char errbuf[AV_ERROR_MAX_STRING_SIZE];
+    int strerr = av_strerror(ret, errbuf, sizeof(errbuf));
+    if (strerr != 0) {
+      throw std::runtime_error("av_frame_get_buffer failed.");
```
ianwal, line 64: Add `#include <stdexcept>` for std::runtime_error

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
ianwal: Yeah the rowStride in fillFloatLumaFromRGB() will also need to be fixed so that it matches the linesize. The fix isn't that hard, but it will be kind of messy and I no longer see a need to fix this right now since the existing solution already forces an alignment of 1.

Let's just add a TODO here to change the alignment to `0` so this fix isn't blocked by that. I filed a bug for this #1918.

Please add this comment above the call to av_frame_get_buffer():

```cpp
// TODO: Alignment is forced to 1, but the alignment should be variable (0).
// This requires updating PDQFrameBufferHasher and VpdqHasher to pass the linesize as the row stride to PDQ (see GH #1918)
```  

Thank you for investigating this!

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
```diff
@@ -47,17 +48,24 @@ void saveFrameToFile(AVFramePtr frame, const std::string& filename) {
...
-          frame->data, frame->linesize, width, height, PIXEL_FORMAT, 1) < 0) {
-    throw std::bad_alloc();
+  // TODO: Alignment is forced to 1, but the alignment should be variable (0).
+  // This requires updating PDQFrameBufferHasher and VpdqHasher to pass the
+  // linesize as the row stride to PDQ (see GH #1918)
+  int ret = av_frame_get_buffer(frame.get(), 1);
+  if (ret < 0) {
+    char errbuf[AV_ERROR_MAX_STRING_SIZE];
+    int strerr = av_strerror(ret, errbuf, sizeof(errbuf));
+    if (strerr != 0) {
```
ianwal, line 63: In addition to this, throw an exception with the errbuf string if strerr == 0 like what you had originally. I wanted both cases so that we weren't trying to print a bad string if av_strerror failed.

vpdq/cpp/vpdq/cpp/hashing/ffmpegutils.cpp
```diff
@@ -47,17 +48,26 @@ void saveFrameToFile(AVFramePtr frame, const std::string& filename) {
...
   frame->width = width;
   frame->height = height;
 
-  if (av_image_alloc(
-          frame->data, frame->linesize, width, height, PIXEL_FORMAT, 1) < 0) {
-    throw std::bad_alloc();
+  // TODO: Alignment is forced to 1, but the alignment should be variable (0).
+  // This requires updating PDQFrameBufferHasher and VpdqHasher to pass the
+  // linesize as the row stride to PDQ (see GH #1918)
+  int ret = av_frame_get_buffer(frame.get(), 1);
```
ianwal:
```suggestion
  int ret = av_frame_get_buffer(frame.get(), 16);
```

Sorry, one last thing. If an alignment of 16 passes all tests let's use that to be safe. Please update the comment above to say `Alignment is forced to 16` to match.

I really want to avoid any potential alignment issues. And even though it was previously 1 I don't have enough knowledge of FFmpeg to know if the behavior will be the same between `av_image_alloc` and `av_frame_get_buffer()` with an alignment of 1.

[A prominent FFmpeg project, pyav, also uses 16 for their alignment](https://github.com/PyAV-Org/PyAV/blob/1746ef49c0f4e9d77eb762e52ffe6a80b90af471/av/video/frame.py#L222-L228), so I think this is a safe bet.

marcodimakis: I have a fix for the entire thing if u want I can commit it. Or change the alignment to 16 if that works then open a new pr up to you. Current implementation matches the original in terms of the buffer and such.

ianwal: If CI passes then updating to 16 in this PR is fine.

marcodimakis:
I'm sorry for all the mess. I ran the tests myself and they all are passing so should have no more issues, I formatted them too. @ianwal I would like if you could review my validity check in `bufferhasher.cpp` I do cannot think of a reason linesize would have to be smaller than the frame's width * 3. This also closes issue #1918.

PR #1924, opened 2026-02-03
[vpdq] [pdq] Misc cleanup to improve portability

ianwal:
Summary
---------

Miscellaneous cleanup that I noticed in the repo following C++ best known practices. Some of this stuff was bugging me for a while since I didn't do a great job on the first pass a few years ago when I was still new to C++.

No functional or behavioral changes intended in this PR beyond improving portability and documentation. Any lines > 2-3 lines are copied/moved from another file untouched.

- Remove using-directives (`using namespace foo`) to avoid namespace clashes
  - This has caused issues in vpdq in the past, e.g., POSIX getline() and std::getline()
- Cleanup includes to include what is used
- Remove `M_PI` usage, define pi variable instead
  - M_PI is not standard C++, and there's no pi until C++20's [std::numbers::pi](https://en.cppreference.com/w/cpp/numeric/constants.html), so defining it directly is the most portable solution by far. M_PI causes serious pain across compilers, especially if used in a header file.
- Split out some unrelated classes into their own header files
- Add Doxygen
- Follow rule-of-five and rule-of-zero
- Remove custom declaration for `uint8_t` and use `<cstdint>` to avoid declaration collisions
  - This can cause issues if <cstdint> has a different declaration for uint8_t. Any C++ compiler in the last 25 years will support `<cstdint>` no problem, even odd embedded compilers.
- Mark class methods const where possible
- Copied vendorized fixes in PDQ files to main PDQ files
- Fix function parameter names

Normally I avoid making PRs this large, but it doesn't make much sense to split up most of these changes.

Test Plan
---------

CI should cover all changes.


pdq/cpp/hashing/pdqhashing.cpp
```diff
@@ -6,28 +6,24 @@
...
-
-using namespace std;
-
 namespace facebook {
 namespace pdq {
 namespace hashing {
 
 namespace {
 
+constexpr double pi = 3.14159265358979323846;
```
Dcallies, line 25: Dumb question - why not use math.h or `<numbers>`?

tmk/cpp/algo/tmkfv.cpp
```diff
@@ -11,17 +11,15 @@
...
-#endif
-
 #include <cmath>
 #include <stdexcept>
 
 namespace facebook {
 namespace tmk {
 namespace algo {
 
+static const double pi = 3.14159265358979323846;
```
Dcallies, line 21: ditto

vpdq/cpp/pdq/cpp/hashing/pdqhashing.cpp
```diff
@@ -6,28 +6,24 @@
...
-
-using namespace std;
-
 namespace facebook {
 namespace pdq {
 namespace hashing {
 
 namespace {
 
+constexpr double pi = 3.14159265358979323846;
```
Dcallies, line 25: tripple ditto

pdq/cpp/hashing/pdqhashing.cpp
ianwal: As crazy as it sounds, there's no standard C++ pi constant until C++20 when they finally added [std::numbers::pi_v](https://en.cppreference.com/w/cpp/numeric/constants.html) (and pi). We are supporting C++14 (which is reasonable), so we can't use it yet.

PR #1929, opened 2026-02-04
Integrate GPT OSS Safeguard into ThreatExchange as chat command

ZhiyLiu:


python-threatexchange/threatexchange/cli/chat_cmd.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python
```
Dcallies: Please the chat command

python-threatexchange/threatexchange/classifier/safeguard/policy/basic_policy.md
```diff
@@ -0,0 +1,38 @@
+# Basic OSS Safeguard policy (hackathon starter)
```
Dcallies, line 1: Please move both the classifier and policy for oss into a new folder nuder classifiers

Issue #1925, opened 2026-02-03
[hma] Add notes to hashbank content


sreeprasad:
## Summary

  Add the ability to attach notes to hashes when adding content to a bank. Notes provide context about why content was
  added (e.g., "Reported by user X on 2024-01-15" or "Part of campaign ABC").

  ## Proposed Changes

  ### Database
  - Add a `note` TEXT field to the `BankContent` model

  ### API
  - Accept optional `note` parameter when adding content via:
    - `POST /bank/<name>/content` (file/URL upload)
    - `POST /bank/<name>/signal` (direct hash)
  - Return `note` in `GET /bank/<name>/content/<id>` response
  - Allow updating `note` via `PUT /bank/<name>/content/<id>`

  ### UI
  - Add note textarea to the "Add to Bank" modal (all three tabs: File, URL, Hash)
  - Display notes when viewing content details
  - Allow editing notes on existing content

  ## Tasks

  1. **DB Schema** - Add `note` column to `bank_content` table + migration
  2. **Storage Layer** - Update `BankContentConfig`, `bank_add_content()`, `bank_content_get()`, `bank_content_update()`
  3. **API Schemas** - Add `note` field to `BankedContentMetadata` and `BankContentResponse`
  4. **API Endpoints** - Update curation.py to handle notes
  5. **Frontend Form** - Add textarea to `add_to_bank_form.html.j2`
  6. **Frontend JS** - Update `bank-management.js` to send/display notes
  7. **View/Edit UI** - Add UI to view and edit notes on existing content
  8. **Tests** - Add test coverage for notes CRUD

  ## Design Decision

  Starting with a single `note` TEXT field on `BankContent`.   This is simplest approach. All notes will be overwritten with latest note.

  ## Questions for Maintainers

  - Should notes have a character limit?
  - Should notes be included when exporting/syncing via exchanges?
  - Any preference on PR structure (stacked small PRs vs single feature PR)?

julietshen:
Hi @sreeprasad ! Any update on this one?

PR #1931, opened 2026-02-05
  [hma] Add note field to BankContent with storage layer support

sreeprasad:
Part of #1925 - Add notes to hashbank content

  Changes:
  - Add  column (VARCHAR 255) to bank_content table
  - Add database migration
  - Update BankContentConfig dataclass with note field
  - Update storage implementation (add, get, update)
  - Update mocked storage
  - Add unit tests for note CRUD operations
  - Fix devcontainer Dockerfile yarn GPG key issue

Summary
---------

Adds a `note` field to BankContent allowing users to attach context when adding content to a bank (e.g., "Reported by user X" or "Part of campaign ABC").

This PR implements the database and storage layer changes:
  - New `note` column (VARCHAR 255) on `bank_content` table
  - Database migration
  - Updated storage interface and implementation
  - Unit tests for note CRUD operations

  Also fixes a yarn GPG key issue in the devcontainer Dockerfile.

  Part of #1925

Test Plan
---------

- All 46 existing tests pass
- Added 2 new tests:
  - `test_bank_content_note` - tests add, read, update, and clear note
  - `test_bank_content_note_max_length` - tests 255 character limit
- Tested locally in devcontainer with `py.test`

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_postgres_storage_impl.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/tests/test_postgres_storage_impl.py
@@ -412,6 +412,7 @@ def test_bank_content_note_max_length(storage: DefaultOMMStore) -> None:
         config=content_config,
     )
     content_configs = storage.bank_content_get([content_id])
+    assert content_configs[0].note is not None
     assert len(content_configs[0].note) == 255
 
     # Test with 256 characters (should fail)


hasher-matcher-actioner/.devcontainer/Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM mcr.microsoft.com/vscode/devcontainers/python:3.13
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.13-bullseye
 
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+RUN rm -f /etc/apt/sources.list.d/yarn.list \
```
Dcallies, line 4: blocking q: What is this change for? Consider documenting what this is for.

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_postgres_storage_impl.py
```diff
@@ -340,3 +340,76 @@ def test_exchange_get_data(storage: DefaultOMMStore):
 
     val = storage.exchange_get_data(cfg.name, example_key)
     assert val == example_metadata
+
+
+def test_bank_content_note(storage: DefaultOMMStore) -> None:
```
Dcallies, line 345: ignorable: You probably could get away with adding setting the note to the existing tests.

A pattern I usually think works and is compact is:

1. Test creating with all defaults
2. Test setting all settable fields
3. Test setting one field, all other fields are unset
4. Test creating with all field set

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_postgres_storage_impl.py
```diff
@@ -340,3 +340,76 @@ def test_exchange_get_data(storage: DefaultOMMStore):
...
+
+    # Test clearing a note (setting to None)
+    content_configs[0].note = None
+    storage.bank_content_update(content_configs[0])
+
+    content_configs = storage.bank_content_get([content_id_with_note])
+    assert content_configs[0].note is None
+
+
+def test_bank_content_note_max_length(storage: DefaultOMMStore) -> None:
```
Dcallies, line 394: blocking: Add a case that demonstrates the behavior creating over the max length. The test as written I don't think tests anymore more than the previous test does

hasher-matcher-actioner/.devcontainer/Dockerfile
sreeprasad: The base image ships with a `yarn apt` source list that's expired/invalid, which causes `apt-get update` to fail
during the Docker build. Removing it before the update fixes the build. Added a comment in the Dockerfile to
document this.

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_postgres_storage_impl.py
sreeprasad: Added a test case that attempts to save a 256-character note and asserts it raises an exception, demonstrating the max length is actually enforced.

hasher-matcher-actioner/.devcontainer/Dockerfile
Dcallies: I was briefly experiencing this for the range of time you authored this, but it I believe it is clear now. If you don't have time to test it, I can test removing it after you land.

Dcallies:
@sreeprasad - I am suspicious of the test failures on bank API add, can you check locally?

hasher-matcher-actioner/src/OpenMediaMatch/migrations/versions/a1b2c3d4e5f6_add_note_to_bank_content.py
```diff
@@ -0,0 +1,26 @@
...
+Create Date: 2024-02-03 00:00:00.000000
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = "a1b2c3d4e5f6"
+down_revision = "21cb8a3df884"
```
Dcallies, line 14: Yoiu are going to collide with #1949 - @juanmrad - do you know offhand how to properly stack these?

PR #1968, opened 2026-04-14
[HMA] Add explicit connection close for signal ops and clear temp refs

juanmrad:
Summary
---------

Attempts to fix memory leaks in the index lifecycle that cause worker RSS to climb steadily with each index refresh (see #1813).

Close `raw_conn` in `load_signal_index()` and `commit_signal_index()` previously these database connections were never returned to the pool, leaking on every refresh cycle. 

Clear `built_index` reference and `signal_list` before calling `trim_process_memory()` so `gc.collect()` can actually reclaim them without waiting for python to clear reference.

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
@@ -84,6 +84,8 @@ def build_index(
         index_store.store_signal_type_index(for_signal_type, built_index, checkpoint)
     finally:
         # Force garbage collection to reclaim memory and attempt to free pages
+        # explicitly free the built index before reclaiming memory
+        built_index = None
         trim_process_memory(logger, "Indexer")
 
     logger.info(
@@ -114,6 +116,9 @@ def _prepare_index(
     # Build index
     index_cls = for_signal_type.get_index_cls()
     built_index = index_cls.build(signal_list)
+
+    # explicitly free the signal list before returning
+    signal_list.clear()
 
     # Create checkpoint
     checkpoint = SignalTypeIndexBuildCheckpoint.get_empty()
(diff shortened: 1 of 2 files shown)


Dcallies:
Have we figured out a way to collect evidence on what the source of memory is for longer running processes? At least internally we have some tools that allow us to inspect the memory usage.

There are other things that we could do to flail at the problem such as rewriting index build to try and make it incremental, and only rebuild once the delta starts to get too high. In practice I think this reduces the need to rebuild the index at all to zero for most workloads.

hasher-matcher-actioner/src/OpenMediaMatch/background_tasks/build_index.py
```diff
@@ -115,6 +117,9 @@ def _prepare_index(
     index_cls = for_signal_type.get_index_cls()
     built_index = index_cls.build(signal_list)
 
+    # explicitly free the signal list before returning
+    signal_list.clear()
```
Dcallies, line 121: I'm very suspicious this does anything, because it requires that the local variable is not properly garbage collected when it leaves scope, which stretches my understanding about how refcounting works. 

Do we have any evidence that this is needed?

Random googling:
1. https://stackoverflow.com/questions/28324084/does-deleting-reassigning-a-python-list-allow-for-garbage-collection
2. https://blog.codingconfessions.com/p/cpython-garbage-collection-internals

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -495,28 +495,34 @@ def commit_signal_index(
...
+                        self.serialized_index_large_object_oid,
+                        level=logging.WARNING,
+                    )
+
+            self.serialized_index_large_object_oid = l_obj.oid
+            db.session.add(self)
+            raw_conn.commit()
+        finally:
+            # explicitly close the raw connection to free memory
+            raw_conn.close()
```
Dcallies, line 525: This might work too: https://docs.python.org/3/library/contextlib.html#contextlib.closing which lets you use a with statement. 

Reading https://docs.sqlalchemy.org/en/21/core/connections.html#working-with-the-dbapi-cursor-directly I think this is very likely missing the close

Dcallies (approved):

PR #1969, opened 2026-04-15
[DRAFT] Test memory leak improvement by moving index to subprocess

juanmrad:
Summary
---------

DO NOT MERGE. Mostly for testing. this moves the index to a subprocess which guarantees the OS will reclaim memory after process exits when it completes index building.

Test Plan
---------


Dcallies:
Thanks for continuing to crack on this one!

PR #1971, opened 2026-04-16
[HMA] Swap to use jemalloc to reduce memory fragmentation

juanmrad:
Summary
---------

Switch the Docker image's memory allocator from glibc malloc to jemalloc to address the steady RSS growth observed during index rebuild cycles (#1813).

FAISS indexes allocate large C++ buffers that are interleaved with small Python objects on glibc's heap. When freed, glibc cannot return those pages to the OS if any live allocation sits on the same page, and `malloc_trim(0)` can only reclaim from the top of the brk heap. 

jemalloc uses per-size-class arenas that avoid this interleaving and aggressively returns pages via `madvise(MADV_DONTNEED)`. It's a drop-in ABI-compatible replacement making this safe while helping memory growth.

### DIFF
--- a/hasher-matcher-actioner/.devcontainer/Dockerfile
+++ b/hasher-matcher-actioner/.devcontainer/Dockerfile
@@ -1,8 +1,11 @@
 FROM mcr.microsoft.com/vscode/devcontainers/python:3.13
 
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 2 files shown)


juanmrad:
@Dcallies are we ok with me merging this?

Dcallies (approved):
Slightly nervous about more dependencies, but I am hoping that this doesn't actively harm us.

Issue #1893, opened 2025-11-19
[hma] Clean up documentation story


cassidyjames:
We've heard from HMA adopters that the current state of HMA documentation is a bit messy to follow. Right now there are a bunch of places to look for information:

- [README.md](https://github.com/facebook/ThreatExchange/tree/main/hasher-matcher-actioner#readme) in the HMA subfolder
- [/docs](https://github.com/facebook/ThreatExchange/tree/main/hasher-matcher-actioner/docs) in the HMA subfolder
- This repo's [wiki](https://github.com/facebook/ThreatExchange/wiki)
- [ThreatExchange GitHub Pages](https://facebook.github.io/ThreatExchange/)
- [ThreatExchange docs on Meta for Developers](https://developers.facebook.com/docs/threat-exchange/)
- API documentation within the web app (#1891)

It could help adoption and adopter productivity to clean this up a bit. :) My proposal, if possible, would be to:

- [ ] Consolidate HMA technical documentation _either_ under `/docs` _or_ on the wiki
  - [ ] Update the README to be a bit more concise, moving its technical documentation over to the rest of the docs
  - [ ] If under `/docs`, optionally spin up a static docs site to make that a bit more browsable
- [ ] Retire the old GitHub Pages site
- [ ] Link to the canonical HMA docs location from the ThreatExchange docs
- [ ] Host the generated API documentation… somewhere, and cross-link between it and the new canonical docs location

I'm almost certainly missing some context, but I'd love to hear from other contributors and adopters if this (or an even better!) plan would make sense and is something we can work towards.

ThisIsMissEm:
> /docs directory

Yeah, if my memory serves, this was largely just to get us *something* as opposed to nothing.

>  so I don't think it makes sense to point to HMA docs from the ThreatExchange API either.

Agreed.

---

On tooling, I'd strongly recommend Vitepress, as it's just markdown but rather easy to configure and theme the built documentation site output. We can also support having documentation for each part of this repository that way too. (Example: [FediMod FIRES site](https://fires.fedimod.org/) and [the source](https://github.com/fedimod/fires/tree/main/docs))

As it's just markdown, pages can also be automatically generated in a build step if necessary, and there's an OpenAPI plugin available too.

cassidyjames:
@ThisIsMissEm that sounds great to me! I feel like we could probably bikeshed which docs generator to use forever, but those Vitepress docs look really nice and the straight-up Markdown source makes me happy. 😁 

[...]

PR #1973, opened 2026-04-22
Scaffold out initial documentation website

ThisIsMissEm:
# Summary

This takes on the next steps of #1893, by starting to build out a documentation website for the entire ThreatExchange Project

# Remaing Tasks
- [x] Fix any remaining style issues (there's a few in the header on mobile)
- [x] Clean up documentation pages
- [x] Add deployment script ([CD Workflow](https://github.com/fedimod/fires/blob/main/.github/workflows/deploy-website.yml), [CI Workflow](https://github.com/fedimod/fires/blob/main/.github/workflows/check-website.yml))

# Future
- There is a plugin for doing OpenAPI: https://vitepress-openapi.vercel.app/guide/getting-started.html
  - I don't know a whole lot about it, but looks like we'd need to generate the openapi.json file for it to consume?
- We can add a [development team page](https://vitepress.dev/reference/default-theme-team-page), if we want

# Screenshots

## Homepage

homepage

## Hashing Algorithms

image

## HMA Documentation

image

Open Questions
----------------

- Do we actually want to use the Facebook Blue on the ThreatExchange project website?
- Do we actually want this fuller website for the project?

Live Preview
-------------

Temporary: https://brandedcode.com/Facebook-ThreatExchange/


cassidyjames:
omg I've lost this comment three times because apparently GitHub doesn't save the buffer anymore when switching between the conversation and files-changed tabs 😭

Thank you so much for taking this on @ThisIsMissEm, it looks great!

I have Opinions™ about the custom styling; I'd personally lean towards just setting the brand color and then leaving the rest alone—otherwise I fear we'll spend forever nit-picking custom CSS in perpetuity, especially along with VitePress upgrades. I'm open to hearing others' thoughts here though and honestly don't feel _that_ strongly. You can preview the default styles on the [VitePress site](https://vitepress.dev/) itself for reference.

Otherwise, my only other major thoughts are around iteration and copy writing/approval process. I'd hate for us to get blocked from shipping a starting HMA docs site because we have too much new/changed copy for someone to sign off on; instead, I wonder if we can break this down into a few phases: 

1. start by just moving the existing HMA docs and shipping the site (optionally linking directly to existing component READMEs if we want), then 

2. start moving each other component over (cleaning up their READMEs for concision and linking to the new docs location) one-by-one in their own PRs, then 

3. finally work on any more general/broad web content we want to include on the site alongside project-specific docs

cassidyjames:
Ah, one more small thought/question (probably for @Dcallies): for the docs site and this repo more generally, would it be helpful to describe it as "ThreatExchange Projects" to help differentiate from the [docs for the hosted ThreatExchange platform](https://developers.facebook.com/docs/threat-exchange)? I know I've personally gotten confused about this more than once, so I'd probably frame these docs and repo like:

> **ThreatExchange Projects**
> Documentation
> _Open source projects for hashing, signal exchange, and content moderation including PDQ, TMK, vPDQ, & HMA_

(or whichever projects/components make sense to list out)

ThisIsMissEm:
@cassidyjames I think what we can initially do on each page being fleshed out is to just have a 

> [!NOTE]
> This page is currently being worked on, view the [original version](#)

Where the link takes you to the GitHub README's or something?

website/.vitepress/config.mts
```diff
@@ -0,0 +1,88 @@
...
+        text: "Hashing Algorithms",
+        items: [
+          { text: "PDQ", link: "/hashing/pdq" },
+          { text: "TMK", link: "/hashing/tmk" },
+        ],
+      },
+      { text: "Python Library", link: "/pytx" },
+      { text: "HMA", link: "/hma" },
+      {
+        text: "ThreatExchange Program",
```
ThisIsMissEm:
```suggestion
        text: "ThreatExchange Platform",
```

@cassidyjames I guess this link should be changed?

cassidyjames: I'm not sure about the proper terminology 😅

cassidyjames:
@ThisIsMissEm yeah placeholder pages could work. I do lean towards just linking out directly at first, though, since that way it will be more obvious it's going to take you somewhere else (with the little `↗` external link icon in the sidebar/navbar). That feels a bit more honest if there's not actually any content for that component on the docs site, yet.

website/.vitepress/config.mts
Dcallies: Platform is probably closer than program - program we have previously used to refer to threatexchange.meta.com

Dcallies (commented):
@ThisIsMissEm - some questions
1. How would you like this reviewed? I've never used a docsite, so other than saying "seems like the right programming language, let's ship it and see what happens" I'm not sure how much I'm able to add.
2. Who will fill out the rest of the content? I'm about to go out on pat leave and might be sparse until the fall.

website/.vitepress/config.mts
```diff
@@ -0,0 +1,88 @@
...
+              text: "PDQ",
+              link: "/",
+              base: "/hashing/pdq/",
+              items: [
+                { text: "Algorithm", link: "/algorithm" },
+                { text: "Implementations", link: "/implementations" },
+              ],
+            },
+            { text: "TMK", base: "/hashing/tmk/", link: "/" },
+            { text: "vPDQ", base: "/hashing/vpdq/", link: "/" },
```
Dcallies: nit: Lets swap these so that tmk is last
```suggestion
            { text: "vPDQ", base: "/hashing/vpdq/", link: "/" },
            { text: "TMK", base: "/hashing/tmk/", link: "/" },

```

ThisIsMissEm:
@Dcallies 

For 1. It's mostly do you like the output, does it look right for the project? If you want I can deploy it to my own GitHub Pages so you can click through if you've not run it directly.

For 2., we can enable "edit this page" links and collaboratively work on it; the goal at this stage is to get the site setup, afterwards we can refine and backfill content.

Dcallies (approved):
I'm fine with landing and iterating. Let me know if you want to merge as is or do something else.

ThisIsMissEm:
@Dcallies the only outstanding question I have is: are we okay using the facebook blue in the navigation bar? That felt contentious potentially but it did feel right. I can also change to a different colour if that'd be better. I did want to anchor back to the Facebook roots a bit with that.

ThisIsMissEm:
I can add the remaining TODOs above in the description either now or later, happy to iterate in smaller chunks

Dcallies:
No concerns about blue, we can always change colors later. Let me know when you are ready to merge.

ThisIsMissEm:
@Dcallies if you can get approval for github pages before your leave, then I can setup the workflow for publishing to that. Otherwise, I'll do it as a follow up PR.

Dcallies:
Is there a repo setting I need to turn on, or are you talking about process?

To my knowledge there is not any internal process for where the repo docs are hosted as long as it's part of GitHub. I'm not super familiar with docsites, are you saying that we need a domain and it's not part of GitHub?

juanmrad:
@Dcallies basically you have to go on the repo settings under pages and you will see like this:

Screenshot 2026-05-01 at 6 00 41 PM

And then you select Actions. And may look like this:
Screenshot 2026-05-01 at 6 00 21 PM

That way deployment of the page is a github workflow that builds and publishes.

It will be a GitHub domain by default unless you want it to be a different domain.

.github/workflows/website-release.yaml
```diff
@@ -0,0 +1,77 @@
...
+  run:
+    working-directory: website
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'ThisIsMissEm/Facebook-ThreatExchange' }}
+    env:
+      PAGES_PATH: /Facebook-ThreatExchange/
+    steps:
```
ThisIsMissEm: These need to be changed before merge.

ThisIsMissEm:
@Dcallies I know we've had previous conversations about the user-interface images, now that we have a documentation website, I think it makes a lot more sense to inline those images in the repository to make editing much easier.

Converting these from png to webp too these from 3.9 MB → 680 KB in total (an 83% reduction). I'm certainly not inclined to be worried about an extra 680 KB in this repository?

ThisIsMissEm:
Just a small disclosure: I did use claude opus 4.7 for helping with automating the inlining & conversion of images to webp, and with fixing the vitepress styles that were broken (it would've taken me much longer due to the structure of the default theme to fix by hand)

.github/workflows/website-ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - website/**
+      - .github/workflows/deploy-website.yaml
+  push:
+    branches-ignore:
```
ThisIsMissEm, line 11: As far as I understand it, this will run on pull requests that target main, and then on pushes to any branch that isn't main.

.github/workflows/website-release.yaml
ThisIsMissEm:
```suggestion
    if: ${{ github.repository == 'Facebook/ThreatExchange' }}
    env:
      PAGES_PATH: /ThreatExchange/
    steps:
```

.github/workflows/website-ci.yaml
juanmrad: No this will actually mean we do not run CI on merge. You can see example of HMA-Ci: https://github.com/facebook/ThreatExchange/blob/main/.github/workflows/hma-ci.yaml#L4-L16 

This runs both on PR and when a push is done i.e. merged.

ThisIsMissEm: Yeah, we run website-release on merge anyway which does a build of the website as part of that process. Hence ignoring this workflow on push to main.

PR #1974, opened 2026-04-30
[hma] Add /status/live 

### LINKED ISSUE #1905: [hma] Split health checkpoint into liveness and readiness
There is one health checkpoint in HMA right now. 

It can be valuable to have two separate health checkpoints for liveness (i'm alive) and readiness (send me stuff). This is helpful for navigating scaling issues and challenges.

Thanks to @ThisIsMissEm for pointing to https://github.com/facebook/ThreatExchange/blob/main/hasher-matcher-actioner/src/OpenMediaMatch/app.py#L260

Context: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

aokolish:
## Summary

Splits the combined `/status` health check into separate `/livez` (liveness) and `/readyz` (readiness) endpoints. Closes #1905.

### Why the previous fix didn't work

#1912 disambiguated the cold-start case (`INDEX-NOT-LOADED`) from staleness (`INDEX-STALE`) in the response *body*, but most health-check clients only inspect the response *status code*. From their perspective both still return `503`, so pointing both a liveness and a readiness check at `/status` is effectively the same check, and during index reloading those `503` responses can cause a restart loop before the index can finish loading.

### What this PR does

- **`/livez`** — minimal liveness. Failing this should mean the instance needs to be restarted.
- **`/readyz`** — readiness. Mirrors the existing `/status` checks
- **`/status`** — left untouched for backwards compatibility.

### Migration

Example (Kubernetes):

```yaml
livenessProbe:
  httpGet:
    path: /livez
    port: 5100
readinessProbe:
  httpGet:
    path: /readyz
    port: 5100
```

## Test plan

- added unit tests
- built dev container and ran curl against all 3 status endpoints

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/app.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/app.py
@@ -265,14 +265,12 @@ def create_app() -> OpenAPI:
             "200": {"description": "Service is alive"},
             "503": {"description": "Service is not ready"},
         },
-        summary="Health check (deprecated)",
+        summary="Health check",
         description=(
-            "Combined liveness/readiness check. Deprecated: prefer /livez for"
-            " liveness checks and /readyz for readiness checks. Health-check"
-            " clients that act on HTTP status codes alone (most load balancers"
-            " and orchestrators) cannot distinguish the failure modes this"
-            " endpoint reports, so using it as a liveness check causes restart"
-            " loops during cold start while the index is still loading."
+            "Returns 200 if the server is healthy and should serve traffic."
+            " If you need a separate check that the server is alive, check"
+            " for a non-empty return from this endpoint or use a 200 return"
+            " from /status/live."
         ),
     )
     def status():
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -279,6 +286,49 @@ def status():
                 return "INDEX-STALE", 503
         return "I-AM-ALIVE", 200
 
+    @app.get(
+        "/livez",
```
Dcallies: Can we nest this under status?

`/status/live`

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -279,6 +286,49 @@ def status():
...
+            " restart the instance during cold start before the index has"
+            " loaded. Failing this should imply the process is wedged and"
+            " needs to be restarted."
+        ),
+    )
+    def livez():
+        return "OK", 200
+
+    @app.get(
+        "/readyz",
```
Dcallies: Can we nest this under status?

`/status/ready`

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -279,6 +286,49 @@ def status():
...
+        tags=[Tag(name="Core")],
+        responses={"200": {"description": "Service is alive"}},
+        summary="Liveness check",
+        description=(
+            "Liveness check. Returns 200 if the process can serve HTTP. Does"
+            " not check index state - this is intentional, so that"
+            " health-check clients acting on status codes alone do not"
+            " restart the instance during cold start before the index has"
+            " loaded. Failing this should imply the process is wedged and"
+            " needs to be restarted."
```
Dcallies: Since this appears to be motivated by kubernetes needs, I would have expected to see that mentioned here.

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -265,8 +265,15 @@ def home():
             "200": {"description": "Service is alive"},
             "503": {"description": "Service is not ready"},
         },
-        summary="Health check",
-        description="Liveness/readiness check",
+        summary="Health check (deprecated)",
```
Dcallies:
```suggestion
        summary="Health check",
```

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -279,6 +286,49 @@ def status():
...
+            "Readiness check. Returns 503 when a matcher has not yet loaded"
+            " its index (INDEX-NOT-LOADED) or when the cached index is stale"
+            " (INDEX-STALE), and 200 READY otherwise. Instances failing this"
+            " should be removed from the load balancer's serving set but not"
+            " restarted - restarting will not refresh a stale cache and may"
+            " cascade if the underlying cause is shared infrastructure (e.g."
+            " the database)."
+        ),
+    )
+    def readyz():
```
Dcallies: blocking: Wait, this seems to be the same implementation as the original status. Why don't we leave the old one so you don't need to deprecate it?

hasher-matcher-actioner/src/OpenMediaMatch/app.py
```diff
@@ -265,8 +265,15 @@ def home():
...
-        summary="Health check",
-        description="Liveness/readiness check",
+        summary="Health check (deprecated)",
+        description=(
+            "Combined liveness/readiness check. Deprecated: prefer /livez for"
+            " liveness checks and /readyz for readiness checks. Health-check"
+            " clients that act on HTTP status codes alone (most load balancers"
+            " and orchestrators) cannot distinguish the failure modes this"
+            " endpoint reports, so using it as a liveness check causes restart"
+            " loops during cold start while the index is still loading."
```
Dcallies:
```suggestion
            "Returns 200 if the server is healthy and should serve traffic. If you need a separate check that the server is alive, check for a non-empty return from this endpoint or use a 200 return from /status/live"
```

Dcallies (approved):
Since the ready endpoint is the same as status, let's keep using that instead of creating a new endpoint, then you don't need to go through the steps of deprecated it. Other comments inline.

ThisIsMissEm (commented):
LGTM

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_api.py
```diff
@@ -72,6 +72,42 @@ def fake_cache() -> matching.IndexCache:
     assert response.data == b"I-AM-ALIVE"
 
 
+def test_status_live_always_alive(client: FlaskClient, monkeypatch: MonkeyPatch):
```
Dcallies, line 75: Yay tests!

Dcallies (approved):

PR #1981, opened 2026-05-13
[HMA] fix: route large object reads through read replica

aokolish:
Summary
---------
I recently tried out the new config to split reads and writes and it didn't have much effect.

That's because `load_signal_index()` and `index_lobj_exists()` bypass read/write session splitting — they use `db.engine.raw_connection()` and `db.session` directly, so the heaviest read path (index downloads) always hits the writer even when a read replica is configured.  during deploys we saw 93% writer CPU while the replica sat at 4-5%. This fixes that by routing both methods through the read engine/session.

Test Plan
---------
Added regression tests that assert `load_signal_index()` calls `db.engines["read"].raw_connection()` and `index_lobj_exists()`
defaults to `get_read_session()`

### DIFF
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
@@ -553,11 +553,11 @@ class SignalIndex(db.Model):  # type: ignore[name-defined]
         Reads the serialized index from the read replica via large object API.
 
         When using read/write replicas, as long as we read this object
-        and its ID from the same db (including replicas), the lobj should be there, 
+        and its ID from the same db (including replicas), the lobj should be there,
         as the lobj is written before the record with the ID.
-        
+
         However, it is still possible to get an error due to races with the
-        garbage collection process that deletes lobjs. 
+        garbage collection process that deletes lobjs.
         """
         oid = self.serialized_index_large_object_oid
         assert oid is not None
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
@@ -455,7 +455,7 @@ class SignalIndex(db.Model):  # type: ignore[name-defined]
 
     serialized_index_large_object_oid: Mapped[int | None] = mapped_column(OID)
 
-    def index_lobj_exists(self, session=None) -> bool:
+    def index_lobj_exists(self, session: t.Optional[Session] = None) -> bool:
         """
         Return true if the index lobj exists and load_signal_index should work.
 
--- a/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
+++ b/hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
@@ -37,6 +37,7 @@ from sqlalchemy import (
 from sqlalchemy.dialects.postgresql import OID
 from sqlalchemy.orm import (
     Mapped,
+    Session,
     mapped_column,
     DeclarativeBase,
     relationship,
@@ -548,6 +549,13 @@ class SignalIndex(db.Model):  # type: ignore[name-defined]
         return self
 
     def load_signal_index(self) -> SignalTypeIndex[int]:
+        """Reads the serialized index from the read replica via large object API.
+
+        Safe against replica lag: commit_signal_index() commits the lobj via
+        raw_conn.commit() before the caller commits the row (with the new OID)
+        via the ORM session. WAL replays in order, so the replica cannot expose
+        the OID without already having the lobj.
+        """
         oid = self.serialized_index_large_object_oid
         assert oid is not None
         # If we were being fully proper, we would get the SignalType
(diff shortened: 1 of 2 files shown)


hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -455,16 +455,21 @@ class SignalIndex(db.Model):  # type: ignore[name-defined]
 
     serialized_index_large_object_oid: Mapped[int | None] = mapped_column(OID)
 
-    def index_lobj_exists(self) -> bool:
+    def index_lobj_exists(self, session=None) -> bool:
```
juanmrad:
```suggestion
    def index_lobj_exists(self, session: t.Optional[Session] = None) -> bool:
```

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -551,7 +556,7 @@ def load_signal_index(self) -> SignalTypeIndex[int]:
         # class no matter which interface we call it on.
         # I'm sorry future debugger finding this comment.
         load_start_time = time.time()
-        raw_conn = db.engine.raw_connection()
+        raw_conn = db.engines["read"].raw_connection()
```
juanmrad, line 570: Replica lag race: a caller can read a `SignalIndex` row whose `oid` was just written on the primary, then hit a replica that hasn't replicated the large object yet, and `lobject(oid, "rb")` will raise. 

It may be worth catching it and returning `None` so `get_signal_type_index` degrades gracefully instead of 500-ing. If the indexer already retries cleanly on the next tick. We could also just document with a docstring on the method.

hasher-matcher-actioner/src/OpenMediaMatch/tests/test_database.py
```diff
@@ -162,3 +163,58 @@ def test_store_content(app: Flask) -> None:
...
+def test_load_signal_index_uses_read_engine(app: Flask) -> None:
+    db_record = _create_index_record()
+    read_engine = database.db.engines["read"]
+
+    with patch.object(
+        read_engine, "raw_connection", wraps=read_engine.raw_connection
+    ) as mock_conn:
+        db_record.load_signal_index()
+        mock_conn.assert_called_once()
+
```
juanmrad, line 202:
```suggestion
def test_load_signal_index_uses_read_engine(app: Flask) -> None:
    db_record = _create_index_record()
    read_engine = database.db.engines["read"]
    write_engine = database.db.engine

    assert read_engine is not write_engine, (
        "Test relies on separate read/write engine instances"
    )

    with patch.object(
        read_engine, "raw_connection", wraps=read_engine.raw_connection
    ) as mock_read, patch.object(
        write_engine, "raw_connection", wraps=write_engine.raw_connection
    ) as mock_write:
        db_record.load_signal_index()

    mock_read.assert_called_once()
    mock_write.assert_not_called()
```

We can ensure we did not call write engine.

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
aokolish: The lobj is committed before the row — `commit_signal_index()` calls `raw_conn.commit()` on the large object, then the caller commits the row via `sesh.commit()`. WAL replays in order, so if the replica shows a row with OID X, the lobj for X is already replicated.

Added a docstring on load_signal_index noting this and lmk if you think I missed something.

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -543,6 +549,13 @@ def commit_signal_index(
         return self
 
     def load_signal_index(self) -> SignalTypeIndex[int]:
+        """Reads the serialized index from the read replica via large object API.
```
Dcallies: nit

```suggestion
        """
        Reads the serialized index from the read replica via large object API.
```

hasher-matcher-actioner/src/OpenMediaMatch/storage/postgres/database.py
```diff
@@ -543,6 +549,13 @@ def commit_signal_index(
...
+
+        Safe against replica lag: commit_signal_index() commits the lobj via
+        raw_conn.commit() before the caller commits the row (with the new OID)
+        via the ORM session. WAL replays in order, so the replica cannot expose
+        the OID without already having the lobj.
```
Dcallies: I am having trouble following this comment, perhaps it could be rephrased

```suggestion
        When using read/write replicas, as long as we read this object
        and its ID from the same db (including replicas), the lobj should be there, 
        as the lobj is written before the record with the ID.
        
        However, it is still possible to get an error due to races with the
        garbage collection process that deletes lobjs. 
```

Dcallies (approved):
No concerns on my end, but @juanmrad since you started the review, want to give you a chance to finish it, go ahead and merge if you are happy.

juanmrad (approved):
LGTM

PR #1982, opened 2026-05-18
[hma] Generate OpenAPI spec during website CI

juanmrad:
Summary
---------

Adds a script that boots the HMA Flask app with a minimal config and dumps the OpenAPI spec served at `/openapi/openapi.json` to a JSON file. 

The website CI and release workflows now run this dump before building VitePress, so the spec lands at `website/content/public/hma/openapi.json` and is bundled into the deployed site as `/hma/openapi.json`. A follow-up PR can render it as an interactive API reference page.

Test Plan
---------

Ran locally `python scripts/dump_openapi.py /tmp/openapi.json` and produced json data of the dump from openapi.


ThisIsMissEm (commented):
We should probably decouple the "generate the openapi.json" from the "build the website" task, we can use github artifacts for this I think.

Dcallies (approved):
Thanks for doing this! Let's see if it works

PR #1987, opened 2026-06-03
Fix mypy type-check errors in pytx CI

juanmrad:
Summary
---------

Fixes Type errors uncovered on last PR. This seems to be some mypy upgrades causing the issue. This pr fixes this.

Test Plan
---------

Mypy is happy :)


Dcallies (approved):
Yaaay!

PR #1992, opened 2026-08-11
[pytx] Drop redundant PDQ hex strings from PDQIndex sidecar

jess-upscrolled:
## Summary
- Stop storing PDQ hex strings in `PDQIndex.local_id_to_entry`; keep only match metadata (typically `bank_content_id`), since FAISS already holds the vectors.
- Migrate legacy pickled sidecars on unpickle via `__setstate__` so existing HMA index blobs keep working until rebuild.
- Add coverage that new pickles store metadata-only and that legacy `(hash, entry)` sidecars still query correctly.

This should lower steady-state matcher RSS for large PDQ indexes (one fewer Python string per hash in the sidecar). It does not address FAISS refresh/leak behavior.

## Test plan
- [x] `pytest threatexchange/signal_type/tests/test_pdq_index.py`
- [x] `black --check .` and `mypy threatexchange`
- [ ] Confirm HMA matcher can load an existing pickled index LOB and return matches
- [ ] After indexer rebuild, confirm new LOBs pickle metadata-only sidecars and memory footprint drops vs baseline


Yisu123:
你好，我是瞿文政，你的邮件我已经收到，我会尽快回复。

PR #1997, opened 2026-09-09
[hma] Support fetching media from S3-compatible storage

### LINKED ISSUE #1828: [hma] Ability to fetch media from S3-compatible storage
Currently when you want to work with media stored on an S3-compatible storage solution (AWS S3, DigitalOcean Spaces, Minio, etc), you need to either use presigned URLs and use the `GET /h/hash` endpoint, or you need to have a intermediary process first download the file from media storage and then send it to HMA via FormData / multipart request (`POST /h/hash`)

It'd be very useful to be able to configure HMA with credentials for a bucket and then pass a payload to HMA saying "fetch this object key from this bucket and hash it", so the API might look something like a JSON `POST /h/hash` with:

```
{
  "bucket": "example-media",
  "key": "/path/to/media/file.png"
}
```

This could also support doing a `POST` request for `url` based hashing too:
```
{
  "url": "https://some.url.example/path/to/media/file.png"
}
```
(potentially worth including an `expires_at` timestamp, such that HMA can quickly assert if the URL is still valid, in the case of queue-based hashing, instead of request-based)

julietshen:
Resolves #1828.

## Why
Today, to hash media in S3-compatible storage you must either presign a URL or run an intermediary that downloads the object and re-uploads it to HMA. This adds `s3://bucket/key` support so HMA can fetch directly.

## What
Rather than a new endpoint, `hash_url_content()` now branches on the `s3` URL scheme, so **every endpoint that already accepts a `url`** (`GET /h/hash`, banking content by URL, matching by URL, the UI) gains S3 support. The `http(s)` path is unchanged.

- `hash_s3_content()`: `head_object` for content-type + size pre-check, then streams `get_object` into a temp file, re-enforcing the size cap mid-stream in case `ContentLength` is absent/wrong.
- boto3 is an **optional extra** (`OpenMediaMatch[s3]`); a guarded import returns a clear 500 if missing, so the base image stays lean.
- Config: `S3_ENDPOINT_URL` / `S3_REGION_NAME` for S3-compatible stores (MinIO, DigitalOcean Spaces), plus an optional `ALLOWED_S3_BUCKETS` allowlist. **Note:** S3 URLs intentionally bypass the `http` SSRF / `ALLOWED_HOSTNAMES` checks (they are authenticated bucket reads), so `ALLOWED_S3_BUCKETS` is the parallel guardrail — flagging this for reviewer scrutiny.
- Started with a single global credential model per the discussion in the issue; per-bucket config can follow if needed.

## Test plan
Ran the CI jobs locally in Docker against a Postgres container:

- `black --check` — clean
- `mypy` — `Success: no issues found in 50 source files`
- `pytest test_hashing.py` — 20 passed (9 existing + 11 new S3 cases: happy path, endpoint/region wiring, invalid URLs, allowlist allow/deny, oversize-by-header, oversize-mid-stream, 404, missing-boto3)

## AI assistance disclosure
This change was developed with AI assistance (Claude). Per the repo's AI Coding Policy, the design decisions and code have been reviewed and the author can vouch for them; the one area to point reviewers to for extra scrutiny is the SSRF-bypass tradeoff noted above.

_Opened as a draft to serve as a reference implementation for a contributor picking up #1828._


julietshen:
Closing in lieu of #1998 !

PR #1998, opened 2026-09-09
[hma] Fetch & hash media from s3:// URLs

### LINKED ISSUE #1828: [hma] Ability to fetch media from S3-compatible storage

reitblatt:
# Summary

Closes #1828

Adds (optional) ability to read directly from S3 buckets by adding support for `s3://` URI schema to existing endpoints that already take in URL. This is cleaner and simpler than extending the POST endpoints to take in `application/json` as originally suggested in #1828.

Note: this approach stolen flagrantly from #1997 (h/t @julietshen!)

# Design choices
- Single credential to keep things simple
- boto3/s3 support is optional to minimize dependencies in base image

# Things to flag

We explicitly omit the SSRF (ALLOWED_HOSTS) checks for S3, and instead add an ALLOWED_S3_BUCKETS check. ALLOWED_HOSTS is meaningless for S3 URIs (no hostname), but we may still need to restrict HMA's reads more than the IAM policy does. This is the knob for that.

Coded with Claude Sonnet

# Test Plan
[X] black --check ./src
[X] mypy src/OpenMediaMatch
[X] pytest - 4 failed, 62 passed, 5 errors (failures + errors exist in main)
  - Passing includes 11 new test cases for S3


hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -141,62 +218,139 @@ def hash_url_content(
...
+    is not well-formed. Also enforces the optional ``ALLOWED_S3_BUCKETS``
+    allowlist (the S3 analogue of ``ALLOWED_HOSTNAMES``).
+    """
+    parsed = urlparse(media_url)
+    bucket = parsed.netloc
+    key = parsed.path.lstrip("/")  # S3 keys are never rooted with "/"
+    if parsed.scheme != "s3" or not bucket or not key:
+        abort(400, "Invalid S3 URL, expected s3://bucket/key")
+
+    allowed = current_app.config.get("ALLOWED_S3_BUCKETS")
```
juanmrad, line 263: We should reject if `ALLOWED_S3_BUCKETS` is not set. this opens up issues of like a potential attack where they can hash any object those credentials can `GetObject` (PDQ / video MD5 of private media) probe key existence via 403 vs 404 or force HMA to pull large objects (bandwidth / worker DoS)

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -141,62 +218,139 @@ def hash_url_content(
...
     """
-    if not is_valid_url(media_url):
-        abort(400, "Invalid or unsafe URL provided")
-
-    try:
-        # Get response with content length tracking
-        max_file_size = (
-            current_app.config.get("MAX_REMOTE_FILE_SIZE")
-            or DEFAULT_MAX_REMOTE_FILE_SIZE
+    if urlparse(media_url).scheme == "s3":
```
juanmrad, line 221: This is not good as `urlparse` could corrupt valid S3 keys. `_s3_split_url` uses `urlparse`. `?` and `#` are legal in S3 keys while `urlparse` treats them as query/fragment.

So for example if we have requests:

Request | Key used
-- | --
s3://bucket/foo#bar | foo
s3://bucket/foo?versionId=abc | foo (version is then ignored)

I think here we do not use `urlparse` for the key. Split `s3://` + first `/` and take the rest as the raw key. Document that query-param users must percent-encode (`%2B`, `%23`, `%3F`). 

Or even better accept bucket and key as separate fields so this does not depend on URL grammar.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -141,62 +218,139 @@ def hash_url_content(
...
+            500,
+            "S3 support requires the 'boto3' package. "
+            "Install it with 'pip install OpenMediaMatch[s3]'.",
+        )
+
+    return boto3.client(
+        "s3",
+        endpoint_url=current_app.config.get("S3_ENDPOINT_URL") or None,
+        region_name=current_app.config.get("S3_REGION_NAME") or None,
+    )
```
juanmrad, line 292: we should define a timeout for s3 this can be configurable ideally have some `botocore.config.Config(connect_timeout=..., read_timeout=...)` when building the client.

hasher-matcher-actioner/pyproject.toml
```diff
@@ -38,13 +38,18 @@ all = [
     "types-psycopg2",
     "types-python-dateutil",
     "gunicorn",
-    "flask_apscheduler"
+    "flask_apscheduler",
+    "boto3"
```
juanmrad, line 42: current published images won't be able to use this as such we should document that the deploy `ghcr.io/facebook/threatexchange/hma` image needs to do “install OpenMediaMatch[s3]”, for s3 support. 

We could just install s3 and it be part of the image even when not used or document the extend image note for this support. IMHO install by default should be acceptable, but if @Dcallies has thoughts they are welcomed too.

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -141,62 +218,139 @@ def hash_url_content(
...
+    from botocore.exceptions import BotoCoreError, ClientError
+
+    try:
+        return fn(**kwargs)
+    except ClientError as e:
+        status = int(e.response.get("ResponseMetadata", {}).get("HTTPStatusCode", 400))
+        if status in (403, 404):
+            abort(status, f"Could not access {media_url}")
+        current_app.logger.warning("S3 error for %s: %s", media_url, e)
+        abort(400, "Failed to fetch S3 object")
```
juanmrad, line 310: this hides any 5xx errors and converts to 400 which is not a right error code. We should map 5xx (and likely 301 wrong-region) to 502, same as `BotoCoreError`

hasher-matcher-actioner/src/OpenMediaMatch/blueprints/hashing.py
```diff
@@ -141,62 +218,139 @@ def hash_url_content(
...
+    s3_content_type = head.get("ContentType", "")
+    current_app.logger.debug("%s is type %s", media_url, s3_content_type)
+    signal_types = _resolve_signal_types(
+        s3_content_type, content_type_hint, signal_type_names
+    )
+
+    body = _s3_call(client.get_object, media_url, Bucket=bucket, Key=key)["Body"]
+    return _hash_chunks_to_signals(
+        body.iter_chunks(chunk_size=8192), signal_types, max_file_size
+    )
```
juanmrad, line 351: `abort(413)` mid-read leaves the `StreamingBody` open (connection leak under gunicorn). Mid-stream `botocore` errors also bypass `_s3_call` today and become an unhandled 500, because `iter_chunks` runs after `_s3_call` returns.

I think we can shape this function like:

```python
def _fetch_and_hash() -> dict[str, str]:
    head = client.head_object(Bucket=bucket, Key=key)
    content_length = head.get("ContentLength")
    if content_length is not None and content_length > max_file_size:
        abort(413, "Content too large")
    signal_types = _resolve_signal_types(
        head.get("ContentType", ""), content_type_hint, signal_type_names
    )
    body = client.get_object(Bucket=bucket, Key=key)["Body"]
    try:
        return _hash_chunks_to_signals(
            body.iter_chunks(chunk_size=8192), signal_types, max_file_size
        )
    finally:
        body.close()
return _s3_call(_fetch_and_hash, media_url)
```

That keeps HEAD on the same error mapper, closes the body even on 413, and maps transport errors during the read the same way as the RPC calls.

juanmrad (changes requested):

reitblatt:
Copying some feedback from @Dcallies in Discord:
> Do we have an end user for this functionality?
> Eg. An actual production deployment or prospective production deployment that depends on this feature?
> If yes, then I would say it's okay to include by defaults and enable it by default, if no, then I would either yagni or put it instead at the lower layer at the threat exchange signal type for URL and make it an optional library of some kind if it's theoretical usage

@ThisIsMissEm you originally requested this feature. Do you have a specific deployment in mind?

PR #1999, opened 2026-09-10
[hma] Fix pytest runs

reitblatt:
Summary
---------

4 tests were failing, and all 5 tests in test_match_api were erroring in my dev environment. The issue was that app fixture was loading OMM_CONFIG and not overriding it with the test config. 

Force OMM_CONFIG to the test config via monkeypatch so tests are independent of the surrounding environment.

Test Plan
---------

All 60 tests now pass in my environment.


reitblatt:
Note: CI is broken due to Debian 11 EOL. This PR is stacked on top of the fix (#2001), so both commits show up. We should merge this after #2001 lands.

juanmrad (approved):

PR #2000, opened 2026-09-10
[hma] Restore CHANGELOG.md

reitblatt:
## Summary

HMA's `CHANGELOG.md` was deleted along with the HMA 1.0 codebase in [#1556](https://github.com/reitblatt/threatexchange/issues/1556), the same day Open Media Match became HMA 2.0 ([#1557](https://github.com/reitblatt/threatexchange/issues/1557)), and never re-created — so 1.0.0 through 1.1.7 shipped without one while the docs kept linking to it. This restores it and fixes those docs.

## Where the entries come from

* 1.1.3–1.1.7 — # Full Changelog items are copied verbatim from each version's [GitHub Release](https://github.com/facebook/ThreatExchange/releases) notes (added in [#1957](https://github.com/reitblatt/threatexchange/issues/1957)). So they cover every PR merged repo-wide in that window, [pytx] included, and aren't sorted into :rocket:/:bug: callouts — categorizing would mean re-deriving them instead of quoting.
* 1.0.0–1.1.2 — predate Releases, so they stay reconstructed from git history in the categorized format, flagged as such in the preamble.

## Docs

* `CONTRIBUTING.md` — "changelog" now links to the restored file. Its stale docs/releasing.md path was fixed independently in [#2002](https://github.com/reitblatt/threatexchange/issues/2002), merged in here.
* `releasing.md` — now explains how to update the changelog, as a step after the release publishes, since the notes it copies from don't exist until the tag is cut.

## Notes for reviewers

* The actual version numbers are confusing here: #1557 calls the new code 'HMA 2.0', but the actual release numbers continue with 1.*. I've adopted the version numbers from the releases and version.txt for consistency. 
* The reconstruction boundary is now 1.1.3 (was 1.1.7 in the last revision).

Note: this diff and much of the PR text itself was generated by Claude Code Opus 5.0


hasher-matcher-actioner/CHANGELOG.md
```diff
@@ -0,0 +1,545 @@
+# Changelog
+
+This document and version tags are for reproducibility and the addition of migration guidance as needed for any breaking changes.
+
+Versions correspond to the contents of `version.txt`, the `ghcr.io/facebook/threatexchange/hma:<version>` container images, and the `hma-v<version>` git tags. See [releasing.md](../website/content/hma/releasing.md) for the release process.
+
+Entries are newest-first. Older entries were reconstructed from git history, so they are less detailed than newer ones. The section template used by each entry is at the [bottom of this file](#entry-template).
```
juanmrad: May be good to just define which tag the entries were reconstructed from git. 1.1.7

reitblatt: Done in 41b17936: the intro now says entries up to and including 1.1.7 (`hma-v1.1.7`) were reconstructed after the fact, and notes they were AI generated from the git commit history.

Dcallies:
I think the reason that we deleted the version.txt is because we replaced it with the automatically generated release notes associated with the Docker version, which are generated by github. I forgot if that took a workflow to enable, but let's stick with one or the other and disable whichever one we're not using.

reitblatt:
Updated the PR. For versions 1.1.3 -> 1.1.7 I pulled the changelogs from the actual github releases, and kept the AI generated notes for prior releases.

One open question: do we want to add an 'unreleased changes' section and have contributors update it with PRs, or keep the current workflow where the releaser pulls the changelog wholesale from the generated github releases?

Dcallies:
I am not sure if we bottomed out on this in the discussion in Discord, but is there value of having this in a file if we have it in github? Rather than having two, can we stick with one and not use whichever one we're not using.

My current intepretation is that this PR is proposing we return to using a file and disable the automatically generated changelogs, is that accurate?

Can also voice chat through this with you on Discord if that is easier.

PR #2001, opened 2026-09-11
Fix EOL debian bullseye base images breaking docker builds

reitblatt:
Debian 11 (bullseye) is EOL, so apt-get against its mirrors fails in CI. Switch to bookworm, matching the fix applied in coop (roostorg/coop#1138).

Only HMA and python-threatexchange needed a fix. Other images use Ubuntu 24, so not EOL/broken.

# Test Plan
CI works again!


juanmrad:
sorry I just saw you also made the pr. I just merged mine 😅 closing this one

PR #2002, opened 2026-09-11
[hma] Fix broken doc links in README and CONTRIBUTING

reitblatt:
Docs moved from hasher-matcher-actioner/docs to website/content/hma in #1973 but the links pointing at the old paths were never updated.

Test Plan
---------

Manually confirmed all the local links work in README.md and CONTRIBUTING.md


juanmrad (approved):
👍🏻

PR #2005, opened 2026-09-18
[pytx][stopncii] Skip malformed FetchHashes pages instead of stalling forever

### LINKED ISSUE #2004: [pytx][stopncii] Fetch wedges permanently when the StopNCII API returns a malformed JSON page
## Summary

Since ~2026-09-16 15:41 UTC, an OpenMediaMatch/HMA deployment's StopNCII exchange fetch fails on every cycle with `requests.exceptions.JSONDecodeError`. The failure is deterministic — the same corrupt page is requested every ~15 minutes for 40+ hours, because the fetch checkpoint only advances on a successful fetch. One malformed response body from the StopNCII API permanently wedges the exchange.

This appears to be **server-side malformed JSON** (HTTP 200 with a corrupt body), not an auth or pagination problem, and upgrading does not help: current `main` of `stopncii/api.py` still parses the response with an unguarded `response.json()`.

## Environment

- HMA image `ghcr.io/facebook/threatexchange/hma:1.0.21` (threatexchange as shipped in that image; Python 3.13 per traceback)
- OpenMediaMatch fetcher running the StopNCII collab (`StopNCIISignalExchangeAPI`)
- Failure started between 2026-09-16 15:41 UTC (last successful fetch) and the next cycle

## Traceback

```
ERROR in fetcher: STOP_NCII[stop_ncii] Failed to fetch!
  File "requests/models.py", line 976, in json
    return complexjson.loads(self.text, **kwargs)
  ...
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 181042 (char 181041)
During handling of the above exception, another exception occurred:
  File "/app/OpenMediaMatch/background_tasks/fetcher.py", line 63, in fetch
    _fetch(collab_store, signal_type_cfgs, collab)
  File "/app/OpenMediaMatch/background_tasks/fetcher.py", line 138, in _fetch
    for delta in api_client.fetch_iter(signal_types, checkpoint):
  File "threatexchange/exchanges/impl/stop_ncii_api.py", line 193, in fetch_iter
    for result in self.api.fetch_hashes_iter(start_timestamp=start_time):
  File "threatexchange/exchanges/clients/stopncii/api.py", line 270, in fetch_hashes_iter
    result = self.fetch_hashes(start_timestamp=start_timestamp, next_page=next_page)
  File "threatexchange/exchanges/clients/stopncii/api.py", line 224, in fetch_hashes
    json_val = self._get(StopNCIIEndpoint.FetchHashes, **params)
  File "threatexchange/exchanges/clients/stopncii/api.py", line 188, in _get
    return response.json()
  ...
requests.exceptions.JSONDecodeError: Expecting ',' delimiter: line 1 column 181042 (char 181041)
```

## Why it wedges permanently

1. `fetch_iter` resumes from `StopNCIICheckpoint.update_time` and requests pages with `startTimestamp=<checkpoint>`.
2. One page in the response window comes back as HTTP 200 with a truncated/corrupt JSON body. The failure is at the **response-body** level (`response.json()` in `_get`, line 188) — *before* the per-record `dacite` deserialization loop in `fetch_hashes`, which already has skip-and-log handling and never runs.
3. Because parsing fails, the client never sees `nextPageToken` or `nextSetTimestamp`, so the checkpoint never advances. The next cycle requests the exact same page and fails identically.

Observed: over 150 consecutive identical failures across two days, every cycle logging:

```
INFO in fetcher: STOP_NCII[stop_ncii] Resuming from 2026-09-16T15:41:20
ERROR in fetcher: STOP_NCII[stop_ncii] Failed to fetch!
... JSONDecodeError: Expecting ',' delimiter: line 1 column 181042 (char 181041)
INFO in fetcher: STOP_NCII[stop_ncii] Completed - 0 seconds
```

The same byte offset (`char 181041`) in every response indicates a single specific page in the StopNCII feed is malformed, not transient corruption.

## Evidence it is server-side

- The error offset is byte-identical across ~150+ attempts over two days.
- After advancing the stored checkpoint (`update_time`) forward by 15 minutes manually, the failure **moved**: the next cycle failed with a different offset (`char 68`), i.e. a different corrupt page. The cursor was no longer stuck, confirming the corruption is in the response data and tied to the requested window, not client state.
- HTTP status handling and retries are in place (`raise_for_status`, `Retry` on 429/5xx) — the corrupt responses are 200s.

## Current impact for operators

Without intervention, a single malformed page stops ingestion of new NCII hashes from StopNCII indefinitely. The only workaround is manually editing the stored checkpoint in the collab store's database (advancing `update_time`), skipping whatever records sit on the corrupt pages. That is operational and error-prone, and it silently drops the records in the skipped window. Fetches are currently still failing on later pages in that window.

## Suggested fix

Catch `requests.exceptions.JSONDecodeError` (and arguably other body-level errors) around the `response.json()` call in `_get` / `fetch_hashes`, and skip the page with a loud log, similar in spirit to the existing per-record skip ("Please open an issue if you see the logging statement below").

One design consideration for where to skip: when a page's body is unparseable, the client cannot read `nextPageToken` or `nextSetTimestamp` from it, so a page-level skip can't simply continue the same iteration. Options:

- Re-request the page (transient corruption heals; persistent corruption still wedges), then
- On repeated failure of the same page, advance `start_timestamp` by some delta and continue, logging the skipped window — accepting that records on the corrupt pages are dropped, which is strictly better than a permanent wedge of the whole exchange.

Alternatively/additionally, it may be worth reporting the malformed page contents (first ~200 bytes around the failure offset) to StopNCII when detected, so the source data can be fixed.

Happy to provide more logs or test a patched build.

gdubicki:
A HTTP 200 with a truncated/corrupt JSON body currently crashes response.json() before the per-record skip path, so the checkpoint never advances and HMA retries the same page every cycle.

Retry the page once, then advance start_timestamp by 10 minutes and continue so ingest can recover. Log nearby body bytes to help report the corruption to StopNCII.

Fixes #2004


Dcallies:
Hey @gdubicki - huge thanks for the PR. I'm emailing the Southwest Grid for Learning (SWGfL) about this issue, who owns StopNCII. My concern with advancing the timestamp when you hit malformed records is that these represent permanently lost pieces of data. There are really no good answers here other than the API needs to work, and for the API to work properly, we need to notify the platform owners. Let me give them a short time to reply, and if for some reason it's not fixable, we'll proceed with your proposed fix. Does that work for you?

